PR #12, opened 2012-10-01
Fix #8. Support Python 2.6

### LINKED ISSUE #8: Support Python versions lower than 2.7, too

mbraak:
Hello,

Thank you for your this useful tool. This pull request adds Python 2.6 support.

Greetings

### DIFF
--- a/bin/pip-dump
+++ b/bin/pip-dump
@@ -7,10 +7,10 @@ import logging
 import tempfile
 from itertools import dropwhile, takewhile
 from functools import partial
-from subprocess import check_call as _check_call, check_output as _check_output
+from subprocess import check_call as _check_call
+import subprocess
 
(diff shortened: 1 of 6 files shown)


bionikspoon:
@nvie 

On the topic of 2.6, is there any particular reason `pip-tools` is not supporting it, or at the very least testing against it?

I'm wondering if there's anything fundamentally breaking that 2.6 can't reasonably do?  If not, I'd love to make this work.  I like this tool, but trying to use it in my build pipeline is putting me in an awkward position when I have to deal with py26.

Currently the error points to a dict comprehension, which is unsupported in Py26. 

I found a few
- https://github.com/nvie/pip-tools/blame/master/piptools/sync.py#L60
- https://github.com/nvie/pip-tools/blame/master/piptools/sync.py#L94

Issue #87, opened 2014-05-03
UnicodeEncodeError in download_package: creates 0-byte cached file


blueyed:
1. I have run `pip-compile` (from the future branch), and it failed on the `dj-database-url` package when writing the file contents.

The problem appears to be that pip's `get_file_content` returns a unicode string.

If have tried encoding it as `utf-8`, but that results in an invalid/broken zip/whl file.

```
===> Resolving full tree
- Finding best package matching ['dj-database-url']
    PIP said: Getting page https://pypi.python.org/simple/dj-database-url/
    PIP said: URLs to search for versions for dj-database-url:
    PIP said: * https://pypi.python.org/simple/dj-database-url/
    PIP said: Analyzing links from page https://pypi.python.org/simple/dj-database-url/
    PIP said:   Found link https://pypi.python.org/packages/2.7/d/dj-database-url/dj_database_url-0.3.0-py2.py3-none-any.whl#md5=58211401b11c49ba27deca8183c3061d (from https://pypi.python.org/simple/dj-database-url/), version: 0.3.0
    PIP said:   Found link https://pypi.python.org/packages/source/d/dj-database-url/dj-database-url-0.1.2.tar.gz#md5=8b285dba950c9ca6e3f6082be64b2faf (from https://pypi.python.org/simple/dj-database-url/), version: 0.1.2
    PIP said:   Found link https://pypi.python.org/packages/source/d/dj-database-url/dj-database-url-0.1.3.tar.gz#md5=175111400fee7c1f6ce4a4829c8b26ee (from https://pypi.python.org/simple/dj-database-url/), version: 0.1.3
    PIP said:   Found link https://pypi.python.org/packages/source/d/dj-database-url/dj-database-url-0.1.4.tar.gz#md5=ee1f46a6b0ca3329e6701b5578b2aee6 (from https://pypi.python.org/simple/dj-database-url/), version: 0.1.4
    PIP said:   Found link https://pypi.python.org/packages/source/d/dj-database-url/dj-database-url-0.2.0.tar.gz#md5=faaf1e1f06f738f72ef964696c46e379 (from https://pypi.python.org/simple/dj-database-url/), version: 0.2.0
    PIP said:   Found link https://pypi.python.org/packages/source/d/dj-database-url/dj-database-url-0.2.1.tar.gz#md5=63ad283bb5931e8d3842e5b9edd7a1fd (from https://pypi.python.org/simple/dj-database-url/), version: 0.2.1
[...]

PR #91, opened 2014-05-03
WIP/TBD: use pip's unpack_http_url and download cache

blueyed:
Using `unpack_http_url` works around issue #87 and makes pip handle the
unpacking logic itself.

The drawback appears to be that the packages in pip-tools cache are not
zipped/compressed anymore!
I am not sure if this is such a bad thing (with regards to disk space usage).

It looks for `['pydist.json', 'metadata.json']` in wheel archives - I found the latter to be used with raven-python.

It adds a warning if no dependency information could be found, and removes the old handling for VCS urls, which appears to have defaulted to setup.py anyway?!:
-                    deps = self.read_package_requires_file(path)
-                    logger.warn("Package has no dependency information: %s" % (path,))

Fixes nvie#87

### DIFF
--- a/piptools/package_manager.py
+++ b/piptools/package_manager.py
@@ -22,10 +22,10 @@ except ImportError:
 from functools import partial
 import urlparse
 
-#from pip.backwardcompat import ConfigParser
-from pip.download import get_file_content, unpack_vcs_link
+from pip.backwardcompat import ConfigParser
+from pip.download import unpack_http_url, unpack_vcs_link
 from pip.index import Link, PackageFinder, package_to_requirement
-#from pip.locations import default_config_file
+from pip.locations import default_config_file
 from pip.req import InstallRequirement
 from pip.util import splitext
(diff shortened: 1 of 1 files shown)


blueyed:
@nvie 
Can you provide some feedback regarding my last comment?
Wouldn't it be good to use pip's download cache? (which certainly has changed / will change with newer pip versions)

nvie:
TBH, I think I want to redesign a few things at the heart of pip-tools. The need for downloading was once very pragmatic: we needed to check which dependencies a certain package version has declared, and there was no way of simply looking that up. It would be awesome if PyPI could offer this API, but it does not, at least not yet. Downloading these packages and persisting them saved us a re-download of them for future runs. But the only thing we really need from the packages is which dependencies there are; not their contents. Since we now have the dep_cache, we could download the package, look at it, record those deps in the dep_cache, throw away the package cache. So I think we can remove the entire package caching entirely.

PR #125, opened 2015-01-08
Fix issues with current pip

blueyed:


blueyed:
Should have been against the "future" branch: see #126.

PR #131, opened 2015-02-01
Handling updated pip freeze output 

michaeljones:
The pip freeze command output has been updated to say 'pip freeze' instead of 'pip --freeze'.

We switch from a string match to a regex so that we can handle both the old and new styles.

I was very interested to see the `cram` tests as I've never heard of it before. I've done my best to include tests which cover the case we're interested in. As pip-dump calls `pip freeze` internally it is quite hard to 'mock' it out as such so we have two tests which install different versions of pip before repeating the pattern.

Hope it is acceptable,
Michael

### DIFF
--- a/bin/pip-dump
+++ b/bin/pip-dump
@@ -2,6 +2,7 @@
 from __future__ import absolute_import
 import os.path
 import glob
+import re
 import argparse
 import logging
 import sys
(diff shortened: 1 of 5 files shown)


michaeljones:
Ah, I removed argparse because it is installed on my system and I cannot get my python/virtualenv setup to ignore my system modules so the tests working working for me. I'll update my new tests to work on travis.

The others were failing anyway though I will see if they can be trivially updated.

michaeljones:
I have done my best. Feel free to ignore the last two commits as they aren't strictly related but they do attempt to fix the other test failures.

nvie:
Thanks for the hard work, @michaeljones!

michaeljones:
No trouble. Thanks for writing the tool, I've found it very useful!

PR #134, opened 2015-02-20
Extract wheel dependencies from metadata.json as well

gandelman-a:
I'm failing to extract dependencies from wheels that have theirs
packaged in a metadata.json instead of a pydist.json.  This updates
extract_dependencies() to also check for a metadata.json.


nvie:
Thanks :)

PR #145, opened 2015-04-10
Fix tests to reflect updated dependencies

mlew:
Tests on master were breaking due to outdated package dependencies


nvie:
Thanks!

PR #146, opened 2015-04-12
Future: fix cram tests

blueyed:
This improves and fixes the cram tests.

`tests/pip-compile.t` is  still failing, but that is because of missing `pip-sync` etc.

### DIFF
--- a/tests/pip-compile-allows-inheritance.t
+++ b/tests/pip-compile-allows-inheritance.t
@@ -15,7 +15,8 @@ Run pip-compile to generate the requirements.txt file. Resulting
 file should contain python-dateutil from base.in and it's dependencies
 as well.
 
-  $ pip-compile develop.in >/dev/null 2>&1
+  $ pip-compile develop.in
+  Dependencies updated.
 
   $ cat develop.txt
   python-dateutil==* (glob)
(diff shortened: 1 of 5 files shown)


nvie:
I've redesigned everything, making all tests useless (see #152). We should add them back in, but it requires a little bit of thought and organization, so not just yet!

blueyed:
@nvie 
Great news! (although you could have announced it before maybe, or was it spontaneous? :))

Anyway, why don't you think that the current cram test suite could be used until there is something better?
At the very least this would show up any regressions/changed caused by the rewrite, too.

nvie:
Hey Daniel. Yeah, sorry for this to be a little sudden. It was indeed rather spontaneous. I was looking at these incoming pull requests from last week, and felt like they were all hacks that were necessary due to some design choices made in the past that weren't the best way forward. (This is not to say the authors were putting hacks together! Instead, the core wasn't simple enough to support many use cases in the first place.)

Then I just thought "well, screw this, I'll try to rewrite it from scratch right now".

I feel that sometimes an issue with spending free time on open source projects is that some work simply needs to be done that requires a burst of attention, rather than a few hours here and there.

blueyed:
Thanks for your reply, I totally agree.

However, I still think that the existing cram tests would be still useful (when fixed / adjusted) for now.

Issue #135, opened 2015-02-23
pip-review suggesting upgrade to alpha/beta/rc


kegan:
pip-review suggesting upgrade to alpha/beta/release candidate. It seems like this happen (not limited to):

* Python 3.4.2
* Python 2.7.9

`$ pip-review`
`Django==1.8a1 is available (you have 1.7.4)`
`ipython==3.0.0rc1 is available (you have 2.4.1)`

carschar:
I can confirm the behavior. I think pip-review should be enhanced to mimic current pip behavior, which is to by default only consider release versions. This means alpha, beta, and rc would not be recommended. This is overriden in pip with the "--pre" flag.

I may take a shot at adding this, if no one else a pull request out yet. It sounds like fun.

b4stien:
Same behaviour here. From a user's perspective it's quite a biggie.

nvie:
I'm happy to merge this is if someone wants to make a pull request for this.

b4stien:
After a quick look: it will not be straightforward.

The fact that a release is a pre-release is not specified on PyPI (see for instance https://pypi.python.org/pypi/SQLAlchemy/1.0.0b5/json). I'm not sure how `pip` is able to handle them. Parsing the version and inferring from there maybe ?

b4stien:
I'm currently working on a PR for this.

It will use https://pypi.python.org/pypi/packaging, do you prefer it vendorized (like pip does) or declared as a dependency?

nvie:
Hi @b4stien, that's great. I'm wondering what "it" refers to? And what do you mean by vendorized/declared-as-dependency exactly?

b4stien:
By "it" I meant "my PR", sorry.

Vendorizing a dependency is the process of copying/pasting its source inside yours. The alternative is to "declare" this dependency for pip/setuptools to grab it at install time (that's what I did in my PR).

nvie:
Hi @b4stien, thanks for that! This weekend I have been trying out a few things that could be changed about the current pip-tools implementation. A _lot_ of the requirement parsing, spec resolving, etc. has been NIH'ed by me when I first wrote this tool and it turns out that most of this logic is also accessible in a reusable fashion inside pip — at least it is now.

One of the current problems with the pip-compile code is that I think it has become way too complicated, and I basically want to rewrite it now that I'm more familiar with pip/packaging/setuptools' code bases, and with the experience of writing it the first time, of course.

[...]

PR #148, opened 2015-04-12
Add pre-release logic

b4stien:
Add a command line flag (`--pre`) to enable pre-releases which are disabled by default. See #135.


b4stien:
Tests look pretty obscure to me (especially when it comes to dependencies, `argparse` seems to bee manually installed in every test...).

I guess `packaging` (the new dependency) should be added somewhere there but no clear idea where.

nvie:
I've included this in the redesign #152, too!

PR #150, opened 2015-04-12
future: use pickle protocol 2 for dependency cache

blueyed:
This makes is usable for both Python 2 and 3.

Fixes https://github.com/nvie/pip-tools/issues/149


nvie:
This is also fixed by the rewrite #152. Thanks for the work nonetheless!

PR #153, opened 2015-04-16
Massive rewrite

nvie:
This rewrites the future branch in its entirety. I deemed this necessary to get back to a simpler source code base that contained fewer hacks and supported more pip use cases. The previous code on the future branch had grown several hacks to support all pip use cases that people hit in practice, whereas my initial work focused mainly on the dependency resolver implementation.

I had also manually implemented, NIH-style, a lot of helper classes (think Spec, SpecSet, etc.) that, as it turns out, also exist in the pip and packaging code bases under a different name. The pip (and pkg_resources, and packaging) has become vastly higher quality over the last years, which is great news! I am now reusing as much of pip/pkg_resources/packaging as possible, and let maintenance for those implementations lie within the PyPA projects.

This new version is what you could call "pip-compile" if it was invented today.

Some highlights:

- Relies on the latest version of pip (>= 6.1)
- Supports all operator styles when resolving, including the previously unsupported `~=` and `==1.*` specifiers
- Supports the `--pre` flag, to allow prereleases (beta, alpha, rc, dev versions). Otherwise, only stable versions will be used
- Output on the terminal is made much more beautiful to the eye (and colorized when terminal supports it)
- Verbose mode now makes the process much easier to inspect/follow along
- The new dependency cache uses JSON to serialize, and will serialize per-Python version, so no more Python 2/3 conflicts
- Major speedup of most things
- Proper support for `--index-url` and `--find-links`
- Proper support for `-e` VCS style packages: they are now 1st-class citizens
- Source annotation has become vastly more succinct and useful (see screenshots below)

I've also dropped some things.

- Tests are gone for now. I want them back in, but need to think about how. They certainly won't be cram tests again, and they certainly should not hit real PyPI. Relying on external state, as the books said, was painfully brittle :)
- Simplified the interface: just pass it a single *.in file, and the output is written to stdout. So a typical invocation is `pip-compile requirements.in > requirements.txt`. I'm sure I want to put in the automatic writer again, but I want to think a little more about the UI I want for this multi-input-multi-output, as the previous design was confusing to people.


jonathanchu:
giphy

klinkin:
2fbf52c3

PR #160, opened 2015-05-12
PyPIRepository: cache return value of get_dependencies

blueyed:
This improves performance, especially for VCS editable dependencies.

Fixes: https://github.com/nvie/pip-tools/issues/159

Before:
pip-compile -v requirements/base.in  5,33s user 0,83s system 38% cpu 16,121 total

After:
pip-compile -v requirements/base.in  3,44s user 0,40s system 45% cpu 8,346 total

I have only a handful of editables, and it took 3 rounds to become stable.

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -41,6 +41,9 @@ class PyPIRepository(BaseRepository):
         # project
         self._available_versions_cache = {}
 
+        # Cache for get_dependencies.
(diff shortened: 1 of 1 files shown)


blueyed:
The test failures are unrelated: see #167 for fixes.

blueyed:
This is included in https://github.com/nvie/pip-tools/pull/169 also.

PR #162, opened 2015-05-15
Perform PyPi version lookups in parallel.

mostlyuseful:
Use a thread pool to lookup multiple packages concurrently.
Packages that block for a long time are printed to the debug log using a decorator.
Needs at least Python 2.6 for built-in multiprocessing.

I wrote this patch because package name guessing takes a long time on some of my virtualenvs. By using parallel workers, guessing can continue while normal lookups go through unhindered (of course only until all workers are blocked with guesswork).

I don't know if the master branch is still supported, please advise.

### DIFF
--- a/bin/pip-review
+++ b/bin/pip-review
@@ -3,10 +3,12 @@ from __future__ import absolute_import
 import os
 import re
 import argparse
-from functools import partial
+from functools import partial, wraps
 import logging
 import sys
 import json
+import threading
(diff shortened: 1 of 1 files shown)


blueyed:
Cool!
Would be probably good to have this with `pip-compile` from the future branch, too.

nvie:
Closing this since `pip-review` is gone with the release of 1.0: http://nvie.com/posts/pip-tools-10-released/

Issue #130, opened 2015-01-30
pip-compile should support the same version specifiers as pip


electrofelix:
Was trying pip-compile out since I agree with the idea that just using the standard approach to requirements is not sufficient but handing to manually track all dependencies is somewhat painful.

Ran into a problem where pip would accept a version specifier that pip-compile raises an exception on, seems that PyPi may not have been overly strict on the versions permitted.

My requirements.in file had the following entry:
`GitPython>=0.3.2.RC1,!=0.3.2`

pip-compile chokes on the 0.3.2.RC1, and although you won't find this in PyPi's current version avialable for GitPython it is still available.

Used the following command to see:

```
$ pip install GitPython==0.3.2.RC2
Downloading/unpacking GitPython==0.3.2.RC2
  Could not find a version that satisfies the requirement GitPython==0.3.2.RC2
  (from versions: 0.1.7, 0.2.0-beta1, 0.3.0-beta1, 0.3.0-beta2, 0.3.1-beta2, 0.3.2.1,
  0.3.2.RC1, 0.3.2, 0.3.3, 0.3.4, 0.3.5, 0.3.6)
Cleaning up...
```

I can work around this by changing my requirements.in to use `GitPython>0.3.1,!=0.3.2` but it still seems like it would make sense to support all versions that PyPI and pip accept even if 0.3.2.RC1 > 0.3.2 following the python version comparison rules.

I was also able to patch around this with the following change:

```
--- a/piptools/version.py
+++ b/piptools/version.py
@@ -191,7 +191,7 @@ class Matcher(object):
         return self._string


-PEP426_VERSION_RE = re.compile(r'^(\d+(\.\d+)*)((a|b|c|rc)(\d+))?'
+PEP426_VERSION_RE = re.compile(r'^(?i)(\d+(\.\d+)*)(\.?(a|b|c|rc)(\d+))?'
                                r'(\.(post)(\d+))?(\.(dev)(\d+))?'
                                r'(-(\d+(\.\d+)?))?$')

```

I figure the correct approach is to work out what pip supports and use that.

davidovich:
This now works as expected:
```
echo "GitPython==0.3.2.rc1" > requirements.in
pip-compile 
```

Note that `--pre` can also be used to allow pre-releases.

Reopen if you feel that this is still a problem.

PR #163, opened 2015-05-15
Future pin list

ssalonen:
I've worked on the future branch and added support for 
- version syntax as supported by pip (#130)
- support for --pre and --index-url
- new "pip-compile-pinned" cli tool for resolving a requirements file against a pin list, as opposed to downloading latest and greatest version out of pypi.

How do you want to go forward with these changes? I noticed that some rebasing is needed in order to merge.


blueyed:
Sadly that's probably not fit so good anymore, after the rewrite that has taken place recently.

ssalonen:
Well actually this is good news...the rewrite implements almost everything in this pull request (like --pre, --index-url and relaxed version patterns).

However, those were just nice-to-have features for me. What I was really looking for is the "PinnedPackageManager", which basically allows the user to have a pin list which is used in the compilation phase. This is quite different from the pip tools compile behaviour which resolves against latest versions found in the pypi/repo. We are planning to use this kind of pinning functionality in our CI environment to test the package with different version sets (e.g. pkg1==1.0 & pkg2==1.2 and pkg==1.1 & pkg==1.3). 

Do you think this kind functionality would be suitable in pip-tools or should it be a separate? Actually when writing this I realized that something similar could be done using devpi perhaps...

I'm happy to review the #169, hope that I have some time next week.

PR #167, opened 2015-06-17
future: fix tests

blueyed:
This fixes three test failures/errors, and adds tests for OutputWriter (for the moved annotation code).

(this is included in #169)

PR #158, opened 2015-05-12
Add constraint to UnsupportedConstraint error in _check_constraints

blueyed:
This provides helpful information, in case you have larger
input files.

This improves https://github.com/nvie/pip-tools/commit/b53f79b1006f87ed213a8b46fdcdd144c073fd1b.

PR #169, opened 2015-06-17
future: re-add support for non-editable VCS urls; fix pinning

blueyed:
This allows for regular/non-editable VCS URLs again, and will pip them
to the current commit.  This gets now done for editable VCS URLs, too.

(This includes #167 to fix the existing tests, and #160 + #158)

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -41,6 +41,9 @@ class PyPIRepository(BaseRepository):
         # project
         self._available_versions_cache = {}
 
+        # Cache for get_dependencies.
+        self._get_dependencies_cache = {}
+
         # Setup file paths
         self.freshen_build_caches()
         self._download_dir = os.path.expanduser('~/.pip-tools/pkgs')
@@ -99,6 +102,10 @@ class PyPIRepository(BaseRepository):
         dependencies (also InstallRequirements, but not necessarily pinned).
         They indicate the secondary dependencies for the given requirement.
         """
+
+        if ireq in self._get_dependencies_cache:
+            return self._get_dependencies_cache[ireq]
+
         if not os.path.isdir(self._download_dir):
             os.makedirs(self._download_dir)
         if not os.path.isdir(self._wheel_download_dir):
@@ -116,4 +123,5 @@ class PyPIRepository(BaseRepository):
                                 ignore_installed=True,
                                 session=self.session)
         dependencies = reqset._prepare_file(self.finder, ireq)
-        return set(dependencies)
+        self._get_dependencies_cache[ireq] = set(dependencies)
+        return self._get_dependencies_cache[ireq]
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -11,7 +11,7 @@ from pip.req import InstallRequirement
 from pip.req.req_set import RequirementSet
 
 from ..exceptions import NoCandidateFound
-from ..utils import is_pinned_requirement, lookup_table
+from ..utils import lookup_table
 from .base import BaseRepository
 
 try:
@@ -76,7 +76,7 @@ class PyPIRepository(BaseRepository):
         Returns a Version object that indicates the best match for the given
         InstallRequirement according to the external repository.
         """
-        if ireq.editable:
+        if ireq.editable or ireq.link:
             return ireq  # return itself as the best match
 
         all_candidates = self.find_all_versions(ireq.name)
@@ -99,9 +99,6 @@ class PyPIRepository(BaseRepository):
         dependencies (also InstallRequirements, but not necessarily pinned).
         They indicate the secondary dependencies for the given requirement.
         """
-        if not (ireq.editable or is_pinned_requirement(ireq)):
-            raise TypeError('Expected pinned or editable InstallRequirement, got {}'.format(ireq))
-
         if not os.path.isdir(self._download_dir):
             os.makedirs(self._download_dir)
         if not os.path.isdir(self._wheel_download_dir):
@@ -109,8 +106,14 @@ class PyPIRepository(BaseRepository):
 
         reqset = RequirementSet(self.build_dir,
                                 self.source_dir,
-                                download_dir=self._download_dir,
+                                # Pass no download_dir for editables, otherwise
+                                # pip uses "export" instead of "obtain"
+                                # (dropping the VCS revision info).
+                                download_dir=None if ireq.editable else self._download_dir,
                                 wheel_download_dir=self._wheel_download_dir,
+                                # Ignore installed packages: This is required
+                                # for the source_dir of VCS URLs to be created.
+                                ignore_installed=True,
                                 session=self.session)
         dependencies = reqset._prepare_file(self.finder, ireq)
         return set(dependencies)
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -49,11 +49,21 @@ class PyPIRepository(BaseRepository):
         self._download_dir = os.path.expanduser('~/.pip-tools/pkgs')
         self._wheel_download_dir = os.path.expanduser('~/.pip-tools/wheels')
 
+        # References to old build dirs (to prevent too early pruning).
+        self._old_build_dirs = []
+
     def freshen_build_caches(self):
         """
         Start with fresh build/source caches.  Will remove any old build
         caches from disk automatically.
         """
+        # Keep old build dirs around: they are required by
+        # format_requirement/get_src_requirement.
+        if hasattr(self, '_build_dir'):
+            self._old_build_dirs.append(self._build_dir)
+        if hasattr(self, '_source_dir'):
+            self._old_build_dirs.append(self._source_dir)
+
         self._build_dir = TemporaryDirectory('build')
         self._source_dir = TemporaryDirectory('source')
 
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -9,6 +9,7 @@ from itertools import chain, count
 import click
 from first import first
 from pip.req import InstallRequirement
+from pip.vcs import vcs
 
 from .cache import DependencyCache
 from .exceptions import UnsupportedConstraint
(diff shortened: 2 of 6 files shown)


blueyed:
I've squashed / cleaned up the commits.

Please let me know if the topics should be split again, but it's mainly only the fix-tests PR (#167) that's not required to be in here.

ssalonen:
Had a look at the commits, I'm afraid I don't have enough in-depth understanding to go through all in detail but generally speaking seems OK to me.

blueyed:
@nvie 
I think this should be re-opened, too.

nvie:
Sorry 'bout that — reopened.

piptools/repositories/pypi.py
```diff
@@ -99,8 +112,9 @@ def get_dependencies(self, ireq):
         dependencies (also InstallRequirements, but not necessarily pinned).
         They indicate the secondary dependencies for the given requirement.
         """
-        if not (ireq.editable or is_pinned_requirement(ireq)):
-            raise TypeError('Expected pinned or editable InstallRequirement, got {}'.format(ireq))
+
+        if ireq in self._get_dependencies_cache:
+            return self._get_dependencies_cache[ireq]
```
nvie: Is this a perf optimization? I really like to avoid adding perf logic in the middle of the processing, as it complicates the logic and makes it hard to follow along what's happening. Let's solve it first, and optimize for performance later. If this is not strictly necessary for the functionality, please remove it from this PR.

piptools/repositories/pypi.py
```diff
@@ -109,8 +123,15 @@ def get_dependencies(self, ireq):
...
+                                download_dir=None if ireq.editable else self._download_dir,
                                 wheel_download_dir=self._wheel_download_dir,
+                                # Ignore installed packages: This is required
+                                # for the source_dir of VCS URLs to be created.
+                                ignore_installed=True,
                                 session=self.session)
         dependencies = reqset._prepare_file(self.finder, ireq)
-        return set(dependencies)
+        self._get_dependencies_cache[ireq] = set(dependencies)
+        return self._get_dependencies_cache[ireq]
```
nvie: Here, too. Please avoid caching things in the middle of the core logic. This will quickly explode the complexity. Untangling this stuff was my main driver behind the rewrite in the first place.

piptools/resolver.py
```diff
@@ -87,14 +88,17 @@ def resolve(self, max_rounds=10):
 
     def _check_constraints(self):
         for constraint in chain(self.our_constraints, self.their_constraints):
-            if constraint.link is not None and not constraint.editable:
```
nvie: This also removes the `not constraint.editable` condition — was this intentional? After trying it out, whether or not `-e` is added to VCS URLs seems to make no difference, but I don't know if it should.

piptools/resolver.py
```diff
@@ -87,14 +88,17 @@ def resolve(self, max_rounds=10):
...
-                msg = ('pip-compile does not support URLs as packages, unless they are editable '
-                       '(perhaps add -e option?)')
-                raise UnsupportedConstraint(msg)
+            if constraint.link and constraint.link.scheme not in vcs.all_schemes:
```
nvie: Is there perhaps a convenient property on a `Link` object in the `pip` project that would expose this is-this-a-vcs-url check (`constraint.link.scheme not in vcs.all_schemes`) publicly without having to rely on implementation details like fetching that `vcs` object? Ideally I don't want to know about this at all.

piptools/resolver.py
```diff
@@ -185,11 +189,7 @@ def get_best_match(self, ireq):
             Flask==0.10.1 => Flask==0.10.1
 
         """
-        if ireq.editable:
-            # NOTE: it's much quicker to immediately return instead of
-            # hitting the index server
-            best_match = ireq
-        elif is_pinned_requirement(ireq):
+        if ireq.editable or ireq.link or is_pinned_requirement(ireq):
```
nvie: I don't understand the need for this logic change.

piptools/resolver.py
```diff
@@ -210,12 +210,12 @@ def _iter_dependencies(self, ireq):
...
         changed at any time.
         """
-        if ireq.editable:
+        if ireq.editable or ireq.link:
             for dependency in self.repository.get_dependencies(ireq):
                 yield dependency
             return
         elif not is_pinned_requirement(ireq):
-            raise TypeError('Expected pinned or editable requirement, got {}'.format(ireq))
+            raise TypeError('Expected pinned or editable requirement, or link, got {}'.format(ireq))
```
nvie: Do other type of requirements even exist?

piptools/utils.py
```diff
@@ -17,10 +21,33 @@ def format_requirement(ireq):
...
+        if ireq.source_dir and os.path.exists(ireq.source_dir):
+            line = get_src_requirement(dist=ireq.get_dist(),
+                                       location=ireq.source_dir,
+                                       find_tags=False)
+            line_link = Link(line)
+
+            # Prepend 'vcs+' which gets left out for 'git://', 'hg://' and
+            # 'bzr://' URLs.  Ref: https://github.com/pypa/pip/pull/2913.
+            if not '+' in line_link.scheme:
+                line = '{}+{}'.format(line_link.scheme, line)
```
nvie: I really want to avoid reimplementing things that should be in pip. If this is currently not in pip, my judgment is we simply cannot use it. Please remove this hack and wait for this to be pulled into pip. I'm happy to bump the requirement to the then-latest pip version — no problem.

piptools/resolver.py
nvie: The proper way of adding this would be to contribute this convenience method on `Link` to pip itself, have that PR pulled, wait for the pip release, and then bump the requirement to the latest pip. (See also the comment below about not relying too much on pip's implementation details.)

piptools/repositories/pypi.py
```diff
@@ -109,8 +123,15 @@ def get_dependencies(self, ireq):
...
                                 self.source_dir,
-                                download_dir=self._download_dir,
+                                # Pass no download_dir for editables, otherwise
+                                # pip uses "export" instead of "obtain"
+                                # (dropping the VCS revision info).
+                                download_dir=None if ireq.editable else self._download_dir,
                                 wheel_download_dir=self._wheel_download_dir,
+                                # Ignore installed packages: This is required
+                                # for the source_dir of VCS URLs to be created.
+                                ignore_installed=True,
```
blueyed: IIRC it is a special case in `RequirementSet()` - without this, the source dir would not be created.

piptools/utils.py
blueyed: It is waiting is https://github.com/pypa/pip/pull/2913.

piptools/resolver.py
blueyed: I wasn't sure.

piptools/resolver.py
blueyed: Yes, that would be `Link.is_artifact` (see desc for https://github.com/blueyed/pip-tools/commit/b6c4a0b12fcc5f5359dcd72b82d7dd3f54a5e20f).

bh:
Any news here? I need this :-)

skolsuper:
:+1:

piptools/repositories/pypi.py
blueyed: IIRC it comes from https://github.com/blueyed/pip-tools/commit/82aec97a10d5af438eff86df1f3921bbf83ff029. It provides both a performance optimization, but is required actually, because otherwise `pip` would complain about an existing build dir.

piptools/repositories/pypi.py
blueyed: I think here the same reason from above applies.

piptools/resolver.py
blueyed: This PR  is also about allowing editable requirements again, so yes, it was intentional.

piptools/resolver.py
blueyed: Should I change this?
I would rather leave it as-is for now, but then fix this later.
But I could also do a fixup-commit for this PR.

piptools/resolver.py
blueyed: It used `best_match = ireq` for both `ireq.editable` and `is_pinned_requirement(ireq)`.  This hunk adds `ireq.link` to it, which includes everything to be downloaded.
I cannot remember if that is correct/required really from the top of my head now, sorry.
Maybe it should also use `Link.is_artifact`?!

blueyed:
I've spent a lot of time on this when creating the PR and have tried to comment as good as possible on @nvie's remarks.

I suggest trying this PR, possibly rebased on / merged with master, and report back if it works for you.

stefanfoulis:
Does this PR also handle installing from a source .tar.gz (from an url)?

bh:
:+1: for merging this PR!

mkleehammer:
Any movement on this?  I could really use this.

blueyed:
@mkleehammer 
Have you tried it?  (you can use it through my branch)
Also feedback on the review from @nvie and my answers might help to push this forward.

blueyed:
For reference: there appear to be conflicts with `master` (this PR is still against `future`)?
I've not looked to closely at them, but would be happy to do so after feedback from @nvie.

nvie:
Hey @blueyed, I've tried understanding this PR multiple times now, but I think there's just too much going on. If you want, you can try submitting a smaller PR against the latest master, and I'm happy to review again. Try breaking things up a little more, so I can review smaller patches at a time. Thanks!

blueyed:
I'd argue that this is a much requested feature, has some additional tests, and does not break existing ones.  Also it is separated in several commits, which can be reviewed separately (instead of looking at the whole diff).

I see your point though that you should not merge something that's not clear, but I won't pick this up again for rewriting, at least in the next days.

Feel free to re-implement this feature in a different way, maybe using this PR's total diff and/or commits as hints.
For my projects it's good to use it as is, and as a result I might rebase this in the future etc, but won't rewrite/restructure it - also because I cannot see how it should be done (apart from the one place where you could use the pip library by now safely, and which was not the case in June).

cancan101:
I would love to see this feature merged.

piotr-dobrogost:
It looks like currently there's even no bug for this. I guess it might be difficult for people to find this thread as this is closed PR now. Should someone create a bug for this feature?

voidlily:
I'd like to see this as well.

blueyed:
> Should someone create a bug for this feature?

I'd say so, please go ahead.

jamesob:
> you can try submitting a smaller PR against the latest master

lol 144 line diff, 42 lines of which are tests. I'm curious what "small" looks like @nvie .

@blueyed you should consider forking in a more official capacity -- but as is, we'll consider using your fork!

PR #181, opened 2015-07-17
Implement pip-sync

hugopeixoto:
I implemented `pip-sync`. It would be nice to have this available, so if there's anything you'd like me to do in order to get this merged, let me know.

Here's an overview of what it does:

1. Parse `requirements.txt` using `pip.req.parse_requirements`
2. Fetch currently installed modules using `pip.get_installed_distributions`
3. Diff these two sets, by module name and version specifier
4. install and uninstall the resulting modules using `pip.main(["(un)install"] + ...)`

I had to create a whitelist to avoid uninstalling the following modules, which are automatically installed when creating a virtualenv and not listed on requirements.txt:
- pip
- wheel
- setuptools

I am assuming that `requirements.txt` contains all direct and indirect dependencies, as if it was generated by `pip-compile`, so if an indirect dependency is not present, it will be uninstalled and installed again.

### DIFF
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -13,7 +13,35 @@ if not tuple(int(digit) for digit in pip.__version__.split('.')[:2]) >= (6, 1):
           'perhaps run `pip install --upgrade pip`?'.format(pip.__version__))
     sys.exit(4)
 
+from .. import sync
+
+DEFAULT_REQUIREMENTS_FILE='requirements.txt'
 
 @click.command()
-def cli():
-    click.echo('TODO: Implement')
+@click.option('--dry-run', is_flag=True, help="Only show what would happen, don't change anything")
+@click.argument('src_file', required=False, type=click.Path(exists=True), default=DEFAULT_REQUIREMENTS_FILE)
+def cli(dry_run, src_file):
+    requirements = pip.req.parse_requirements(src_file, session=True)
+    installed = pip.get_installed_distributions()
+
+    to_be_installed, to_be_uninstalled = sync.diff(requirements, installed)
+
+    if not dry_run:
+        sync.sync(to_be_installed, to_be_uninstalled, verbose=True)
+    else:
+        show_dry_run(to_be_installed, to_be_uninstalled)
+
+
+def show_dry_run(to_be_installed, to_be_uninstalled):
+    if not to_be_uninstalled and not to_be_installed:
+        print("Everything up-to-date")
+
+    if to_be_uninstalled:
+        print("Would uninstall:")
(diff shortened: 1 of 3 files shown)


nvie:
This is great — thanks, @hugopeixoto! I got one error trying this out locally. I had an environment where I had installed `Jinja2==2.6` manually, then ran pip-sync with the following input file:

    #
    # Make changes in flask.in, then run this to update:
    #
    #    pip-compile flask.in
    #
    flask==0.10.1
    itsdangerous==0.24        # via flask
    jinja2==2.7.3             # via flask
    markupsafe==0.23          # via jinja2
    werkzeug==0.10.4          # via flask

Then I got the following error:

    > pip-sync --dry-run examples/flask.txt
      File "/Users/nvie/Projects/pip-tools/.direnv/python-2.7.9/bin/pip-sync", line 9, in <module>
        load_entry_point('pip-tools==1.0dev1', 'console_scripts', 'pip-sync')()
      File "/Users/nvie/Projects/pip-tools/.direnv/python-2.7.9/lib/python2.7/site-packages/click/core.py", line 664, in __call__
      File "/Users/nvie/Projects/pip-tools/.direnv/python-2.7.9/lib/python2.7/site-packages/click/core.py", line 644, in main
        rv = self.invoke(ctx)
      File "/Users/nvie/Projects/pip-tools/.direnv/python-2.7.9/lib/python2.7/site-packages/click/core.py", line 837, in invoke
      File "/Users/nvie/Projects/pip-tools/.direnv/python-2.7.9/lib/python2.7/site-packages/click/core.py", line 464, in invoke
      File "/Users/nvie/Projects/pip-tools/piptools/scripts/sync.py", line 27, in cli
        to_be_installed, to_be_uninstalled = sync.diff(requirements, installed)
      File "/Users/nvie/Projects/pip-tools/piptools/sync.py", line 25, in diff
        elif module.version in requirements[key].specifier:
    TypeError: argument of type 'SpecifierSet' is not iterable

hugopeixoto:
Thank you for giving this a try :)

I was testing with python 3.4 and pip 7.1.0. I see that you are using python 2.7.x.

Can you tell me which version of pip do you have installed?
That's probably the cause of the problem. I didn't test for 2.x.
If you tell me which version of pip you're using I'll try to figure out a way to do it in a more compatible way.

hugopeixoto:
It seems that SpecifierSet.__contains__ (which allows us to do `version in specifier`) was only added in pip 7.0.0. I changed it to `specifier.contains(version)`, which was added in pip 6.0. This should fix the issue.

nvie:
This is AWESOME. Thanks, will merge immediately :)

PR #186, opened 2015-07-22
Accept env variables for pip-inspired options

pquentin:
pip allows passing options as environment variables [0]. pip-tools has
three pip-inspired options: passing envvar to `@click.option` allows to
accept those environment variables too. click also supports
`auto_envvar_prefrix` [1], but this would apply to all options, not only
pip-inspired options.

[0] https://pip.pypa.io/en/stable/user_guide.html#environment-variables
[1] http://click.pocoo.org/4/options/#values-from-environment-variables


nvie:
Great, thanks!

PR #190, opened 2015-07-24
Add pip-alike "trusted-host" option to pip-compile.

tonyseek:
With this option, we could use self-host PyPI index without HTTPS.


tonyseek:
@nvie Could you review/merge it?

davidovich:
#208 is an attempt to fix this in a more general way (by supporting pip.conf files). Please feel free to comment.
Both approaches might be needed (and combined) if command line support is wanted.

nvie:
Thanks! I've merged this in with the changes as made by @davidovich in #208. Thank you both!

nvie:

tonyseek:
Thank you! :smile:

PR #201, opened 2015-09-09
Copy 'mising -e' error from compile to sync

pquentin:
Not sure if this is the best way to do it? But the `AttributeError` that I was getting before was not really informative. Thanks!

### DIFF
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -3,7 +3,7 @@ import collections
 from . import click
 import pip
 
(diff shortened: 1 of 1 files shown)


frgtn:
:+1: I think it's a good idea.

nvie:
Thanks!

nvie:

pquentin:
Thank you!

Issue #154, opened 2015-04-23
pip-compile does not handle extra requirements


lseek:
I have a requirements.in file with an "extra" requirement:
```
foopkg[FEATURE1]
```

When I `pip-compile` this requirement file what I get is:
```
foopkg==X.Y
```

What I would expect is:
```
foopkg==X.Y [FEATURE1]
```

In other words, `FEATURE1` and its requirements are missing and therefore doesn't get installed when I `pip install -r requirements.txt`.

Is there any way to work around this at the moment?

nvie:
Hi @lseek, thank for this. I'll have a look into what it takes to support this. I've personally never used them, but pip-compile should definitely support these use cases too. It may have a bit of an impact on the internals, but I'll have a look and see if I can add this.

nvie:
Do you have an example of a package that uses extras, so I can look into this?

piotr-dobrogost:
For instance requests library uses them – https://github.com/kennethreitz/requests/blob/master/setup.py#L71

nvie:
I'm going to work on this as far as time permits me. I've added a message that this is not yet supported in the meantime.

hale:
I ran into this when including `ipython` in `requirements.in`.

So, that's an example of a package that uses extras.  

(The [install docs](http://ipython.org/ipython-doc/dev/install/install.html) say to install like `ipython[all]`)

gavinwahl:
Is this expected to be difficult to implement? I might be able to do it with some guidance.

paulkernfeld:
I added a pull request, #202, that adds support for extras in pip-compile. Please let me know what I need to do to get this merged in!

paulkernfeld:
Pull request #202 is merged now, so this issue can be closed.

coriolinus:
I'm encountering this right now.

Sample from requirements.in:
```
pip-tools               # needed for pip-compile and pip-sync
hypothesis[django]      # testing framework
```

Sample from requirements.txt:

```
hypothesis==3.6.0
pip-tools==1.7.0
```

paulkernfeld:
@coriolinus can you open a new issue for this, please? In the new issue, can you describe what you are expecting to happen, compared with what is actually happening? Are you seeing an error, for example? If you tag me on the new ticket I can take a look.

coriolinus:
@paulkernfeld No problem; I created the issue last night but forgot to mention you then. Either way, it's up now: #408 .

PR #202, opened 2015-09-09
Add support for extras to pip-compile

paulkernfeld:
This addresses issue #154, "pip-compile does not handle extra requirements."


mattrobenolt:
fwiw, I just commented out the check, and it appeared to work...

I was just about to submit a patch for:

```diff
diff --git a/piptools/resolver.py b/piptools/resolver.py
index 8892175..811edb2 100644
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -98,10 +98,6 @@ class Resolver(object):
                 msg = ('pip-compile does not support URLs as packages, unless they are editable. '
                        'Perhaps add -e option?')
                 raise UnsupportedConstraint(msg, constraint)
-            elif constraint.extras:
-                msg = ('pip-compile does not yet support packages with extras. '
-                       'Support for this is in the works, though.')
-                raise UnsupportedConstraint(msg, constraint)
 
     def _group_constraints(self, constraints):
         """
```

then saw this. Not entirely sure why it worked for me without your patch.

paulkernfeld:
Hey @mattrobenolt, it works in some situations without my patch, but it will fail in other cases. I think that it will fail in any of these cases:
1. If my project requires package A, and package A requires package B with with extras, pip-compile will not put the extras for package B into `requirements.txt`.
2. pip-compile doesn't know how to merge extras correctly. For example, it knows that it should merge `ipython>=2` and `ipython<3` into `ipython>=2,<3`, but it doesn't know that `ipython[notebook]` and `ipython[nbconvert]` should be merged into `ipython[nbconvert,notebook]`.
3. The pip-compile dependencies cache is keyed on only name and version (not on extras), so the cache doesn't know that `ipython[notebook]` and `ipython` have different dependencies. I actually just realized that my pull request does not fix this problem, so I need to update it.

mattrobenolt:
Ah, I tested against a very simple case which worked fine. :)

tests/fixtures/fake-index.json
```diff
@@ -29,6 +29,24 @@
       "Six"
     ]
   },
+  "gnureadline": {
+    "6.3.3": []
+  },
+  "ipython": {
+    "2.1.0": {
```
nvie: I see you changed the data type under the version key to a dict here, which makes sense. To not further complicate the logic in the FakeRepository helper, let's just use this convention everywhere in this JSON file, so make all the other lists dicts (with only a `''` key), too.

nvie:
Thanks, @paulkernfeld! Your analysis is spot-on, and I really look forward to properly support these. If you can address your point 3 (i.e. change the format of the dependency cache), then I'm happy to merge this in!

paulkernfeld:
All right, @nvie, let me know what you think of this update! Sorry it's so long; I can split it into smaller commits if that's helpful.

habnabit:
Found a bug in this PR, which I haven't written tests for but at least resolved: habnabit/pip-tools@973e62391656e20a26c724f9eca67f52caf26a43

paulkernfeld:
@habnabit you're right, thanks for finding and fixing that! That is what I get for making changes without testing them :(

paulkernfeld:
I have added some tests for the cache, and incorporated habnabit's fix. Right now I think Travis is broken, but the build passes on my computer.

paulkernfeld:
Hey @nvie, is there anything I can do to make it easier to get this patch in?

tests/test_cache.py
```diff
@@ -0,0 +1,96 @@
...
+    # Create the file and write to it
+    cache_file = NamedTemporaryFile(mode="w", delete=False)
+    cache_file.write(to_write)
+    cache_file.close()
+
+    # Yield the path to the file
+    yield cache_file.name
+
+    # Delete the file on exit
+    remove(cache_file.name)
```
nvie: We should probably wrap this yield in a `try: ... finally:` block.

tests/test_cache.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    A good cache file.
+    """
+    with _read_cache_file_helper('{"__format__": 1, "dependencies": "success"}') as cache_file_name:
+        assert "success" == read_cache_file(cache_file_name)
+
+
+def test_reverse_dependencies(from_line):
+    # Since this is a test, make a temporary directory
+    tmp_dir_path = mkdtemp()
```
nvie: py.test offers a super nice way of creating temp dirs: you can simply have it inject the fixture `tmpdir` to this function, and you can use it by saying `tmp_dir_path = str(tmpdir)`. It'll be cleaned up automatically by pytest.

nvie: See https://pytest.org/latest/tmpdir.html

tests/test_resolver.py
```diff
@@ -15,13 +15,57 @@
...
+             # Note that the extras should be sorted
+             'ipython[nbconvert,notebook]==2.1.0',
+             'pyzmq==2.1.12',
+             'jinja2==2.7.3',
+             'tornado==3.2.2',
+             'markupsafe==0.23',
+             'gnureadline==6.3.3',
+             'pygments==1.5',
+             'sphinx==0.3']
+         ),
```
nvie: This looks great!

nvie:
This is great work—thanks a lot, @paulkernfeld! Sorry I didn't find the time to review this a bit earlier. I had one case locally that behaved a bit strange. If I create a test file with the following contents, it still compiles all the extras, where I expected just the main `ipython` package:

    ipython#[notebook]

Note: there's no space between the 'n' and the '#'. I'd assume this would be considered a comment, but it wasn't. If you add a space, it acts how I would expect (the `#[notebook]` is considered a comment).

Otherwise, this looks fine, and I'm happy to merge it in.

paulkernfeld:
Oh, very interesting point about the hash symbol. This is something that I know nothing about. Basically, I'm calling `pip.req.InstallRequirement.from_line`, and it ignores the hash symbol before the extras. We can definitely work around this, e.g. by only sending it the text before the hash symbol, but are you sure that's what we want to do?

I sort of think that this is the intended behavior, because when I run `pip install -r requirements.txt`, and `requirements.txt` has the one line, `ipython#[notebook]`, pip installs the notebook extra! Is this perhaps a bug in pip's parsing code? Or... a feature?

paulkernfeld:
Oh boy. Actually, it looks like pip 6 considers `ipython#[notebook]` to mean `ipython`, whereas pip 7 considers it to mean `ipython[notebook]`.

tests/test_cache.py
paulkernfeld: Thanks for the tip! I implemented this.

tests/test_cache.py
paulkernfeld: You're right. Good point.

nvie:
OK, I don't care in that case. If this is a bug in pip, it'll probably be fixed. If it's not, then we're just pip-compliant. No big deal.

Thanks for the work, I think it's awesome :)

paulkernfeld:
Yay! Thanks @nvie!

PR #208, opened 2015-10-05
Honor pip's pip.conf (or pip.ini on Windows).

### LINKED ISSUE #199: pip-compile honor pip.conf

davidovich:
This ground work allows specifying sensible defaults in pip.conf for things like trusted-host, isolated index-url, etc.

The strategy is to use the pip OptionParser to honor the defaults (which supports pip.conf).

We then translate click options to pip options which are used to instantiate the PackageFinder.

Fixes #199


nvie:
Thanks!

nvie:

davidovich:
Thank you!

PR #211, opened 2015-10-07
remove __future__ unicode import to allow non unicode os.environ keys

### LINKED ISSUE #209: python 2.7 TypeError: environment can only contain strings
I have this error on python 2.7 on Windows when at the resolve cycle.

The resolver.py module sets its unicode literals from the \_\_future\_\_ import, and also wants to inject PIP_EXISTS_ACTION in the os.environ which does not accept unicode data as keys, nor values.

This makes subprocess calls issued in pip fail, because os.environ cannot contain unicode data.

Other projects have had this problem:

https://github.com/mitsuhiko/werkzeug/issues/467
and
https://github.com/celery/django-celery/issues/344

[Other technique seen also on stackoverflow.](http://stackoverflow.com/questions/31003437/python-2-3-os-environ-and-unicode)

One suggestion is to remove the unicode_literals \_\_future\_\_ import and this seems the easiest fix as it supports both python 2.7 and 3.

davidovich:
Fixes #209

Using unicode_literals as \_\_future\_\_ breaks os.environ on python 2.7, when used in subprocess invocations (that forward the env) which does not allow unicode data.

As this module uses os.environ['PIP_EXISTS_ACTION'] = 'i' and pip copies this in a subprocess call, it breaks with this stack trace (actual path root removed from output):

```
  File "pip-tools-venv\lib\site-packages\piptools\scripts\compile.py", line 108, in cli
    results = resolver.resolve()
  File "pip-tools-venv\lib\site-packages\piptools\resolver.py", line 78, in resolve
  File "pip-tools-venv\lib\site-packages\piptools\resolver.py", line 162, in _resolve_one_round
    for best_match in best_matches
  File "pip-tools-venv\lib\site-packages\piptools\resolver.py", line 163, in <genexpr>
    for dep in self._iter_dependencies(best_match))
  File "pip-tools-venv\lib\site-packages\piptools\resolver.py", line 234, in _iter_dependencies
  File "pip-tools-venv\lib\site-packages\piptools\repositories\pypi.py", line 127, in get_dependencies
    dependencies = reqset._prepare_file(self.finder, ireq)
  File "pip-tools-venv\lib\site-packages\pip\req\req_set.py", line 505, in _prepare_file
    abstract_dist.prep_for_dist()
  File "pip-tools-venv\lib\site-packages\pip\req\req_set.py", line 123, in prep_for_dist
    self.req_to_install.run_egg_info()
  File "pip-tools-venv\lib\site-packages\pip\req\req_install.py", line 410, in run_egg_info
    command_desc='python setup.py egg_info')
  File "pip-tools-venv\lib\site-packages\pip\utils\__init__.py", line 702, in call_subprocess
    cwd=cwd, env=env)
  File "F:\Python2.7.8\Lib\subprocess.py", line 710, in __init__
    errread, errwrite)
  File "F:\Python2.7.8\Lib\subprocess.py", line 958, in _execute_child
    startupinfo)
TypeError: environment can only contain strings
```

This PR fixes this error.

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -1,6 +1,5 @@
 # coding: utf-8
-from __future__ import (absolute_import, division, print_function,
-                        unicode_literals)
+from __future__ import (absolute_import, division, print_function)
 
 import os
 from functools import partial


nvie:
I'd rather not remove this directive, since it may have effects elsewhere (i.e. where we print out log lines for packages with non-ascii names). Instead, wouldn't this simply fix the problem, too? https://github.com/nvie/pip-tools/compare/master...5e7fcb3581e372a731dacf3ef2273aa646b18b23 Let me know if this works for you.

davidovich:
Unfortunately, this breaks python3 (others have had this issue, as mentioned in the bug ticket reference #209):

```
  ...
  File "f:\git\pip-tools\piptools\scripts\compile.py", line 124, in cli
    results = resolver.resolve()
  File "f:\git\pip-tools\piptools\resolver.py", line 68, in resolve
    os.environ['PIP_EXISTS_ACTION'] = b'i'  # ignore existing packages
  File "F:\Python35-32\lib\os.py", line 686, in __setitem__
    value = self.encodevalue(value)
  File "F:\Python35-32\lib\os.py", line 740, in check_str
    raise TypeError("str expected, not %s" % type(value).__name__)
TypeError: str expected, not bytes
```

As an alternative, I may be able to set the PIP_EXISTS_ACTION option through the new pip options support introduced in #208, which defers the setting to pip, which does not use unicode_literals. But to me this seems overkill.

Chronial:
I think the correct fix for this issue is to do `os.environ[str('PIP_EXISTS_ACTION')] = str('i')`.

nvie:
Thanks @Chronial, this is fixed in 3a439045c71851151c51953b79f5994d3a12d9fe now.

PR #212, opened 2015-10-07
Add support for client ssl certificates

hlawrenz:
This adds a --client-cert option as seen in the pip command line so that you can use an alternate index which requires a client certificate for authentication.

I also got rid of two bits of empty line whitespace.


nvie:
Thanks!

nvie:
This is now released as part of pip-tools==1.1.6. Thanks for the patch!

hlawrenz:
Excellent. Thanks for pip-tools, it's been very helpful!

PR #213, opened 2015-10-07
Allow wheel as requirements source

davidovich:
It can be practical to use a wheel to calculate dependencies. This PR implements this in a simple manner (leveraging wheel support from pip).

Allows

```
pip-compile my-wheel.whl
```

In our use case, we use this to pin requirements **after** the top level package (wheel) has been created. This helps in our automation of deployments, without the dependency of an external requirements.txt file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ htmlcov
 build
 dist
 *.egg-info
+.cache/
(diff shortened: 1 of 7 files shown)


swethaHub:
+1 This would be quite useful.

PR #265, opened 2015-12-04
Add support for multiple requirements files

### LINKED ISSUE #104: Support for stdin / stdout
It would be useful if pip-compile would accept input from stdin and allowed to write to stdout.

`pip-compile` might detect if stdin/stdout is used, but an explicit option should work, too:

    git show :requirements.in | pip-compile > requirements.txt
    git show :requirements.in | pip-compile - -o > requirements.txt

This would allow to easily compile only what's in Git's index, for example.

khwilson:
This commit does four things:
* Allows you to specify multiple requirements input files,
* Specify an output file (required if you have multiple input files),
* Maintains the standard behavior for 0 or 1 command line arguments, and
* Allows for stdin as an input (closes #104)

There is a caveat on that last one: In click 6.0, an `allow_dash` kwargs was added to `click.Path`. However, this feature is broken currently. I submited a [pull request](https://github.com/mitsuhiko/click/pull/479) to click, to fix this, at which point in time, this commit simply works. In the interim, the stdin is broken.

Alternatively, I could create a kludge click type that does the stdin checking in the meantime, but I leave that up to you.

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -4,7 +4,10 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
 import optparse
+import os
 import sys
+import tempfile
+
 import pip
 
 # Make sure we're using a reasonably modern version of pip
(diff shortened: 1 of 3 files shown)


nvie:
Thanks, I've modified the messaging slightly and released this as 1.3.0!

keimlink:
Awesome! :+1: Thanks a lot @khwilson and @nvie!

PR #271, opened 2015-12-12
Add '-n' option as shorthand of '--dry-run' at pip-compile, pip-sync

raimon49:
I hope use '-n' option as the short hand of '--dry-run' option.

Because these are used with a tool of major Linux.

e.g.)

* [make](http://linux.die.net/man/1/make)
* [rsync](http://linux.die.net/man/1/rsync)
* [git-add](https://www.kernel.org/pub/software/scm/git/docs/git-add.html)


nvie:
Thanks!

PR #273, opened 2015-12-15
Add a new --force-range switch

amitsaha:
This commit adds support for a new switch named --force--range which when specified to ``pip-compile`` will preserve the range specifications for packages instead of finding the latest package
which matches and pinning that.

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -11,7 +11,8 @@ from pip.req.req_set import RequirementSet
 
 from ..cache import CACHE_DIR
 from ..exceptions import NoCandidateFound
-from ..utils import is_pinned_requirement, lookup_table, make_install_requirement
+from ..utils import (is_pinned_requirement, lookup_table,
+                     make_install_requirement, is_range_pinned_requirement)
 from .base import BaseRepository
 
 try:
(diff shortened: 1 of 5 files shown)


nvie:
I dont understand this feature yet. What is the use case? Is it that if you put django>=1.8,<1.9 in *.in, then compile, that this same spec (django>=1.8,<1.9) would remain ("be copied") to the output *.txt file (rather than getting 1.8.7 pinned or so)?

amitsaha:
@nvie Thanks for taking a look. Yes, you are absolutely right. I am keen to hear what you think? I know I need to fix some existing tests.

nvie:
I don't think pip-tools is the project for this (it's quite the opposite of the philosophy). If your use case requires it, why not simply have a requirements.txt and no requirements.in at all?

amitsaha:
Well, if I just had one requirement which needed to have a range of dependencies preserved, but others would be just fine to use the latest available, I would still have the use for requirements.in.

Anyway, thanks for taking a look.

nvie:
If you did not want to upgrade to Django 1.9 yet, you can put `django<1.9` in your *.in file, and it would never compile to a version that's 1.9 or higher. Still, it would put you on the latest version in the available range. If you don't want that, you can even pin the version in your *.in file, i.e. `django==1.8.7`, and it would never go to Django 1.8.8 or higher.

I'm not sure what use case I'm missing.

amitsaha:
The reason we have a range in our requirements and why pinning it to the latest won't work for us is:

On every commit to master (which is for every commit), a package is built and is pushed to our own PyPi. Hence, we want the latest package between ``min_version``  and ``max_version``, but do not want to pin it to a certain version ``min_version.XXYYZZ``, because then we will have to compile the requirements.in again else we will be picking up the non-latest package.

nvie:
I have briefly experimented with the idea of allowing a "directive" syntax in the *.in files, e.g.

```
django<1.8,>=1.7    # pip-tools:verbatim
```

Or something similar. I think this is too much of an edge case to support generically in a tool like this. I think you can solve this problem yourself easily by scripting around pip-tools here. I.e. where you invoke pip-compile now, instead do the following:

```
$ (pip-compile --output -; echo "django>1.8,<=1.9") > requirements.txt
```

amitsaha:
Sure, no worries. I felt that since we let pinned dependencies as they are, a switch for range pinned would make sense.

PR #274, opened 2015-12-15
support writing multiple output files

scottbelden:
Previously when you read in multiple files, you had to specify an output file. The dependencies were all grouped together into the single output file. However, in some cases it would be nice to have each input file generate its own output file (e.g. if you have a normal requirements.txt and a requirements.txt for running tests).


nvie:
This is certainly a valid use case, but I'd simply run pip-compile twice in a row. What's the added value of adding an extra option for this?

scottbelden:
It turns out that I put this in before I realized that there was in fact a cache. The main reason was for the caching, but I now see that that's probably not necessary.

nvie:
Yeah, the cache is on-disk, so you should get the perf benefit, even though it takes two runs. This just keeps the already large amount of CLI options a bit more managable.

PR #276, opened 2015-12-22
Add option to omit footer from the output of pip-compile

jonafato:
When compiling requirements, the commented unsafe packages can create
noise. setuptools, for example, is updated frequently. The proposed
``no-footer`` option allows suppression of these comments (e.g. to
reduce noise in source control diffs or simply because a user may not
care about those packages).

The option is called ``--no-footer`` to match the existing
``--no-header`` option.

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -46,6 +46,7 @@ class PipCommand(pip.basecommand.Command):
               help="Mark this host as trusted, even though it does not have "
                    "valid or any HTTPS.")
 @click.option('--header/--no-header', is_flag=True, default=True, help="Add header to generated file")
+@click.option('--footer/--no-footer', is_flag=True, default=True, help="Add footer to generated file")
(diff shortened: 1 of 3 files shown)


nvie:
The idea behaviour here:

- Don't name option `--footer`. It's not necessarily a footer. I just happened to have grouped them there for convenience.
- The real noise is caused by version changes in these comment lines. I'm fine removing the version spec for these lines from the output. Basically, it'd just say that "setuptools" is ignored.

This way, we'd solve the problem of those lines not remaining stable over time and causing noise (which I agree is a good idea), but we don't introduce an extra flag.

What do you think, @jonafato? Could you change this PR to reflect that?

jonafato:
@nvie Thanks for the feedback. You're right that this pull request added a lot more than it needed to address the core issue I was having. Because the changes are so different, I'm closing this PR in favor of a new one created at #294.

Issue #277, opened 2015-12-22
Mark appnope as an unsafe package


achauve:
or provide a way to configure unsafe packages via cli option on config file.

Ipython requires appnope on OSX only, so generated requirements.txt. fails on linux afterwards.

see PR #278

khwilson:
ipython's reliance on appnope is annoying for this toolchain. If it's any help, we get around this with a bit of sed in our tox.ini, which solves the problem "well enough" for our purposes:

```
[testenv:pip-compile]
deps=
    pip-tools==1.2.0
whitelist_externals=
    sed
    rm

; iPython 4 has a darwin only platform dependency on appnope. pip-compile is not platform aware and
; outputs this unconditionally. We use sed to patch in the dependency until pip-compile is
; platform aware.
commands=
    pip-compile requirements.in -v
    sed -i.bak -e's/^appnope[^[:space:]]*/&;sys_platform=="darwin"/' requirements.txt
    rm requirements.txt.bak
```

nvie:
Agreed, this is annoying. I don't think we can guarantee desired behaviour here under all circumstances. In a way, pip-tools has to work with the output of pip, which on a single machine, is quite platform-specific. However, I would support a command line flag like `--ignore-package appnope` or something like that if that would practically solve this issue. If you want to make a PR for this, I'm happy to pull it.

khwilson:
tbh, I'm really not sure you *should* be trying to get around it because who knows how deep the rabbit hole goes. But maybe an `--ignore-platform-specific` flag (or perhaps something with a better name)? Maybe pip-compile would go ahead and resolve the dependencies for the local platform and then comment them out in the resulting requirements file, or would just leave them tagged with `;sys_platform=sys.platform`? I think either of those options would be better than an outright `--ignore`.

Thoughts?

nvie:
How would you detect those without residing to parsing the package's metadata yourself? Right now, I'm basically letting pip do all the heavy lifting by simply asking it what the dependencies are. Not sure if we can detect from there is certain deps are platform specific.

khwilson:
The `appnope` dependency arises from an `extras_require[':sys_platform == "darwin"] = ['appnope']` type situation. So there are choices you could make, e.g., monkey patching `sys.platform` or, perhaps better, changing the PEP-0425 tags passed to pip. (I haven't looked at the pip source in a while, but I'm pretty sure there's an option for this.)

[...]

PR #278, opened 2015-12-22
Mark appnope as unsafe package

achauve:


nvie:
At the very least, I'll need to have some motivation for this.

achauve:
Yes sorry I created an issue with some context: see #277 

Appnope is required by ipython only on OSX, so when you develop on OSX and then deploy on linux using pip-tools generated requirements.txt breaks deployment.

achauve:
@nvie What do you think?

nvie:
We have now concluded that appnope itself is not an issue (https://github.com/nvie/pip-tools/issues/285#issuecomment-169847863), so I'm closing this one.

Issue #166, opened 2015-06-16
Add `--no-upgrade` flag to `pip-compile`


nvie:
This should prevent executing version updates, but shoud add/remove dependencies as they have (dis)appeared in *.in files. This would be useful to add a new dependency to in *.in file, and have `pip-compile` only resolve that new package and add it to the *.txt file.

There are a lot of edge cases though, so getting the semantics right here is non-trivial.

jgonggrijp:
I'm glad this was tackled, it greatly improves interoperability with pip-review because it separates updating from pinning.

PR #281, opened 2015-12-28
Feature - "no upgrade" flag for only upgrading as necessary

ejames:
Resolve issue #166 by adding a "--no-upgrade" flag to pip-compile.

If there's already an existing file that would be overwritten by running pip-compile, read in that file, and remember the existing pinned requirement.  Whenever looking for the best match for a package, take the existing pinned requirement if it satisfies the constraints we're applying.

A test is added for the resolver.  The new dependency `frobnicator==1.0.0` is output by the resolver, and the preexisting `widgetizer==1.0.0` pin is changed to `widgetizer==2.0.0` because `frobnicator` requires an upgraded version of `widgetizer`, but the preexisting `flux-capacitor` pin is not moved even though a later version of `flux-capacitor` is available in the fixture.


ejames:
Grovelling into the structure of the `InstallRequirement` object to ask if `existing_dependency.req.specs[0][1] in ireq.req` feels naughty, but isn't fundamentally any different than the operations performed in `utils.as_tuple` to get the `version`.  It seems like a style question - opinions?

The description of #166 suggests that there are a lot of edge cases, but I frankly couldn't think of any; if the pinned version in the existing file fits the constraint we're examining, use that version instead of going to the repo.  Is my solution brilliantly simple, or stupidly naive?  I haven't spent enough time in the guts of Python dependency management to have examples of pathological requirements files, so all I can say is that this PR works for the requirements.txt we're actually using at my company.

nvie:
One of the edge cases is if you have existing dependencies, i.e. suppose you start out with an *.in file like:

    werkzeug

And at any point in time that happens to be compiled (so in *.txt):

    werkzeug==0.9  # an old version

Then, you add `flask` to your *.in file:

    flask
    werkzeug

If we then compile this using `--no-upgrade`, this would only update Flask, right? But now it happens to be the case that the latest version of Flask (which this compiles to) wants the latest version of werkzeug, i.e. it has a dependency of `werkzeug>=0.11`. Ignoring werkzeug's required upgrade here will lead to a broken *.txt file.

So either we should fix this (by also updating werkzeug (and its outdated dependencies, recursively) too in this case), or simply letting this command error in such cases, and advise the user to run it without the `--no-upgrade` flag after all. Obviously the former approach is preferred :)

nvie:
Perhaps a clearer example / test case.

Suppose you are already using jinja2 and werkzeug, but not Flask yet.

    $ cat requirements.in
    jinja2
    werkzeug

And suppose your werkzeug is on a (very) old version, but you use a recent (but not _most_ recent Jinja2 version):

    $ cat requirements.txt
    #
    # Make changes in foo.in, then run this to update:
    #
    #    pip-compile foo.in
    #
    jinja2==2.7.3          # NOTE: reasonably recent, 2.8 is newer
    markupsafe==0.23       # via jinja2
    werkzeug==0.6          # NOTE: pretty old version of werkzeug

Next, you decide to also use Flask for your project. So you add "flask" to the *.in file, and then run `pip-compile --no-upgrade`. Since the latest version of Flask requires `werkzeug>=0.7` (we don't have it, but it still fits our specs) and `jinja2>=2.4` (we already satisfy that). So what _should_ happen is:

    $ cat requirements.txt
    #
    # Make changes in foo.in, then run this to update:
    #
    #    pip-compile foo.in
    #
    flask==0.10.1
    jinja2==2.7.3          # UNTOUCHED
    markupsafe==0.23       # UNTOUCHED
    werkzeug==0.11.3          # bumped to latest version matching our specs
    ...

The point being that werkzeug should still be updated, even though we specified `--no-upgrade`. But Jinja2 isn't. If we had `werkzeug<0.10` in our original *.in file, this upgrade should fail since it could not satisfy the new requirement with the existing requirements. Simply ignoring werkzeug in this process because it's already in our output file misses these cases.

ejames:
That case is already handled.  I've updated the test case in my PR to match your example:

```python
        # Add Flask to an existing requirements.in, using --no-upgrade
        (['flask', 'jinja2', 'werkzeug'],
         [
            # Add flask and upgrade werkzeug from incompatible 0.6
            'flask==0.10.1',
            'itsdangerous==0.24',
            'werkzeug==0.10.4',
            # Other requirements are unchanged from the original requirements.txt
            'jinja2==2.7.3',
            'markupsafe==0.23'
            ],
         False,  # no prereleases
         True,  # --no-upgrade flag set
         [
            # The requirements.txt from a previous round
            'jinja2==2.7.3',
            'markupsafe==0.23',
            'werkzeug==0.6']
         )
```

The test passes.

Essentially the algorithm is to act as if requirements.txt is a package repository, and always check the "listing" there first before going to the real package repository on PyPi.  The existing pin is used if and only if it satisfies the requirement we were about to resolve.

In the case of the requirement `werkzeug>=0.7`, the existing pin `werkzeug==0.6` doesn't satisfy the specifier, so we go to PyPi and get the latest version.

In the case of the requirement `jinja2>=2.4`, the existing pin `jinja2==2.7.3` specifies a version which satisfies `>=2.4`, so we keep `jinja2==2.7.3` instead of going to PyPi to pick up a later version of `jinja2`.

If a package is subtracted from the requirements.in - for example, if you delete the requirement for flask after running the above example - then when compile is run again with `--no-upgrade`, we never try to resolve a requirement for `flask`, and therefore the existing pin for `flask==0.10.1` is never examined, and flask drops out of the resulting requirements.txt.  The `werkzeug==0.10.4` pin would remain because the requirements.txt still lists `werkzeug`.

jonafato:
@ejames This build seems to have failed because of the error fixed in #295. If you rebase, things should be back to normal.

nvie:
Interesting, @ejames! I do agree this is a sound approach. Strangely, I did not pick up from the diff that this is how it really worked. This may have been a bit more clear if the code reflected how you just explained it.

Could we therefore change the implementation so that we instead instantiate a Repository subclass that does just that proxying? I.e. first check the requirements.txt, then the _real_ Repository? If we then just pass that Repository instance along, no other parts of the resolver are affected, right? There would have to be no special casing anywhere else, nor passing around of extra kwargs to several pieces. I think it'd make the code simpler (and thus simpler to read/understand).

I think the test case is sound. If we can keep that green, we're good.

ejames:
Thanks for help on the PR, @jonafato, @nvie.

Rewritten as #301.

PR #286, opened 2016-01-06
Also add the stdlib to the "already installed" list

palfrey:
Without this change, on Python 3.5 `pip-sync -n` always (if argparse is in the requirements.txt) returns `Would install: argparse==1.4.0` despite argparse being already installed as part of the standard library in modern Python versions. With this change, it also considers the "stdlib pkgs" as installed, and so doesn't try to reinstall them for Python versions that have merged packages.


nvie:
Thanks!

PR #291, opened 2016-01-14
Add --no-index option to hide index URL

daradib:
Adds --index/--no-index flag to include or hide index URL
from generated file. Also adds --no-* options to generated header.


nvie:
I like the idea of echoing the command flags to the output file, but I don't feel like adding an `--index/--no-index` flag just for this. IMHO, this does not have to be configurable. Ideally, if any non-default index is used, I'm fine just always echoing that in the output's header. If it's just using PyPI (the default), then don't emit it. What do you think @daradib? Could you change this PR to reflect that?

daradib:
@nvie I understand the hesitation to add another configuration option. However, some users have a different PyPI server configured as `index-url` in `~/.pip/pip.conf`, for example, a local mirror or cache of the PyPI server or a private, password-protected PyPI server, in which case I think the credentials get stored in `requirements.txt`. So there are some use cases in my opinion. What do you think?

nvie:
I'm sorry, I think I mis-read your change (I thought this only reflected the emitted comment line in the header, I missed the actual emission of the `--index-url` line itself). Yeah, your argument makes sense now. I'll pull this in—thanks, @daradib!

daradib:
Thanks that was quick!

PR #294, opened 2016-01-15
Remove specifiers for unsafe requirements

jonafato:
As discussed in #276, version specifiers for unsafe requirements create
noise in diffs. This change omits versions from those requirements, so
the noise is silenced.


nvie:
Thanks! Much better this way!

nvie:
Can't merge this now (I'm on mobile). Will merge when I'm on a desktop machine.

jonafato:
@nvie: Rebased against master.

nvie:
Thanks!

PR #295, opened 2016-01-15
Fix missed header->emit_header name change

jonafato:
The changes made in 5b2861d334b6cb866444d8e6cb37e7a5d1ce4e61 rename
``OutputWriter.header`` to ``OutputWriter.emit_header``. An instance was
missed, causing the ``write_header`` method to fail and break tests.
This change fixes the attribute name.


nvie:
Oops, thanks!

PR #300, opened 2016-01-18
Fix index-url flags in pip-sync

pwmarcz:
The flags are being passed to `pip uninstall` where they
aren't recognized.

Also, consider not passing the flags at all when the options come from environment variables (`PIP_INDEX_URL` etc.) - `pip` will see the environment anyway, and it seems that using command-line index-url flags disables the cache.


nvie:
Thanks!

PR #301, opened 2016-01-19
Feature - "no upgrade" flag implemented as proxy repository

ejames:
Updated from #281.

@nvie - Test case passes, although I've moved it to a different file, since it's no longer a test for the resolver.

While working on this, I realized that `--minimal-upgrade` would be a better name than `--no-upgrade`, since there are still upgrades possible - simply no more than are necessary.  Using the word "minimal" clarifies the intent.

I would personally prefer for the setting to default to `True` but didn't want to change the default behavior in a pull request.


ejames:
It occurs to me that this is also a good way to get started with pip-compile by capturing a good environment state.

By running these commands on a known good environment:

```
$ mv requirements.txt requirements.in
$ pip freeze > requirements.txt
$ pip-compile requirements.in --minimal-upgrade
```

You now have a requirements.in that captures all of the working knowledge from your old requirements.txt, and a requirements.txt that captures a proven list of pinned versions that form a good environment.

palfrey:
Tried using this and got
```
  File "/Users/palfrey/src/potboiler/python/ENV/bin/pip-compile", line 9, in <module>
    load_entry_point('pip-tools==1.4.5.dev0', 'console_scripts', 'pip-compile')()
  File "/Users/palfrey/src/potboiler/python/ENV/lib/python3.5/site-packages/click/core.py", line 716, in __call__
  File "/Users/palfrey/src/potboiler/python/ENV/lib/python3.5/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/Users/palfrey/src/potboiler/python/ENV/lib/python3.5/site-packages/click/core.py", line 889, in invoke
  File "/Users/palfrey/src/potboiler/python/ENV/lib/python3.5/site-packages/click/core.py", line 534, in invoke
  File "/Users/palfrey/src/potboiler/python/ENV/lib/python3.5/site-packages/piptools/scripts/compile.py", line 132, in cli
    for index_url in repository.finder.index_urls:
AttributeError: 'MinimalUpgradeRepository' object has no attribute 'finder'
```

which looks to be a thing the PyPIRepository has (https://github.com/ejames/pip-tools/blob/feature-no_upgrade_proxy_repository/piptools/repositories/pypi.py#L41)

ejames:
I might have been in the process of coming down with the flu while rewriting the PR this morning.  Let me take some aspirin and read the diff again.

ejames:
@palfrey - That fixes the issue.

palfrey:
Previous issue no longer occurs, and mostly works, although I did have this one odd diff

```
-# setuptools==19.2          # via logilab-common, voluptuous
+# setuptools                # via logilab-common, voluptuous
```

nvie:
@ejames I'm so happy with this PR! The way the implementation ended up is very elegant, and it works like a charm. I definitely want to make this a major release to call out some attention to this feature. However, I don't really like the command line flag's name (and the class name) yet.

That said, I'm thinking about making this behaviour the default, and having it emit a console message to stderr to indicate people can do a "full upgrade" using a `--upgrade` option. How about that?

pwmarcz:
@nvie As a user I agree, holding back upgrades as a default would be friendlier: if I'm adding a new package, I don't want to be bothered about upgrades until I ask for them explicitly.

Very much looking forward to using this!

ejames:
`MinimalUpgradeRepository` is an awkward name, since it describes the "why" of the class instead of the "what", and the "minimal" part is a function of the way other code uses the class, and not of the class itself.  But I'm not sure what a better name would be; I settled on that name because it at least makes it clear that the class exists solely for the purpose of supporting the similarly-named flag.  Feel free to rename it.

I would be in favor of making the behavior default with an `--upgrade` flag to get new versions - especially since `--upgrade` is a flag that works on Pip itself.  In practice I almost always want to avoid upgrading pinned requirements.

nvie:
I'll merge this in now as-is and I'll add another patch to make this the default option before releasing this as 1.6. Thanks, this is an awesome improvement to pip-tools I'm sure everyone will love :)

nvie:
Done in 72ac64c...3a2de92.

PR #326, opened 2016-03-01
Feature/make snapshot

fmarco:
Add a flag to pip-compile to handle absence of requirements file (it generates .in and .txt)

Add a snapshot command to capture the situation at a certain point inside an env and maybe let external tool to handle a system to tracking requirements changes.

Also, adding --no-use-wheel as flag.

ps: I'm thinking to add some test, some hints? :)

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -15,7 +15,7 @@ from ..exceptions import PipToolsError
 from ..logging import log
 from ..repositories import LocalRequirementsRepository, PyPIRepository
 from ..resolver import Resolver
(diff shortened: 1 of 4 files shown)


nvie:
Rather than coding up an entire command for this, couldn't you just run `pip freeze > requirements.in`? This tool is about as complex as I can/want to maintain for now. I'd like to keep pip-tools as simple and focused as possible, as pip contains a lot of inherent complexity.

fmarco:
@nvie I see and i agree, but i tought to avoid keep excluded requirements even in that phase using pip-tools functions, anyway thanks for review.

PR #327, opened 2016-03-01
Added command line flag "--unsafe" to allow the pinning of

cyrusd:
"unsafe" dependencies; unfortunately there are some valid
instances of needing to enforce this, it seems like setuptools
in particular goes through various phases of backwards
compatibility.

I wasn't 100% certain how to run the tests ("python setup.py test"
didn't seem to do the trick) but I did add a best guess at how
to test the new option.

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -9,7 +9,7 @@ from .utils import comment, format_requirement
 
 
 class OutputWriter(object):
-    def __init__(self, src_file, dst_file, dry_run, emit_header, emit_index, annotate,
+    def __init__(self, src_file, dst_file, unsafe, dry_run, emit_header, emit_index, annotate,
                  default_index_url, index_urls):
         self.src_file = src_file
         self.dst_file = dst_file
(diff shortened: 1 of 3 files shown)


piptools/writer.py
```diff
@@ -69,13 +70,22 @@ def _iter_lines(self, results, reverse_dependencies, primary_packages):
...
-                line = self._format_requirement(ireq, reverse_dependencies, primary_packages, include_specifier=False)
-                yield comment('# ' + line)
+            if self.unsafe:
+                yield ''
+                yield comment('# The following packages are included because pip-tools was invoked with --unsafe')
+                yield comment('# However, they may be considered unsafe in a requirements file:')
+
+                for ireq in unsafe_packages:
+                    line = self._format_requirement(ireq, reverse_dependencies, primary_packages)
+                    yield line
```
nvie: I like the idea for this flag. Rather than putting in an explicit comment here and repeat some of the code, I'd be fine simply setting the `unsafe_packages` set to the empty set (on line 62) if this flag is active. This will not make setuptools stand out from the rest (but that's the intention if this flag is used anyway).

nvie:
To run the tests, run:

    $ tox

cyrusd:
Ok, updated the code to reflect your suggestions, I wasn't able to get the test for the 'unsafe' option to pass (indexing issues I think) but kind of ran out of time to look into it...

alekzvik:
Implemented in #377

PR #330, opened 2016-03-04
Add command line alias '-U' as '--upgrade'

raimon49:
The `-U` option is famous as alias of `--upgrade` for pip user. e.g.) `pip install -U pip`

When it is supported in pip-tools 1.6 or later, it is more convenient.


nvie:
Thanks!

PR #337, opened 2016-03-17
Adding format_control support

markphilpot:
Update OutputWriter to support passing through `--no-binary` and `--only-binary` parameters


nvie:
TIL about format control. Ideally we could just pass through these flags in a way that would not require pip-tools to specifically add support for it, but I don't know a better way for now.

nvie:
Thanks!

PR #339, opened 2016-04-05
OutputWriter gains support for listing all input filenames in the output file header.

mbaltaks:
Without this change, the header is actually incorrect when using multiple input files.


nvie:
I've generalized the logic a little bit, but thanks for initiating this!

PR #341, opened 2016-04-11
Use pip's mechanics for creating a session

davidovich:
Instead of making an empty PipSession object, use pip's implementation.

This allows for using pip config for client-cert, trusted-hosts etc.

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -89,13 +89,13 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
     # Use pip's parser for pip.conf management and defaults.
     # General options (find_links, index_url, extra_index_url, trusted_host,
     # and pre) are defered to pip.
-    pip_options = PipCommand()
+    pip_command = PipCommand()
     index_opts = pip.cmdoptions.make_option_group(
         pip.cmdoptions.index_group,
-        pip_options.parser,
+        pip_command.parser,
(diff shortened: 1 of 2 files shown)


davidovich:
@nvie This PR addresses problems with pip-tools not respecting values like trusted-hosts in the pip.{conf,ini}.

piptools/scripts/compile.py
```diff
@@ -114,9 +114,10 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
         for host in trusted_host:
             pip_args.extend(['--trusted-host', host])
 
-    pip_options, _ = pip_options.parse_args(pip_args)
+    pip_options, _ = pip_command.parse_args(pip_args)
 
-    repository = PyPIRepository(pip_options)
+    session = pip_command._build_session(pip_options)
```
nvie: I do think using private pip APIs can be a bit problematic, but I can see the usecase here though. Still, relying on these private APIs means at any time in the future pip-tools can break when pip decides to change its internals. We'll see what happens, I guess :)

Thanks for the PR!

davidovich:
@nvie I really agree with your private implementation usage comment, but short of maintaining a duplicated implementation, I thought this was the least problematic of the two.

That said, as it is pip doing the heavy lifting, we must comply to it's configuration pre-conditions as much as possible... 

Thanks for the fast merge.

PR #344, opened 2016-04-16
Update README.md to match new pip-compile behavior

rfleschenberg:
Before I read the changelog, I was a little confused about why `pip-compile` would not update my packages. This should clarify the behavior for future users.


nvie:
Thanks for the PR, @rfleschenberg. This has however already been fixed in the latest master, so this branch does not merge cleanly anymore. I'll close this PR now.

PR #354, opened 2016-05-09
Use ignore_installed=True with PyPIRepository.get_dependencies

blueyed:
This ignores installed packages, which seems to be desired behaviour in
general, but also works around a bug in pip/setuptools/pkg_resources.
(https://github.com/pypa/pip/issues/3661)

TEST CASE:

```
1. rm ~/.cache/pip-tools/depcache*
2. echo "django-autocomplete-light[gfk]==3.1.5" > test.in
3. pip-compile test.in
=> correct
4. pip install django-autocomplete-light
5. rm ~/.cache/pip-tools/depcache*
6. pip-compile test.in
=> incorrect
```

EXPECTED:
It should have `django-querysetsequence` and `django` as additional
dependencies, but this is not the case when `django-autocomplete-light`
is installed.

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -128,6 +128,7 @@ class PyPIRepository(BaseRepository):
                                 self.source_dir,
                                 download_dir=self._download_dir,
                                 wheel_download_dir=self._wheel_download_dir,
+                                ignore_installed=True,
                                 session=self.session)
         dependencies = reqset._prepare_file(self.finder, ireq)
         return set(dependencies)


nvie:
I don't understand this one. I tried running your test case on `master` and got a `test.txt` including django and django-queryset, no matter if I had django-autocomplete-light installed or not. When running this on your branch, those 2 extra dependencies disappear for me.

So it seems that the expected behaviour you describe is already the case on `master` and is **not** present on your branch.

Can you clarify?

blueyed:
Hmm, sounds like you might have mixed up the tests?
There does not appear to be a relevant change between master and this branch apart from that, but I'll rebase it just to be sure.

blueyed:
Just tried it again (on master), and still can confirm it, using Python 3.5.2.

```
% mktmpenv
Using base prefix '/usr'
New python executable in /home/daniel/.pyenv/versions/tmp-system-UtVk4B/bin/python3
Also creating executable in /home/daniel/.pyenv/versions/tmp-system-UtVk4B/bin/python
Installing setuptools, pip, wheel...done.
Ignoring indexes: https://pypi.python.org/simple
Requirement already satisfied (use --upgrade to upgrade): setuptools in /home/daniel/.pyenv/versions/tmp-system-UtVk4B/lib/python3.5/site-packages
Requirement already satisfied (use --upgrade to upgrade): pip in /home/daniel/.pyenv/versions/tmp-system-UtVk4B/lib/python3.5/site-packages

Entering a new shell session with virtualenv 'tmp-system-UtVk4B' (system).
It will be removed when exiting (ctrl+d or 'exit').

% pip install -e .
Obtaining file:///home/daniel/Vcs/pip-tools
Collecting click>=6 (from pip-tools==1.7.1.dev0)
Collecting first (from pip-tools==1.7.1.dev0)
  Using cached first-2.0.1-py2.py3-none-any.whl
Collecting six (from pip-tools==1.7.1.dev0)
  Using cached six-1.10.0-py2.py3-none-any.whl
Installing collected packages: click, first, six, pip-tools
  Running setup.py develop for pip-tools
Successfully installed click-6.6 first-2.0.1 pip-tools six-1.10.0

% cat test.in
django-autocomplete-light[gfk]

% rm ~/.cache/pip-tools/depcache*
% pip-compile test.in
#
# To update, run:
#
#    pip-compile --output-file test.txt test.in
#
django-autocomplete-light[gfk]==3.1.6
django-querysetsequence==0.6  # via django-autocomplete-light
django==1.9.7             # via django-querysetsequence

% pip install django-autocomplete-light
Collecting django-autocomplete-light
Installing collected packages: django-autocomplete-light
Successfully installed django-autocomplete-light-3.1.6

 λ m@o✗❯❯ pip-compile test.in
#
# To update, run:
#
#    pip-compile --output-file test.txt test.in
#
django-autocomplete-light[gfk]==3.1.6
django-querysetsequence==0.6  # via django-autocomplete-light
django==1.9.7             # via django-querysetsequence

% rm ~/.cache/pip-tools/depcache*
% pip-compile test.in
#
# To update, run:
#
#    pip-compile --output-file test.txt test.in
#
django-autocomplete-light[gfk]==3.1.6
```

Then on my branch (rebased on master):
```
% gco -
Switched to branch 'pypi-ignore_installed'
Your branch is up-to-date with 'blueyed/pypi-ignore_installed'.

% grbiom
Successfully rebased and updated refs/heads/pypi-ignore_installed.

# It will still fail first, because of the existing cache:

% pip-compile test.in
#
# To update, run:
#
#    pip-compile --output-file test.txt test.in
#
django-autocomplete-light[gfk]==3.1.6

% rm ~/.cache/pip-tools/depcache*

% pip-compile test.in
#
# To update, run:
#
#    pip-compile --output-file test.txt test.in
#
django-autocomplete-light[gfk]==3.1.6
django-querysetsequence==0.6  # via django-autocomplete-light
django==1.9.7             # via django-querysetsequence

% pip uninstall django-autocomplete-light
Uninstalling django-autocomplete-light-3.1.6:
  …
Proceed (y/n)? y
  Successfully uninstalled django-autocomplete-light-3.1.6

% rm ~/.cache/pip-tools/depcache*
% pip-compile test.in
#
# To update, run:
#
#    pip-compile --output-file test.txt test.in
#
django-autocomplete-light[gfk]==3.1.6
django-querysetsequence==0.6  # via django-autocomplete-light
django==1.9.7             # via django-querysetsequence
```

If I remember correctly this might be related to django-autocomplete-light using "GFK" instead of "gfk" in the `extras_require`.  But the main difference is that two separate code paths are used in the beginning.

blueyed:
btw: it behaves differently when installing django-autocomplete-light using `--editable`: then it works on master, too.  Probably because it's not picked up in that case?!

blueyed:
Replacing "GFK" with "gfk" in `…/pyenv/tmp-system-UtVk4B/lib/python3.5/site-packages/django_autocomplete_light-3.1.6.dist-info/METADATA` and `…/pyenv/tmp-system-UtVk4B/lib/python3.5/site-packages/django_autocomplete_light-3.1.6.dist-info/metadata.json` fixes it on master, too.

vphilippon:
FTR, this is still a thing right now. I reproduce the test (with sliglty different commands):

```
$ echo "django-autocomplete-light[gfk]==3.1.5" > test.in
$ pip-compile -n --rebuild test.in
=> the right dependencies are listed.
$ pip install django-autocomplete-light==3.1.5
$ pip-compile -n --rebuild test.in
=> Missing dependencies.
```

One thing I noted though, with the `-v` output when `django-autocomplete-light==3.1.5` is installed
```
$ pip-compile -nv --rebuild test.in
Using indexes:
  https://pypi.python.org/simple/

                          ROUND 1
Current constraints:
  django-autocomplete-light[gfk]==3.1.5

Finding the best candidates:
  found candidate django-autocomplete-light[gfk]==3.1.5 (constraint was ==3.1.5)

  django-autocomplete-light[gfk]==3.1.5 not in cache, need to check index
No handlers could be found for logger "pip.req.req_set"
  django-autocomplete-light[gfk]==3.1.5 requires -

#
# To update, run:
#
#    pip-compile --output-file test.txt test.in
#
--trusted-host pypi.rdv.geo.ubisoft.onbe

django-autocomplete-light[gfk]==3.1.5
Dry-run, so nothing updated.
```
Note the `No handlers could be found for logger "pip.req.req_set"` line.

Setting `ignore_installed=True` does solve the issue, but I'd like to understand why, and what the impacts are.

@blueyed If you have some insight on what's going on behind the scenes, I would appreciate. I'm fully aware this was a long time ago, so if you've moved on to other things, I'd understand that as well.

blueyed:
@vphilippon 
Yes, it is old already and I do not remember the details, only that I digged into this deeply likely back then for the fix that is being ignored for so long.
It is likely because of the case of the `extras` dep getting lost somewhere, and IIRC it is inside of pip itself.
I think it picked up the already installed dep in one part of the code, but not another - and ignoring it altogether (or rather where this patch applies to) fixed it then.

vphilippon:
@blueyed Thanks for the info, that'll help me out

atugushev:
Fixed in #657.

PR #359, opened 2016-05-11
Fix reading from stdin with Python 3

blueyed:
This fixes the following error:

> TypeError: a bytes-like object is required, not 'str'

Source: http://stackoverflow.com/a/32282458/15690

TEST CASE:
1. `cat test.in | pip-compile --output-file test.txt -`


nvie:
Closing since it's superseded by #403.

PR #361, opened 2016-05-12
Improve pip-compile with regard to stdout

blueyed:
I also had this, but it seems like `-o -` should be given explicitly?!

Or should `-o -` be the default for `-` as input file (which makes sense to me)?
(Done in 31cbab8)

See also 1aa91b5 where `--header` gets disabled for this.
Should it be still possible to pass in `--header` to force it, as with `pip-comfile --header -`?

```
diff --git a/piptools/scripts/compile.py b/piptools/scripts/compile.py
index 9b94809..fd075e6 100755
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -65,10 +65,6 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
                                       "the default is {}").format(DEFAULT_REQUIREMENTS_FILE))
         src_files = (DEFAULT_REQUIREMENTS_FILE,)

-    if len(src_files) == 1 and src_files[0] == '-':
-        if not output_file:
-            raise click.BadParameter('--output-file is required if input is from stdin')
-
     if len(src_files) > 1 and not output_file:
         raise click.BadParameter('--output-file is required if two or more input files are given.')
```

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -52,11 +52,12 @@ class OutputWriter(object):
                 yield ''  # extra line of whitespace
 
     def write_format_controls(self):
-        for nb in self.format_control.no_binary:
-            yield '--no-binary {}'.format(nb)
-        for ob in self.format_control.only_binary:
-            yield '--only-binary {}'.format(ob)
-        yield ''
+        if self.format_control.no_binary or self.format_control.only_binary:
+            for nb in self.format_control.no_binary:
+                yield '--no-binary {}'.format(nb)
+            for ob in self.format_control.only_binary:
+                yield '--only-binary {}'.format(ob)
+            yield ''
 
     def _iter_lines(self, results, reverse_dependencies, primary_packages):
         for line in self.write_header():
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -89,13 +89,15 @@ class OutputWriter(object):
     def write(self, results, reverse_dependencies, primary_packages):
         with ExitStack() as stack:
             f = None
-            if not self.dry_run:
+            if not self.dry_run and self.dst_file != '-':
                 f = stack.enter_context(AtomicSaver(self.dst_file))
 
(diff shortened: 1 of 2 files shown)


piptools/writer.py
```diff
@@ -89,14 +90,15 @@ def _iter_lines(self, results, reverse_dependencies, primary_packages):
     def write(self, results, reverse_dependencies, primary_packages):
         with ExitStack() as stack:
             f = None
-            if not self.dry_run:
+            if not self.dry_run and self.dst_file != '-':
                 f = stack.enter_context(AtomicSaver(self.dst_file))
 
             for line in self._iter_lines(results, reverse_dependencies, primary_packages):
-                log.info(line)
```
cjerdonek: Instead of removing logging, shouldn't the log be going to stderr like most logging does? If it goes to stderr, it wouldn't be interfering with the stdout issue.

blueyed: @cjerdonek 
> shouldn't the log be going to stderr like most logging does?

Agreed.  Do you want to create an issue for it?
But then this log line is still a bit too verbose (move it to debug? / add a prefix like "writing line"?).

cjerdonek: Sure, I can make an issue. (I would need to confirm first that `log` does indeed go to stdout.)

blueyed: It does not use the logging module, but click: https://github.com/nvie/pip-tools/blob/d521d766113e69ea0a10017bb968d3b3bb242e5f/piptools/logging.py#L21-L22.

cjerdonek: Done here: https://github.com/nvie/pip-tools/issues/362

blueyed:
@piotr-dobrogost 
Here we are.. ;)

atugushev:
Close in favour of #727.

PR #364, opened 2016-05-19
Write error message to the stderr.

firebirdberlin:
Some users may want to filter log messages using pipes. Error messages
should be written to stderr so that errors remain visible.


tysonclugg:
@firebirdberlin This looks like a great start, but sadly it's been a long time and this PR is now stale. We'd love to see a fresh pull request if you've got some time!

PR #365, opened 2016-05-19
Add ability to process CA Bundles

khwilson:
One of the options that's missing from pip-compile that pip allows for is `--cert`, which has been around longer than `--client-cert` (at least since pip 5).

These commits attach the cert to the PipSession in the same way that pip does and also exposes the option in the CLI.


vphilippon:
Closing, #612 is the rebased version.
Thanks for your original PR, I'm sorry it stayed there for so long.

PR #368, opened 2016-06-01
Add -markerlib and pkg-resources to the list of PACKAGES_TO_IGNORE

cworth-gh:
On Debian unstable, mkvirtualenv 15.0.0 is installing two pseudo
packages (each with a version number of 0.0.0) as can be seen here:

```
$ mkvirtualenv foo
Running virtualenv with interpreter /usr/bin/python2
New python executable in /home/cworth/.virtualenvs/foo/bin/python2
Also creating executable in /home/cworth/.virtualenvs/foo/bin/python
Installing setuptools, pkg_resources, _markerlib, pip, wheel...done.
```

But then pip-sync will try to uninstall these packages (since nothing
explicitly depends on them) which will then cause pip stop working as
can be seen here:

```
$ pip-sync reqs.txt
Uninstalling pkg-resources-0.0.0:
  Successfully uninstalled pkg-resources-0.0.0

$ pip-sync reqs.txt
  File "/home/cworth/.virtualenvs/foo/bin/pip-sync", line 5, in <module>
    from pkg_resources import load_entry_point
ImportError: No module named pkg_resources
```

This commit adds these two package names to the PACKAGES_TO_IGNORE
list so that pip-sync will not try to uninstall them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,14 @@
-# 1.7.1
+# 1.7.0+nimbis.1
+
+- Bug fix to work with Debian mkvirtualenv (this has been submitted
+  upstream here: https://github.com/nvie/pip-tools/pull/368 )
 
 - Add `--allow-unsafe` option (#377)
-
 
 # 1.7.0
 
 - Add compatibility with pip >= 8.1.2 (#374)
   Thanks so much, @jmbowman!
-
 
 # 1.6.5
 
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -8,9 +8,11 @@ from .exceptions import IncompatibleRequirements, UnsupportedConstraint
 from .utils import flat_map, key_from_req
 
 PACKAGES_TO_IGNORE = [
+    '-markerlib',
     'pip',
     'pip-tools',
     'pip-review',
+    'pkg-resources',
     'setuptools',
     'wheel',
 ]
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ from setuptools import find_packages, setup
 
 setup(
     name='pip-tools',
-    version='1.7.1dev0',
+    version='1.7.0+nimbis.1',
     url='https://github.com/nvie/pip-tools/',
     license='BSD',
     author='Vincent Driessen',


tekacs:
Currently having to add these to every requirements.in file as extras - would be great to see this merged!

palfrey:
I'd also like to see this in. Any thoughts on getting it merged? Is it worth setting up a separate PR with just the `sync.py` changes?

amadornimbis:
I know this repository recently changed ownership, so I'm bumping this to see if anything needs to be changed in to get the first commit of this series merged?

vphilippon:
@amadornimbis It definitely needs a rebase on master. The version update in setup.py and CHANGELOG.md should probably be remove too, as that'll be handled when doing the release (CHANGELOG entry itself should be kept though).

`pkg_resources` seems fine to add to the list, as it usually comes from setuptools anyway.
`_markerlib` is not a valid package name to install anyway (`pip install _markerlib` leads to and `InvalidRequirement` error), so I feel like it can be ignored too.

@davidovich Could you take a look at this? It seems like this would fix some error related to `pip-sync` uninstalling `pkg_resources`.

oxplot:
As a workaround, you can setup the env with no setuptools and install it after activation:

```sh
virtualenv --no-setuptools venv
. venv/bin/activate
pip install setuptools
pip-sync requirements.txt
```

vphilippon:
Will be completed in #613

graingert:
instead of ignoring `-markerlib` what about also ignoring anything without a valid package name:

`^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$`

graingert:
@vphilippon as defined: https://github.com/pypa/warehouse/blob/2b54d1e6a004ced695745d1f38a4304d81d6f8f7/warehouse/packaging/models.py#L82

vphilippon:
@graingert Hm, maybe. but that should be opened in its own issue, that'll be lost on here.

graingert:
@vphilippon https://github.com/jazzband/pip-tools/issues/618 done

PR #371, opened 2016-06-21
338: Trusted hosts not written to requirements.txt

gtoonstra:


nvie:
Thanks for the contribution!

Issue #355, opened 2016-05-09
Add support for non-editable vcs requirements


voidlily:
Previous attempt at #169

simon-weber:
I'm interested in taking a stab at this.

@blueyed is it correct that your original changes don't require https://github.com/pypa/pip/pull/2913 if vcs requirements.in are of the form ` git+https` or `git+ssh`? I'll be more hesitant if this would also involve pushing things through pip 😁

jmbowman:
Should this be closed now that https://github.com/jazzband/pip-tools/pull/807 has been merged?  The resolution of https://github.com/jazzband/pip-tools/pull/372 implies that it should.

blueyed:
Yeah, thanks @jmbowman.

PR #372, opened 2016-06-29
support pinned vcs dependencies [#355]

simon-weber:
This includes a minimal set of changes from https://github.com/nvie/pip-tools/pull/169 to support pinned vcs dependencies (#355).

For example, a requirements.in like:

```
git+https://github.com/simon-weber/gmusicapi.git@develop#egg=gmusicapi==10.0.2rc1
```

will output a requirements.txt with all the dependencies, keeping that line exactly the same. Without the version pinned, you'll get an error:

```
pip-compile does not support non-editable vcs URLs that are not pinned to one version.
```


simon-weber:
Bump 😁 The failing tests seem to be a known issue.

simon-weber:
Ok, both compile and sync seem solid now with our gnarly set of prod dependencies. I've also rebased onto master and fixed an unrelated linting error.

Ready for a look, @nvie.

piptools/resolver.py
```diff
@@ -119,9 +119,9 @@ def resolve(self, max_rounds=10):
 
     def _check_constraints(self):
         for constraint in chain(self.our_constraints, self.their_constraints):
-            if constraint.link is not None and not constraint.editable:
-                msg = ('pip-compile does not support URLs as packages, unless they are editable. '
-                       'Perhaps add -e option?')
+            if ((constraint.link and not constraint.editable
+                 and not constraint.link.is_artifact and not is_pinned_requirement(constraint))):
```
nvie: I'm having a hard time understanding this condition. Why is the `is_pinned_requirement()` in there?

piptools/repositories/pypi.py
```diff
@@ -91,7 +91,7 @@ def find_best_match(self, ireq, prereleases=None):
         Returns a Version object that indicates the best match for the given
         InstallRequirement according to the external repository.
         """
-        if ireq.editable:
+        if ireq.editable or (ireq.link and not ireq.link.is_artifact):
```
nvie: Seems like this `x.link and not x.link.is_artifact` is a recurring pattern. Can you make a helper function for that? I.e. `is_vcs_link(ireq)`?

nvie: Then you can also make another helper method `is_editable_vcs_link(ireq)`, which also does the `.editable` check. I think having those functions will make the rest of the changes below more readable.

simon-weber: Sure thing.

piptools/resolver.py
simon-weber: Ah, yeah, this is something I wanted to run by you. The motivation is to restrict the input so pip-compile can take it at face value. In particular, supporting unpinned vcs links seems tough since:
* pip doesn't seem to expose the "true" version of a vcs link unless it's been cloned and installed (it just seems to assume the provided version is accurate for the `@ref` in the link)
* internal logic assumes that constraints are pinned later on (ie, removing _check_constraints raises a "expected pinned or editable..." TypeError later on)

Thoughts?

simon-weber:
(github hid the unresolved discussion about is_pinned_requirement; here's the link: https://github.com/nvie/pip-tools/pull/372#discussion_r70166343)

simon-weber:
@nvie got a chance to look at this? For what it's worth, we've been running off my pr fork in prod without problems.

simon-weber:
@nvie bump! 😁

simon-weber:
@nvie bump again! 😁 We're looking into contributing more, but it's upstream of this.

nvie:
Sorry about the long delay here. I've been incredibly involved lately and having a hard time keeping up with issues/PRs on my open source projects at the moment.

Re the ticket: I don't understand what the pin in that example line is doing:

    git+https://github.com/simon-weber/gmusicapi.git@develop#egg=gmusicapi==10.0.2rc1

When I change the `10.0.2rc1` to `0.1` here locally and recompile, I get the exact same output, same dependencies. I can even change it to `foobar` and it still works. So it doesn't seem to do anything?

Here's what I had in mind for the feature.

Unfortunately Pip is making all of this hard since there are 3 different concepts at play here:

- Name vs URL, so `my_package` vs `https://example.org/my_package.tar.gz`.
- "Just" URLs vs VCS-based URLs, which are treated specially. So `https://example.org/my_package.tar.gz` vs `-e git+git://github.com/foo/bar.git#egg=bar`.
- Pinning to a specific symbolic reference for VCS URLs: So `-e git+git://github.com/nvie/pip-tools.git#egg=pip-tools` vs `-e git+git://github.com/nvie/pip-tools.git@example-branch#egg=pip-tools`.

In this latter case, the symbolic name `example-branch` points to a specific commit at a specific time. When you call `pip-compile` on an input file that has this branch name in the URL, it should "compile" this specification to the following URL:

    -e git+git://github.com/nvie/pip-tools.git@1db806f2f92375bd37b5e5c2b4fe4481e43deb3c#egg=pip-tools

In other words, it "resolves" the branch to the commit it points to at compile time, and pins it to that. When you invoke the same command later, the commit may be different. Normal upgrade semantics should be at play here. If the branch has advanced, but `requirements.txt` still exists, don't upgrade it, unless `--upgrade` is passed along, in which case we should re-resolve.

VCS URLs should not be dealing with version numbers. Their branch names as given after the `@` symbol in the URL should be considered to be the "spec", very much akin to `foo`, or `foo<2.0`, or `foo>=1.4`. The "actual commit" they point should go into the output file and is very much the same as a "pinned version", i.e. `foo==1.7.9`.

Does that make sense?

nvie:
**tl;dr**

Input:

    -e git+git://github.com/nvie/pip-tools.git@example-branch#egg=pip-tools

Output:

    -e git+git://github.com/nvie/pip-tools.git@1db806f2f92375bd37b5e5c2b4fe4481e43deb3c#egg=pip-tools

simon-weber:
> Sorry about the long delay here.
 
No problem! Totally understand.

> When I change the 10.0.2rc1 to 0.1 here locally and recompile, I get the exact same output...

Oh, whoops, that was a bad example. I should have used the release tag (@10.0.2rc1) instead. I wouldn't recommend using a non-static ref for the reasons you laid out (ie, it may change and the version will then not match).

> I can even change it to foobar and it still works. So it doesn't seem to do anything?

It doesn't do anything at pip-tool's level, but it does for pip: since pip doesn't know the version of the package that's checked in, it will believe the version you give it via the spec. For example, try these examples in order:
* create new venv
* `pip install git+https://github.com/kennethreitz/requests.git@v2.11.1#egg=requests==2.11.1` -> installs requests as expected
* `pip install git+https://github.com/kennethreitz/requests.git@v2.11.1#egg=requests==foobar` -> reclones and reinstalls requests, even though it was already up to date with the code from that revision
* `pip install git+https://github.com/kennethreitz/requests.git@v2.11.1#egg=requests==2.11.1` -> no clone; requirement already satisfied
* ` pip install git+https://github.com/kennethreitz/requests.git@foobar#egg=requests==2.11.1` -> no clone; requirement already satisfied
* `pip install git+https://github.com/kennethreitz/requests.git@v2.9.2#egg=requests` -> no clone; requirement already satisfied
* `pip install git+https://example.com/foo.git@v2.11.1#egg=requests` -> no clone; requirement already satisfied

So, pip treats the provided egg==version pair as truth of what it will find if it were to clone url@ref (which seems reasonable, if not well documented; it took some time to figure this behavior out). That's my reasoning behind requiring a pinned version: it pushes people towards a url pip will behave well with.

We use this format (static ref, pinned version) at Venmo pretty extensively for two different use cases:
* internal packages installed from GitHub Enterprise
* forks and/or contributed changes waiting on merges (eg `git+https://github.com/venmo/pip-tools.git@1.7.1a1#egg=pip-tools==1.7.1a1`)

> Here's what I had in mind for the feature.

I really like this behavior. But, it does seem like a good amount of work to implement, and not something I'm likely to have time for. Would you consider merging the behavior in this PR as a stop-gap so vcs urls are at least functional? Maybe with added docs so people know to only use a static ref, and why the version is required?

piotr-dobrogost:
@nvie 
> Would you consider merging the behavior in this PR as a stop-gap so vcs urls are at least functional?

Would you? :)

> In this latter case, the symbolic name example-branch points to a specific commit at a specific time. 

What if there's a tag name not a branch name given after '@' in the vcs url? Do you also want to treat it like a *moving target* and pin it to the commit hash?

s0undt3ch:
I'm interested in this support landing into the repo. Been bit by the lack of VCS support recently and the compiled order being wrong because of requirements from the VCS sources not being computed.

Daenyth:
Any update here?

atugushev:
Fixed in #807

PR #374, opened 2016-07-06
Add support for pip 8.1.2

jmbowman:
Proposed fix for https://github.com/nvie/pip-tools/issues/358 . pip 8.1.2 switched from [pkg_resources](http://setuptools.readthedocs.io/en/latest/pkg_resources.html#requirement-objects) to [packaging](https://packaging.pypa.io/en/latest/requirements/) for its internal representation of requirements.  This PR adds a few utility functions and a wrapper class to handle the differences that pip-tools cares about so it can work with pip 8.1.2 in addition to all the previously supported versions.

I looked into switching over from pip to `packaging` directly, but it lacks some of the required features (like [parsing editable requirements](https://github.com/pypa/packaging/issues/64)) and they've been deemed out of scope for that package.  In the long term, there will probably be a new package specifically for parsing requirements files (which utilizes `packaging` where it can, and could have a vendored version included in pip for use there).

I also fixed a minor whitespace issue in `tests/test_cli.py` that was causing a flake8 failure.

### DIFF
--- a/piptools/repositories/local.py
+++ b/piptools/repositories/local.py
@@ -2,7 +2,23 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
+from piptools.utils import key_from_req
 from .base import BaseRepository
+
+
+def ireq_satisfied_by_existing_pin(ireq, existing_pin):
+    """
+    Return True if the given InstallationRequirement is satisfied by the
+    previously encountered version pin.
+    """
+    if hasattr(existing_pin.req, 'specs'):
+        # pip < 8.1.2
+        version = existing_pin.req.specs[0][1]
+        return version in ireq.req
+    else:
+        # pip >= 8.1.2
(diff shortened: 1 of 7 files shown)


nvie:
Wow, @jmbowman! This is fantastic! Thanks for making this effort. I like how you even managed to keep the stuff backward-compatible! 🎩 

I just tried it out myself and discovered a few more places that had to be touched, but those were easy fixes. You can check them out in 90349b55e66431037199597bbeed9c659325cd7d.

This is so very appreciated, thanks so much! ❤️

jmbowman:
Ah-ha, sounds like there's room for improvement in test coverage. 😃  At least the changes should be set up so that such oversights won't break any already-working code with the previous versions of pip, they'd just be bugs when trying it against the new version.

fuhrysteve:
Thanks @jmbowman !

PR #377, opened 2016-07-08
Add option to pin packages considered unsafe in pip-compile

alekzvik:
Due to some specifics I had to include some version of `pip` in dependencies in several projects.
I am not sure it is needed for everyone, so I added a flag to have an option to enable that.


nvie:
Can you remove the short option (`-u`) and rename this option (both internally and externally) to `--allow-unsafe`? Otherwise, looks good to me. Thanks!

nvie:
Thanks!

alekzvik:
Thank you for a great project.

When do you plan to release this on PyPI?

borgstrom:
@nvie Is there anything else holding up 1.7.1 from landing in PyPI with the addition of `--allow-unsafe`?  I too have run into a situation where I want to include pip.

Issue #333, opened 2016-03-15
Exclude package?


ariscn:
I'm running pip-compile under OS X and have a `requirements/dev.in` file which includes IPython. This pulls in `gnureadline`. When I build a pex (https://github.com/pantsbuild/pex) from this requirements file for a Linux system, it's not usable, because gnureadline is only available for OS X.

I can work around this by removing or commenting out the gnureadline dependency, but this is a bit hacky. It would be great to be able to do: `pip-compile requirements/base.in && pip-compile requirements/dev.in && pip-compile --remove gnureadline requirements/stage.in`.

jeffwidman:
FYI--latest version of `ipython` removed the `gnureadline` dependency in favor of the cross-platform `prompt-toolkit`.

The feature request is still valid in the abstract though...

nvie:
If someone can make a PR for a `--exclude <pkg>` flag, I'd be happy to merge the PR. Please make sure to allow multiple occurrances of this flag, i.e. `--exclude foo --exclude bar ...`

lwm:
@nvie @ariscn 

I've started working on this but have questions.

Should this exclusion happen only for packages described in a `requirements.in` or should also exclude dependencies?

If we start to exclude dependencies, then we might break some packages. The user might not know at which level some package they want to exclude will be excluded for some other package.

Groxx:
Yeah, this would be great.  We've had to add a lint step to prevent `enum` from sneaking in because `enum34` collides with it (why the heck did they think that was a good idea?!), ideally we could `--exclude enum` or similar right in the .in file so it's unmissable.

lwm:
@Groxx could you check out the branch (add-exclude-flag) of https://github.com/lwm/pip-tools/commit/f5771a3aa855887f22dad2ba4a58eef1f158f34f and give it a check if it works for you? I am afraid #388 is going to rot without review or progress.

milin:
@lwm Would this fix issues like this https://github.com/nvie/pip-tools/issues/366? I was thinking along with exclude,  one could also make pip-compile flexible, so if issues like that happen, we could fall back to a backup version.

lwm:
@milin I am not sure (new to the code here). Try it and let's find out? Not sure what you mean by a 'backup version'? How would the interface look like?

milin:
@lwm I ended up fixing it since I needed it for a project. https://github.com/nvie/pip-tools/pull/394 as more info on the problem and the proposed solution.

[...]

PR #388, opened 2016-09-05
add `--exclude` option

lwm:
Something for #333.

It excludes dependencies only. It's not very smart (we still fetch from local / external) and I am not sure it's in the right place. But anyway, a start. https://github.com/nvie/pip-tools/issues/333#issuecomment-242063237 is probably still a valid concern!? This change would need some documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.7.2
+
+- Add `--exclude` option (#333)
+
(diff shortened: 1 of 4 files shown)


lwm:
Closing. The code is there if anyone needs a reference later on.

piotr-dobrogost:
Could this be reopend and possible merged/reviewed?

PR #392, opened 2016-09-19
Issue 376 fix duplicate package entries

milin:
Fixes https://github.com/nvie/pip-tools/issues/376


davidovich:
Will close in favor of #427 which fixes the duplication problem related to normalization. Thanks for having a stab at this!

PR #403, opened 2016-10-20
Added mode='wt' to tempfile creation.

rolando:
This allows to write text into the file directly (as stdin.read()
returns a text type and by default tmpfile.write expects bytes).


rolando:
The use case I had is that I wanted to get the results for a single package:

```
$ echo jsonschema | pip-compile - -o -
```

Duplicate of #359.

nvie:
Thanks!

PR #406, opened 2016-11-04
Phased install

cpennington:
In some cases, a package A may required that package B has been fully installed before the installation for package A can start (numpy and scipy, for some set of versions, for instance). In this case, we need a way to get consistent versions across a number of requirements files, without forcing installation from a single requirements file.

To that end, this PR adds `--phased` to both `pip-compile` and `pip-sync`, which compiles and installs sets of requirements files **in the order they are listed on the commandline**.

### DIFF
--- a/piptools/cache.py
+++ b/piptools/cache.py
@@ -80,6 +80,9 @@ class DependencyCache(object):
         like so:
 
         ("ipython", "2.1.0[nbconvert,notebook]")
+
+        For a requirement with a link (either editable or not), the link will be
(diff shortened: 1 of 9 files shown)


cpennington:
This builds on top of #405

vphilippon:
(Catching up over a year later)
A lots has changed. Can you tell if this is still required in the latest release (1.10.2)?
If yes, could you give an example showing why its needed? I haven't any reported cases where this seemed to cause issue.

atugushev:
Closed due to lack of response.

PR #409, opened 2016-11-11
Support specifying which packages and versions to update.

nnja:
The current behavior for `pip-compile --upgrade` updates all requirements. This becomes complicated for large `.in` files when that's not always the desired behavior.

This patch introduces a new flag, `--upgrade-package` which has the following features:

Upgrade only a specific package to the latest version:
```console
$ pip-compile --upgrade-package flask
```

Upgrade a specific package to a specific version:
```console
$ pip-compile --upgrade-package flask==0.10.1 
```

Upgrade multiple packages at a time, by passing in multiple arguments:
```console
$ pip-compile --upgrade-package flask==0.10.1 --upgrade-package werkzeug
```

Additional output was added when using the `--verbose` flag to list which packages were updated, and to which version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,17 @@ werkzeug==0.10.4          # via flask
 
 And it will produce your `requirements.txt`, with all the Flask dependencies
 (and all underlying dependencies) pinned.  Put this file under version control
-as well and periodically re-run `pip-compile --upgrade` to update the packages.
+as well.
 
+To update all packages, periodically re-run `pip-compile --upgrade`.
+
+To update a specific package to the latest or a specific version use the `--upgrade-package` or `-P` flag:
+
+```console
+$ pip-compile --upgrade-package flask  # only update the flask package
+$ pip-compile --upgrade-package flask --upgrade-package requests  # update both the flask and requests packages
+$ pip-compile -P flask -P requests==2.0.0  # update the flask package to the latest, and requests to v2.0.0
+```
 
 Example usage for `pip-sync`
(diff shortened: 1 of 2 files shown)


nnja:
Pushed some updates - the build is now green, except for https://travis-ci.org/nvie/pip-tools/jobs/176202798 `piptools/utils.py:22:1: E305 expected 2 blank lines after class or function definition, found 1` which seems unrelated to my changes, since I have not modified utils.py

nvie:
I love this idea, thanks for your contribution, @nnja!

The way pip-compile's default invocation vs the upgrade logic works is quite clever.  Basically, what happens is that _if_ an output file exists (and the `--upgrade` flag is not given), it will add the current pins from the output file into the list of *input* specs, and then "just" trigger the resolver.

In that light, using this new `-P <pkg>` flag really means "don't perform the above-mentioned process for this package `<pkg>`".  Or, when a pin is given: "ditto, but also pin it to `<pin>` while you're at it".

So really all you need to do here is to filter the list of "existing pins" and therefore discarding those explicit pins in favor of the stuff provided on the command line.  I've gone ahead and implemented that in d0dd26a79fc66758af134e9687ed2a9a8cbb1d29, which as you can see does not require any special handling for this feature at all.  (It's perhaps best to look at [the net-diff against master](https://github.com/nvie/pip-tools/compare/nvie:dc49d01...nvie:d0dd26a) for the best view of the total change required for this.)

Thanks a lot!

PR #410, opened 2016-11-14
Update LocalRequirementsRepository to retain extras when returning best match

### LINKED ISSUE #408: Extras not propagated into requirements.txt after insertion
Extras, in this context, are specific functional extensions to a particular package.

This is similar to #154 but differs because this triggers when adding an extra to an existing requirement.

The cached result appears to prevent appropriate detection that the extra was added.

##### Steps to replicate

1. Create a `requirements.in` as follows:
  ```text
pip-tools
hypothesis
   ```
2. Run `pip-compile` from the command line.
3. Edit the requirements.in file to the following:
  ```text
pip-tools
hypothesis[django]
  ```
4. Re-run `pip-compile` from the command line.
3. Examine `requirements.txt`

##### Expected result

```text
#
# To update, run:
#
#    pip-compile --output-file reqs.txt reqs.in
#
click==6.6                # via pip-tools
django==1.10.3            # via hypothesis
fake-factory==0.7.2       # via hypothesis
first==2.0.1              # via pip-tools
hypothesis[django]==3.6.0
pip-tools==1.7.0
python-dateutil==2.6.0    # via fake-factory
pytz==2016.7              # via hypothesis
six==1.10.0               # via fake-factory, pip-tools, python-dateutil
```

##### Actual result

```text
#
# To update, run:
#
#    pip-compile --output-file reqs.txt reqs.in
#
click==6.6                # via pip-tools
first==2.0.1              # via pip-tools
hypothesis==3.6.0
pip-tools==1.7.0
six==1.10.0               # via pip-tools
```

##### Workaround

Simply deleting the existing `requirements.txt` file and installing from scratch appears to work, but shouldn't be a necessary part of the pip-tools workflow.

acdimalev:
This is behavior that clearly varies between `PyPIRepository` and `LocalRequirementsRepository`.

This patch should resolve #408.

I did write a unit test for this issue, but it's not yet suitable for a pull request.

https://gist.github.com/acdimalev/c3219ac4aaad44e6e77a4f7429e48e29

### DIFF
--- a/piptools/repositories/local.py
+++ b/piptools/repositories/local.py
@@ -2,7 +2,7 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
-from piptools.utils import key_from_req
+from piptools.utils import as_tuple, key_from_req, make_install_requirement
 from .base import BaseRepository
 
 
@@ -57,7 +57,10 @@ class LocalRequirementsRepository(BaseRepository):
         key = key_from_req(ireq.req)
         existing_pin = self.existing_pins.get(key)
         if existing_pin and ireq_satisfied_by_existing_pin(ireq, existing_pin):
-            return existing_pin
+            project, version, _ = as_tuple(existing_pin)
+            return make_install_requirement(
+                project, version, ireq.extras
+            )
         else:
             return self.repository.find_best_match(ireq, prereleases)
 


nvie:
Thanks for this great patch, @acdimalev!

PR #411, opened 2016-11-15
Update README.md

realsby:
some packages in your requirements.in need this upgrade, to compile without error (like ipython)


nvie:
I'd love to leave that to the README of IPython instead of pip-tools.  pip is the exception here, since pip-tools itself needs it. Thanks!

PR #424, opened 2016-12-13
Allow editable dependencies with hashes

orf:
I don't think the `--generate-hashes` flag should give up and quit if a single dependency is editable/not pinned. This MR just skips those that don't have hashes.

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -75,7 +75,17 @@ class Resolver(object):
         """
         Finds acceptable hashes for all of the given InstallRequirements.
         """
-        return {ireq: self.repository.get_hashes(ireq) for ireq in ireqs}
+        hashes = {}
+        for ireq in ireqs:
+            try:
+                ireq_hash = self.repository.get_hashes(ireq)
+            except TypeError:
+                continue
+            else:
+                if ireq_hash is not None:
+                    hashes[ireq] = ireq_hash
+
+        return hashes
 
     def resolve(self, max_rounds=10):
         """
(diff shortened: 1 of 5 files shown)


orf:
Fixed up the tests, I used the github editor but didn't indent properly :-1:

orf:
Hi, is there any problem or comment on this MR? Some hashes are better than no hashes, and if you need to add a requirement from a git repo it's a shame to loose all other hashes for non-git requirements.

orf:
Didn't mean to close. I've added tests, and it's working well.

vphilippon:
@orf Sorry for the long wait.
If you still want to pick this up, you can go and rebase this on master.
You can remove the changelog entry, and put it in the description of your PR. Maintainers will take care of writting up the changelog afterwars (I realised this was an incredible source of conflicts).

If not, feel free to say no, I will completely understand. I'll likely take up the work to get it in myself.
In any case, thanks for your contribution 👍

piptools/resolver.py
```diff
@@ -75,7 +75,17 @@ def resolve_hashes(self, ireqs):
...
         """
-        return {ireq: self.repository.get_hashes(ireq) for ireq in ireqs}
+        hashes = {}
+        for ireq in ireqs:
+            try:
+                ireq_hash = self.repository.get_hashes(ireq)
+            except TypeError:
+                continue
+            else:
+                if ireq_hash is not None:
```
vphilippon: I could not find a case where the result would be `None`. Di you find one?
Also, I'm not sure I'd like to silently skip hashes, except for clear, defined cases, so I would rather see it fail here. Otherwise, that might bring up a false sense of security, which is really not nice.

vphilippon:
Finally covered by #524

Issue #376, opened 2016-07-07
Duplicate package entries in compiled requirements with pip 8.1.2 and pip-compile 1.7.0


bwells:
I'm unclear if this a bug within pip or pip tools. If multiple valid names for a package are included in a requirements list then the canonical name is included twice in the compiled requirements file. In our case we had 'factory-boy' and 'factory_boy' both included in separate files of a requirements file chain. 

This is not true with pip==8.1.1 for either pip-tools==1.6.5 or 1.7.0. It is only true with pip==8.1.2 and pip-tools==1.7.0
##### Steps to replicate
1. Create `requirements.in` with `factory_boy` and `factory-boy` on separate lines
2. `pip-compile requirements.in`
##### Expected result

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file requirements.txt requirements.in
#
factory-boy==2.7.0
fake-factory==0.5.8       # via factory-boy
python-dateutil==2.5.3    # via fake-factory
six==1.10.0               # via fake-factory, python-dateutil
```
##### Actual result

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file requirements.txt requirements.in
#
factory-boy==2.7.0
factory-boy==2.7.0
fake-factory==0.5.8       # via factory-boy
python-dateutil==2.5.3    # via fake-factory
six==1.10.0               # via fake-factory, python-dateutil
```

See that `factory-boy==2.7.0` is included twice.

dimka665:
Same issue

requirements.in:
```
django-libsass == 0.7
django_compressor == 2.0
```
pip-compile requrements.in
```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file t.txt t.in
#
django-appconf==1.0.2     # via django-compressor, django_compressor
django-compressor==1.6    # via django-libsass
django-libsass==0.7
django_compressor==2.0    # via django-libsass
libsass==0.11.1           # via django-libsass
rcssmin==1.0.6            # via django_compressor
rjsmin==1.0.12            # via django_compressor
six==1.10.0               # via libsass
```

gbataille:
exactly the same for me :(

Cabalist:
Same here for Flask-Migrate:

```
flask-script==2.0.5       # via flask-migrate
flask-script==2.0.5       # via flask-migrate
```

milin:
Can confirm.

```
[...]

PR #427, opened 2016-12-20
Fix the usage of ireq keys for comparison purposes

jonafato:
Per [PEP 426](https://www.python.org/dev/peps/pep-0426/):

>  All comparisons of distribution names MUST be case insensitive, and
>  MUST consider hyphens and underscores to be equivalent.

Without this normalization, dependencies can become unintentionally
upgraded if the existing pinned `requirements.txt` contains a package
specified in a different style from another packages `setup.py`.

Example of the current behavior without this change:

`requirements.in`

    aiohttp==1.0.1

`requirements.txt` (before running `pip-compile`)

    #
    # To update, run:
    #
    #
    aiohttp==1.0.1
    async-timeout==1.0.0      # via aiohttp
    chardet==2.3.0            # via aiohttp
    multidict==2.1.4          # via aiohttp

`requirements.txt` (after running `pip-compile`)

    #
    # To update, run:
    #
    #
    aiohttp==1.0.1
    async-timeout==1.1.0      # via aiohttp
    chardet==2.3.0            # via aiohttp
    multidict==2.1.4          # via aiohttp

Desired behavior: `async-timeout` is pinned here and should not be
automatically upgraded, as `aiohttp` does not place a minimum version
requirement on `async-timeout` in the `1.0.1` release
(https://github.com/KeepSafe/aiohttp/blob/v1.0.1/setup.py#L57).

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -35,10 +35,12 @@ def key_from_req(req):
     """Get an all-lowercase version of the requirement's name."""
     if hasattr(req, 'key'):
         # pip 8.1.1 or below, using pkg_resources
-        return req.key
+        key = req.key
     else:
         # pip 8.1.2 or above, using packaging
-        return req.name.lower()
+        key = req.name.lower()
+    key = key.replace('_', '-')
+    return key
 
 
 def name_from_req(req):


davidovich:
Will go without tests right now, because we need to make a 1.8.1 release, and manual tests are conclusive.

davidovich:
Thanks ! for this @jonafato

davidovich:
fixes #376

PR #429, opened 2016-12-21
Check the pip version is compatible ASAP

### LINKED ISSUE #484: pip-compile fails if output file already exists
Given a requirements.in file which consists only of the line:
```
-e ..
```
(that is, pull in requirements from the setup.py file one folder up from my requirements.in file)

The first pip-compile run works fine. The second one throws an error. If I delete the file that was generated by the first run, it will work again. There are no permissions issues (tried it with full perms on the file to check).

##### Environment Versions

1. os: OSX 10.11.6
1. python version: 2.7
1. pip version: 9.0.1
1. pip-tools version: 1.8.2

##### Steps to replicate

1. pip-compile -v -o requirements.txt
**SUCCESS**
2. pip-compile -v -o requirements.txt 
**FAILURE**

##### Expected result
An updated (or identical) requirements.txt file

##### Actual result
```python
  File "/Users/<username>/.virtualenvs/<virtualenv>/bin/pip-compile", line 11, in <module>
    sys.exit(cli())
  File "/Users/<username>/.virtualenvs/<virtualenv>/lib/python2.7/site-packages/click/core.py", line 722, in __call__
  File "/Users/<username>/.virtualenvs/<virtualenv>/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/<username>/.virtualenvs/<virtualenv>/lib/python2.7/site-packages/click/core.py", line 895, in invoke
  File "/Users/<username>/.virtualenvs/<virtualenv>/lib/python2.7/site-packages/click/core.py", line 535, in invoke
  File "/Users/<username>/.virtualenvs/<virtualenv>/lib/python2.7/site-packages/piptools/scripts/compile.py", line 131, in cli
    key = key_from_req(ireq.req)
  File "/Users/<username>/.virtualenvs/<virtualenv>/lib/python2.7/site-packages/piptools/utils.py", line 41, in key_from_req
    key = req.name
AttributeError: 'NoneType' object has no attribute 'name'
```

alexwlchan:
This avoids getting a nasty traceback if we hit an error between importing pip and running the compatibility check.

Additionally, it seems like the supported version of pip was incremented but the README wasn’t. This patch fixes the README.

I considered adding a test for this, but that would require an entirely new set of tests for what is hopefully an edge case scenario.

Resolves #484.

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -8,6 +8,11 @@ import sys
 import tempfile
 
 import pip
+
+# Make sure we're using a compatible version of pip
+from ..utils import assert_compatible_pip_version
+assert_compatible_pip_version()
+
 from pip.req import InstallRequirement, parse_requirements
 
 from .. import click
@@ -15,12 +20,8 @@ from ..exceptions import PipToolsError
 from ..logging import log
 from ..repositories import LocalRequirementsRepository, PyPIRepository
 from ..resolver import Resolver
-from ..utils import (assert_compatible_pip_version, is_pinned_requirement,
-                     key_from_req)
+from ..utils import is_pinned_requirement, key_from_req
 from ..writer import OutputWriter
-
-# Make sure we're using a compatible version of pip
-assert_compatible_pip_version()
 
 DEFAULT_REQUIREMENTS_FILE = 'requirements.in'
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Installation
 ============
 
 ```console
-$ pip install --upgrade pip  # pip-tools needs pip==6.1 or higher (!)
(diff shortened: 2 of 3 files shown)


piptools/scripts/compile.py
```diff
@@ -8,20 +8,21 @@
 import tempfile
 
 import pip
+
+# Make sure we're using a compatible version of pip
+from ..utils import assert_compatible_pip_version
+assert_compatible_pip_version()
```
alexwlchan: There is still a possibility of error if an `ImportError` is thrown in `utils.py`, but the fix for that is more destructive, and would mean moving this function into its own file. I don’t mind doing that, but I’ll wait to see if reviewers think that would be necessary.

Issue #380, opened 2016-07-18
--trusted-host causes pip-sync to fail


Groxx:
pip-compiling with `pip-tools==1.7.0` produces a file with:

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --no-index --output-file tmp_all_requirements.txt requirements.txt requirements-docs.txt requirements-dev.txt requirements-test.txt
#
--trusted-host pypi.local.internal
--trusted-host archive.local.internal
...
```

which then fails to sync with an error:

```
Usage: pip-sync [options]

Traceback (most recent call last):
  File "/mnt/jenkins/workspace/test-irslib/env/bin/pip-sync", line 11, in <module>
    sys.exit(cli())
  File "/mnt/jenkins/workspace/test-irslib/env/local/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/mnt/jenkins/workspace/test-irslib/env/local/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/mnt/jenkins/workspace/test-irslib/env/local/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/mnt/jenkins/workspace/test-irslib/env/local/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/mnt/jenkins/workspace/test-irslib/env/local/lib/python2.7/site-packages/piptools/scripts/sync.py", line 52, in cli
    requirements = sync.merge(requirements, ignore_conflicts=force)
  File "/mnt/jenkins/workspace/test-irslib/env/local/lib/python2.7/site-packages/piptools/sync.py", line 70, in merge
    for ireq in requirements:
  File "/mnt/jenkins/workspace/test-irslib/env/local/lib/python2.7/site-packages/pip/req/req_file.py", line 89, in parse_requirements
    for req in req_iter:
  File "/mnt/jenkins/workspace/test-irslib/env/local/lib/python2.7/site-packages/pip/req/req_file.py", line 119, in process_line
    opts, _ = parser.parse_args(shlex.split(options_str), defaults)
  File "/usr/lib/python2.7/optparse.py", line 1402, in parse_args
    self.error(str(err))
  File "/usr/lib/python2.7/optparse.py", line 1584, in error
[...]

PR #430, opened 2016-12-28
Add trusted host option to pip-sync

stBecker:
Possible fix for #380


davidovich:
pip will honour the flags output by pip-compile, including --trusted-host. This PR is not needed anymore since pip 8.0 can read the flag.

See: https://pip.pypa.io/en/stable/news/

PR #437, opened 2017-01-17
Ignore pkg-resources Fixes #389

### LINKED ISSUE #389: pip-sync removes pkg-resoources
##### Steps to replicate

``` shell
$ cat requirements-dev.txt 
#
# To update, run:
#
#    pip-compile --output-file requirements-dev.txt requirements-dev.in
#
click==6.6                # via pip-tools
first==2.0.1              # via pip-tools
mypy-lang==0.4.4
pip-tools==1.7.0
pluggy==0.3.1             # via tox
py==1.4.31                # via tox
six==1.10.0               # via pip-tools
tox==2.3.1
virtualenv==15.0.3        # via tox
wheel==0.29.0

# The following packages are commented out because they are
# considered to be unsafe in a requirements file:
# pip
# setuptools
```
##### Expected result

``` shell
```
##### Actual result

``` shell
$ pip-sync requirements-dev.txt 
Uninstalling pkg-resources-0.0.0:
  Successfully uninstalled pkg-resources-0.0.0
```

graingert:


piptools/sync.py
```diff
@@ -13,6 +13,7 @@
     'pip-review',
     'setuptools',
     'wheel',
+    'pkg-resources',
```
das-g: Should this be marked as a workaround for an [alleged Ubuntu bug](https://bugs.launchpad.net/ubuntu/+source/python-pip/+bug/1635463) or would ignoring `pkg-resources` be a good idea even independent of that?

graingert: probably to ignore all the blacklisted packages from PyPI

vphilippon:
Done with PR #555

PR #444, opened 2017-01-28
Catch CalledProcessError when pip install/uninstall is run

crdoconnor:


piptools/sync.py
```diff
@@ -13,6 +13,7 @@
     'pip-review',
     'setuptools',
     'wheel',
+    'pkg-resources',
```
davidovich: pkg-resources does not seem to be a pypi package, I believe this should be removed from the PR because it is also irrelevant to the change.

vphilippon:
Nothing happened here for a good while. I'll go ahead an close.
Also, if anyone would actually need this, please indicate how/why that's needed.

crdoconnor:
Catching the subprocess error was to stop it from spewing ugly and unnecssary stacktrace for decidedly non-exceptional behavior.

I included pkg-resources because it was a workaround that prevented the tool from being completely broken on ubuntu.

I've since switched to pipenv though, so, moot points.

Issue #431, opened 2016-12-29
pip-compile forcefully converts package names to lowercase on second run


IlyaSemenov:
On a first run, `pip-compile` creates `requirements.txt` where package names are properly cased. However on subsequent runs, it converts them all to lower case.

##### Steps to replicate

Please see the full log below:

```console
semenov@daedra ~ $ pyenv virtualenv 3.5.2 piptools-test                                                                                         "53e50feeIgnoring indexes: https://pypi.python.org/simple
Requirement already satisfied (use --upgrade to upgrade): setuptools in /Users/semenov/.pyenv/versions/3.5.2/envs/piptools-test/lib/python3.5/site-packages
Requirement already satisfied (use --upgrade to upgrade): pip in /Users/semenov/.pyenv/versions/3.5.2/envs/piptools-test/lib/python3.5/site-packages
semenov@daedra ~ $ mkdir -p tmp/piptools-test
semenov@daedra ~ $ cd tmp/piptools-test/
semenov@daedra ~/tmp/piptools-test $ pyenv local piptools-test
(piptools-test) semenov@daedra ~/tmp/piptools-test $ pip install pip-tools
Collecting pip-tools
  Using cached pip_tools-1.8.0-py2.py3-none-any.whl
Collecting first (from pip-tools)
  Using cached first-2.0.1-py2.py3-none-any.whl
Collecting six (from pip-tools)
  Using cached six-1.10.0-py2.py3-none-any.whl
Collecting click>=6 (from pip-tools)
  Using cached click-6.6-py2.py3-none-any.whl
Installing collected packages: first, six, click, pip-tools
Successfully installed click-6.6 first-2.0.1 pip-tools-1.8.0 six-1.10.0
You are using pip version 8.1.1, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
(piptools-test) semenov@daedra ~/tmp/piptools-test $ echo 'Django>=1.9,<1.10' > requirements.in
(piptools-test) semenov@daedra ~/tmp/piptools-test $ pip-compile
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file requirements.txt requirements.in
#
Django==1.9.12
(piptools-test) semenov@daedra ~/tmp/piptools-test $ pip-compile
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file requirements.txt requirements.in
#
django==1.9.12
```

##### Expected result

[...]

PR #452, opened 2017-02-09
Fix capitalization

klaplong:
Fixed https://github.com/nvie/pip-tools/issues/431

### DIFF
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -36,7 +36,7 @@ def test_format_requirement_annotation(from_line, writer):
             'test==1.2                 ' + comment('# via xyz'))
 
 
-def test_format_requirement_annotation_case_sensitive(from_line, writer):
+def test_format_requirement_annotation_lower_case(from_line, writer):
     ireq = from_line('Test==1.2')
     reverse_dependencies = {'test': ['xyz']}
 
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -43,7 +43,7 @@ def test_format_requirement_annotation_case_sensitive(from_line, writer):
     assert (writer._format_requirement(ireq,
                                        reverse_dependencies,
                                        primary_packages=[]) ==
-            'Test==1.2                 ' + comment('# via xyz'))
+            'test==1.2                 ' + comment('# via xyz'))
 
 
 def test_format_requirement_not_for_primary(from_line, writer):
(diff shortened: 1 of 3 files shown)


barrywhart:
Any idea why the build is failing? I can't access the Travis log.

davidovich:
The project has moved to the jazzband organization and it seems that past builds have not followed from nvie. 

Pushing another commit (or amending the last commit) should trigger a build in the new organization.

rpkilby:
If you change the url to jazzband from nvie, it should show the build.

davidovich:
https://travis-ci.org/jazzband/pip-tools/builds/199895230

Note however that the "This branch has no conflicts with the base branch" message is flawed here. Rebasing on master should help.

barrywhart:
This looks like a good fix for the capitalization behavior. In the original issue (https://github.com/jazzband/pip-tools/issues/431), I had raised some questions about some possible cache-related weirdness. Should I create another GitHub issue for that? Either way (new issue or no), I will need to go back to try and reproduce the issue.

klaplong:
Yes I have read a lot of code, but I could not really find where the caching actually takes place. This seems to be part of another package.

tests/test_writer.py
```diff
@@ -43,7 +43,7 @@ def test_format_requirement_annotation_case_sensitive(from_line, writer):
     assert (writer._format_requirement(ireq,
                                        reverse_dependencies,
                                        primary_packages=[]) ==
-            'Test==1.2                 ' + comment('# via xyz'))
```
davidovich: This test is made to explicitly validate that case is NOT changed. (test is named test_format_requirement_annotation_case_sensitive). Are we sure this is ok ?

klaplong: Case _sensitivity_ is not the same as _unchanged_ case. On the other hand, neither pip, nor pypi, nor pip-tools (as it uses pip itself) are case sensitive. Maybe it's better to either just get rid of this test, or rename it, or even write another one?

davidovich: Yes that makes sense. 

Going forward, I think adding and maintaining tests is really important, particularly in this new "distributed" maintainership model (jazzband) that pip-tools is using. I know it is more work, but it is better to spend time locally than distribute the debugging effort on the internet :-). 

Maybe this should be included in the 1.8.1 release I am preparing.  A release candidate 1 is out already.

klaplong:
I do have a feeling we should document this case-insensitivity somewhere, although I have no idea where.

davidovich:
Thanks @klaplong !

PR #458, opened 2017-02-28
Remove special casing of pip specifier attribute.

davidovich:
The use of req.specs in the resolver causes the resolver to not
stabilize.

This happens on long lists of subdependency. The old conditionnal was
introducing discrepancies in the comparison function, under pip < 8.1.2.

This change removes the pip version dependant check and
unconditionnally uses req.specifier. Now that pip-tools only supports
pip > 8.0, and that this attribute is present in pip
>= 8.0, the special case is not needed.

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -36,12 +36,7 @@ class RequirementSummary(object):
         self.req = req
         self.key = key_from_req(req)
         self.extras = str(sorted(req.extras))
-        if hasattr(req, 'specs'):
-            # pip < 8.1.2
-            self.specifier = str(req.specs)
-        else:
-            # pip >= 8.1.2
-            self.specifier = str(req.specifier)
+        self.specifier = str(req.specifier)
 
     def __eq__(self, other):
         return str(self) == str(other)
--- a/tests/fixtures/fake_package/setup.py
+++ b/tests/fixtures/fake_package/setup.py
@@ -1,0 +1,22 @@
+from setuptools import setup
+
+setup(
+    name='fake_with_deps',
+    version=0.1,
+    install_requires=[
+        "python-dateutil>=2.4.2,<2.5",
+        "colorama<0.4.0,>=0.3.7",
+        "cornice<1.1,>=1.0.0",
+        "enum34<1.1.7,>=1.0.4",
+        "six>1.5,<=1.8",
+        "futures<3.1,>=3.0.3",
+        "ipaddress<1.1,>=1.0.16",
+        "jsonschema<3.0,>=2.4.0",
+        "pyramid<1.6,>=1.5.7",
+        "pyzmq<14.8,>=14.7.0",
+        "simplejson>=3.5,!=3.8,>3.9",
+        "SQLAlchemy!=0.9.5,<2.0.0,>=0.7.8,>=1.0.0",
+        "python-memcached>=1.57,<2.0",
+        "xmltodict<=0.11,>=0.4.6",
+        ],
+)
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,5 +1,6 @@
 import os
 from textwrap import dedent
+import subprocess
 
 from click.testing import CliRunner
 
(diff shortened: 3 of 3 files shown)


davidovich:
@jmbowman Hi, could you have a look at this PR, I believe you were the original author which introduced the conditional. 

This happens on 1.8.1rc1, but is probably present in earlier versions.

The problem is that the req.specs and req.specifier are sometimes not equivalent. Here is a snapshot of a RequirementSummary() construction:

```
self.key: u'sqlalchemy'
str(req.specifier): '!=0.9.5,<2.0.0,==1.0.8,>=0.7.8,>=1.0.0'
srt(req.specs): "[(u'!=', u'0.9.5'), (u'>=', u'0.7.8')]"
```

Clearly, these are not equivalent, and produce unstable resolving.

What do you think? I would really like to ship a 1.8.1 ASAP.

davidovich:
Also tested manually on pip 8.0, 8.1.1 and 9.0.1

jmbowman:
The condition was needed for pip versions less than 8.1.2, since versions before that got their requirement objects from `pkg_resources` which has a `specs` attribute, and ones after that get `specifier` from objects created by `packaging`; you can see the switch at https://github.com/pypa/pip/pull/3307.  I'm not sure why you're seeing both attributes at the same time; none of the code in the master branches of `pip` or `packaging` seems to set or reference a `specs` attribute.  Is the source code for your problem case available?  I'm a little worried that dropping the condition may break with pip between 8.0 and 8.1.1 in cases you haven't tested.

davidovich:
You made me doubt this for a moment. But the `req` does have a specifier **and** a specs, it is a pkg_resources.Requirement.parse(req) object: https://github.com/pypa/pip/blob/8.0.0/pip/_vendor/pkg_resources/__init__.py#L2990-L2993

When I remove the conditional, the resolver stabilises much faster, and does not break on my example requirement topology (which is admittedly a local set, but is bound to happen again in the wild).

I have setup a local testing rig with the 1.8.1rc1 version. This minimal setup exhibits the breakage:

```
cd pip-tools
mkdir -p localtests/fake_package
cd localtests
```
save the following setup.py to fake_package/setup.py
```
from setuptools import setup

setup(
    name='fake_with_deps',
    version=0.1,
    install_requires=[
        "python-dateutil>=2.4.2,<2.5",
        "colorama<0.4.0,>=0.3.7",
        "cornice<1.1,>=1.0.0",
        "enum34<1.1,>=1.0.4",
        "functools32<3.3,>=3.2.3",
        "futures<3.1,>=3.0.3",
        "ipaddress<1.1,>=1.0.16",
        "jsonschema<3.0,>=2.4.0",
        "psutil<3.2,>=3.1.1",
        "pyramid<1.6,>=1.5.7",
        "pyzmq<14.8,>=14.7.0",
        "SQLAlchemy==1.0.8",
        "python-memcached==1.57",
        "xmltodict<=0.11,>=0.4.6",
        ],
)
```
and invoke compile
```
rm -f fake_package/fake_with_deps-0.1-py2-none-any.whl && pip wheel --no-deps -w fake_package/ ./fake_package/ && pip-compile --rebuild -v -n -f fake_package/
```

Note that this PR makes the resolver converge very fast on this example (5 rounds).

cjauvin:
I have been studying what is proposed, and it makes sense to me. Basically it boils down to:

1. `req.specifier` is better/more stable than `req.specs` (it's the one we want to use) 
2. It was clearly introduced in pip 8.0.0, as the source shows

Since pip-tools only supports pip > 8.0.0, then it makes sense to remove any reference to `req.specs`, and forget that it even exists!

jmbowman:
Ok, I'd been looking at when `specs` was removed, not when `specifier` was added; I hadn't realized there was a sequence of versions where both were present.  You're right in that dropping support for pip < 8, we no longer need this condition.  Note that there's a virtually identical condition in LocalRequirementsRepository that should also be removed: https://github.com/jazzband/pip-tools/blob/master/piptools/repositories/local.py#L14 .

davidovich:
@jmbowman Thanks for the info, I'll be removing the duplicated condition.

PR #461, opened 2017-03-03
Add support for running as Python module

### LINKED ISSUE #402: Add support for `pythonX.Y -m piptools ...`.
I work with both Python 2.7 and Python 3.5 regularly.  The dependency resolution process uses the Python version in multiple different ways and I'm afraid that running with both might not give the same result depending on the Python version.

For this reason, I don't have a global alias to `pip-compile` _et al._.  I could of course use a virtualenv just for running piptools, but that's quite inconvenient (especially for projects that don't use Tox).   What I do have is a global alias to `pythonX.Y` (Linux) and `py -X.Y` (Windows).

The most convenient way to use pip-tools in this context is to support the `python -m piptools ...` CLI syntax, which requires a `piptools.__main__` module.

That way, when I work on a Python 2.7 project:

```
py -2.7 -m piptools compile requirements.txt  # compiles for Python 2.7
```

And when I switch to a Python 3.5 project:

```
py -3.5 -m piptools compile requirements.txt  # compiles for Python 3.5
```

Are you open to such a change if I'm willing to put some time into it?

Thanks!

AndreLouisCaron:
This change makes it easier to run pip-tools on systems which use multiple Python versions by enabling workflows like `py -X.Y -m piptools ...` on Windows and `pythonX.Y -m piptools ...` elsewhere.

Fixes #402.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,10 @@ $ pip-compile --upgrade-package flask --upgrade-package requests  # update both 
 $ pip-compile -P flask -P requests==2.0.0  # update the flask package to the latest, and requests to v2.0.0
 ```
 
+If you use multiple Python versions, you can run ``pip-compile`` as ``py -X.Y
+-m piptools compile ...`` on Windows and ``pythonX.Y -m piptools compile ...``
+on other systems.
+
 Example usage for `pip-sync`
 ============================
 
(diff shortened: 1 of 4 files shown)


davidovich:
Thanks @AndreLouisCaron !

PR #464, opened 2017-03-05
sync: Use options from the txt file

suutari-ai:
The requirements.txt file might have index or find-links options which
affects finding of the installable packages.  Parse those options from
the txt file before doing the sync.

Utilize the pip options parsing code from the compile command which is
now moved to a separate file.

### DIFF
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -11,6 +11,7 @@ from .. import click, sync
 from ..exceptions import PipToolsError
 from ..logging import log
 from ..utils import assert_compatible_pip_version, flat_map
+from ._repo import get_pip_options_and_pypi_repository
 
 # Make sure we're using a compatible version of pip
 assert_compatible_pip_version()
@@ -47,8 +48,15 @@ def cli(dry_run, force, find_links, index_url, extra_index_url, no_index, quiet,
             log.error('ERROR: ' + msg)
             sys.exit(2)
 
-    requirements = flat_map(lambda src: pip.req.parse_requirements(src, session=True),
-                            src_files)
+    (pip_options, repository) = get_pip_options_and_pypi_repository(
+        index_url=index_url, extra_index_url=extra_index_url,
+        no_index=no_index, find_links=find_links)
+
+    def parse_req_file(filename):
+        return pip.req.parse_requirements(
+            filename, session=True, finder=repository.finder)
+
+    requirements = flat_map(parse_req_file, src_files)
 
     try:
         requirements = sync.merge(requirements, ignore_conflicts=force)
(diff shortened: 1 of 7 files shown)


vphilippon:
(Catching up)
The feature looks good to me.
This would need a rebase on master and some tests too, and that should be good.

suutari:
Rebased and added some tests too.

piptools/scripts/sync.py
```diff
@@ -60,15 +68,15 @@ def cli(dry_run, force, find_links, index_url, extra_index_url, no_index, quiet,
     to_install, to_uninstall = sync.diff(requirements, installed_dists)
 
     install_flags = []
-    for link in find_links or []:
+    for link in repository.finder.find_links or []:
         install_flags.extend(['-f', link])
-    if no_index:
```
davidovich: With this change the no_index flag is inhibited. Is this wanted?

piptools/scripts/sync.py
```diff
@@ -60,15 +68,15 @@ def cli(dry_run, force, find_links, index_url, extra_index_url, no_index, quiet,
...
+    if not repository.finder.index_urls:
         install_flags.append('--no-index')
-    if index_url:
-        install_flags.extend(['-i', index_url])
-    if extra_index_url:
-        for extra_index in extra_index_url:
-            install_flags.extend(['--extra-index-url', extra_index])
+    for (i, index_url) in enumerate(repository.finder.index_urls):
```
davidovich: This is not quite the same intent. I see that you are reconstructing the parameter invocation based upon the index_urls, but that exposes too much of the implementation detail. I don't quite grasp why you are doing this.

piptools/scripts/sync.py
suutari: The `no_index` flag is passed to the `get_pip_options_and_pypi_repository` call which then passes it along to the `finder` in the constructed repository. That'll make the `repository.finder.index_urls` to be empty if `no_index` is `True`. It'll also consider `--no-index` specified in the parsed txt file.

piptools/scripts/sync.py
suutari: The point here is that the call to `pip.req.parse_requirements` above (line 56) will update the flags in the `repository.finder` (which is passed to it) according to the flags specified in the txt file. Therefore the `repository.finder.index_urls` will contain combination of command line options and txt file options and this is then used here to construct the options given to `pip install`.

suutari: I don't understand what you mean by the "implementation detail", since there wasn't really any separation between implementation and interface here before my chagnes either. If implementation details should be hidden, then there should be a class like `RequirementsFileParser` which then could be used to parse the txt files and as a result would output the list of requirements and pip options from the parsed files. That would be nice improvement, but completely another issue.

Can you elaborate what do you mean by exposing the implementation detail here?

atugushev:
Superseded by #824.

Issue #316, opened 2016-02-05
Running without --update duplicates --extra-index-url lines


Chronial:
If there is an --extra-index-url line at the beginning of the `requirements.in`, it will be duplicated in the `requirements.txt` on each run of `pip-compile` without the `--update` switch.

jbbarth:
Confirmed here. It was not the case in pip-tools 1.4.4, but I see this now with pip-tools 1.6. In my case the "--extra-index" option sits in `~/.pip/pip.conf` for what is worth, so not specific to the `requirements.in` parsing.

For what is worth everything still works normally despite of this bug, so not a blocker.

nvie:
Can you give an example *.in file that does not have the problem and one that does? Thanks!

Chronial:
It's as simple as it sounds: 
**Good**

```
chardet
```

**Bad**
```
--extra-index-url https://pypi.python.org
chardet
```

hugoboos:
Same here, with version 1.6

I have one `--extra-index-url <url>` line at the top of my `requirements.in`. On every run of `pip-compile` the same `--extra-index-url <url>` is added to `requirements.txt`.

yakky:
Confirmed here
Maybe just extract index_urls from finder beyond this point https://github.com/nvie/pip-tools/blob/master/piptools/scripts/compile.py#L132 ?
Tested with pip >= 7
@nvie I can work a on PR if a solution is agreed

mchristopher:
@yakky @nvie any update on this? I'm running into this problem as well.

anthrotype:
I have the same issue with --extra-index-url occurring multiple times, when I include a pip-compile-generated requirements.txt inside my requirements.in file, e.g. by adding `-r some_requirements.txt` at the top of it.

davidovich:
Closed by #468

PR #468, opened 2017-03-10
remove duplicate index urls

majuscule:
closes issue https://github.com/jazzband/pip-tools/issues/316

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -144,7 +144,7 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
 
     log.debug('Using indexes:')
     # remove duplicate index urls before processing
-    repository.finder.index_urls = set(repository.finder.index_urls)
+    repository.finder.index_urls = sorted(set(repository.finder.index_urls))
     for index_url in repository.finder.index_urls:
         log.debug('  {}'.format(index_url))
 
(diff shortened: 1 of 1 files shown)


piptools/scripts/compile.py
```diff
@@ -143,6 +143,8 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
         repository = LocalRequirementsRepository(existing_pins, repository)
 
     log.debug('Using indexes:')
+    # remove duplicate index urls before processing
+    repository.finder.index_urls = sorted(set(repository.finder.index_urls))
```
davidovich: Thanks for your contribution!

As it is, this has the possibility of changing the index-url order, which is a property that needs to be maintained.

As a user, I might use index urls to explicitly set desired index usage ( A -> B -> C ).

See http://stackoverflow.com/a/39835527/28275 for a clean reference implementation that keeps order.

majuscule: That's fine by me. I have pushed an update with the collections implementation as per your link.

davidovich:
Thanks for this @majuscule

majuscule:
Happy to help! If you have time, I would appreciate your thoughts on my comment to you here: https://github.com/jazzband/pip-tools/pull/418#issuecomment-285823330.

PR #469, opened 2017-03-11
Fix doctests and run them with tox

suutari-ai:
There's some tests in the sources that are not utilized.  Let's enable them.


suutari:
Rebased this on to current master. Should be good for merging. Any reviewers?

suutari:
In other Python projects I know fixtures are usually *code* that create some object instances for testing. The reason for the change of that name is that a directory named `fixtures` could also contain some Python code which could have tests/doctests and therefore shouldn't be ignored by pytest, but the current contents of the fixtures directory seems to be only data which can be ignored, and the `setup.py` files there *must* be ignored.

Anyway, now that I look more carefully, it seems that there is also some other data than the "fake PyPI", so you're right, `fake_pypi` isn't good choice for the name. How about `test_data` then?

davidovich:
Yes I think test_data is a good compromise.

suutari:
@davidovich How about now?

davidovich:
Thanks @suutari

PR #472, opened 2017-03-22
Added --max-rounds argument to pip-compile

derek-miller:
With large constraint/requirement sets the default of 10 rounds is not enough to reach a stable solution.


davidovich:
Thanks @derek-miller !

Issue #466, opened 2017-03-09
Second-level requirements for editable reqs disappear after round 1


voidlily:
In 1.8.0, running `pip-compile` would include requirements from editable reqs in the output file.

This is a possible regression introduced by 444fba5, when the multi-round requirement calculation was introduced.

##### Environment Versions

1. OS Type - linux
1. Python version: 3.6.0
1. pip version:  9.0.1
1. pip-tools version: 1.8.1rc3

##### Steps to replicate

requirements.in:

```
-e git+git://github.com/pyca/cryptography@master#egg=cryptography
```

##### Expected result

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file requirements.txt requirements.in
#
-e git+git://github.com/pyca/cryptography@master#egg=cryptography
appdirs==1.4.3            # via setuptools
asn1crypto==0.21.1
cffi==1.9.1
idna==2.5
packaging==16.8           # via setuptools
pycparser==2.17           # via cffi
pyparsing==2.2.0          # via packaging
six==1.10.0               # via packaging, setuptools

# The following packages are considered to be unsafe in a requirements file:
# setuptools
```

(Note that this includes setuptools reqs as in #445, but unrelated to the issue at hand )

##### Actual result

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file requirements.txt requirements.in
#
-e git+git://github.com/pyca/cryptography@master#egg=cryptography
```

#### Debug output

```
Using indexes:
  https://pypi.python.org/simple

                          ROUND 1                           
Current constraints:
  cryptography from git+git://github.com/pyca/cryptography@master#egg=cryptography (from -r requirements.in (line 1))

Finding the best candidates:
  found candidate -e git+git://github.com/pyca/cryptography@master#egg=cryptography (constraint was <any>)

Finding secondary dependencies:

New dependencies found in this round:
  adding ['asn1crypto', '>=0.21.0', '[]']
  adding ['cffi', '>=1.4.1', '[]']
  adding ['idna', '>=2.1', '[]']
  adding ['packaging', '', '[]']
  adding ['setuptools', '>=11.3', '[]']
  adding ['six', '>=1.4.1', '[]']
Removed dependencies in this round:
------------------------------------------------------------
Result of round 1: not stable

                          ROUND 2                           
Current constraints:
  asn1crypto>=0.21.0
  cffi>=1.4.1
[...]

PR #476, opened 2017-03-23
Keep second-level requirements for editable reqs after round one

mattlong:
This is a straw-man proposal to address #466 based on my comments there.

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -57,6 +57,11 @@ class PyPIRepository(BaseRepository):
         # project
         self._available_candidates_cache = {}
 
+        # stores InstallRequirement => list(InstallRequirement) mappings
+        # of all secondary dependencies for the given requirement, so we
+        # only have to go to disk once for each requirement
+        self._dependencies_cache = {}
+
         # Setup file paths
         self.freshen_build_caches()
         self._download_dir = os.path.join(CACHE_DIR, 'pkgs')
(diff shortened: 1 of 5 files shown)


piptools/repositories/pypi.py
```diff
@@ -125,18 +130,20 @@ def get_dependencies(self, ireq):
...
-            os.makedirs(self._wheel_download_dir)
-
-        reqset = RequirementSet(self.build_dir,
-                                self.source_dir,
-                                download_dir=self._download_dir,
-                                wheel_download_dir=self._wheel_download_dir,
-                                session=self.session)
-        dependencies = reqset._prepare_file(self.finder, ireq)
-        return set(dependencies)
+        if ireq not in self._dependencies_cache:
```
mattlong: It might be better to key this cache off some derived key of ireq rather than using ireq itself, but I was unclear what would be the optimal key to use. Open to suggestions!

davidovich:
I like the relatively simple approach of caching the dependency, and it fixes in an unobtrusive manner the dependencies of an editable. 

Can you put up a test case so we can catch this in the future (preferably in a commit of the failing situation and then the commit fix)?

For the test case, you could probably use **-e** on the fake_package in tests/fixtures.

piptools/repositories/pypi.py
davidovich: I don't see a problem for this right now, but I am not the authority on pip-tools code either.

davidovich:
@mattlong For your info, I would like to make a 1.8.2 release to fix this regression. The only thing missing is tests. :-)

mattlong:
@davidovich Great! Will try to get a test case in today!

piptools/scripts/compile.py
```diff
@@ -93,16 +93,7 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
     # Setup
     ###
 
-    # Use pip's parser for pip.conf management and defaults.
```
mattlong: Extracted this bit of code into a separate function to facilitate testing

davidovich:
@mattlong outstanding! thank you!
fixes #466

majuscule:
Awesome! Thanks from me also @mattlong, started working on adding the tests myself and was getting very confused.

PR #479, opened 2017-03-23
add requirement for minimal version of setuptools

vadimg:
This tools doesn't work with old verisons of setuptools, due to an old version of the pkg_resources library.

Here's an error I get after running: `sudo pip install --upgrade setuptools==20.1.1`

```
  File "/usr/local/bin/pip-compile", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 895, in invoke
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
  File "/usr/local/lib/python2.7/dist-packages/piptools/scripts/compile.py", line 223, in cli
    reverse_dependencies = resolver.reverse_dependencies(results)
  File "/usr/local/lib/python2.7/dist-packages/piptools/resolver.py", line 273, in reverse_dependencies
    return self.dependency_cache.reverse_dependencies(non_editable)
  File "/usr/local/lib/python2.7/dist-packages/piptools/cache.py", line 139, in reverse_dependencies
    return self._reverse_dependencies(ireqs_as_cache_values)
  File "/usr/local/lib/python2.7/dist-packages/piptools/cache.py", line 163, in _reverse_dependencies
    for name, version_and_extras in cache_keys
  File "/usr/local/lib/python2.7/dist-packages/piptools/utils.py", line 192, in lookup_table
    for value in values:
  File "/usr/local/lib/python2.7/dist-packages/piptools/cache.py", line 164, in <genexpr>
    for dep_name in self.cache[name][version_and_extras])
  File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 3047, in parse
    req, = parse_requirements(s)
  File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2991, in parse_requirements
    "version spec")
  File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2967, in scan_list
    raise RequirementParseError(msg, line, "at", line[p:])
pkg_resources.RequirementParseError: Expected ',' or end-of-list in funcsigs>=1; python_version < "3.3" at ; python_version < "3.3"
```

With the next available version of setuptools (20.2.2), it works normally.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ setup(
         'click>=6',
         'first',
         'six',
+        'setuptools>=20.2.2',
     ],
     zip_safe=False,
     entry_points={


setup.py
```diff
@@ -17,6 +17,7 @@
         'click>=6',
         'first',
         'six',
+        'setuptools>=20.2.2',
     ],
```
graingert: it's bad to depend on setuptools: https://github.com/pypa/setuptools/blob/master/CHANGES.rst#v3400

graingert:
Might be better to use packaging https://packaging.pypa.io/en/latest/requirements/ instead, which is safe to depend on.

vphilippon:
This is something that has been in the back of my head for a while, but I'm not sure it's that straightforward.

By depending on the internals of `pip`, and not being able to specify the required `pip` version, we end up depending on `setuptools` in the cases where the version of `pip` installed uses `pkg_resources` instead of `packaging` (see that function `key_from_req` we have in `utils.py`).

I'm worried about the effect of setting a lower bound on `setuptools`, especially for users with older versions of `pip` that might not be able to update `setuptools`.

The only "easy" way I see out of this is when we'll drop the support of old `pip` versions, and we could then move to `packaging` and try to be as clean as possible. But as a lot of users use the version of `pip` distributed by their linux distribution, many of which are lagging behind.

Now if I missed something that actually makes all of this way easier than that, correct me, I'll be pleased.

vphilippon:
Closing this, as a simple lowerbound will not cut it.
If we end up vendoring `pip 9.0.1` we might be able to do something like that.

PR #481, opened 2017-03-27
pr418 testing

majuscule:
This branch builds on https://github.com/jazzband/pip-tools/pull/418 and adds a basic test case to ensure that a `setup.py` file is parsed when present and no requirements.txt is offered.

@davidovich


davidovich:
@majuscule can you rebase on master ? I added Windows testing, and I suspect that the `tempfile.NamedTemporaryFile()` usage will break (because of the lack of delete=False parameter). On Windows, the file is deleted at context manager exit, breaking subsequent processing.

majuscule:
Rebased and it does look like tests on Windows are failing. I'm afraid I don't quite follow what the error is here. In fact the only `tempfile.NamedTemporaryFile()` instance I see *does* use the delete=False parameter:

```
tests/test_cache.py
4:from tempfile import NamedTemporaryFile
22:        cache_file = NamedTemporaryFile(mode="w", delete=False)
```

Regardless that doesn't seem related to this error:

```
<Result InstallationError("Could not open requirements file: [Errno 13] Permis...ed: 'c:\\\\users\\\\appveyor\\\\appdata\\\\local\\\\temp\\\\1\\\\tmpg362ei'",)>
```

I do see that in the pip_conf fixture, it explicitly prepends a tmpdir to the written file, however my code takes place within the isolated_filesystem, and needs to be relative to the cli invocation.

Do you have any further ideas as to what is going wrong?

piptools/scripts/compile.py
```diff
@@ -158,12 +163,19 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
...
     for src_file in src_files:
-        if src_file == '-':
+        is_setup_file = os.path.basename(src_file) == 'setup.py'
+        if is_setup_file or src_file == '-':
             # pip requires filenames and not files. Since we want to support
             # piping from stdin, we need to briefly save the input from stdin
-            # to a temporary file and have pip read that.
+            # to a temporary file and have pip read that.  also used for
+            # reading requirements from install_requires in setup.py.
             with tempfile.NamedTemporaryFile(mode='wt') as tmpfile:
```
davidovich: This would need delete=False, especially on Windows.

majuscule:
I don't quite follow why that would be causing the error, but I appreciate your pointing it out. I've made the change and also removed it's use as a context manager which imo does imply it's auto-deletion. Might even be able to keep delete=True this way? I'm afraid I can't reproduce the error locally here without a windows machine to test on so will have to wait for AppVeyor.

majuscule:
Looks like it's all passing now. Please let me know if there's anything else I can do for this ticket or any of the others on the 1.9 milestone to get it released!

davidovich:
Good, thanks !

@majuscule In fact, if you have a little time, I want one bugfix for the 1.8.2 release, before 1.9, but @mattlong has not responded to add tests on https://github.com/jazzband/pip-tools/pull/476. Are you up for it ? :-)

davidovich:
Thanks @tysonclugg and @majuscule

PR #492, opened 2017-04-12
Improve setup.py parsing

myw:
Since requirements.txt should hold the reqs for the whole system, we should read from `setup_requires` _and_ `tests_require` when making the list of requirements.

Also avoid erroring out in an ugly if the user doesn't happen to have any requirements. Now, we throw a meaningful error if that's the case.

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -165,7 +165,15 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
             if is_setup_file:
                 from distutils.core import run_setup
                 dist = run_setup(src_file)
-                tmpfile.write('\n'.join(dist.install_requires))
+                reqs = (
+                  (dist.setup_requires or [])
+                  + (dist.install_requires or [])
+                  + (dist.tests_require or [])
+                )
+                if reqs:
+                    tmpfile.write('\n'.join(reqs))
+                else:
+                    raise PipToolsError('No requirements specified')
             else:
                 tmpfile.write(sys.stdin.read())
             tmpfile.flush()


davidovich:
I'm not certain that the default would be to include test_requires and setup_requires.

setup_requires list dependencies that is needed at install time only (and is managed solely by setuptools)
test_requires are not related to the runtime of the application.

myw:
but if requirements.txt represents the complete state of things in a
virtualenv, i think it should be nailing that stuff down. I agree that you
can make arguments either way about a reasonable set of defaults—i'm happy
to add flags for the other two so users can make their own calls.

but the functionality should be there, IMO

--
Misha

etiktin:
I think including `setup_requires` and `tests_require` is useful, especially for cases when you want to bundle all of your dependencies for offline installation (without `setup_requires` you can't install some components, e.g. `protobuf==2.6.1`).
That said, I agree it should be exposed through flags, since not everyone will need this.

vphilippon:
Doing `pip install` does not include the setup_requires and test_requires, so by default, I would suggest to not include those for sure.

> but if requirements.txt represents the complete state of things in a
> virtualenv, i think it should be nailing that stuff down. 

Here's the catch: tests_require and setup_requires dependencies are never installed in the virtualenv. Running `python setup.py test` does not install the packages in the virtualenv, in only puts them in a `.eggs` directory. It's the same for setup_requires.

The more I think about this, the more I feel like we should not bring that kind of behaviour, even as an option.

I'm definitly interested in clearer error message though!

myw:
My use case is inside a Docker container, not with `venv` on a regular system—in that scenario, it's very helpful to have one pinned-down requirements.txt file from which you install all dependencies before installing the app you are deploying.

A common pattern looks something like this:

```Dockerfile
COPY requirements.txt .
RUN pip install -r requirements.txt
COPY . ./
RUN pip install -e .  # or python setup.py develop 
```

Here, all of the requirements are installed in their pinned-down versions first, and then the second `pip install` doesn't actually install any additional packages (unless there are discrepancies between the `install_requires` field in `setup.py` and `requirements.txt`, but that's a bug).

The benefit of specifying setup.py as the single place for all abstract requirements  that generate the specific, pinned-down requirements in requirements.txt is that no data is duplicated (there's no need to add an additional `requirements.in` file), but your abstract dependencies cover all of the use cases you'd need, much like what would happen with `requirements.in`.

Now in that scenario, if I want to upgrade a package, `pip-compile -U foo --include_tests_require` will do it for me, regardless of whether it's a test framework or an install dependency. This is, imo, the way it should be.

I can definitely see why, given the venv behavior, enabling `tests`, `setup`, and potentially `extras` should _not_ be the default. But I still think adding those options is a valid use case for a Docker environment, which is now common in app deployment, and only growing more popular.

I won't push for adding it in if there's wide-spread consensus that even behind an option, this is the wrong direction. But I think I've outlined why it is useful, and I'll happily do the work to make this a full-fledged feature (the first commit was basically just a thought experiment).

merwok:
in that scenario, it's very helpful to have one pinned-down requirements.txt file from which you install all dependencies

Are there downsides to installing all dependencies from two requirement files?  In a previous project using docker, we would install only the runtime dependencies in the app container, and the test dependencies only in the test container that was built `FROM` the app container.

myw:
The need to create an entire new container for testing is not really the kind of constraint you want to impose on your users. In our use case, for example, this does not work.

Multiple requirement files can be OK, but the problem is that every `pip install` invocation is independent, so a downstream `pip install` could overwrite packages installed by the upstream one instead of resolving the of the requirements correctly down to a single, explicitly specified satisfactory version of all the packages you need, which is another big benefit of pip-tools.

Ideally, `pip-tools` would basically make `pip` work like `yarn`, with a single place that defines the abstract dependencies, which programmatically get resolved and converted into specific dependencies for a system. It almost does that. This potential feature, once implemented with the right defaults and cleaned up, could help it accomplish  that goal, without resorting to additional files for duplicate specification of dependencies (`setup.py` vs `requirements.in`).

tysonclugg:
@myw What you've put forward makes sense to me, so I'm +1 on adding options to support the behaviour you're requesting.  But certainly not as the default behaviour.

tysonclugg:
@myw If you're still interested, I suggest you create a new PR (based on current master) with options to support the use case you have described.

IvanAnishchuk:
I would like to revisit this issue, I think situation has slightly changed
since May and I think I could see the way for pip-compile to effectively
replace all other dependency management tools I tried with few simple changes.
I could open a new issue but reusing this seems appropriate. Tell me if it's
not.

Layering productions and dev requirements is something almost every project
needs but doesn't have. Competing tools like pipenv and poetry provide support
for it by default but there isn't yet a standard way to do it with pinned
requirements.txt files.

I think currently-recommended way is to produce two pinned files:
requirements.txt and requirements-dev.txt (with the latter usually including
everything from the former and adding some things) and there's a widely
adopted practice to use two separate in-files e.g. requirements.in and
dev-requirements.in with `-r requirements.in` in the latter one. It seems
vaguely consistent with implementation and purpose of `install_requires` and
`tests_require` and named `extras_require` in setup.py/setup.cfg (using an
extra named `dev` is a way recommended for dev requirements at least
[here](https://snarky.ca/clarifying-pep-518/))

I'd propose to add a way to support compiling multiple sets of requirements be
they multiple in-files like requirements.in, requirements-test.in, and
requirements-tests.in or multiple sections in setup.{cfg,py} or pyproject.toml
or some other project config format (let's not focus on particular formats for
now).

In my head, the ideal workflow should be like this:

1. User specify sets of requirements and _their order_ somehow (two easy ways
   are setup.cfg with install_requires, tests_require, extras_require and a
   number of in-files spicfied in pip-compile command line, both are ordered,
   it's important).

2. Some tool iterates over those sets and compiles each set of requirements
   into one resulting requirements-something.txt file (using in-file name or
   requirements.txt, requirements-tests.txt, requirements-<extra>.txt pattern)
   with added constraints for already-compiled files i.e. if we are compiling
   req.in, req-dev.in, and req-docs.in we would actually be compiling
   `-r req.in`, `-c req.txt -r req-dev.in`, and `-c req.txt -c rec-dev.txt -r
   rec-docs.in` respectively.

Does it make sense to build such functionality right into pip-compile or would
we be better off with having it in a separate tool (I mean one I'd support
myself)? I think it is possible to implement it in pip-compile without breaking
current default behaviour (unless we decide that we want to break it). Like,
perhaps, `-m` or `--multi` option that would enable generating multiple
resulting .txt files from multiple source .in files or multiple sections in
setup.cfg

Using `-c` means we don't duplicate requirements in generating unless there's
a reason to and the generated sets are all compatible with each other while
they don't necessarily include each other. For example, for compiling docs you
can theoretically install _only_ things in requirements-docs.txt without
installing all your project deps. And you can always install -r requirements.txt
-r requirements-tests.txt when your tests require project deps. Granular control
and consistent results, ain't that good?

I had a thought whether it makes sense to just add explicit `-c` lines to the
in-files themselves and not to change any code but that coould break if the
user doesn't compile files in the right order (results could be very
inconsistent). So, unless some robust automation wrapper is provided, I'd guess
it would break pretty often in this real world. So why not to include it
with pip-tools?

Are there any reasons not to do it like this? Any ideas of how to implement
requirements-layering better?

IvanAnishchuk:
Ah, I can't edit pull requests to switch forks :( I guess it should be separate ticket after all

PR #500, opened 2017-04-30
add ability to override compile command in generated req file

dschaller:


dschaller:
cc @nvie

dschaller:
@jazzband-bot

dschaller:
cc @davidovich

ulope:
This would be great. We're currently running sed on the generated files to change this comment and point to the right command.

davidovich:
This PR is missing some tests.

Issue #204, opened 2015-09-17
pip-compile absolute links when given relative path


ColtonProvias:
requirements.in:

```
-e .
-e ./sqlalchemy_jsonapi
```

compiles to:

```
-e file:///Users/coltonprovias/MyProject
-e file:///Users/coltonprovias/MyProject/sqlalchemy_jsonapi
```

It should compile to

```
-e .
-e ./sqlalchemy_jsonapi
```

It's not really useful for me to provide a requirements.txt file then to anybody unless they use a Mac and have the same name as me.  Although if that person does exist, I would like to meet them.

cdwilson:
I ran into the same issue today.  Why are relative paths are expanded?

michaeljones:
This has affected me too as I'm trying to install packages which are submodules of my project. It seems the issues arises here: https://github.com/pypa/pip/blob/develop/pip/req/req_install.py#L1131

I suspect it is hard to avoid as `pip-tools` uses the `pip` library and the `pip` library has no reason to keep a relative path around. It wants to deal with absolute paths.

Which means that it is unlikely that we'd get the `pip` library changed to help us so which suggests that we'd have to get the output from the `pip` library and go back and make any paths relative again which is an option but not incredibly desirable as it might be hard to know exactly when to do it and when not to.

Personally, I'm going to try switching from using submodules and relative paths to having `git+git://` paths in my `requirements.in` file but at the same time it seems very reasonable to try to fix this as it makes little sense to have absolute paths in a `requirements.txt` file when there is little chance that the paths in different development environments or the production environment is going to be the same. That is certainly what affects me. Of course, the `requirements.txt` should also be in git so it should be possible to `pip-compile` and then edit the `requirements.txt` before committing it.

breerly:
Looks like https://github.com/nvie/pip-tools/issues/325 is related

jredwards:
I've been planning to move all of the requirements handling across all of our repos to using pip-tools and just realized it had this issue, which may completely shelve that plan. This is a pretty significant problem.

orf:
So the core of this comes [down to this function here](https://github.com/pypa/pip/blob/2e534b0c923c96c6c3f7949e23df2573078cb9d7/pip/download.py#L470), inside `pip`. I don't think it would be easy to fix in `pip` itself, so I will see if I can manipulate the result while outputting it to the file by making the result relative.

ztane:
[...]

PR #507, opened 2017-05-12
Make editable requirements use relative paths where appropriate.

orf:
#204

When `pip-compile` is given a relative editable dependency, like `-e ./some_module/`, it is resolved to the *full absolute path* by `pip` under the hood. This is not what we want or need in most cases.

As the path resolution happens inside `pip` and it might be a bit complicated to add a change in that project, this merge request works around the issue by modifying the `format_requirement` to check that if the editable dependency uses the `file:` scheme and if the *location* of the editable dependency exists within a subdirectory **of the current working directory** then we format the dependency using the relative path **from the current working directory**.

I think this is the best place to add this change and requires less potential breakage and internal changes. I'm not too sure on if we should use the current directory *or* the path from the **output file directory**, but that's a small detail.

I also did a small refactoring of the tests by adding fixtures to return the paths of the packages within the `tests/fixtures/` directory rather than duplicating various `os.path.join` calls everywhere.

##### Contributor checklist

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -2,6 +2,7 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
+import os
 import sys
 from itertools import chain, groupby
 from collections import OrderedDict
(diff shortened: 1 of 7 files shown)


orf:
It's also common to have a requirements file in a directory like `project/requirements/{debug,prod}.in`. In this case the editable requirements are not in a sub-directory and so the current implementation wouldn't work.

I guess to cover this case we could just say 'it resolves to a relative path if it is in a sub-directory of the requirements file' and be done with it (providing we can fix the issue you described). People who need this fix can then move the files as needed to make it work?

piptools/utils.py
```diff
@@ -61,13 +62,30 @@ def make_install_requirement(name, version, extras, constraint=False):
...
 
 
+def is_subdirectory(base, directory):
+    """
+    Return True if directory is a child directory of base
+    """
+    base = os.path.join(os.path.realpath(base), '')
+    directory = os.path.join(os.path.realpath(directory), '')
+
+    return os.path.commonprefix([base, directory]) == base
```
merwok: From os.path docs: «Note that this may return invalid paths because it works a character at a time.»

Python 3.5 added os.path.commonpath to avoid that, which pip-tools may want to backport.

AndydeCleyre:
I'll just add my opinion that the sanest choice is to use the path relative to the input file.

PR #508, opened 2017-05-13
Fix allow unsafe false pinning bug

dschaller:
Fix a bug introduced in [441](https://github.com/jazzband/pip-tools/pull/441) that pinned unsafe dependencies in generated requirements files when `--allow-unsafe` was `False`.

##### Contributor checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Features:
 
 Bug Fixes:
 - Fixed bug where unsafe packages would get pinned in generated requirements files
-when `--allow-unsafe` was set. ([#508](https://github.com/jazzband/pip-tools/pull/508)).
+when `--allow-unsafe` was not set. ([#508](https://github.com/jazzband/pip-tools/pull/508)).
 
 # 1.9.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 1.10.0rc1
+Features:
+-
+
+Bug Fixes:
+- Fixed bug where unsafe packages would get pinned in generated requirements files
+when `--allow-unsafe` was set. ([#508](https://github.com/jazzband/pip-tools/pull/508)).
+
 # 1.9.0
 
 Features:
@@ -25,7 +33,7 @@ Bug Fixes:
 # 1.8.1
 
 - Recalculate secondary dependencies between rounds (#378)
-- Calculated dependencies could be left with wrong candidates when 
+- Calculated dependencies could be left with wrong candidates when
   toplevel requirements happen to be also pinned in sub-dependencies (#450)
 - Fix duplicate entries that could happen in generated requirements.txt (#427)
 - Gracefully report invalid pip version (#457)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,4 @@
-# 1.10.0rc1
-Features:
--
-
+# 1.9.1
 Bug Fixes:
 - Fixed bug where unsafe packages would get pinned in generated requirements files
 when `--allow-unsafe` was not set. ([#508](https://github.com/jazzband/pip-tools/pull/508)).
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -106,12 +106,14 @@ class OutputWriter(object):
             yield comment('# The following packages are considered to be unsafe in a requirements file:')
 
             for ireq in unsafe_packages:
-                line = self._format_requirement(ireq,
-                                               reverse_dependencies,
-                                               primary_packages,
-                                               include_specifier=self.allow_unsafe,
-                                               marker=markers.get(ireq.req.name),
-                                               hashes=hashes if self.allow_unsafe else None)
+                line = self._format_requirement(
+                    ireq,
+                    reverse_dependencies,
+                    primary_packages,
+                    include_specifier=self.allow_unsafe,
+                    marker=markers.get(ireq.req.name),
+                    hashes=hashes if self.allow_unsafe else None
+                )
                 if not self.allow_unsafe:
                     yield comment('# {}'.format(line))
                 else:
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -11,7 +11,8 @@ from .utils import comment, format_requirement, dedup, UNSAFE_PACKAGES
 class OutputWriter(object):
     def __init__(self, src_files, dst_file, dry_run, emit_header, emit_index,
                  emit_trusted_host, annotate, generate_hashes,
-                 default_index_url, index_urls, trusted_hosts, format_control):
+                 default_index_url, index_urls, trusted_hosts, format_control,
+                 allow_unsafe=False):
         self.src_files = src_files
         self.dst_file = dst_file
         self.dry_run = dry_run
@@ -24,6 +25,7 @@ class OutputWriter(object):
         self.index_urls = index_urls
         self.trusted_hosts = trusted_hosts
         self.format_control = format_control
+        self.allow_unsafe = allow_unsafe
 
     def _sort_key(self, ireq):
         return (not ireq.editable, str(ireq.req).lower())
@@ -96,7 +98,7 @@ class OutputWriter(object):
         for ireq in packages:
             line = self._format_requirement(
                 ireq, reverse_dependencies, primary_packages,
-                markers.get(ireq.req.name), hashes=hashes)
+                marker=markers.get(ireq.req.name), hashes=hashes)
             yield line
 
         if unsafe_packages:
@@ -104,12 +106,16 @@ class OutputWriter(object):
             yield comment('# The following packages are considered to be unsafe in a requirements file:')
 
             for ireq in unsafe_packages:
-
-                yield self._format_requirement(ireq,
+                line = self._format_requirement(ireq,
                                                reverse_dependencies,
                                                primary_packages,
+                                               include_specifier=self.allow_unsafe,
                                                marker=markers.get(ireq.req.name),
-                                               hashes=hashes)
+                                               hashes=hashes if self.allow_unsafe else None)
+                if not self.allow_unsafe:
+                    yield comment('# {}'.format(line))
+                else:
+                    yield line
 
     def write(self, results, reverse_dependencies, primary_packages, markers, hashes):
         with ExitStack() as stack:
@@ -124,9 +130,12 @@ class OutputWriter(object):
                     f.write(unstyle(line).encode('utf-8'))
                     f.write(os.linesep.encode('utf-8'))
 
-    def _format_requirement(self, ireq, reverse_dependencies, primary_packages, marker=None, hashes=None):
-        line = format_requirement(ireq, marker=marker)
+    def _format_requirement(self, ireq, reverse_dependencies, primary_packages, **kwargs):
+        include_specifier = kwargs.get('include_specifier', True)
+        marker = kwargs.get('marker')
+        line = format_requirement(ireq, include_specifier=include_specifier, marker=marker)
 
+        hashes = kwargs.get('hashes')
         ireq_hashes = (hashes if hashes is not None else {}).get(ireq)
         if ireq_hashes:
             for hash_ in sorted(ireq_hashes):
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -1,4 +1,4 @@
-from pytest import yield_fixture
+from pytest import fixture
 
 from pip.index import FormatControl
 from piptools.utils import comment, UNSAFE_PACKAGES
@@ -18,7 +18,7 @@ class Writer(object):
                             **kwargs)
 
 
-@yield_fixture
+@fixture
 def writer():
     yield Writer()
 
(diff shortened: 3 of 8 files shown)


dschaller:
cc @jdufresne

jdufresne:
Apologies if I have introduced a regression.

> Fixed bug where unsafe packages would get pinned in generated requirements files when `--allow-unsafe` was not set

Do you have additional details on this? I do not use the `--allow-unsafe` option and I do no have pinned unsafe packages in my output. So I'm curious how you triggered this. Do you have an example, minimal `requirements.in`? What is the command line you're using?

Here is the command I use:

```
$ ./venv/bin/pip-compile --version
pip-compile, version 1.9.0
$ pip-compile --upgrade --generate-hashes -o requirements.txt requirements.in
...
[no unsafe packages]
```

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 1.10.0rc1
+Features:
+-
+
+Bug Fixes:
+- Fixed bug where unsafe packages would get pinned in generated requirements files
+when `--allow-unsafe` was not set. ([#508](https://github.com/jazzband/pip-tools/pull/508)).
```
jonathanburns: I think you mean WAS set, right?

jdufresne:
If I run with `--allow-unsafe` setuptools it appears in my `requirements.txt` as: 

```
setuptools==35.0.2 \
    --hash=sha256:19d753940f8cdbee7548da318f8b56159aead279ef811a6efc8b8a33d89c86a4 \
    --hash=sha256:1e55496ca8058db68ae12ac29a985d1ee2c2483a5901f7692fb68fa2f9a250fd \
    # via html5lib
```

Are you saying this behavior is incorrect? setuptools should not have a pinned version?

davidovich:
I believe the initial implementation was correct in removing the unsafe packages at runtime. This PR introduces a similar kind of logic in the Output writer. I am on @jdufresne side here, and I am not sure there is a regression. Can you write a test that demonstrates the failure?

tests/test_writer.py
```diff
@@ -1,50 +1,56 @@
...
+from pytest import yield_fixture
 
 from pip.index import FormatControl
-from piptools.utils import comment
+from piptools.utils import comment, UNSAFE_PACKAGES
 from piptools.writer import OutputWriter
 
 
-@fixture
+class Writer(object):
```
davidovich: I don't understand this refactoring of correct code.

For what it's worth, pytest now doesn't discriminate between a yield and an ordinary fixture. @fixture is only needed.

dschaller:
@jdufresne no worries! I wanted to tag you in case I was missing a use case that isn't tested.

The command I'm running is `pip-compile --allow-unsafe --no-index -o piptools_requirements.txt` which correctly pins pip.
```
pip==9.0.1
```

But if I run `pip-compile --no-index -o requirements.txt` pip also gets pinned when it shouldn't.
```
pip==9.0.1
```

@davidovich not trying to imply there are "sides" here. Just wanted to fix a bug ;) happy to fork and fix in that.

tests/test_writer.py
dschaller: By yielding a class that I can invoke later I am allowing for keyword arguments to be passed in so that I can reuse this fixture while setting `allow_unsafe`.

I'll update the fixture annotation

CHANGELOG.md
dschaller: Unsafe packages were being pinned in requirements files even when `--allow-unsafe` wasn't set instead of being comments

jdufresne:
> But if I run `pip-compile --no-index -o requirements.txt requirements.in` pip also gets pinned when it shouldn't.

Thanks for following up. Can you provide the `requirements.in` file that triggers this? Or a minimal version of it?

dschaller:
@jdufresne 
```
pip==9.0.1
```

piptools/writer.py
```diff
@@ -124,9 +132,12 @@ def write(self, results, reverse_dependencies, primary_packages, markers, hashes
                     f.write(unstyle(line).encode('utf-8'))
                     f.write(os.linesep.encode('utf-8'))
 
-    def _format_requirement(self, ireq, reverse_dependencies, primary_packages, marker=None, hashes=None):
-        line = format_requirement(ireq, marker=marker)
+    def _format_requirement(self, ireq, reverse_dependencies, primary_packages, **kwargs):
```
davidovich: Keep the named arguments here.

dschaller: Can you provide a reason why?

jdufresne:
> pip==9.0.1

I'm having trouble understanding your use case. Can you expand on it?

Why do you put `pip` in your `requirements.in` if you don't want it in your `requirements.txt`?

What is the point to putting `pip` in a requirements file anyway. As you use pip to install requirements, it won't be updated until _after_ it is invoked.

dschaller:
@jdufresne I do a first pass with pip pinned and then a second pass including the generated file from the first pass.

Even if this seems odd the functionality is broken. If someone has a unsafe library in their requirements and `--allow-unsafe` is not set it should not be pinned.

ryan-lane:
@jdufresne I think the idea is that if an end-user puts pip into requirements.in, that we strip it unless `--allow-unsafe` is set. People sometimes put unsafe things into their .in file and we're ensuring that by default we try to do the right thing. The `--allow-unsafe` is there to let the end-user say "I know what I'm doing".

Basically, this PR just makes piptools work as advertised, based on the arguments.

davidovich:
The spirit of --allow-unsafe is to control if unsafe sub-dependencies are output to the requirements.txt, not to filter a user input which is the user desired state of wanted dependencies. requirements.in state what the user wants. The --allow-unsafe flag is to remove unwanted side effects of sub-dependencies which are `unsafe`. 

Even unsafe here is a bit hard to define. 

But if the said unsafe requirements are in the input which YOU control in the requirements.in, pip-tools cannot decide to remove them for you.

I used this setup to convince myself there was no bug here. The following creates a fake package with only setuptools as sub-dependency.

```
mktmpenv
pip install pip-tools
mkdir pkg && cd pkg
cat <<EOF > setup.py
from setuptools import setup
setup(install_requires=['setuptools'])
EOF
cd ..
echo "-e pkg" | pip-compile -n -o r.txt -
```
The result:
```
#
# To update, run:
#
#    pip-compile --output-file r.t -
#
-e file:///Users/david/.virtualenvs/tmp-ef044366e6edf64/pkg
Dry-run, so nothing updated.
```
With --allow-unsafe:

```
david$ echo "-e pkg" | pip-compile --allow-unsafe -n -o r.t -
#
# To update, run:
#
#    pip-compile --output-file r.t -
#
-e file:///Users/david/.virtualenvs/tmp-ef044366e6edf64/pkg
six==1.10.0               # via packaging, setuptools

setuptools==35.0.2
```

dschaller:
> The spirit of --allow-unsafe is to control if unsafe sub-dependencies are output to the requirements.txt, not to filter a user input which is the user desired state of wanted dependencies. 

@davidovich this is actually wrong. If you were to look at the PR to add the flag (https://github.com/jazzband/pip-tools/pull/377) the description says verbatim,

> Due to some specifics I had to include some version of pip in dependencies in several projects.
I am not sure it is needed for everyone, so I added a flag to have an option to enable that.

This is exactly what we are doing...

tests/test_writer.py
```diff
@@ -64,8 +70,58 @@ def test_format_requirement_environment_marker(from_line, writer):
...
+    ireq = from_line(unsafe_dep)
+    w = writer(allow_unsafe=False)
+    mocker.patch.object(w, 'write_header', return_value=['header'])
+    mocker.patch.object(w, 'write_flags', return_value=['flags'])
+    mocker.patch.object(w, '_sort_key', return_value=(ireq.editable, str(ireq.req).lower()))
+    format_requirement = mocker.patch.object(
+        w, '_format_requirement', return_value='{}'.format(unsafe_package)
+    )
+    comment = mocker.patch(
+        'piptools.writer.comment', side_effect=['foobar', '# {}'.format(unsafe_package)])
```
dschaller: Unit tests should test functionality in isolation of other modules assuming that those modules are tested independently.

tests/test_writer.py
```diff
@@ -64,8 +70,58 @@ def test_format_requirement_environment_marker(from_line, writer):
...
+
+    expected_results = ['header', 'flags', '', 'foobar', '# {}'.format(unsafe_package)]
+    result = w._iter_lines([ireq], False, [], {unsafe_package: 'marker'}, 'hashes')
```
dschaller: `'marker'` is an  arbitrary value used for testing purposes.

piptools/writer.py
dschaller: For the sake of argument it makes it harder for discoverability but it makes usage much clearer because you aren't allowing for usages to rely on position.

In terms of discoverability of arguments there should be documentation on the function with these values.

piptools/writer.py
```diff
@@ -96,20 +98,26 @@ def _iter_lines(self, results, reverse_dependencies, primary_packages, markers,
...
-                                               hashes=hashes)
+                line = self._format_requirement(
+                    ireq,
+                    reverse_dependencies,
+                    primary_packages,
+                    include_specifier=self.allow_unsafe,
+                    marker=markers.get(ireq.req.name),
+                    hashes=hashes if self.allow_unsafe else None
+                )
+                if not self.allow_unsafe:
```
dschaller: I think it's safer to fall back to the unmodified line at the catch all.

dschaller:
Given that all the criteria in the description are satisfied, none of the feedback is critical, and most importantly this fixes a bug with intended functionality I propose that we merge this.

davidovich:
@dschaller You are the only one saying the feedback is not critical. I feel the overall gist of this is that you want the change in, without regards to style or existing code. 

I have made comments and tried to understand the nature of what this PR trying to fix something that I am still not convinced is broken, but they seem to be ignored.

The response I got from the example code was that I didn't read a previous PR. I understand what that PR brought to the table, but I believe we are now in a better position than at that time.

The problem is that if you put something in the input, it is not for pip-tools to decide to remove that wanted state. If you want to remove pip from the output, don't put it in the input as a wanted requirement in the first place. I sill fail to understand your use case.

I can perfectly see that I would like to put `pip` in the requirements.txt **AND** not wanting to have unsafe dependencies end up in the compiled requirements. This PR would break that.

For example, with the same setup as above, imaging I would like to have `pip`, so with this PR, I would need to add `--allow-unsafe`, but at the same time, I would not want appdirs, packaging, etc:
```
david$ echo -e "-e pkg\npip" | pip-compile --allow-unsafe -n -o r.t -
#
# To update, run:
#
#    pip-compile --output-file r.t -
#
-e file:///Users/david/.virtualenvs/tmp-ef044366e6edf64/pkg
six==1.10.0               # via packaging, setuptools

pip==9.0.1
setuptools==35.0.2
```

dschaller:
@davidovich happy to chat more about the changes. My apologies if that is the way my comment came off as it was not my intent. By critical I meant the inline comments left by you were mostly style ones and they seem awfully like a personal preference.

Please help me understand how I can I better help you understand how this is broken.

> The problem is that if you put something in the input, it is not for pip-tools to decide to remove that wanted state.

Pip-tools, up until the last release with the change in it, would remove unsafe dependencies from the generated requirements files *even if* they were part of the original requirements *unless* the `--allow-unsafe` flag was set. I guess a better question here is was this functionality unintended then?

As far as your example goes it seems counter-intuitive that you would want pip-tools, a tool whose entire purpose is to pin all underlying dependencies, to not pin the underlying dependencies of unsafe packages when they are allowed. Can you help me understand why you wouldn't want this?

vphilippon:
I went through the discussion on wheter this is a bug or not, and I think I can summarize this:

@dschaller, you're pointing out that prior to 1.9.0, `--allow-unsafe` was an "all or nothing" option, but with the change of PR #441, we brought an unexpected change where giving an unsafe pin directly in ``requirements.in`` now stays in the output, which goes against the original idea.

@davidovich, you're saying that in retrospective, maybe thats a good change, because it allows us to say "Yes, I want `pip` to be pinned, but not the other unsafe ones (like `setuptools`)".

IMHO, I can see use cases where both behaviour would be good. And I think I have an idea to keep the best of both worlds: having a way allow _specific unsafe packages_, in the same way we do with `--upgrade` and `--upgrade-package`.
By default, `--allow-unsafe` would let all unsafe go through, but if I set `--allow-unsage-package pip`, then only pip is allowed.
We keep the previous behaviour that our users expects, while giving more control to the users.

Whether we should do that in one or many PR is debatable (oh no... what have I done ;) ), but I feel like this is the way to go.

What do you say about this?


On another subject, @dschaller, I haven't fully analysed the code change (sorry, I'll try to!), but the feeling I got is that we went back at filtering the output rather than filtering on the input. Am I right? If so, do you think its possible to change that? I think its a better way to handle the case, and I also got a feeling that the unsafe dependencies of unsafe requirements directly in the `requirements.in` are going to get through, which goes against the work in PR #441.

As always, thanks everyone for your time!

dschaller:
@davidovich @vphilippon @jdufresne opened https://github.com/jazzband/pip-tools/pull/517/ to hopefully fix this bug while preserving the additions in #441. Please take a look when you get a chance

jdufresne:
WIth 4e2f928902ac30878b4e989c843d00fa07085c4c merged, should this now be closed?

PR #513, opened 2017-05-25
Replace custom logging with built-in version.

### LINKED ISSUE #362: pip-tools logger should always go to stderr
I noticed that pip-tools's logger (called [`LogContext` here](https://github.com/nvie/pip-tools/blob/d521d766113e69ea0a10017bb968d3b3bb242e5f/piptools/logging.py#L10)) sometimes writes to stdout (depending on the log level).

It would be best if the logger always writes to stderr since a logger is for monitoring / diagnostic purposes, and its messages can interfere if one is trying to capture the "output" portion of running the script. This would help in addressing some of the other pip-tools issues related to stdout and stderr.

In cases where stdout is needed, a separate function can be used (e.g. `print()`). This would help in separating pip-tools's "API" (stdout) from its diagnostic / progress reporting (stderr).

cmc333333:
Rather than write to stdout via click.echo, use Python's built-in logging
mechanism. This is a bit more conventional and could allow users to configure
logs more thoroughly. It should also resolve #362.

To match the ANSI colors, this now imports the coloredlogs module.

##### Contributor checklist

- [ ] ~Provided the tests for the changes~ (N/A, refactor)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Master
+
+Bug Fixes:
+- Logs are now emitted to stderr instead of stdout.
+
 # 1.9.0
 
 Features:
--- a/piptools/logging.py
+++ b/piptools/logging.py
@@ -2,34 +2,20 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
-import sys
+import logging
 
-from . import click
-
-
-class LogContext(object):
-    def __init__(self, verbose=False):
-        self.verbose = verbose
-
-    def log(self, *args, **kwargs):
-        click.secho(*args, **kwargs)
-
-    def debug(self, *args, **kwargs):
-        if self.verbose:
-            self.log(*args, **kwargs)
-
-    def info(self, *args, **kwargs):
-        self.log(*args, **kwargs)
-
-    def warning(self, *args, **kwargs):
-        kwargs.setdefault('fg', 'yellow')
-        kwargs.setdefault('file', sys.stderr)
-        self.log(*args, **kwargs)
-
-    def error(self, *args, **kwargs):
-        kwargs.setdefault('fg', 'red')
-        kwargs.setdefault('file', sys.stderr)
-        self.log(*args, **kwargs)
+import coloredlogs
 
 
-log = LogContext()
+def configure_logging(verbose=False):
+    """Setup a colored logger to write to the console."""
+    logger = logging.getLogger('piptools')
+    if verbose:
+        level = logging.DEBUG
+    else:
+        level = logging.INFO
+    level_styles = dict(coloredlogs.DEFAULT_LEVEL_STYLES)
+    level_styles.update(debug={})   # reset debug to be the default style
+    fmt = '%(message)s'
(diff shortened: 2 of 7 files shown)


jdufresne:
I like the idea of moving away from click and towards using the stdlib for logging. In doing so, can we avoid introducing a new dependency, coloredlogs? It seems as pip-tools is a very low level tool used to build projects instead of consumed directly by projects it should strive for a minimal number of dependencies.

cmc333333:
While I'm generally in favor of reducing dependencies, the `coloredlogs` library is tailor-fit to this situation. Python's logging system isn't aware of ANSI colors (they don't make sense when outputting to a file, constructing JSON for logstash, etc.). `coloredlogs` allows us to add that back when writing to the console (while avoiding it for other outputs) to match the existing functionality.

I think that leaves us with a few possibilities:

1. Continue to write everything to stdout and skip Python's logging system. This is not a great solution as it mixes different types of messages and doesn't allow the logs to be configured.
2. Use Python's logging system, but don't attempt to colorize the stdout/stderr logs. This is a regression in terms of functionality.
3. Use Python's logging system and `coloredlogs`. This solves the problems of 1 and 2, but adds a new dependency.
4. Use Python's logging system and re-implement `coloredlogs` in pip-tools. This is also a less-than-optimal solution as it requires new tests and maintenance for the pip-tools team. It's also trading a custom logging system for a custom coloring system, which isn't as significant a step forward.

jdufresne:
Thanks for the analysis and reasoning. I agree that 1 & 4 are definitely not wanted. Your considerations of the trade offs are very fair.

vphilippon:
I'm like the idea. The only thing bugging me is the effect of a pip-sync for an app also using `coloredlogs`.
I did a simple test with your branch on Windows:
requirements.in:
```
coloredlogs<6.0
```
then
```
$ pip-compile
$ pip-sync
```
and it went well, no problem with modifying `coloredlogs` while `pip-sync` used it itself.
(If only it could go that well all the time... *sigh*).
That was a really simple test, it would probably be a good idea to do test the behaviour a bit more.

Note, I actually managed to "break" `pip-sync` doing this:
requirements.in:
```
coloredlogs<1.0
```
then
```
$ pip-compile
$ pip-sync
```
ending up with `coloredlogs==0.8` installed, which broke pip-sync.
So to be nice, we should set a lower bound on coloredlogs, so that running `pip check` will tell that something is wrong in that case.

I'll make a more complete review after the branch is rebased with the latest changes.

jdufresne:
Maybe it would be acceptable to just not have colored logging? What are your thoughts?

vphilippon:
Thinking about this over again, this is the same situation with any of of `pip-tools` dependencies (`click` or anything else). I did have to go out of my way a bit to break it. I think a reasonable idea here is to use `coloredlogs`, and put a version specifier, like we should, in `install_requires`. Let's try to put that lower bound as low as possible to minimize the conflict zone. I don't feel like this is lib with a high risk of conflict though.

Sorry for taking so long on this. Its a bit hard to get motivation when you know you're somewhat blocked ahead by being unable to make releases and give value to the time spent. /rant.

ssbarnea:
Too old to be salvageable.

PR #515, opened 2017-05-27
follow PEP 257 docstrings & quote consistency

dschaller:
- Follow [PEP 257](https://www.python.org/dev/peps/pep-0257/) docstring guidelines
- Follow [PEP 8](https://www.python.org/dev/peps/pep-0008/#string-quotes) guideline of consistency with quotes across the project

##### Contributor checklist


rpkilby:
> Going to merge this by EOD since, as the contributing doc states, it's just documentation updates but I would love feedback if anyone has some.

I would wait, since this touches a large number of files. It's erring on the side of *more than trivial*. 

Either way, I'm generally -1 on these kinds of changes unless there's style enforcement in the build. Without enforcement, the changes are kind of meaningless since nothing prevents future PR's from breaking the style guidelines. For consideration:

- There is a [pep257 plugin](https://gitlab.com/pycqa/flake8-docstrings) for flake8.
- I'd gather more feedback from other people, but I'm 👎  on changing string quotes in code since it  negatively affects git-blame. That said, it would be more acceptable if there was enforcement of quote consistency. Maybe [this](https://github.com/zheller/flake8-quotes/) plugin?

davidovich:
I don't see the value this PR adds. Adding a automatic checker would bring value for future contributions, but aesthetic changes like these are counter productive, and produce unnecessary conflicts for every pending PR.

dschaller:
@rpkilby a style checker is a great idea. However, even if that were added these changes would be required.

@davidovich the value is consistency which is the entire point of PEP standards. Maybe give [this page](http://python-guide-pt-br.readthedocs.io/en/latest/writing/style/) a read. :D

rpkilby:
> the value is consistency which is the entire point of PEP standards. Maybe give this page a read. :D

I'm sure @davidovich understands the benefits of style consistency 😉. The implication is that the PR does not have a net positive value on the codebase. ie, even though style consistency is a plus, is it worth mucking up git-blame and creating conflicts for existing PRs?

@davidovich - would simplifying this to just doc strings/pep 257 make sense? It would be a lot less disruptive.

PR #516, opened 2017-05-27
Use utils.UNSAFE_PACKAGES as source of truth

dschaller:
Update cli help for --allow-unsafe to use utils.UNSAFE_PACKAGES so this is only defined in one place.

```bash
  --allow-unsafe                  Pin packages considered unsafe: distribute,
                                  pip, setuptools
```
##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -58,7 +58,7 @@ class PipCommand(pip.basecommand.Command):
               help=('Output file name. Required if more than one input file is given. '
                     'Will be derived from input file otherwise.'))
 @click.option('--allow-unsafe', is_flag=True, default=False,
-              help="Pin packages considered unsafe: {}".format(list(UNSAFE_PACKAGES)))
+              help="Pin packages considered unsafe: {}".format(', '.join(sorted(UNSAFE_PACKAGES))))
 @click.option('--generate-hashes', is_flag=True, default=False,
               help="Generate pip 8 style hashes in the resulting requirements file.")
 @click.option('--max-rounds', default=10,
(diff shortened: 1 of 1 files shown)


dschaller:
@jdufresne @vphilippon @davidovich any chance one of you could take a look?

piptools/scripts/compile.py
```diff
@@ -58,7 +58,7 @@ class PipCommand(pip.basecommand.Command):
               help=('Output file name. Required if more than one input file is given. '
                     'Will be derived from input file otherwise.'))
 @click.option('--allow-unsafe', is_flag=True, default=False,
-              help="Pin packages considered unsafe: pip, setuptools & distribute")
+              help="Pin packages considered unsafe: {}".format(list(UNSAFE_PACKAGES)))
```
rpkilby: Maybe use `sorted` instead of `list`, since sets are unordered.

jdufresne: This is displaying as, `['distribute', 'pip', 'setuptools']` where previously it was displaying as `pip, setuptools & distribute`. Did you intend to include the full repr value of the list? Maybe do the following instead:

```python
', '.join(sorted(UNSAFE_PACKAGES)
```

dschaller: I did consider formatting the output but figured the overhead wasn't worth it since we wouldn't be able to have it identical to the original and the list output is still consumable. Happy to update if you think it's worth it @jdufresne

jdufresne: Just my personal preference, the commas are more readable than the repr. I'm open to other people's preference if they'd like to see something else.

rpkilby: maybe `', '.join(sorted(UNSAFE_PACKAGES))`? Predictable ordering and better presentation than a simple list repr.

dschaller:
Updated @jdufresne @rpkilby

PR #523, opened 2017-05-31
Remove redundant check for editable

jdufresne:
The first line of is_pinned_requirement() checks the value of editable, don't need to do so on the same line as the call.


vphilippon:
Thanks for the PR!

PR #527, opened 2017-06-01
Remove prints used during debugging from tests

jdufresne:
Removes spurious output when running tests.


vphilippon:
Thanks for the PR!

PR #385, opened 2016-08-16
Use download_dir=None for editable PyPI dependencies

blueyed:
Using a `download_dir` will make `pip` export the cloned repo via
`reqset._prepare_file` and `git-checkout-index`, which then makes
`python setup.py egg_info` fail, when
[setuptools_scm](https://github.com/pypa/setuptools_scm) is used.

Fixes https://github.com/nvie/pip-tools/issues/369.

PR #538, opened 2017-06-25
No download dir for editable

dfee:
A rebase of https://github.com/jazzband/pip-tools/pull/385.


dfee:
@vphilippon I'd like to get this in the next release as it replaces #385. Can we get this done?

tests/test_cli.py
```diff
@@ -222,6 +222,19 @@ def test_editable_package(tmpdir):
         assert 'six==1.10.0' in out.output
 
 
+def test_editable_package_vcs(tmpdir):
+    vcs_package = 'git+git://github.com/pytest-dev/pytest-django#egg=pytest-django'
```
vphilippon: I would pin this to the current latest commit of `pytest-django` master branch.
The `pytest-django` package could stop using `setuptools_scm`.
Also, if they break the master branch for a while, I'd rather not have our tests fail on that.

tests/test_cli.py
```diff
@@ -222,6 +222,19 @@ def test_editable_package(tmpdir):
...
+    runner = CliRunner()
+    with runner.isolated_filesystem():
+        with open('requirements.in', 'w') as req_in:
+            req_in.write('-e ' + vcs_package)
+        out = runner.invoke(cli, ['-n',
+                                  '--rebuild'])
+        print(out.output)
+        assert out.exit_code == 0
+        assert vcs_package in out.output
```
vphilippon: Add a check that `pytest` is also in the output, as it's one of `pytest-django`'s `install_requires`.
We want to make sure that we still collect the subdependencies.

tests/test_cli.py
dfee: Sounds good.

tests/test_cli.py
dfee: Sounds good.

dfee:
@vphilippon updated.

vphilippon:
Thanks for the contribution!

PR #540, opened 2017-06-28
Fixing OutputWriter._iter_lines for pip 8.1.1 or below

Lucas-C:
##### Contributor checklist

- [x] Provided the tests for the changes:
An existing test ensures this does not introduces any regression: https://github.com/jazzband/pip-tools/blob/master/tests/test_writer.py
I don't think this is needed as it is a fix.
Nope. Anyone please ? :)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,16 +2,20 @@ environment:
 
     matrix:
         - TOXENV: py27-pip8
+        - TOXENV: py27-pip8.1.1
         - TOXENV: py27-pip9
         - TOXENV: py27-piplatest
         - TOXENV: py34-pip8
+        - TOXENV: py34-pip8.1.1
         - TOXENV: py34-pip9
         - TOXENV: py34-piplatest
-        - TOXENV: py35-pip9
         - TOXENV: py35-pip8
+        - TOXENV: py35-pip8.1.1
+        - TOXENV: py35-pip9
         - TOXENV: py35-piplatest
-        - TOXENV: py36-pip9
         - TOXENV: py36-pip8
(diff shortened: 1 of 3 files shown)


Lucas-C:
I'm totally willing to add a test, but I don't understand what you want to test here ?

My change is in `OutputWriter._iter_lines` while you asked me to add a test for `OutputWriter._format_requirement` which is not affected by my changes (`marker=None` is going to be passed as an argument).

Also this aims to fix a compatibility issue with pip 8.1.1 or below, so maybe the best way to test it would be to add a `pip==8.1.1` testenv in https://github.com/jazzband/pip-tools/blob/master/tox.ini#L7 ?

vphilippon:
You're right, I was under the impression that the formatting was changed, but it's actually the value passed that might be different. That would be a different test.

Adding a test suite in tox using `pip 8.1.1` seems reasonable to me, as this version (while having a more recent bugfix version) seems to still be common in the wild.
You can go ahead and add a test environment with `pip 8.1.1`.

Also, sorry for the delay, life happened. Thank you for your time!

Lucas-C:
I added a test tox env,
but I'm not 100% sure I got it right :)

vphilippon:
@Lucas-C You also have to add the new `pip 8.1.1` tox environment in `.appveyor.yml`, for each python version already listed. That's for the Windows tests.
We're almost there, pretty sure we'll be done after that.
Thanks again.

Lucas-C:
No problem: done !

vphilippon:
Thanks for your contribution!

Lucas-C:
Thanks for your help :)

PR #542, opened 2017-07-10
Remove unwanted annotations from primary packages

quantus:
`primary_packages` contains lower case package names where as `ireq.name` may contain upper case letters. This causes the primary packages to have the unwanted `# via XXX` annotations. In my case this issue was seen with packages `SQLAlchemy` and `Flask`.

##### Contributor checklist

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -137,7 +137,7 @@ class OutputWriter(object):
             for hash_ in sorted(ireq_hashes):
                 line += " \\\n    --hash={}".format(hash_)
 
-        if not self.annotate or ireq.name in primary_packages:
+        if not self.annotate or key_from_req(ireq.req) in primary_packages:
             return line
 
         # Annotate what packages this package is required by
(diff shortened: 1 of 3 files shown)


quantus:
I forgot to tell in the PR how this problem can be reproduced.

Setup:
```
mkvirtualenv --python=$(which python2.7) test 
pip install Flask Flask-Debugtoolbar pip-tools
echo "Flask" > requirements.in
echo "Flask-Debugtoolbar" >> requirements.in
```

After the setup:
```
(test) ~/Documents/tmp pip-compile requirements.in 
#
# To update, run:
#
#
blinker==1.4              # via flask-debugtoolbar
click==6.7                # via flask
flask-debugtoolbar==0.10.1
flask==0.12.2             # via flask-debugtoolbar
itsdangerous==0.24        # via flask, flask-debugtoolbar
jinja2==2.9.6             # via flask
markupsafe==1.0           # via jinja2
werkzeug==0.12.2          # via flask, flask-debugtoolbar
(test) ~/Documents/tmp pip-compile --version
pip-compile, version 1.9.0
```

As we can see from the output, the `flask` dependency was listed to come from `# via flask-debugtoolbar`. As the `flask` dependency was our primary dependency, that comment should not be there.

With this information I feel this PR is ready to be reviewed.

piptools/writer.py
```diff
@@ -137,7 +137,7 @@ def _format_requirement(self, ireq, reverse_dependencies, primary_packages, mark
             for hash_ in sorted(ireq_hashes):
                 line += " \\\n    --hash={}".format(hash_)
 
-        if not self.annotate or ireq.name in primary_packages:
+        if not self.annotate or ireq.name.lower() in primary_packages:
```
vphilippon: Good catch
While at it, as `primary_packages` was generated using `key_from_req`, we should use that here too for consistency.
`key_from_req(ireq.req) in primary_packages`

quantus:
@vphilippon: I've made the change you requested.

vphilippon:
Thanks for the contribution!

Issue #563, opened 2017-09-20
Environment markers do not carry out to transitive dependencies


tuukkamustonen:
As in title.

##### Environment Versions

Ubuntu 14.04, pip-tools 1.9.0, pip 9.0.1, python 2.7 + 3.4

##### Steps to replicate

Imagine some package 'foo` that requires:

    enum34; python_version < '3.4'

Require `foo` in `requirements.txt`:

    foo

With python 2.7, when you run `pip-compile requirements.txt`...

##### Expected result

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file requirements.txt requirements.in
#
enum34==1.1.6; python_version < '3.4'           # via foo
foo==1.0.0
```

##### Actual result

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file requirements.txt requirements.in
#
enum34==1.1.6           # via foo
foo==1.0.0
```

So environment marker is not carried on.

If you run with python3.4+, whole `enum34` is omitted.

suutari-ai:
It's not actually only the environment markers that will affect the dependencies. Also different distribution packages for the same requirement might have different dependencies.

An example of this is provided in issue #558: `django-auth-ldap==1.2.15` has three different packages:
 * `django_auth_ldap-1.2.15-py2-none-any.whl` requiring `django` and `python-ldap>=2.0`;
 * `django_auth_ldap-1.2.15-py3-none-any.wh` requiring `django` and `pyldap`; and
 * `django-auth-ldap-1.2.15.tar.gz` requiring `django` and either `python-ldap>=2.0` or `pyldap` depending on which python version is used to run the `setup.py`.

In ideal world everyone would use environment markers, but in reality there are many distributions in PyPI which have different dependencies for the different packages of a release.

vphilippon:
Carrying over the environment marker would't be easy, because that's handled directly by `pip` and we simply get the resulting dependencies after their evaluation.

Even if we'd manage to do this in a maintainable way, we'd have to compute the dependencies for *each condition branching* from those markers for the output to be exact.

So unfortunately, the generated `requirements.txt` are not guaranteed to be "environment independent", and IMHO, trying to achieve this would be a really hard task that would require a **lot** of testing.

tuukkamustonen:
[...]

PR #520, opened 2017-05-30
Generate hashes for all available distributions

jdufresne:
Previously would only generate hashes for the current platform. Allows
running pip-compile to generate a requirements.txt for any other
platform. Uses monkey patch approach recommend by the PIP maintainers
at:

https://github.com/pypa/pip/issues/4422#issuecomment-292219599

Fixes #414

##### Contributor checklist

- [x] Provided the tests for the changes
- [x] Added the changes to CHANGELOG.md
- [ ] Requested (or received) a review from another contributor

PR #571, opened 2017-09-28
Respect platform's wheels for dependencies (Fixes #558)

### LINKED ISSUE #558: Python 2 package is added to Python 3 dependency list using pip-compile
Issue occurs in conjunction with [django-auth-ldap](https://bitbucket.org/illocution/django-auth-ldap).

`django-auth-ldap` requires different LDAP packages, depending on whether Python 2 or 3 is used. As defined in `setup.py`:

```
[...]

install_requires=[
    "django",
    "pyldap" if PY3 else "python-ldap >= 2.0",
],

[...]
```

When using `pip-compile` to compile a list of dependencies on a Python 3 environment, the Python 2-compatible package (`python-ldap`) is added to the list. This results in a broken application when all dependencies are installed through `pip`.

##### Environment Versions

1. Fedora 26 (`4.12.9-300.fc26.x86_64 GNU/Linux`)
1. Python version: 3.6.2
1. pip version: 9.0.1
1. pip-tools version: 1.9.0

##### Steps to replicate

1. Create a file called `base.in` and add at least the following dependencies (non-related packages have been omitted):
```
Django==1.10.6
django-auth-ldap==1.2.10
pyldap==2.4.28

[...]
```
2. Run the following command in a Python 3 environment to compile a list of dependencies:
```
pip-compile --output-file base.txt base.in
```
Note: The resulting file should now contain a line like `python-ldap==2.4.42       # via django-auth-ldap`.

3. Install packages through `pip install -r /path/to/base.txt`.

##### Expected result

I expect all packages to install without errors.

##### Actual result

The following error occurs:

```
Collecting python-ldap==2.4.42 (from -r /requirements/dev.txt (line 18))
  Downloading python-ldap-2.4.42.tar.gz (297kB)
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-32v49evs/python-ldap/setup.py", line 53
        print name + ': ' + cfg.get('_ldap', name)
                 ^
    SyntaxError: invalid syntax
```

Note here that my `base.txt` file is called `dependencies.txt` in the snippet above. This is because the file is renamed and used in a Docker environment.

When I remove the line `python-ldap==2.4.42       # via django-auth-ldap` from by dependency file manually, all packages are installed without any error.

vphilippon:
Keep pip.Wheel monkey patch only in the get_hashes context.

Fixes #558 

##### Contributor checklist

### DIFF
--- a/piptools/repositories/base.py
+++ b/piptools/repositories/base.py
@@ -3,6 +3,7 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
 from abc import ABCMeta, abstractmethod
+from contextlib import contextmanager
 
 from six import add_metaclass
 
@@ -37,4 +38,11 @@ class BaseRepository(object):
         Given a pinned InstallRequire, returns a set of hashes that represent
         all of the files for a given requirement. It is not acceptable for an
         editable or unpinned requirement to be passed to this function.
+        """
+
+    @abstractmethod
+    @contextmanager
+    def allow_all_wheels(self):
+        """
+        Monkey patches pip.Wheel to allow wheels from all platforms and Python versions.
         """
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -28,24 +28,6 @@ except ImportError:
     from .._compat import TemporaryDirectory
 
 
-# Monkey patch pip's Wheel class to support all platform tags. This allows
-# pip-tools to generate hashes for all available distributions, not only the
-# one for the current platform.
-
-def _wheel_supported(self, tags=None):
-    # Ignore current platform. Support everything.
-    return True
-
-
-def _wheel_support_index_min(self, tags=None):
-    # All wheels are equal priority for sorting.
-    return 0
-
-
-Wheel.supported = _wheel_supported
-Wheel.support_index_min = _wheel_support_index_min
-
-
 class PyPIRepository(BaseRepository):
     DEFAULT_INDEX_URL = 'https://pypi.python.org/simple'
 
@@ -182,12 +164,13 @@ class PyPIRepository(BaseRepository):
 
         # We need to get all of the candidates that match our current version
         # pin, these will represent all of the files that could possibly
-        # satisify this constraint.
-        all_candidates = self.find_all_candidates(ireq.name)
-        candidates_by_version = lookup_table(all_candidates, key=lambda c: c.version)
-        matching_versions = list(
-            ireq.specifier.filter((candidate.version for candidate in all_candidates)))
-        matching_candidates = candidates_by_version[matching_versions[0]]
+        # satisfy this constraint.
+        with self.allow_all_wheels():
+            all_candidates = self.find_all_candidates(ireq.name)
+            candidates_by_version = lookup_table(all_candidates, key=lambda c: c.version)
+            matching_versions = list(
+                ireq.specifier.filter((candidate.version for candidate in all_candidates)))
+            matching_candidates = candidates_by_version[matching_versions[0]]
 
         return {
             self._get_file_hash(candidate.location)
@@ -200,6 +183,37 @@ class PyPIRepository(BaseRepository):
             for chunk in iter(lambda: fp.read(8096), b""):
                 h.update(chunk)
         return ":".join([FAVORITE_HASH, h.hexdigest()])
+
+    @contextmanager
+    def allow_all_wheels(self):
+        """
+        Monkey patches pip.Wheel to allow wheels from all platforms and Python versions.
+
+        This also saves the candidate cache and set a new one, or else the results from the
+        previous non-patched calls will interfere.
+        """
+        def _wheel_supported(self, tags=None):
+            # Ignore current platform. Support everything.
+            return True
+
+        def _wheel_support_index_min(self, tags=None):
+            # All wheels are equal priority for sorting.
+            return 0
+
+        original_wheel_supported = Wheel.supported
+        original_support_index_min = Wheel.support_index_min
+        original_cache = self._available_candidates_cache
+
+        Wheel.supported = _wheel_supported
+        Wheel.support_index_min = _wheel_support_index_min
+        self._available_candidates_cache = {}
+
+        try:
+            yield
+        finally:
+            Wheel.supported = original_wheel_supported
+            Wheel.support_index_min = original_support_index_min
+            self._available_candidates_cache = original_cache
 
 
 @contextmanager
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -165,12 +165,11 @@ class PyPIRepository(BaseRepository):
         # We need to get all of the candidates that match our current version
         # pin, these will represent all of the files that could possibly
         # satisfy this constraint.
-        with self.allow_all_wheels():
-            all_candidates = self.find_all_candidates(ireq.name)
-            candidates_by_version = lookup_table(all_candidates, key=lambda c: c.version)
-            matching_versions = list(
-                ireq.specifier.filter((candidate.version for candidate in all_candidates)))
-            matching_candidates = candidates_by_version[matching_versions[0]]
+        all_candidates = self.find_all_candidates(ireq.name)
+        candidates_by_version = lookup_table(all_candidates, key=lambda c: c.version)
+        matching_versions = list(
+            ireq.specifier.filter((candidate.version for candidate in all_candidates)))
+        matching_candidates = candidates_by_version[matching_versions[0]]
 
         return {
             self._get_file_hash(candidate.location)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 1.10.2 (UNRELEASED)
+
+Bug Fixes:
+- Fixed bug causing dependencies from invalid wheels for the current platform to be included ([#571](https://github.com/jazzband/pip-tools/pull/571)).
+
 # 1.10.1 (2017-09-27)
 
 Bug Fixes:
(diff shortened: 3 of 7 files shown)


vphilippon:
Requesting review from @suutari-ai and @jdufresne 
(As a side note, any reason why I wans't able to add you as "official reviewers"?)

piptools/repositories/base.py
```diff
@@ -38,3 +39,10 @@ def get_hashes(self, ireq):
         all of the files for a given requirement. It is not acceptable for an
         editable or unpinned requirement to be passed to this function.
         """
+
+    @abstractmethod
+    @contextmanager
+    def allow_all_wheels(self):
+        """
+        Monkey patches pip.Wheel to allow wheels from all platoforms and Python versions.
```
jdufresne: Typo platoforms -> platforms (multiple docstrings)

piptools/repositories/pypi.py
```diff
@@ -201,6 +183,36 @@ def _get_file_hash(self, location):
...
+
+        original_wheel_supported = Wheel.supported
+        original_support_index_min = Wheel.support_index_min
+        original_cache = self._available_candidates_cache
+
+        Wheel.supported = _wheel_supported
+        Wheel.support_index_min = _wheel_support_index_min
+        self._available_candidates_cache = {}
+
+        yield
```
jdufresne: I would wrap this in a `try`/`finally` in case an exception is raised and then caught somewhere else in the stack.

```py
try:
    yield
finally:
     Wheel.supported = original_wheel_supported
     Wheel.support_index_min = original_support_index_min
     self._available_candidates_cache = original_cache
```

vphilippon: Well I learned something today, I though the exiting part would always be called, even with an exception, without the `try`. Just tried it out and you're right, will fix this, thanks.

suutari-ai:
Thanks @vphilippon. Good work!

Nevertheless, don't get me wrong, but I think this is still quite hacky, and I don't see the point. Why don't you just revert the commit 4900c7c which is causing this? Why would I ever want hashes of non-compatible wheels to my requirements.txt?

If you're trying to generate a `requirements.txt` which works for all platforms and Python versions, then this would make sense, but since the dependencies aren't calculated for all platforms, but only for the current one, the `requirements.txt` doesn't work for another platform anyway. Creating a `requirements.txt` which works for all platforms is requested in #563 and if that is ever implemented this hack would cause more harm than good.

As I said in PR #460, my opinion is that requirements.txt should be locking a single set of the requirements which is tested to work. The point of locking the versions is to get reproducible environment and if the platform or Python version is different, then it's no more reproducible: Those packages might act very differently on another platform and, as seen here, even the required packages might be different.

The whole point why I want locked requirements is *confidence* and having platform independent `requirements.txt` doesn't help there.

If I want to install a project for another platform (which is untested), then I should be using the `requirements.in` to generate a new lock file (`requirements-{new_platform}.txt`) and then install from that. If it works, the new lock file can be committed as a sign of "this is also a supported and tested platform".

suutari-ai:
> As a side note, any reason why I wasn't able to add you as "official reviewers"?

Because I wasn't member of @jazzband yet. I joined now. Didn't know it's that easy to join.

jdufresne:
> Nevertheless, don't get me wrong, but I think this is still quite hacky, and I don't see the point. Why don't you just revert the commit 4900c7c which is causing this? Why would I ever want hashes of non-compatible wheels to my requirements.txt?

A lot of this is described in the commit message and issue #520. But I'll summarize.

I work on a project that has a diverse group of environments:

* Different Linux distros
* MacOS
* All Pythons are 3.4+ but could be on a different version

My Project is Python 3 only. For security purpose, I want to generate package hashes for some additional verification that the download is correct.

In 1.9.0, when generating requirements.txt on one machine it might fail when running on another machine. It failed because the hashes generated were not compatible with the running OS & Python
version. As I saw it, I had some options:

1. Update pip-tools to generate all hashes 
2. Stop generating hashes
3. Unify all developers on one platform

I saw 1 as the best tradeoff at the time (and still do).

> If you're trying to generate a requirements.txt which works for all platforms and Python versions, then this would make sense, but since the dependencies aren't calculated for all platforms, but only for the current one, the requirements.txt doesn't work for another platform anyway. Creating a requirements.txt which works for all platforms is requested in #563 and if that is ever implemented this hack would cause more harm than good.

Maybe it is just "luck" of my dependencies, but I'm not seeing an issue regarding transitive packags in my project.

> The whole point why I want locked requirements is confidence and having platform independent requirements.txt doesn't help there.

I understand. But with hashes specific to binary platforms, this means the requirements.txt is only useful for identical environments to which it was generated. This seems limitng to me.

vphilippon:
@suutari-ai  Thanks for the review and detailed explanation of your concern. As @jdufresne mentionned, while a compiled `requirements.txt` is not guaranteed to be cross-platform compatible, it can be in some cases, like his.

> Why would I ever want hashes of non-compatible wheels to my requirements.txt?

In the case where a `requirements.txt` is cross-platform (i.e. there are no dependencies difference between platforms), but the packages still provides different wheels per platform (i.e. these packages requires compilation), having all hashes would keep the happy case of having a `requirements.txt` that is cross-platform compatible.

Currently, I see no harm in having `--generate-hashes` including all hashes.
I agree with you that reusing a `requirements.txt` on multiple platforms is something that should be done really carefully, and on a case-by-case basis after making sure there are no dependencies difference. That's the case of @jdufresne. Following the "we are all adult here" mindset of Python, I wouldn't feel justified to remove this feature as I was able to make it work and it provides some value to some users.

If it had been impossible (or way too hard) to keep the feature without breaking the dependencies resolving, I would have suggested to revert of course.

@suutari-ai I think a case justifying the feature has been provided. If you have a case or example that would prove this feature to be harmful to the users, outweighing the gain, please provide it.

jdufresne:
> If it had been impossible (or way too hard) to keep the feature without breaking the dependencies resolving, I would have suggested to revert of course.

> @suutari-ai I think a case justifying the feature has been provided. If you have a case or example that would prove this feature to be harmful to the users, outweighing the gain, please provide it.

I think this is very fair. If an argument or evidence can be provided to show that this will be too difficult to maintain or is breaking previously working environments. Then I agree, reverting the feature is justified.

suutari-ai:
@jdufresne:
>  1. Update pip-tools to generate all hashes
>  2. Stop generating hashes
>  3. Unify all developers on one platform

You're missing couple options:

   4. Unify all developers to use one of N pre-specified platforms and generate a `requirements-{platform_tag}.txt` for each supported platform
   5. Let CI and deployment scripts use requirements.txt, but developers on a different platform use requirements.in.

IMHO this would make sense, since having only the hashes of the main platform in the txt file should help ensuring that CI is testing the same stuff as is installed to production.

But anyway. I'm not against merging this. Even though it's not something that I want, it's still a feature that someone finds useful. And merging this should make pip-compile work better and would only make it easier to add an option to disable the 'generate hashes for all platforms'.

suutari-ai:
@vphilippon:
> Currently, I see no harm in having --generate-hashes including all hashes.
> I agree with you that reusing a requirements.txt on multiple platforms is something that should be done really carefully, and on a case-by-case basis after making sure there are no dependencies difference.

For me the 'generate hashes for all platforms' feature is more harmful than useful, because I don't want hashes of another platform to my txt files. It's also encouraging users to reuse the same requirements.txt on different platforms, which IMHO is not a good idea.

 * There could be dependencies which are missing from the txt file. Happens when the generating platform didn't have those dependencies, but the using platform has. (E.g. enum34 on python 2). Then those dependencies are not pinned and things might break when they get upgraded silently.
 *  There could be extra dependencies in the txt file. Happens when the generating platform has some dependencies which the using platform doesn't.
 * It's also possible that some version of a dependency is not available for all platforms which would affect which version is pinned.

But the feature is there already. Fixing it is wise. :+1:

suutari-ai:
>> Unify all developers to use one of N pre-specified platforms and generate a `requirements-{platform_tag}.txt` for each supported platform.
>
> This seems like a mess to manage, honestly.

Yeah, this certainly needs some tooling and a new feature so that it would be possible to generate a txt for another platform.

>>  Let CI and deployment scripts use `requirements.txt`, but developers on a different platform use `requirements.in`.
>
> A problem with this is the requirements in `requirements.in` aren't pinned. So different development environment will be using different versions of the dependencies, causing subtle behavior differences.

Different platforms will also cause behavior differences. Only way to get exactly same behavior is to remove that factor too.

Also, if it's not working with a certain version of a requirement, you should either (a) make it work with it by changing your code or (b) limit out the non-working versions in `requirements.in` (using specifiers like `<ver` or `~=`). The locked `requirements.txt` should be easily upgradeable anyway, so the project should usually just work after regenerating the `requirements.txt`.

>>    For me the 'generate hashes for all platforms' feature is more harmful than useful, because I don't want hashes of another platform to my txt files.
>
> Hmm, this is interesting to me. @suutari-ai I think you make a convincing argument. I may be persuaded that generating all hashes was a bad idea for a requirements.txt that will be used on a production server. I honestly don't know what is best at this point. If there is large agreement to revert the feature, I won't stand in the way.

Maybe it should be optional at least? (Which would be another PR, of course.)

For Prequ, I just reverted it (https://github.com/suutari/prequ/pull/12), since my plan there is to implement the 'generate a txt for another platform' feature, but it seems that pip-tools is going to another direction. That's OK too, since I'll probably just use my own tool anyway. :smile:

If the feature is not reverted, then this fix should be merged at least.

vphilippon:
@suutari-ai @jdufresne Thanks you both for your input on this.
I'll go ahead and merge the fix.

PR #573, opened 2017-10-06
Support directories as input to pip-compile (fixes #207)

### LINKED ISSUE #207: pip-compile cant take a directory.
When I input `pip-compile` with a directory,

``` shell
❯ pip-compile requirements 
  File "/home/sanket/.virtualenvs/djangohub/bin/pip-compile", line 11, in <module>
    sys.exit(cli())
  File "/home/sanket/.virtualenvs/djangohub/local/lib/python2.7/site-packages/click/core.py", line 700, in __call__
  File "/home/sanket/.virtualenvs/djangohub/local/lib/python2.7/site-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/home/sanket/.virtualenvs/djangohub/local/lib/python2.7/site-packages/click/core.py", line 873, in invoke
  File "/home/sanket/.virtualenvs/djangohub/local/lib/python2.7/site-packages/click/core.py", line 508, in invoke
  File "/home/sanket/.virtualenvs/djangohub/local/lib/python2.7/site-packages/piptools/scripts/compile.py", line 73, in cli
    for line in parse_requirements(src_file, finder=repository.finder, session=repository.session):
  File "/home/sanket/.virtualenvs/djangohub/local/lib/python2.7/site-packages/pip/req/req_file.py", line 75, in parse_requirements
    filename, comes_from=comes_from, session=session
  File "/home/sanket/.virtualenvs/djangohub/local/lib/python2.7/site-packages/pip/download.py", line 416, in get_file_content
    'Could not open requirements file: %s' % str(exc)
pip.exceptions.InstallationError: Could not open requirements file: [Errno 21] Is a directory: 'requirements'
```

It would be great if `pip-compile` would find all the *.in files and compile them when a directory is given as a parameter.

simlun:
A first jab at implementing issue #207, passing a directory as input to pip-compile in addition to just files.

Any feedback is much appreciated :) Yay for [#hacktoberfest](https://hacktoberfest.digitalocean.com/)!

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -86,6 +86,9 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
         if not output_file:
             raise click.BadParameter('--output-file is required if input is from stdin')
 
+    if any(map(os.path.isdir, src_files)) and not output_file:
+        raise click.BadParameter('--output-file is required if an input directory is given.')
+
     if len(src_files) > 1 and not output_file:
         raise click.BadParameter('--output-file is required if two or more input files are given.')
 
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -70,7 +70,11 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
         client_cert, trusted_host, header, index, emit_trusted_host, annotate,
         upgrade, upgrade_packages, output_file, allow_unsafe, generate_hashes,
         src_files, max_rounds):
-    """Compiles requirements.txt from requirements.in specs."""
+    """
+    Compiles requirements.txt from requirements.in specs.
+
+    SRC_FILES may include directories of *.in files.
+    """
     log.verbose = verbose
 
     if len(src_files) == 0:
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -3,6 +3,8 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
 import optparse
+from glob import glob
+
 import os
 import sys
 import tempfile
@@ -151,6 +153,17 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
         log.debug('Configuration:')
         for find_link in repository.finder.find_links:
             log.debug('  -f {}'.format(find_link))
+
+    ###
+    # Find source files in input directories
+    ###
+
+    src_dirs = list(filter(os.path.isdir, src_files))
+    src_files = list(set(src_files) - set(src_dirs))
+
+    for d in src_dirs:
+        more_files = glob('{}/*.in'.format(d))
+        src_files += more_files
 
     ###
     # Parsing/collecting initial requirements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,12 @@
 Bug Fixes:
 - Fixed bug causing dependencies from invalid wheels for the current platform to be included ([#571](https://github.com/jazzband/pip-tools/pull/571)).
 
+
 # 1.10.1 (2017-09-27)
 
 Bug Fixes:
 - Fixed bug breaking `pip-sync` on Python 3, raising `TypeError: '<' not supported between instances of 'InstallRequirement' and 'InstallRequirement'` ([#570](https://github.com/jazzband/pip-tools/pull/570)).
+
 
 # 1.10.0 (2017-09-27)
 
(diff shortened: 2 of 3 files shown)


simlun:
Thanks @vphilippon for the quick feedback :) The changelog is now updated and edits from maintainers is allowed.

Cheers!

vphilippon:
Allright, thanks!
I was hopping to get a 1.10.2 release (including a pretty important bugfix), so I was keeping master stable without merging additional PR for now, but its unfortunately out of my hands. It might take a little while before we merge this, but it should go in without any issue later on.

Anyway, even if I was to merge it right away, we still can't make any release.

simlun:
Ok @vphilippon, include it whenever it's suitable - no stress :)

vphilippon:
@simlun We should be good to get stuff moving again.
You can rebase on master.

piptools/scripts/compile.py
```diff
@@ -86,6 +92,9 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
         if not output_file:
             raise click.BadParameter('--output-file is required if input is from stdin')
 
+    if any(map(os.path.isdir, src_files)) and not output_file:
+        raise click.BadParameter('--output-file is required if an input directory is given.')
```
merwok: I was hoping for a different behaviour: for each `.in` file found in the directory (or directories), use `{basename}.txt` as output file.  The idea is to have different files for runtime requirements (direct dependencies), test deps (additional things for CI, not installed during deployment), extra deployment deps (not installed locally or duing CI).  At the moment I have a wrapper script that iterates over `in` files to create separate `txt` files.

vphilippon: @merwok If I'm following your use correctly, you would have something like:
requirements.in:
```
myapp
requests[security]
```
and then test-requirements.in:
```
pytest
flake8
```
and you would want separate requirements.txt and test-requirements.txt, which are the `pip-compile` output of each?
If so, then you might have duplicate dependencies in each, which would force you to do `pip install` of each separately, but that breaks the guarantee that the dependencies of each will be met (your test-requirements.txt and requirements.txt `requests` requirement may differ, as an example).

suutari-ai: This can be solved, if you define which of the files is the "master".

I have solved this in Prequ by making the main `requirements.txt` as the master which is always compiled first. When compiling the other requirements files, Prequ will add
```
-c requirements.txt
```
line to the input lines so that the compiled `requirements-*.txt` won't have any conflicting dependencies. Here's the relevant code: https://github.com/suutari/prequ/blob/v1.3.0/prequ/configuration.py#L182

Though, there might still be conflicting dependencies in requirements-a.txt and requirements-b.txt, if those are not in requirements.txt, but that's not very common case (and it's not yet handled by Prequ).

vphilippon: @suutari-ai Thank you for the input. I like where you're going with this. I will have to revisit this idea after my coffee, but it seems to make sens to me right now.

@simlun Did you have a specific use case for the way you implemented this, or where you simply (and kindly!) taking care of an existing open issue?

Let's make sure we're all on the same page 😄

merwok: @vphilippon Yes that’s right.  I don’t have issues with duplicate dependencies; typically I have Pyramid or Django and various libs in the runtime requirements, test deps that work with Pyramid or Django but don’t depend on them (see https://github.com/nedbat/django_coverage_plugin/issues/44 for example), CI dependencies which are an other disjoint set (tox, coverage and pip-tools).  To run tests I call `pip-sync runtime.txt tests.txt` (see https://github.com/tox-dev/tox/issues/149#issuecomment-290263985 ), and the only problem I’ve had was with six, which I solved by re-compiling the files so that the latest version gets picked up in both files.  So far I haven’t used test deps that need requests, which could easily cause a version conflict since third-party API clients will probably depend on a specific version range of requests.

This has been working very well on many projects in the last year!  Depending on the task, different sets are installed: ci.txt to bootstrap, runtime.txt + tests.txt inside tox testenvs, runtime.txt + deploy.txt for deployment.  I was looking forward to this ticket to be able to do `pip-compile deps/` and remove my wrapper script.

If it is agreed that this is about taking many sources and creating a unique frozen requirements file, then I’ll probably open another ticket to ask for an option to make pip-compile accept multiple source files (or a directory) and compile each one to its matching frozen version.  Then the two use cases would be satisfied :)

simlun: @vphilippon I just love this project and was simply supplying a PR for an existing open issue. If I misunderstood, or someone wants to take it in another direction - feel free! :)

merwok:
This could be an inspiration: https://github.com/peterdemin/pip-compile-multi (hi @peterdemin)

peterdemin:
Hi :-)
From my experience, I'd like to add, that when processing directories you naturally hit a few obstacles.
1. You need to filter input files because some files should not be compiled (for example py2 vs py3 separation).
2. You need to configure input extension and output extension (to enable `.in` => `.txt` => `.hash` pipeline).
3. You need a way to check, that `pip compile` was run after changes to `.in` files.

If `pip-tools` supported all these cases (and compatible releases, but that's not about directory input), I'd be happy to retire `pip-compile-multi`. Also, I think that it's fine to have `pip-compile-multi` as a wrapper around `pip-tools` and keep it separated.

IMHO, the most complex part of `pip-compile-multi` is conflicts detection. I didn't review your PR, so I just want to highlight this part.

atugushev:
`pip-compile-multi --directory` is the best choice. Thanks to @peterdemin!

PR #574, opened 2017-10-07
support URLs as packages, fix #272

### LINKED ISSUE #272: Error with pip : pip-compile does not support URLs as packages
I'm using `pip` with `pip-compile` (installed this way: `pip install pip-tools`)

I got the following error when I run the `pip-compile -v` command:

> **pip-compile does not support URLs as packages**, unless they are editable. Perhaps add -e option? (constraint was:
> aldryn-django==1.8.7.0 from
> https://control.aldyn.com/api/v1/apps/serve/aldryn-django/1.8.7.0/592213b1-e515-4447-8ef0-850713571a42/aldryn-django-1.8.7.0.tar.gz#egg=aldryn-django==1.8.7.0
> (from -r requirements.in (line 2)))

I have tried with the `-e` option, but this causes another problem.

Below is an short extract of my `requirements.in` file:

    \# <INSTALLED_ADDONS>  # Warning: text inside the INSTALLED_ADDONS tags is auto-generated. Manual changes will be overwritten.
    https://control.aldryn.com/api/v1/apps/serve/aldryn-django/1.8.7.0/592213b1-e515-4447-8ef0-850713571a42/aldryn-django-1.8.7.0.tar.gz#egg=aldryn-django==1.8.7.0
    ...
    \# </INSTALLED_ADDONS>

I'm using Docker container based on the `python:2.7-slim` image.

The `requirements.in` work well on one other similar docker container. 

I don't know why on mine, `pip-compile` does not work...

Have you any idea?

I case of bad use of pip, I have open a ticket on stackoverflow : http://stackoverflow.com/questions/34267352/error-with-pip-pip-compile-does-not-support-urls-as-packages

nim65s:
Hi,

This PR adds a `is_url_requirement` util, which is used to let pip find the right dependencies for a package, as if the package was editable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.12.0
+
+- Support URLs as packages ([#574](https://github.com/jazzband/pip-tools/pull/574)).
+
 # 1.11.0 (2017-11-30)
 
 Features:
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -15,7 +15,7 @@ from .cache import DependencyCache
 from .exceptions import UnsupportedConstraint
 from .logging import log
 from .utils import (format_requirement, format_specifier, full_groupby,
-                    is_pinned_requirement, key_from_ireq, key_from_req, UNSAFE_PACKAGES)
+                    is_pinned_requirement, is_url_requirement, key_from_ireq, key_from_req, UNSAFE_PACKAGES)
 
 green = partial(click.style, fg='green')
 magenta = partial(click.style, fg='magenta')
@@ -120,8 +120,8 @@ class Resolver(object):
     @staticmethod
     def check_constraints(constraints):
         for constraint in constraints:
-            if constraint.link is not None and not constraint.editable:
-                msg = ('pip-compile does not support URLs as packages, unless they are editable. '
+            if constraint.link is not None and constraint.link.url.startswith('file:') and not constraint.editable:
+                msg = ('pip-compile does not support file URLs as packages, unless they are editable. '
                        'Perhaps add -e option?')
                 raise UnsupportedConstraint(msg, constraint)
 
@@ -244,11 +244,7 @@ class Resolver(object):
             Flask==0.10.1 => Flask==0.10.1
 
         """
-        if ireq.editable:
-            # NOTE: it's much quicker to immediately return instead of
-            # hitting the index server
-            best_match = ireq
-        elif is_pinned_requirement(ireq):
+        if ireq.editable or is_url_requirement(ireq, self.repository) or is_pinned_requirement(ireq):
             # NOTE: it's much quicker to immediately return instead of
             # hitting the index server
             best_match = ireq
@@ -262,14 +258,14 @@ class Resolver(object):
 
     def _iter_dependencies(self, ireq):
         """
-        Given a pinned or editable InstallRequirement, collects all the
+        Given a pinned, an url, or editable InstallRequirement, collects all the
         secondary dependencies for them, either by looking them up in a local
         cache, or by reaching out to the repository.
 
         Editable requirements will never be looked up, as they may have
         changed at any time.
         """
-        if ireq.editable:
+        if ireq.editable or is_url_requirement(ireq, self.repository):
             for dependency in self.repository.get_dependencies(ireq):
                 yield dependency
             return
@@ -293,5 +289,5 @@ class Resolver(object):
             yield InstallRequirement.from_line(dependency_string, constraint=ireq.constraint)
 
     def reverse_dependencies(self, ireqs):
-        non_editable = [ireq for ireq in ireqs if not ireq.editable]
+        non_editable = [ireq for ireq in ireqs if not (ireq.editable or is_url_requirement(ireq, self.repository))]
         return self.dependency_cache.reverse_dependencies(non_editable)
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -71,6 +71,18 @@ def make_install_requirement(name, version, extras, constraint=False):
         constraint=constraint)
 
 
+def is_url_requirement(ireq, repository=None):
+    """
+    Finds if a requirement is a URL
+    """
+    if not ireq.link or 'pypi.python.org' in ireq.link.url or ireq.link.url.startswith('file'):
+        return False
+    if repository is not None and hasattr(repository, 'finder'):
+        if any(index_url in ireq.link.url for index_url in repository.finder.index_urls):
+            return False
+    return True
+
+
 def format_requirement(ireq, marker=None):
     """
     Generic formatter for pretty printing InstallRequirements to the terminal
@@ -78,6 +90,8 @@ def format_requirement(ireq, marker=None):
     """
     if ireq.editable:
         line = '-e {}'.format(ireq.link)
+    elif is_url_requirement(ireq):
+        line = ireq.link.url
     else:
         line = str(ireq.req).lower()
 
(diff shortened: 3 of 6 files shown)


piptools/utils.py
```diff
@@ -71,13 +71,22 @@ def make_install_requirement(name, version, extras, constraint=False):
         constraint=constraint)
 
 
+def is_url_requirement(ireq):
+    """
+    Finds if a requirement is a URL
+    """
+    return bool(ireq.link and 'pypi.python.org' not in ireq.link.url and not ireq.link.url.startswith('file'))
```
vphilippon: Is there a reason to exclude `pypi.python.org` here? 
If its needed to exclude pypi-hosted packages, the we need to exclude the configured pypi url too for users using their local devpi instances and such.

piptools/resolver.py
```diff
@@ -85,9 +84,6 @@ def resolve(self, max_rounds=10):
             self.dependency_cache.clear()
             self.repository.clear_caches()
 
-        self.check_constraints(chain(self.our_constraints,
```
vphilippon: We still have a limitation on constraints: We don't support file link (as checked in `is_url_requirement`), so we still need to make a check to warn the user, both here and in `pip-sync`.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Features:
 - `--generate-hashes` now generates hashes for all wheels,
 not only wheels for the currently running platform ([#520](https://github.com/jazzband/pip-tools/pull/520)). Thanks @jdufresne
 - Added a `-q`/`--quiet` argument to the pip-sync command to reduce log output.
+- Support URLs as packages.
```
vphilippon: This will go in a `1.11.0` section on top.

nim65s:
Hi,

Thanks for the review :)

I tried to address your comments, but after that I noticed that there are other PR working on a similar goal, thanks to the label. What should we do about that ?

vphilippon:
Thanks for the comeback.
For the other PRs, I intended to go through them, check what each did, and cross-check them to check if one of them covered a corner case another one didn't and making sure we got the best fix/solution at the end.

I'll get on re-reviewing this as soon as I got time to. Also, if you want to go ahead and make that cross-check right yourself, that will certainly help out.

piptools/utils.py
```diff
@@ -71,13 +71,27 @@ def make_install_requirement(name, version, extras, constraint=False):
         constraint=constraint)
 
 
+def is_url_requirement(ireq, repository=None):
+    """
+    Finds if a requirement is a URL
+    """
+    if not ireq.link or 'pypi.python.org' in ireq.link.url or ireq.link.url.startswith('file'):
```
vphilippon: I might be missing something, but `ireq.link` will only have a value when using files uri or vcs urls, right?
If that's the case, why check for the presence of index urls and discard?

nim65s: `ireq.link` can also give the link to download a requirement. For example, for `six==1.10.0`, it can be `https://pypi.python.org/packages/c8/0a/b6723e1bc4c516cb687841499455a8505b44607ab535be01091c0f24f079/six-1.10.0-py2.py3-none-any.whl#md5=3ab558cf5d4f7a72611d59a81a315dc8`

nim65s: And if I remove `or 'pypi.python.org' in ireq.link.url`, I get some tests failures  because of that:
```
E           AssertionError: assert 'six==1.10.0' in '#\n# This file is autogenerated by pip-compile\n# To update, run:\n#\n#    pip-compile --output-file requirements.txt...5be01091c0f24f079/six-1.10.0-py2.py3-none-any.whl#md5=3ab558cf5d4f7a72611d59a81a315dc8\nDry-run, so nothing updated.\n'
E            +  where '#\n# This file is autogenerated by pip-compile\n# To update, run:\n#\n#    pip-compile --output-file requirements.txt...5be01091c0f24f079/six-1.10.0-py2.py3-none-any.whl#md5=3ab558cf5d4f7a72611d59a81a315dc8\nDry-run, so nothing updated.\n' = <Result okay>.output
```

nim65s:
If I understand correctly,  #405 should handle urls like `https://github.com/django/django/archive/master.zip#egg=django` but it doesn't (see https://github.com/jazzband/pip-tools/pull/405#issuecomment-350464378), and #372 supports only vcs urls.

I checked with django's zip, and it puts `django==2.0` in `requirements.txt`, no matter which branch was in `requirements.in`.

Therefore, I think that this PR handles more cases, and correctly, but I should add two tests instead of one: 
`test_url_package` & `test_url_package_vcs`, in order to check the same things as in `test_editable_package` & `test_editable_package_vcs`

piptools/utils.py
vphilippon: Ok, I'll have to do some more tests on this, as `ireq.link` seemed to be always `None` when I was installing packages from PyPi. 
If the Pypi url can be present like you said, then the logic here cannot simply exclude `pypi.python.org`, because it's not the only possible PyPi server. We have `pypi.org`, as well as any private pypi server defined by the user.

vphilippon:
I did some more tests with the PR and noted something: The URLs in the resulting `requirements.txt` are not "pinned" in any way. This means that if someone puts `git+https://github.com/jazzband/pip-tools.git@master#egg=pip-tools` in the `requirements.in`, it stays as-is in the `requirements.txt`. `master` is not converted to the SHA of the current HEAD of master. We can't have that in the `requirements.txt`: When the HEAD of master will change, the `requirements.txt` will become invalid, as if we had any unpinned dependencies in it.

The behavior of `pip` is that if you do 
`pip install git+https://github.com/jazzband/pip-tools.git@master#egg=pip-tools`
then `pip freeze` will give something like
```
click==6.7
first==2.0.1
pip-tools==1.11.1.dev3+ga2446fa
six==1.11.0
```
which gives the "version" of `pip-tools` on master at that moment. But `pip` looses the "source" of the dependency in that case, which is not ok in our case. That means `pip-tools` might need its own logic to determine the SHA (and implement a logic for the various supported VCS, `git` is not the only one out there). We might be able to reuse the logic that `pip` has that pins the SHA when installing with `-e`.
(This suddenly seems to explain why only the editable mode was ever supported in `pip-tools` with VCS URLs)

It looks like there's a fair amount of stuff that needs to be dealt with for this feature.

piptools/utils.py
nim65s: Aren't those private pypi servers in `index_urls` ? This case is  adressed two lines later

nim65s:
I see, you're right.
I don't see any easy solution right now :/

twig:
I'll add my 2c here as this one is more recent activity and tests (although I prefer the simplicity of the older attempt https://github.com/jazzband/pip-tools/pull/372)

I was digging around the pip-tools src because of the same frustrations here.

Would it be sufficient if piptools...:

* checked if the `constraint.link.egg_fragment` had the right information for a pinned version `u'package-name==1.2.3'`
* downloaded the archive using `_vendored.pip.download.unpack_url`

> def unpack_url(link, location, download_dir=None,
>               only_download=False, session=None, hashes=None):

* look at `setup.py` in the extracted folder and grab the requirements from there (I couldn't find a function for doing this)
* add the requirements to the generated `requirements.txt` file
* add the CVS/URL requirements to `requirements.txt` so pip would install it as it normally would
* now `pip-sync` should have what it needs to check if a pinned package is installed at the right version or not

Don't have much experience with python package managers so my comment may come across as naive, but feel free to correct me if I'm wrong.

jcushman:
This work was included in #807 -- thanks!

PR #577, opened 2017-10-16
Rename [wheel] section to [bdist_wheel] as the former is legacy

jdufresne:
For additional details, See:

https://bitbucket.org/pypa/wheel/src/54ddbcc9cec25e1f4d111a142b8bfaa163130a61/wheel/bdist_wheel.py?fileviewer=file-view-default#bdist_wheel.py-119:125

http://pythonwheels.com/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,2 @@
-[wheel]
+[bdist_wheel]
 universal = 1


vphilippon:
Thanks @jdufresne, small steps forward!

PR #578, opened 2017-10-16
Include BSD license at the top level of the project

jdufresne:
It is customary for open source projects to include the license at the top level of the project so users can easily know integration is compatible with other projects.

Also include license file in the generated wheel package. The wheel package format supports including the license file. This is done using the `[metadata]` section in the `setup.cfg` file. For additional information on this feature, see:

https://wheel.readthedocs.io/en/stable/index.html#including-the-license-in-the-generated-wheel-file

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,26 @@
+Copyright (c). All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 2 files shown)


vphilippon:
Thanks @jdufresne !

PR #584, opened 2017-10-24
Recommend installation into virtual environment

### LINKED ISSUE #582: Clarification re: installation
The README suggests `pip install pip-tools`, but it's not clear what environment that should be run in. Some of the options that I see:

1. System Python (e.g., distributed with OS): generally not recommended
2. Third-party Python (e.g., Homebrew, python.org, pyenv, Anaconda)
3. Virtual environment dedicated to pip-tools
4. Each project's virtual environment

In my case, I chose option 3, using [pipsi](https://github.com/mitsuhiko/pipsi) to streamline adding it to my PATH. I've installed other tools like fabric and cookiecutter in a similar manner, whereas I install things like flake8 using option 4.

I think option 1 or 2 would be safer with `pip install --user`.

bhrutledge:
Fix #582. Does this need tests or a CHANGELOG entry?

##### Contributor checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,16 @@ even when you've pinned them.  [You do pin them, right?][0]
 Installation
 ============
 
-As part of a Python project's environment tooling (similar to `pip`), it's recommended to install `pip-tools` in each project's [virtual environment](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments):
+As part of a Python project's environment tooling (similar to `pip`), it's recommended to install `pip-tools` in each
+project's [virtual environment][1]:
 
 ```console
 $ source /path/to/venv/bin/activate
 (venv) $ pip install --upgrade pip  # pip-tools needs pip==8.0 or higher (!)
 (venv) $ pip install pip-tools
 ```
+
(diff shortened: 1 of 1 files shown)


bhrutledge:
@vphilippon Any suggestions on what else this PR needs to be merge-ready? I haven't dug into why the Travis build failed.

vphilippon:
@bhrutledge Sorry for the delay, trying to catch up on everything.
What you wrote seems pretty good to me.

No need for changelog/tests on this.
For the build, that's fixed in master now.

You'll have to rebase on master, and we converted the readme to .rst meanwhile, so it essentially means you'll have to rewrite your thing, sorry about that.

Ping me as soon as your done, it should get right in.
Thanks!

README.md
```diff
@@ -120,7 +126,7 @@ Example usage for `pip-sync`
 ============================
 
 Now that you have a `requirements.txt`, you can use `pip-sync` to update your
-virtual env to reflect exactly what's in there.  Note: this will
+virtual environment to reflect exactly what's in there.  Note: this will
 install/upgrade/uninstall everything necessary to match the `requirements.txt`
 contents.
```
merwok: To make things explicit, you could add here «except for setuptools, pip and pip-tools itself.»

PR #588, opened 2017-10-26
--exclude support

s0undt3ch:
Describe the changes here.

##### Contributor checklist

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -42,7 +42,8 @@ class RequirementSummary(object):
 
 
 class Resolver(object):
-    def __init__(self, constraints, repository, cache=None, prereleases=False, clear_caches=False, allow_unsafe=False):
+    def __init__(self, constraints, repository, cache=None, prereleases=False,
+                 clear_caches=False, allow_unsafe=False, exclude=None):
         """
         This class resolves a given set of constraints (a collection of
         InstallRequirement objects) by consulting the given Repository and the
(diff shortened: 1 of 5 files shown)


piptools/resolver.py
```diff
@@ -289,6 +291,15 @@ def _iter_dependencies(self, ireq):
         dependency_strings = self.dependency_cache[ireq]
         log.debug('  {:25} requires {}'.format(format_requirement(ireq),
                                                ', '.join(sorted(dependency_strings, key=lambda s: s.lower())) or '-'))
+
+        if self.exclude:
+            dep_set = {str(self.get_best_match(InstallRequirement.from_line(e))) for e in dependency_strings}
```
vphilippon: Calling `get_best_match` here means that the lookup on PyPi will still be done, and an error will occur if the packages isn't found. One of the main use case I see of `--exclude` would be to deal with the lack of proper environment marker handling we have right now, but it would't work in this case.

Is there a way to get the same logic without doing the lookup? Alternatively, catching and ignoring the `NoCandidateFound` error could be enough for now.

piptools/resolver.py
```diff
@@ -289,6 +291,15 @@ def _iter_dependencies(self, ireq):
...
+            exclude_set = {str(self.get_best_match(InstallRequirement.from_line(e))) for e in self.exclude}
+            log.debug('Excluding following dependencies:')
+            for exclude in (dep_set & exclude_set):
+                log.debug('  {}'.format(exclude))
+            dependency_strings = dep_set - exclude_set
```
vphilippon: By doing this, the dependencies becomes strictly pinned version to the best current candidate, which will make the resolving fail if there's a need of changing candidate.

Try the following requirements.in:
```
docker-compose
requests
```
It will fail with something like `Could not find a version that matches requests==2.11.1,==2.18.4`.

atugushev:
Closed due to a lack of response.

PR #605, opened 2017-11-20
Making README.rst PyPi compliant

### LINKED ISSUE #594: README broken on PyPI (must be reStructuredText)
The [package description](https://pypi.python.org/pypi/pip-tools/) on PyPI is unreadable since PyPI expects the README in [reStructuredText](http://www.sphinx-doc.org/en/stable/rest.html) file format and we use MarkDown.

Solution A: Convert to reST
---------------------

1. Rename the current `README.md` to `README.rst`
1. Replace the markdown of the badges and the code samples ([example](https://github.com/Organice/djangocms-maps/blob/master/README.rst))
1. Add a `long_description=read_file('README.rst')` line to `setup.py` ([example](https://github.com/Organice/djangocms-maps/blob/master/setup.py#L50))

Solution B: Process before Upload
-------------------

1. Integrate [pypandoc](https://pypi.python.org/pypi/pypandoc) in `setup.py` ([example](https://github.com/jrief/djangocms-cascade/blob/master/setup.py#L7-L14))
1. Add a `long_description=convert('README.md', 'rst')` line to `setup.py` ([example](https://github.com/jrief/djangocms-cascade/blob/master/setup.py#L49))

------------

Both solutions above will render a nicely formatted, HTML-styled package description on PyPI.

Quality Assurance
--------------

Optionally, you may check your README with [checkdocs](https://github.com/Organice/djangocms-maps/blob/master/tox.ini#L13-L19) before uploading the package to PyPI, because sometimes the reST-to-HTML conversion that PyPI uses fails -- and renders a still hard-to-read, broken, unformatted package description.

vphilippon:
Should fix the PyPi README.rst render.
Fixes #594

##### Contributor checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,7 +22,7 @@ even when you've pinned them.  `You do pin them, right?`_
 Installation
 ============
 
-.. code-block:: bash
+.. code-block::
 
     $ pip install --upgrade pip  # pip-tools needs pip==8.0 or higher (!)
     $ pip install pip-tools
@@ -38,7 +38,7 @@ Suppose you have a Flask project, and want to pin it for production.
 If you have a ``setup.py`` with ``install_requires=['Flask']``, then run
 ``pip-compile`` without any arguments:
 
-.. code-block:: bash
+.. code-block::
 
     $ pip-compile
     #
@@ -64,14 +64,14 @@ Without setup.py
 If you don't use ``setup.py`` (`it's easy to write one`_), you can instead
 write the following line to a file:
 
-.. code-block:: ini
+.. code-block::
 
     # requirements.in
     Flask
 
 This time, run ``pip-compile requirements.in``:
 
-.. code-block:: bash
+.. code-block::
 
     $ pip-compile requirements.in
     #
@@ -99,7 +99,7 @@ Using hashes
 If you would like to use *Hash-Checking Mode* available in *pip* since version
 8.0, ``pip-compile`` offers ``--generate-hashes`` flag:
 
-.. code-block:: bash
+.. code-block::
 
     $ pip-compile --generate-hashes requirements.in
     #
@@ -138,7 +138,7 @@ To update all packages, periodically re-run ``pip-compile --upgrade``.
 To update a specific package to the latest or a specific version use the
 ``--upgrade-package`` or ``-P`` flag:
 
-.. code-block:: bash
+.. code-block::
 
     $ pip-compile --upgrade-package flask  # only update the flask package
     $ pip-compile --upgrade-package flask --upgrade-package requests  # update both the flask and requests packages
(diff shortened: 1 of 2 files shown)


merwok:
rst2html is part of docutils; rst2html5 looks like an independent project.

The `setup.py check` command provided by distutils can check the validity of long_description if docutils is installed: https://github.com/caravancoop/configstore/blob/master/tox.ini#L52

vphilippon:
Oddly, `python setup.py --restructuredtext --strict` would fail with a parse error before my change. But now I'm testing this on the master branch (without this change) from home, and it passes.

I'll look deeper into this.

vphilippon:
Ha! I think I found it: `..code-block:: <bash|tox>` requires to have pygments installed, which isn't installed on the PyPi server. We have to fall back to the default `.. code-block`.

That falls in that category of issues:
> Usage of non-builtin lexers (e.g. bibtex) will pass locally but not be recognized/parsable on PyPI

jezdez:
@vphilippon I recommend adding a tox test env that installs [readme_renderer](https://github.com/pypa/readme_renderer) which is the official renderer of readmes on PyPI. That way you'll know that it'll work before you release. Here's [an example](https://github.com/mozilla-services/python-dockerflow/blob/6c1005befa3058262f6ad40ce00463e7bfed5acf/tox.ini#L35-L37).

vphilippon:
So I tested with `readme_renderer`, and it did not find any error in the README.rst in master (which fails), so it's not enough.

Looking at the `readme_renderer` ReadMe, it's said that the tool is for upload toward Warehouse, while we're still using the legacy-pypi (unless I've missed an update). Is it still worth adding another flow for the future here? Should we wait? I'm not sure, lets address this separately.

Here's what I'll do: I'll merge this and make another RC to check if it fixes this.
If not, we'll assemble again! *superhero pose*

Thanks for your help everyone. Please keep coming back :)

jezdez:
@vphilippon Just a FYI: it's also used in the "legacy-pypi" project: https://github.com/pypa/pypi-legacy/blob/f45232ddf7c4c501d0b2f9a6e88f136973199718/webui.py#L2001-L2008

vphilippon:
@jezdez Thanks, that's good that's good to know. Then again, I'm left wondering why the rendering failed if this same official tool is used on the PyPi server.

vphilippon:
So I just tried out doing the same as legacy-pypi with `render_readme.rst.render`, and if I uninstall `pygments`, it returns `None`, which would fallback to `render_readme.txt.render`, which matches pretty well what we see on the page.

As a bonus, I've also checked on Warehouse: No rendering either.
https://pypi.org/project/pip-tools/1.10.2rc1/

I'm not sure why `pygments` wouldn't be installed on legacy-pypi or Warehouse. I'll try to find where to report issues/questions about the live instances. That is if this PR fixes the rendering.

PR #614, opened 2017-11-24
Slightly improved NoCandidateFound error message.

### LINKED ISSUE #352: pip-compile gives misleading error message if no internet
If there is no internet access, `pip-compile` will fail with a misleading message. This is with pip-compile 1.6.1, pip 8.1.1, and Python 3.5.1.
##### Steps to replicate
1. Create a `requirements.in` file with a single project name.
2. Disable internet access (e.g. by disabling wi-fi).
3. Run `pip-compile --verbose requirements.in`.
##### Expected result

The command should fail with an error message relating to the fact that there is no internet connection.
##### Actual result

Instead, the output looks something like this:

```
Using indexes:
  https://pypi.python.org/simple

                          ROUND 1                           
Current constraints:
  MyPackage

Finding the best candidates:
Could not find a version that matches MyPackage
Tried: (no version found at all)
```

vphilippon:
This should help a little, making the message less cryptic.
Still no "source of conflict" yet.
Fixes #352

**Changelog-friendly one-liner**: Slightly improved the NoCandidateFound error message.

##### Contributor checklist

### DIFF
--- a/piptools/exceptions.py
+++ b/piptools/exceptions.py
@@ -3,9 +3,10 @@ class PipToolsError(Exception):
 
 
 class NoCandidateFound(PipToolsError):
-    def __init__(self, ireq, candidates_tried):
+    def __init__(self, ireq, candidates_tried, index_urls):
         self.ireq = ireq
         self.candidates_tried = candidates_tried
+        self.index_urls = index_urls
 
     def __str__(self):
         sorted_versions = sorted(c.version for c in self.candidates_tried)
@@ -13,6 +14,12 @@ class NoCandidateFound(PipToolsError):
             'Could not find a version that matches {}'.format(self.ireq),
             'Tried: {}'.format(', '.join(str(version) for version in sorted_versions) or '(no version found at all)')
         ]
+        if sorted_versions:
+            lines.append('There are incompatible versions in the resolved dependencies.')
+        else:
+            lines.append('{} {} reachable?'.format(
+                'Were' if len(self.index_urls) > 1 else 'Was', ' or '.join(self.index_urls))
+            )
         return '\n'.join(lines)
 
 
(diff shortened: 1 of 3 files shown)


piptools/exceptions.py
```diff
@@ -13,6 +13,10 @@ def __str__(self):
             'Could not find a version that matches {}'.format(self.ireq),
             'Tried: {}'.format(', '.join(str(version) for version in sorted_versions) or '(no version found at all)')
         ]
+        if sorted_versions:
+            lines.append('You probably have incompatible dependencies.')
```
davidovich: I would try to take a more neutral tone, as this seems to be accusing. How about: 
There are incompatible versions in solved versions of dependencies

vphilippon: You're right, I'll change that

vphilippon: Updated, I rephrased a bit

piptools/exceptions.py
```diff
@@ -13,6 +13,10 @@ def __str__(self):
             'Could not find a version that matches {}'.format(self.ireq),
             'Tried: {}'.format(', '.join(str(version) for version in sorted_versions) or '(no version found at all)')
         ]
+        if sorted_versions:
+            lines.append('There are incompatible versions in the resolved dependencies.')
+        else:
+            lines.append('Was the PyPi repository reachable?')
```
davidovich: We know which pypi repository was tried, I would add it here. If it is an internal server, this will help diagnose errors vs public pypi. Small nit:

`Is the $(pypi) repo reachable?`

vphilippon: There we go

davidovich:
Much better ;-) LGTM

PR #615, opened 2017-11-25
Recommend installation into virtual environment

### LINKED ISSUE #582: Clarification re: installation

bhrutledge:
And other assorted formatting/language tweaks.

Fix #582, supersedes #584.


bhrutledge:
@vphilippon Here's the updated PR.

vphilippon:
Thank you, this is really appreciated!

PR #621, opened 2017-12-16
Always use _compat to import TemporaryDirectory

jdufresne:
`_compat/__init__.py` already handles the differences between Python 2 & 3 with the code:

```py
if six.PY2:
    from .tempfile import TemporaryDirectory
    ...
else:
    from tempfile import TemporaryDirectory
    ...
```

There is no reason to redo this logic in files importing TemporaryDirectory.


vphilippon:
Thanks for the code improvement 👍

Issue #521, opened 2017-05-30
pip-compile with generate hashes is very slow


graingert:
Could probably do with a progress bar

##### Environment Versions

1. OS Type Ubuntu
1. Python version: `$ python -V` 3.6
1. pip version:  `$ pip --version` n/a
1. pip-tools version: `$ pip-compile --version` master

##### Steps to replicate

use geenrate-hashes with a fairly big package graph

##### Expected result

generated hashes turn up in requirements.txt and pip-compile does not appear to hang
##### Actual result

generated hashes turn up in requirements.txt but pip-compile appears to hang

proinsias:
I agree that a progress bar would be useful for normal operations. I would also suggest verbose output similar to the verbose output of `hashin` when `-v` is enabled.

vphilippon:
The situation seems to have improved to a reasonable level with #557.
In `--verbose` mode, we could use some output during the hashing though.

kaeptmblaubaer1000:
A progress bar would still be nice.

atugushev:
Hello @kaeptmblaubaer1000,

I've added a PR #743 for that feature. I hope it'll be merged.

RonRothman:
Question: Is there currently a way to see the progress _without_ turning on verbose output (which contains a lot of details beyond what's needed to indicate progress).

atugushev:
Hello @RonRothman,

Unfortunately there is no way to see the progress without --verbose flag.

graingert:
Would it be possible to add in-between verbosity levels?

graingert:
@RonRothman do you have an example verbose output with what you want to see Vs what you don't want to see?

PR #641, opened 2018-03-08
Cache hashes for `--generate-hashes`

justicz:
Improves the speed of `pip-compile --generate-hashes` by caching the hashes we already have.

If hashes for a particular dependency have already been generated, and are available in the output file, this PR skips pulling all of the packages again from PyPI, dramatically increasing the speed of `pip-compile` when the `--generate-hashes` flag is passed.

This has the added benefit of improving security; if `pip-compile --generate-hashes` is run regularly as part of a build pipeline, hashes will be guaranteed stay the same unless the version has been explicitly updated.

Helps resolve https://github.com/jazzband/pip-tools/issues/521

**Maintainer Edit** - **Changelog**:
Improved the speed of `pip-compile --generate-hashes` by caching the hashes from an existing output file.

### DIFF
--- a/piptools/repositories/local.py
+++ b/piptools/repositories/local.py
@@ -6,6 +6,10 @@ from contextlib import contextmanager
 
 from piptools.utils import as_tuple, key_from_req, make_install_requirement
 from .base import BaseRepository
+try:
+    from pip.utils.hashes import FAVORITE_HASH
+except ImportError:
+    FAVORITE_HASH = 'sha256'
 
 
 def ireq_satisfied_by_existing_pin(ireq, existing_pin):
@@ -64,6 +68,16 @@ class LocalRequirementsRepository(BaseRepository):
         return self.repository.get_dependencies(ireq)
 
     def get_hashes(self, ireq):
+        key = key_from_req(ireq.req)
+        existing_pin = self.existing_pins.get(key)
+        if existing_pin and ireq_satisfied_by_existing_pin(ireq, existing_pin):
+            hashes = existing_pin.options.get('hashes', {})
+            hexdigests = hashes.get(FAVORITE_HASH)
+            if hexdigests:
+                return {
+                    ':'.join([FAVORITE_HASH, hexdigest])
+                    for hexdigest in hexdigests
+                }
         return self.repository.get_hashes(ireq)
 
     @contextmanager
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -218,6 +218,16 @@ def dedup(iterable):
     return iter(OrderedDict.fromkeys(iterable))
 
 
+def name_from_req(req):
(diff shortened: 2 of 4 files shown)


vphilippon:
Thanks for that first contribution @justicz, nice work!

graingert:
does this include the new hash of the new wheel in the case that someone uploads a .tar.gz, then later a wheel?

justicz:
This change doesn’t handle the case where new distributions are uploaded
later for the same version. Is that going to be a problem?
It’s definitely a good observation, but IMO that sounds like it will  be
a pretty uncommon issue.

graingert:
I often hit this scenario. Because I regularly find packages without wheels and I go and poke the maintainer to go release them

justicz:
OK. In that particular case, would it be acceptable for you to just delete the hashes in the generated output file for that package? Then they'll just be regenerated the usual, slow way. I would tend to think that this change provides the better UX 99% of the time.

justicz:
Also, and maybe this is a discussion for somewhere else -- I think it's worth mentioning that if `--generate-hashes` is run regularly, and the hashes are coming from fresh downloads of every distribution each time like they were before this change, there's not much of a security benefit to pinning hashes.

If your threat model is PyPI or a maintainer's machine getting compromised, this PR improves the security benefit of `--generate-hashes` significantly, since it means that the code associated with a particular package/version pair will never change.

Appending hashes for new distributions dynamically actually cancels out that security benefit, since an attacker could just upload a different distribution type containing their malware.

vphilippon:
A `pip-compile` with an existing `requirements.txt` would not have updated the versions (pins) unless required to, or explicitly told to do so with `--upgrade` or `--upgrade-package`. I think this PR actually aligns `--generate-hashes` with that: No changes unless required or explicitly told to do so.
Now we might want to add and option analog to  `--upgrade`/`--upgrade-package` regarding hashes. Maybe something like  `--regenerate-hashes`. Manual edits to the compiled `requirements.txt` can lead to a lot of trouble, and I usually advise against it (warranty void if touched 😛 ) .

(btw @graingert thanks for the observation, and for poking maintainers for wheels too 😄 )

justicz:
> A pip-compile with an existing requirements.txt would not have updated the versions (pins) unless required to, or explicitly told to do so with --upgrade or --upgrade-package.

Good to know! My mistake, I guess the security benefit I mentioned earlier is actually already the default behavior (modulo the fact that the new change won't add hashes for new distributions).

vphilippon:
@justicz Your point on increased security sounds right to me, but don't quote me on this, I don't consider myself enough of a security guy to argue that it's better or worse in terms of security.
What I can say though is that I prefer this new behavior regarding how and when the `requirements.txt` and its content gets to be modified by a `pip-compile`.

PR #644, opened 2018-03-21
Vendor pip 9.0.3

### LINKED ISSUE #580: `pip` 10 Coming - Lots of API Changes
We should double check pip-compile etc. still works with the announced upcoming pip 10 release. A lot of APIs have been moved to internal. We should see what breaks with the pip-tool suiete.

##### Environment Versions

OS Type - Any
Python version: 3.6.3
pip version:  HEAD
pip-tools version: HEAD

##### Steps to replicate

1. Install HEAD build of pip and see wht happens

##### Expected result

Some API moves and maybe change what APIs are called from pip-tools.

vphilippon:
Vendoring pip 9.0.3 inside of pip-tools.
This will ensure the dependency resolution code will work even if users install pip 10.
The install/uninstall operations of `pip-sync` as still done through shell out, so they will benefit directly from pip 10, as those command will still be available.

Fixes #580
This should be release in a Major 2.0.0 version.

**Changelog-friendly one-liner**: Vendored pip 9.0.3 to keep compatibility for users with pip 10

##### Contributor checklist


vphilippon:
I didn't change the test matrix to stop testing on multiple pip version right away. I want to make sure that it still works with earlier pip versions (as I took away some conditional code). If it passes, I could clear the matrix to make CI faster.

vphilippon:
@pradyunsg FYI

vphilippon:
There seems to be some issue when running through `coverage`, I'll get digging (likely tomorrow).

vphilippon:
- [ ] Don't forget to Vendor pip 9.0.3 now that its out, while at it.

piptools/utils.py
```diff
@@ -45,10 +27,10 @@ def key_from_ireq(ireq):
 def key_from_req(req):
     """Get an all-lowercase version of the requirement's name."""
     if hasattr(req, 'key'):
-        # pip 8.1.1 or below, using pkg_resources
+        # from pkg_resources, such as installed dists for pip-sync
```
davidovich: I believe this will be stable now that we use vendored pip

piptools/utils.py
```diff
@@ -221,10 +203,10 @@ def dedup(iterable):
 def name_from_req(req):
     """Get the name of the requirement"""
     if hasattr(req, 'project_name'):
-        # pip 8.1.1 or below, using pkg_resources
```
davidovich: same: I believe this will be stable now that we use vendored pip

piptools/utils.py
vphilippon: It thought so too, but unfortunately there's another condition unrelated to the pip version.
When parsing the *installed* packages (in pip-sync), the returned object by `pip.get_installed_distributions` is a `DistributionEggInfo` or something alike, which comes from `pkg_resources`, and those object do not have a `.name` attribute.
The `if hasattr(req, 'key')` condition masked that case, making us believe it was only a pip8-9 difference.

piptools/utils.py
vphilippon: See comment above: same applies.

piptools/utils.py
davidovich: ok

piptools/utils.py
davidovich: ok

vphilippon:
For the record: I installed pip 10 (commit https://github.com/pypa/pip/tree/b20235d7c9e6a7ece969567ad0a21b55aea96624), and both pip-compile and pip-sync work fine so far. I went as far as to uninstall pip to test pip-compile, and it keeps working.

Look like we'll survive through this!

PR #647, opened 2018-04-03
Filter pip environment markers

JoergRittinger:
Add filter for pip environment markers

**Changelog-friendly one-liner**: Bugfix: Handling of pip environment markers

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -174,6 +174,10 @@ def cli(verbose, dry_run, pre, rebuild, find_links, index_url, extra_index_url,
             constraints.extend(parse_requirements(
                 src_file, finder=repository.finder, session=repository.session, options=pip_options))
 
+    # Filter out pip environment markers which do not match (PEP496)
+    constraints = [req for req in constraints
+                   if req.markers is None or req.markers.evaluate()]
+
     # Check the given base set of constraints first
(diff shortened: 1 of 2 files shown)


JoergRittinger:
Thx @vphilippon  for you feedback. I tried an implementation which puts the package with the environment marker into the requirements.txt but excludes it from repository lookups and dependency searches. But it will get inconsistent anyway.

If I have the requirments.in with:
`unknown_python_3_package; python_version < '3'`
I will end up with the requirements.txt in python 2:
`unknown_python_3_package==1.0 ; python_version < '3'`
and with the requirements.txt in python 3:
`unknown_python_3_package ; python_version < '3'`

Note that in python I can find a version on my Repository server for python 2, but for python 3 it is not available and therefore can't be pinned.

If you want I can push my changes for further discussion.

vphilippon:
@JoergRittinger Thanks for the detailed example.

Your original idea seems to be the best one in the end.
Plus, it will keep the environment markers in the `requirements.txt` when the marker evaluates as true, which should cover some easy cross-environment cases, as long as the pip-compile is done on the python version that includes the additional packages (FYI @barrywhart for your usage).

vphilippon:
You're right, that's what happens with late night releases. Thanks for pointing it out

PR #651, opened 2018-04-16
Added cross-environment usage documentation to README.rst

### LINKED ISSUE #563: Environment markers do not carry out to transitive dependencies

### LINKED ISSUE #585: Ability to Cross Compile
Support the ability to run `pip-compile` specifying the OS / architecture that should be used for resolving dependencies. Currently it uses the OS where the `pip-compile` is run. This causes issues such as https://github.com/jazzband/pip-tools/issues/333. It also means that if a package does not exist on the current OS (eg for `tensorflow-gpu` on MacOS`), then compile fails.

##### Environment Versions

1. OS Type
MacOS
1. Python version: `$ python -V`
`Python 3.5.3`
1. pip version:  `$ pip --version`
`pip 9.0.1`
1. pip-tools version: `$ pip-compile --version`
`pip-compile, version 1.9.0`
##### Steps to replicate

1. Add `tensorflow-gpu>=1.2` to requirements.in
2. `pip-compile`

##### Expected result
requirements.txt file with pinned deps. (assuming that `--arch manylinux1_x86_64` was set)

##### Actual result
```
Could not find a version that matches tensorflow-gpu>=1.2
Tried: 0.12.0rc1, 0.12.0, 0.12.1, 1.0.0, 1.0.1, 1.1.0rc1, 1.1.0rc2, 1.1.0
```

### LINKED ISSUE #635: 1.11.0 fails to compile mypy on python 2.7
I believe this is related to how environment markers are handled.

With `requirements.in` of:

    mypy >= 0.560; python_version >= '3.2'

On python 2.7, run:

    pip-compile --no-index --rebuild -v requirements.in

Produces:

```shell
Using indexes:
  ---snip---

                          ROUND 1                           
Current constraints:
  mypy>=0.560

Finding the best candidates:
  found candidate mypy==0.560 (constraint was >=0.560)

  mypy==0.560 not in cache, need to check index
  File "/home/musttu/.virtualenvs/wires/bin/pip-compile", line 11, in <module>
    sys.exit(cli())
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/click/core.py", line 722, in __call__
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/click/core.py", line 895, in invoke
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/click/core.py", line 535, in invoke
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/piptools/scripts/compile.py", line 187, in cli
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/piptools/resolver.py", line 102, in resolve
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/piptools/resolver.py", line 199, in _resolve_one_round
    for dep in self._iter_dependencies(best_match):
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/piptools/resolver.py", line 285, in _iter_dependencies
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/piptools/repositories/local.py", line 64, in get_dependencies
    return self.repository.get_dependencies(ireq)
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/piptools/repositories/pypi.py", line 157, in get_dependencies
    self._dependencies_cache[ireq] = reqset._prepare_file(self.finder, ireq)
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/pip/req/req_set.py", line 634, in _prepare_file
    abstract_dist.prep_for_dist()
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/pip/req/req_set.py", line 129, in prep_for_dist
    self.req_to_install.run_egg_info()
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/pip/req/req_install.py", line 439, in run_egg_info
    command_desc='python setup.py egg_info')
  File "/home/musttu/.virtualenvs/wires/local/lib/python2.7/site-packages/pip/utils/__init__.py", line 707, in call_subprocess
    % (command_desc, proc.returncode, cwd))
pip.exceptions.InstallationError: Command "python setup.py egg_info" failed with error code 1 in /tmp/tmpreRyRhbuild/mypy/
```

On python 3.5, it works, and output even retains the environment marker:

    ...
    mypy==0.560 ; python_version >= "3.2"
    ...

This works on `1.10.x` but not on `1.11.0` or `1.9.x`.

##### Environment Versions

1. OS Type: `linux`
1. Python version: `2.7.12`
1. pip version:  `9.0.1`
1. pip-tools version: `1.11.0`

vphilippon:
Added some documentation regarding cross-environment usage of `requirements.txt`.
In short: Run `pip-compile` on each targeted environment.

Resolves #585
Resolves #563

**Changelog-friendly one-liner**: Add cross-environment usage documentation to `README`.

##### Contributor checklist

- ~Provided the tests for the changes.~

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,8 +88,7 @@ If you have a ``setup.py`` with ``install_requires=['django']``, then run
     sqlparse==0.3.0           # via django
 
 ``pip-compile`` will produce your ``requirements.txt``, with all the Django
-dependencies (and all underlying dependencies) pinned.  You should put
-``requirements.txt`` under version control.
+dependencies (and all underlying dependencies) pinned.
 
 Without ``setup.py``
 --------------------
@@ -119,8 +118,7 @@ Now, run ``pip-compile requirements.in``:
     sqlparse==0.3.0           # via django
 
 And it will produce your ``requirements.txt``, with all the Django dependencies
-(and all underlying dependencies) pinned.  You should put both
-``requirements.in`` and ``requirements.txt`` under version control.
+(and all underlying dependencies) pinned.
 
 .. _it's easy to write one: https://packaging.python.org/guides/distributing-packages-using-setuptools/#configuring-your-project
 
@@ -377,6 +375,42 @@ If you use multiple Python versions, you can run ``pip-sync`` as
 **Note**: ``pip-sync`` will not upgrade or uninstall packaging tools like
 ``setuptools``, ``pip``, or ``pip-tools`` itself. Use ``pip install --upgrade``
 to upgrade those packages.
+
+Should I commit ``requirements.in`` and ``requirements.txt`` to source control?
+===============================================================================
+
+Generally, yes. If you want a reproducible environment installation available from your source control,
+then yes, you should commit both ``requirements.in`` and ``requirements.txt`` to source control.
+
+Note that if you are deploying on multiple Python environments (read the section below),
+then you must commit a seperate output file for each Python environment.
+We suggest to use the ``{env}-requirements.txt`` format
+(ex: ``win32-py2.7-requirements.txt``, ``macos-py3.6-requirements.txt``, etc.).
+
+
+Cross-environment usage of ``requirements.in``/``requirements.txt`` and ``pip-compile``
+=======================================================================================
+
+The dependencies of a package can change depending on the Python environment in which it
+is installed.  Here, we define a Python environment as the combination of Operating
+System, Python version (2.7, 3.6, etc.), and Python implementation (CPython, PyPy,
+etc.). For an exact definition, refer to the possible combinations of `PEP 508
+environment markers`_.
+
+As the resulting ``requirements.txt`` can differ for each environment, users must
+execute ``pip-compile`` **on each Python environment separately** to generate a
+``requirements.txt`` valid for each said environment.  The same ``requirements.in`` can
+be used as the source file for all environments, using `PEP 508 environment markers`_ as
+needed, the same way it would be done for regular ``pip`` cross-environment usage.
+
+If the generated ``requirements.txt`` remains exactly the same for all Python
+environments, then it can be used across Python environments safely. **But** users
+should be careful as any package update can introduce environment-dependant
+dependencies, making any newly generated ``requirements.txt`` environment-dependant too.
+As a general rule, it's advised that users should still always execute ``pip-compile``
+on each targeted Python environment to avoid issues.
+
+.. _PEP 508 environment markers: https://www.python.org/dev/peps/pep-0508/#environment-markers
 
 Other useful tools
 ==================


vphilippon:
@tuukkamustonen @barrywhart @cancan101 @taion @JoergRittinger
FYI as you've been the ones discussing mostly discussing the cross-environment topic.
Do tell me if something isn't clear or if you have any questions!

README.rst
```diff
@@ -41,6 +41,29 @@ project's virtual environment.
 
 .. _virtual environment: https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments
 
+
+Cross-environment usage of ``requirements.in``/``requirements.txt`` and ``pip-compile``
```
tuukkamustonen: Maybe this should go to the bottom of the file instead (just before "Note about pip")? Let's not confuse readers by this advanced stuff until they actually know the basics?

README.rst
```diff
@@ -41,6 +41,29 @@ project's virtual environment.
...
+=======================================================================================
+
+The dependencies of a package can change depending on the Python environment in which it is installed.
```
tuukkamustonen: How about an example here, just to make it clear what we are talking about:

Consider this ```requirements.in```:

```
enum34; python_version < '3.4'
distro; platform_system == 'Linux'
```

Maybe not needed, but shouldn't hurt to show some real "code".

tuukkamustonen: Uh, scratch this. I was thinking about self-added environment markers now only, but this example would be just misleading. So nevermind.

taion:
Yeah, giving Python packaging limitations this does make sense.

This is incompatible with checking in `requirements.txt`, though, right?

README.rst
vphilippon: Agreed, I'll move that down

vphilippon:
@taion Good point about source control. That depends on the user's case I guess. I they deploy an app/service through CI/CD on a single defined Python environment (ex: Only Py3.6 on linux), then committing `requirements.txt` to source control makes sense. If they deploy on multiple environments from a single repo, then the requirements.txt shouldn't be commited to source control.

I'll figure a way to point that out w/o having to write a dissertation on CI/CD.

taion:
It's tricky, though. For example, we mostly develop on OS X, but our CI runs in a Linux Docker container.

vphilippon:
@taion Argh, yeah I see. If you don't mind me asking: what are you currently doing to deal with that?

taion:
Nothing – just running pip compile on OS X gives good-enough results. Not perfect, but even when there are mismatches, it’s good enough. This may well be due to specific pecularities with our setup, though (we now get `tensorflow-gpu` from the Docker image whenever we run on Linux, so we never need to put it in our requirements files).

vphilippon:
(Pushed but I still need to take care of the source control recommendations)

davidovich:
@vphilippon

> If they deploy on multiple environments from a single repo, then the requirements.txt shouldn't be commited to source control.

I'm not sure you should cristallise in doc not commiting requirements.txt to source control. Not doing so opens the door to non deterministic builds (because a compile phase is needed to re-create).

If you want to talk about this, I would instead mention that the files on different platforms should have different annotation (a platform suffix) corresponding to the source environment. That way you keep deterministic behavior for the different platforms.

taion:
To be more concrete, now that I'm at a laptop, in the one repo where this even comes up for us, we do:

```
tensorflow; 'linux' not in sys_platform
```

On Linux we use a Docker image with TensorFlow pre-installed.

It so happens to work, but it's obviously far from perfect.

In practice, though, with the frequency with which people develop on OS X but deploy on Linux, it's not clear to me that the recommendation here taken literally would improve DX.

Really this is the fault of Python packaging; it's not clear there's much that's possible here

JoergRittinger:
Looks good to me.

vphilippon:
@davidovich @tuukkamustonen @taion Updated, with a distinct section regarding commiting.

README.rst
```diff
@@ -226,6 +225,40 @@ If you use multiple Python versions, you can run ``pip-sync`` as
...
+Here, we define a Python environment as the combination of Operating System, Python version (2.7, 3.6, etc.),
+and Python implementation (CPython, PyPy, etc.). For an exact definition, refer to the possible combinations of
+`PEP 508 environment markers`_.
+
+As the resulting ``requirements.txt`` can differ for each environment, users must execute ``pip-compile``
```
taion: I would soften this to "may want to". I would imagine the vast majority of developers these days are using flows where they develop on OS X but deploy on Linux.

While it's technically true that there may be issues, it's very, very uncommon for these issues to actually arise. We have a pretty broad array of web, data engineering, data science, and machine learning projects, and this has only come up for a single project for us – and it's a project that has a sufficiently complicated deploy routine that we're definitely in a "warranties are void" place.

I would instead downgrade this to something like "what if my dependencies behave differently on different platforms?".

While in principle this can be a problem whenever people work across multiple platforms, in practice it's very seldom an actual problem.

vphilippon: > While it's technically true that there may be issues, it's very, very uncommon for these issues to actually arise.

My experience contributing and offering support with `pip-tools` and `pipenv`, which reached a wider part of Python users, was quite the opposite. OS mismatch issues did happen often enough to be a pain, and then there were the **Python version based mismatch** which are surprisingly common. If you feel like it, take a dive in `pip-tools` and `pipenv` past issues: its and adventure 😁 .

I expect that the number of libs that will only support py3 will increase over time as we're getting closer to 2020, which I also expect to increase the number of cases of difference in requirements between Python versions as the Py2-Py3 transition continues.

I'm not confortable switching this as a "what-if" warning. I think that the statement a few lines below stated the possibility of reusing the same `requirement.txt` across environments, with the pitfalls. Though, I can add information about what's expected to be more or less risky  (Py2/Py3 vs OS X/Linux vs package tightly bound to the OS, etc.) to avoid being too alarming.

taion: I've seen (and contributed) to a number of those issues.

On OS issues, my impression and experience is that they come from a small, vocal minority. If there's an overwhelming majority of users for whom everything "just works", you're unlikely to hear from them. My impression is that the majority of users of pip-tools (and Pipenv) are more in the web dev branch of Python development.

In those cases, where you're mostly just spinning up a Django or Flask app with attendant other dependencies, the OS compatibility issues really do almost never come up. And it really is ubiquitous in that environment to develop on OS X and deploy on Linux.

Python version mismatches are terrible, though, and... yeah, those cases should call for different "compiled" requirements files.

README.rst
```diff
@@ -226,6 +225,40 @@ If you use multiple Python versions, you can run ``pip-sync`` as
 to upgrade those packages.
 
 
+Should I commit ``requirements.in`` and ``requirements.txt`` to source control?
+===============================================================================
+
+Generally, yes. If you want a reproducible environment installation available from your source control,
+they yes, you should commit both ``requirements.in`` and ``requirements.txt`` to source control.
```
atugushev: There is a typo: `they yes`.

bhrutledge:
@vphilippon This was a handy read. Are you game to rebase so that this can be merged? If not, should somebody else open a new PR?

I'd like to confirm that the recommendation is to generate and commit a requirements file for each environment. For example, I'm currently in the process of migrating a Django application from Python 2 to Python 3. We have base and dev requirements. Based on this PR, here's a tox configuration that I came up with:

```ini
[tox]
envlist = {2.7,3.7}
skipsdist = true

[testenv]
basepython =
    2.7: python2.7
    3.7: python3.7
envdir =
    {toxworkdir}/requirements-{envname}
deps =
    pip-tools
commands =
    pip-compile -o requirements-{envname}.txt \
        requirements.in
    pip-compile -o requirements-{envname}-dev.txt \
        requirements-{envname}.txt requirements-dev.in
```

Resulting in these files:

```
requirements.in
requirements-dev.in
requirements-2.7.txt
requirements-2.7-dev.txt
requirements-3.7.txt
requirements-3.7-dev.txt
```

If I understand correctly, I should commit all of those files?

bhrutledge:
For anyone interested, I put together an example repo of how to manage cross-environment requirements files using Tox or Nox: https://github.com/bhrutledge/requirements-py2-py3

Feedback welcome.

atugushev:
@auvipy 

> please rebase

Addressed in 2b6892fd7624be3d8c61a22003efe1adb09a26d6.

atugushev:
Thanks to @vphilippon and all reviewers!

PR #652, opened 2018-04-19
Update all pypi.python.org URLs to pypi.org

jdufresne:
For details on the new PyPI, see the blog post:

https://pythoninsider.blogspot.ca/2018/04/new-pypi-launched-legacy-pypi-shutting.html

Update pypi.python.org URLs to pypi.org

##### Contributor checklist

- [ ] Provided the tests for the changes (N/A)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ even when you've pinned them.  `You do pin them, right?`_
    :target: https://jazzband.co/
 .. |pypi| image:: https://img.shields.io/pypi/v/pip-tools.svg
    :alt: PyPI
-   :target: https://pypi.python.org/pypi/pip-tools/
+   :target: https://pypi.org/project/pip-tools/
 .. _You do pin them, right?: http://nvie.com/posts/pin-your-packages/
 
 


piptools/repositories/pypi.py
```diff
@@ -22,7 +22,7 @@
 
 
 class PyPIRepository(BaseRepository):
-    DEFAULT_INDEX_URL = 'https://pypi.python.org/simple'
+    DEFAULT_INDEX_URL = 'https://pypi.org/simple/'
```
vphilippon: The error is caused by adding a trailling `/` here, in conjuction with this: https://github.com/jazzband/pip-tools/blob/fe8c30e14201abf46f5639a79b65f56398b6ff9f/piptools/writer.py#L58

The result is that the index-url is added to the output as if it wasn't the default one. Removing the trailling `/` here will restore the behavior.

jdufresne: Thanks for looking at the failure. I've made this change. However, it looks like tests are still failing.

vphilippon: I guess the configured index on the BMs is still the old one, or it's in the vendored pip code.

vphilippon:
Basing myself on this `pip` PR: https://github.com/pypa/pip/pull/5215
We have to make a similar change to our vendored `pip` code too.

Changing this line seems to be sufficient: https://github.com/jazzband/pip-tools/blob/5c26e9fde8b560085cc89d0a2dd7c6709f1b68cf/piptools/_vendored/pip/models/index.py#L16
but we should make the change in the doc and pip._vendor libs too. Also, put a note in `piptools/_vendored/README.rst`.

jdufresne:
Thanks for investigating. I've updated the vendored files and `README.rst`. Tests are now green.

vphilippon:
Hi @jdufresne,

Looking at things, we're going to unvendor `pip` in PR #657 , which will have `pip-tools` use the default PyPI URL matching what `pip` gives us.

I'll keep this PR open until the `pip` unvendoring is confirmed and merged.
As always, thanks for contributing, it's much appreciated!

jdufresne:
Rebased to latest master with the pip10 work. This PR now only affects the README.

PR #655, opened 2018-04-25
Added clearer error reporting when skipping pre-releases

WoLpH:
Added clearer error reporting when skipping pre-releases by splitting the output by regular/pre-release versions.

### DIFF
--- a/piptools/exceptions.py
+++ b/piptools/exceptions.py
@@ -9,14 +9,31 @@ class NoCandidateFound(PipToolsError):
         self.index_urls = index_urls
 
     def __str__(self):
-        sorted_versions = sorted(c.version for c in self.candidates_tried)
+        sorted_candidates = sorted(self.candidates_tried)
+        versions = []
+        pre_versions = []
+
+        for candidate in sorted(self.candidates_tried):
+            version = str(candidate.version)
+            if candidate.version.is_prerelease:
+                pre_versions.append(version)
+            else:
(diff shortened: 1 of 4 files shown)


WoLpH:
This pull request is to fix: https://github.com/pypa/pipenv/issues/2046

piptools/exceptions.py
```diff
@@ -9,14 +9,31 @@ def __init__(self, ireq, candidates_tried, index_urls):
...
             'Could not find a version that matches {}'.format(self.ireq),
-            'Tried: {}'.format(', '.join(str(version) for version in sorted_versions) or '(no version found at all)')
         ]
-        if sorted_versions:
+
+        if versions:
+            lines.append('Tried: {}'.format(', '.join(versions)))
+
+        if pre_versions:
+            lines.append('Skipped pre-versions: {}'.format(', '.join(pre_versions)))
```
vphilippon: That statement isn't true in all cases: if the user did provide `--pre`, then the pre-release versions where actually "tried".
Incompatible requirements can still occur with `--pre` (ex: `mypackage<1.2.3a1,>=1.2.3a2`), so we still want to give the proper message.

From the top of my head, I'd say we could pass the `PackageFinder` instance (`self.finder`) through `NoCandidateFound.__init__` instead of `index_urls`, and use the info from the finder instance to make the proper message (getting the index URLs and check `allow_all_prereleases`).

PR #666, opened 2018-06-11
Removed README section mentionning pip vendoring (was un-vendored)

vphilippon:
Removed README section mentionning pip vendoring (as we're un-vendoring it in 3.0.0)

##### Contributor checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -225,17 +225,6 @@ If you use multiple Python versions, you can run ``pip-sync`` as
 ``setuptools``, ``pip``, or ``pip-tools`` itself. Use ``pip install --upgrade``
 to upgrade those packages.
 
-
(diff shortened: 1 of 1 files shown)


suutari:
For other Pip 10 fixes see also following two commits: https://github.com/suutari/prequ/commit/deeac2cb5467e487b66c997c43675ca6f86ffcca and https://github.com/suutari/prequ/commit/3a8dd240a63c685bd4b3ee0e0cedfd7bfd4882a4

vphilippon:
@suutari Thanks for the pointers!
Now that I think about it, I think I did notice some odd "via comment" differences between pip9 and pip10 yesterday, but only once. I'll have to re-test, making sure to clear the cache this time.

Issue #664, opened 2018-06-01
when running with --allow-unsafe it's not added to the 'To update, run' instructions


rmwthorne:
when running with --allow-unsafe it's not added to the 'To update, run' instructions

##### Environment Versions

1. Mac OSX
1. Python version: `$ python -V` Python 3.6.3
1. pip version:  vendored
1. pip-tools version: 2.0.2

##### Steps to replicate

1. pip-compile with --allow-unsafe

##### Expected result

--allow-unsafe is added to the instructions

##### Actual result

--allow-unsafe is not added to the instructions

atugushev:
Closed in #708. Thanks for the issue!

PR #708, opened 2019-01-12
Add --allow-unsafe to update instructions

richafrank:
Prior to this change, running `pip-compile` with `--allow-unsafe` affected the output file, in that `setuptools` et al would become uncommented. However, the output file's instructions on what command to re-run to produce it did not include `--allow-unsafe`, so `setuptools` would end up commented out again. See https://github.com/jazzband/pip-tools/issues/664

This PR includes that flag in the instructions in the output file's header.

**Changelog-friendly one-liner**: Add --allow-unsafe to update instructions

cc @vphilippon as a reviewer

##### Contributor checklist


piptools/writer.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, src_files, dst_file, dry_run, emit_header, emit_index,
     def _sort_key(self, ireq):
         return (not ireq.editable, str(ireq.req).lower())
 
-    def write_header(self):
+    def write_header(self, allow_unsafe):
```
atugushev: Maybe it would be better to injected `allow_unsafe` to the `OutputWriter` constructor instead?

tests/test_writer.py
```diff
@@ -92,10 +93,16 @@ def test_iter_lines__unsafe_dependencies(from_line, writer):
...
-    for line in lines:
-        str_lines.append(line)
+                               None,
+                               allow_unsafe=allow_unsafe)
+    str_lines = list(lines)
     assert comment('# The following packages are considered to be unsafe in a requirements file:') in str_lines
-    assert comment('# setuptools') in str_lines
+    if allow_unsafe:
+        assert any('#    pip-compile --allow-unsafe --output-file ' in str_line
+                   for str_line in str_lines)
```
atugushev: I would prefer:
```
assert comment('#    pip-compile --output-file dst_file src_file src_file2') in str_lines
```

richafrank: Sounds reasonable to me. I had been trying not to assert much beyond the part of the output that changes, but that's not so bad.

piptools/writer.py
richafrank: Certainly can do that. 

Right now `allow_unsafe` gets here from being passed to each call of `OutputWriter.write`. Is there any big benefit from the current model? I could see re-using a single instance across `write` calls where this value varies, if it's slow to construct the instance. It looks like both the constructor and `write` take a lot of arguments, and I'm not sure how they are meant to be split; however, I only see one caller of `write`, which is right after instantiation, so maybe there isn't an intentional divide...

Ah, the other caller is in the tests, so I'll need either a new "unsafe writer" fixture, or I can update the tests to expect the fixture gives them a class (or partial), instead of an instance. Or maybe there's a third, better way? Preferences there?

atugushev: My point is it looks a bit hacky, because the `allow_unsafe` injected now across two layers `_iter_lines` and `write_header` and it could go deeper. A quick research showed **it was already in the constructor earlier** (see #377 PR) and later was deleted by @jdufresne (see 68533537 commit). Current `allow_unsafe` was appeared in `_iter_lines` (see #517 PR).

Of course it'll affect on tests, but we could create writer instance and write a test for each behavior, allow_unsafe or not allow_unsafe, separately which looks better to me.

atugushev: Now it looks pretty good to me. Thanks!

atugushev:
@richafrank thanks!

richafrank:
Great - thanks @atugushev !

PR #710, opened 2019-01-17
Setup coveralls

atugushev:
- Run coveralls after success build in travis-ci
- Add badge to the README

**Changelog-friendly one-liner**: Setup coveralls

##### Contributor checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,10 @@ env:
 
 cache: pip
 install:
-  - travis_retry pip install tox-travis
+  - travis_retry pip install tox-travis coveralls
 script:
   - tox
+after_success: coveralls
 
 jobs:
   exclude:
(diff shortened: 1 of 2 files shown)


blueyed:
Hmm, also no failed request in https://travis-ci.org/jazzband/pip-tools/requests.. :confused:

.travis.yml
```diff
@@ -36,6 +36,9 @@ jobs:
     - stage: readme
       python: 2.7
       env: TOXENV=readme
+    - stage: coveralls
+      python: 3.7
+      env: TOXENV=coveralls
```
blueyed: It should upload to coveralls after running (test) jobs.
With this it will have a separate (new) job that would run tests, and upload coverage for this only.

atugushev: Do you mean upload on each test job?

atugushev:
> Hmm, also no failed request in https://travis-ci.org/jazzband/pip-tools/requests.. 😕

It's really strange.

.travis.yml
blueyed: Yes, but not for qa ones etc.
Coveralls allows for multiple/parallel uploads itself.
If codecov would be used it could have different flags per job.

atugushev: That's a nice feature in codecov, indeed. I don't have a strong opinion about coveralls, so we could use codecov instead.

blueyed:
Should also get .coveragerc fixed from #711 .

I suggest to integrate the reporting to coveralls into #711 also - it is OK to use both IMHO.

atugushev:
Should both badges be presented in README?

atugushev:
Okay, i'll close this PR then and move changes for coveralls to #711.

PR #735, opened 2019-02-15
Show default index url provided by pip

### LINKED ISSUE #705: index-url option help misleading
Help message describing pip-compile index-url option does not seem compliant with the expected behavior if I refer to #208: the default could be different of PyPI in that case.

Am I right?

atugushev:
Closes #705.

**Changelog-friendly one-liner**: Show default index url provided by pip

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -2,7 +2,6 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
-import optparse
 import os
 import sys
 import tempfile
@@ -10,23 +9,18 @@ import tempfile
 from .._compat import (
     install_req_from_line,
     parse_requirements,
-    cmdoptions,
-    Command,
 )
 
 from .. import click
 from ..exceptions import PipToolsError
 from ..logging import log
+from ..pip import get_pip_command, pip_defaults
 from ..repositories import LocalRequirementsRepository, PyPIRepository
 from ..resolver import Resolver
 from ..utils import (dedup, is_pinned_requirement, key_from_req, UNSAFE_PACKAGES)
 from ..writer import OutputWriter
 
 DEFAULT_REQUIREMENTS_FILE = 'requirements.in'
-
-
-class PipCommand(Command):
-    name = 'PipCommand'
 
 
 @click.command()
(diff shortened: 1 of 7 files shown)


piptools/scripts/compile.py
```diff
@@ -37,7 +31,7 @@ class PipCommand(Command):
 @click.option('-p', '--pre', is_flag=True, default=None, help="Allow resolving to prereleases (default is not)")
 @click.option('-r', '--rebuild', is_flag=True, help="Clear any caches upfront, rebuild from scratch")
 @click.option('-f', '--find-links', multiple=True, help="Look for archives in this directory or on this HTML page", envvar='PIP_FIND_LINKS')  # noqa
-@click.option('-i', '--index-url', help="Change index URL (defaults to PyPI)", envvar='PIP_INDEX_URL')
+@click.option('-i', '--index-url', help="Change index URL (defaults to {})".format(pip_defaults.index_url), envvar='PIP_INDEX_URL')  # noqa
```
blueyed: This should be asserted/checked in a test.
Or are there existing, unchanged ones already for it?

atugushev: I've found that there is no test for that case. I've added a test. Check it out.

tests/test_cli.py
```diff
@@ -440,3 +440,12 @@ def test_no_candidates_pre():
...
         assert 'Tried pre-versions:' in out.output
+
+
+@pytest.mark.usefixtures('pip_conf')
+def test_default_index_url():
+    status, output = _invoke(['pip-compile', '--help'])
+    output = output.decode('utf-8')
+
+    assert status == 0
+    assert 'http://example.com' in output
```
blueyed: I think this should be more specific, and could probably also be asserted from a test that runs `--help` again (less overhead).

blueyed: Re "more specific": it should match other parts of the line printing this, so that it would not match e.g. a copyright line etc.. ;)

atugushev: > Re "more specific": it should match other parts of the line printing this, so that it would not match e.g. a copyright line etc.. ;)

Good point! Though i'm not sure how to test properly multiline string in output. See:

```
         -i, --index-url TEXT            Change index URL (defaults to
                                         http://example.com)
```

Do you have any suggestions?

> also be asserted from a test that runs `--help` again (less overhead).

I'd rather to make a separate test for that. In ideal world i'd prefer one assert per one test :)

atugushev: What about this one? Looks a bit hacky though.
```python
assert "Change index URL (defaults to http://example.com)" in " ".join(output.split(" "))
```

atugushev: No, it doesn't work.

atugushev: Anyways, i've tested a whole string.  Check it out please.

atugushev: @blueyed finally managed to fix failing test on win+py27. Reported an issue  https://github.com/pallets/click/issues/1231.

atugushev:
I've rebased and squashed commits. I'll merge this as soon as tests are passed.

@blueyed as always, appreciate your feedback!

atugushev:
`pip-tools v3.5.0` is released

PR #736, opened 2019-02-15
ci: Travis: skip deploy stage without tag

blueyed:
It would get skipped anyway, but with this it does not show up in the
jobs for PR builds.


blueyed:
Looks good: https://travis-ci.org/jazzband/pip-tools/builds/493877860 - canceled the build.

PR #742, opened 2019-02-25
Split test_cli.py

atugushev:
Splits test_cli.py into two files for compile and sync modules. I'm going to add more tests to test_cli_sync.py, let's split them at first.


atugushev:
I've resolved conflicts. It's ready to merge now.

PR #745, opened 2019-02-27
Run codecov in appveyor ci

atugushev:
Increases the test coverage.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -98,4 +98,4 @@ test_script:
     - tox
 
 after_test:
-    - codecov
+    - codecov -F windows -n %TOXENV%
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,6 @@
 environment:
+    global:
+      PYTHON: "C:\\Python36"
 
     matrix:
         - TOXENV: py27-pip8.1.1
@@ -87,9 +89,13 @@ matrix:
     - PIP: master
 
 install:
-    - C:\Python36\python -m pip install tox
+    - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+    - pip install tox codecov
 
 build: false
 
 test_script:
-    - C:\Python36\Scripts\tox
+    - tox
+
+after_test:
+    - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
 script:
   - tox
 after_success:
-  - codecov -e PIP
+  - codecov -F $TRAVIS_OS_NAME -n "py$TRAVIS_PYTHON_VERSION-pip$PIP"
   - "COVERALLS_PARALLEL=true coveralls"
 
 stages:


.appveyor.yml
```diff
@@ -87,9 +89,13 @@ matrix:
...
+    - pip install tox codecov
 
 build: false
 
 test_script:
-    - C:\Python36\Scripts\tox
+    - tox
+
+after_test:
+    - codecov
```
blueyed:
```suggestion
    - codecov --flags windows --name %TOXENV%
```

.appveyor.yml
```diff
@@ -87,9 +89,13 @@ matrix:
     - PIP: master
 
 install:
-    - C:\Python36\python -m pip install tox
+    - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
```
blueyed:
```suggestion
    - "SET PATH=%PYTHON%\\Scripts;%PATH%"
```
Shouldn't that be enough?  (but just a guess)

blueyed:
This will likely cause problems with codecov timing out though - if there are non already (given the huge number of builds for linux already).
We're having trouble (likely due to flags and number of builds) in pytest (and other projects) since months.

So in general it would be better to only run selective envs with coverage, by using a coverage factor in tox (see pytest's tox.ini).

.appveyor.yml
atugushev: The `python` binary in `C:\Python36\` folder. That's why `%PYTHON%` was added.

.appveyor.yml
atugushev: Good point. I think `--flag=linux` should be added also in the `.travis.yml`. What do you think?

blueyed: Yes, in general - but flags might trigger timeouts with codecov.io, so I would rather only use name here for now.
It does not post comments anyway, does it?  (where the flags would be visible)

.appveyor.yml
blueyed: Ok.

atugushev:
> We're having trouble (likely due to flags and number of builds) in pytest (and other projects) since months.

@blueyed could you please tell me what kind of problems do you mean?

blueyed:
@atugushev 
When looking at PRs via codecov comments or the UI in general it might time out on their backend ("connection reset").
See https://github.com/pytest-dev/pytest/pull/4846#issuecomment-467949404.

blueyed:
Here https://codecov.io/gh/jazzband/pip-tools/pull/745?src=pr&el=h1 is not timing out, although there are a lot of builds.
Source code size might also be a factor - I don't know.

blueyed:
So I suggest adding flags ("windows" / "linux" (or rather `$TRAVIS_OS_NAME`)) here after all.

atugushev:
@blueyed 
> So in general it would be better to only run selective envs with coverage, by using a coverage factor in tox (see pytest's tox.ini).
> See [pytest-dev/pytest#4846 (comment)](https://github.com/pytest-dev/pytest/pull/4846#issuecomment-467949404).

I see. Running selective envs with coverage sounds reasonable to me now. I'll try to test it and propose a PR soon.

> So I suggest adding flags ("windows" / "linux" (or rather `$TRAVIS_OS_NAME`)) here after all.

Builds section looks pretty good now, see the latest [codecov report](https://codecov.io/gh/jazzband/pip-tools/commit/1bc0bb195add121a2460cf91af1cc3551672afe0/build). Thanks for the suggestion, Daniel!

blueyed:
For reference: extra coverage due to this: https://codecov.io/gh/jazzband/pip-tools/compare/c45862bd3e99df71fc020f828395fc7f6766be7c...fc69f3a2eab8d54c1a9850dcb2fdd9ec3d6930c5/changes

But it could be achieved with a single (or few) jobs only probably also.

PR #731, opened 2019-02-08
Fix the output file for pip-compile with an explicit setup.py as source file

atugushev:
The command `pip-compile setup.py` generates a `setup.txt`, but
it should be a requirements.txt file.

Fixes #730.

**Changelog-friendly one-liner**: Fix the output file for pip-compile with an explicit setup.py as source file

##### Contributor checklist

- [X] Provided the tests for the changes.
- [X] Requested a review from another contributor.
- [X] Gave a clear one-line description in the PR (that the maintainers can add to CHANGELOG.md on release).
- [x] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #749, opened 2019-02-28
Fix a regression bug came from #731 PR

atugushev:
Fix the `pip-compile setup.py --output-file=output.txt` ignores `--output-file` option and compiles to a `requirements.txt` file.

Refs #731.


tests/test_cli_compile.py
```diff
@@ -90,13 +87,42 @@ def test_command_line_setuptools_read(pip_conf, cli_args):
...
+        package = open('setup.py', 'w')
+        package.write(dedent("""\
+            from setuptools import setup
+            setup(install_requires=[])
+        """))
+        package.close()
+
+        out = runner.invoke(cli, options)
+        assert out.exit_code == 0
+        assert os.path.exists(expected_output_file)
```
blueyed: Just a nitpick: the asserts could be dedented.

atugushev: Agreed. Let's clean it up in a following PR.

atugushev: In addition to that, i would also make the runner a fixture and clean up all the tests (which use the runnrt) like this one: https://github.com/jazzband/pip-tools/blob/67edc7d227507f3902090d2db8e129715279ae04/tests/test_cli_compile.py#L442-L449

IMO the tests look more concise in that way.

PR #750, opened 2019-03-01
Use pytest-cov plugin

atugushev:
The benefit is the [subprocess support](https://pytest-cov.readthedocs.io/en/latest/readme.html?highlight=subprocess). Since the piptools uses subprocess module in the tests it allows to cover `__main__.py`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,9 +15,9 @@ deps =
     pip10.0.1: pip==10.0.1
     pip18.0: pip==18.0
     pip19.0: pip==19.0
-    coverage
     mock
     pytest
+    pytest-cov
 setenv =
     piplatest: PIP=latest
     pipmaster: PIP=master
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -30,9 +30,8 @@ setenv =
 install_command= python -m pip install {opts} {packages}
 commands =
     pip --version
-    coverage run -m pytest --strict --doctest-modules {posargs:tests/ piptools/}
-    coverage report -m
-    coverage html
+    pytest --strict --doctest-modules --cov piptools --cov tests \
+        --cov-report html --cov-report term-missing {posargs:tests/ piptools/}
```
blueyed: Just `--cov` should be enough - then it uses the config (which as source="." though).  I think those should be in line.

atugushev: Good call!

blueyed:
Same time on AppVeyor compared to master.

But slower on Travis: 2h06 vs. 1h49 - might not be relevant though.

atugushev:
@blueyed 

i've just noticed that since https://travis-ci.org/jazzband/pip-tools/jobs/483526623 the tests are not run for pip==19.0.

atugushev:
> But slower on Travis: 2h06 vs. 1h49 - might not be relevant though.

I'm inclined to think it's not related, because time for `tox` command stand at the same level ±35 sec.

blueyed:
> @blueyed
> 
> i've just noticed that since https://travis-ci.org/jazzband/pip-tools/jobs/483526623 the tests are not run for pip==19.0.

https://github.com/jazzband/pip-tools/pull/750#issuecomment-468903117

PR #751, opened 2019-03-01
ci: codecov: remove flags, add OS name to job name

blueyed:
While pip-tools is not (yet?) that much affected by timeouts on
codecov.io's backend, it is certainly rather slow in general.

As it turned out with pytest (where it was constantly timing out), this
is due to the usage of flags.

Therefore this removes flags for now, but adds TRAVIS_OS_NAME to the
upload's name.

This also moves the after_success section to the end of the config
(Where you would expect it), and only installs codecov/coveralls when
necessary.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -100,4 +100,4 @@ test_script:
 
 after_test:
     - pip install codecov
-    - codecov -n %TOXENV%
+    - codecov -n %TOXENV%-windows
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -91,7 +91,7 @@ matrix:
 
 install:
     - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-    - pip install tox codecov
+    - pip install tox
 
 build: false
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -86,5 +83,10 @@ jobs:
     - env: PIP=master
     - python: 3.8-dev
 
+after_success:
+  - travis_retry pip install codecov coveralls
+  - codecov -n "py${TRAVIS_PYTHON_VERSION}-pip${PIP}-${TRAVIS_OS_NAME}"
```
atugushev: Could you fix it the same way in the `.appveyour.yml`?

blueyed: Done.  Sorry for missing it in the first place..

.appveyor.yml
```diff
@@ -91,12 +91,13 @@ matrix:
...
 
 build: false
 
 test_script:
     - tox
 
 after_test:
-    - codecov -F windows -n %TOXENV%
+    - pip install codecov
+    - codecov -n %TOXENV%
```
atugushev:
```suggestion
    - codecov -n %TOXENV%-windows
```
just a nitpick.

PR #752, opened 2019-03-02
ci: Travis: test pypy/pypy3 only with PIP=latest

blueyed:
This is mainly meant to reduce the number of jobs (58 currently!).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ python:
   - "3.5"
   - "3.6"
   - "3.7"
-  - "pypy2.7-6.0"
-  - "pypy3.5-6.0"
 
 env:
   - PIP=8.1.1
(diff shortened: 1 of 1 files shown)


blueyed:
Travis jobs are down to 45 (from 58) with this: https://travis-ci.org/jazzband/pip-tools/builds/500616125.

blueyed:
Canceled AppVeyor, to save the environment and because it is not necessary to check.
But hopefully Codecov complains about dropped coverage.

atugushev:
Unfortunately, I don't clearly understand why it should be tested only on the latest PIP. Could you explain that?

atugushev:
@blueyed probably that's because on a pypy test are too slow (x2 time)?

blueyed:
It is meant to reduce the number of jobs.

But it is also true that pypy is slower with coverage (so much that pytest does not run it with coverage in the first place, but here it is still okish).

atugushev:
@blueyed I'd happy to see this merged. Could we get this done?

blueyed:
@atugushev 
I think it is good (AppVeyor was canceled manually).
I don't remember why I've changed the title to WIP - sorry for the confusion.

blueyed:
(closed accidentally.. :/)

PR #754, opened 2019-03-04
Revisit coverage reporting

blueyed:
* tox: add coverage factor

This is used via tox-travis automatically.
With AppVeyor, only selected envs are run with coverage.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,7 +3,7 @@ environment:
       PYTHON: "C:\\Python36"
 
     matrix:
-        - TOXENV: py27-pip8.1.1
+        - TOXENV: py27-pip8.1.1-coverage
           PIP: 8.1.1
         - TOXENV: py27-pip9.0.1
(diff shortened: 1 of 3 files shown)


blueyed:
TODO: squash-merge

.appveyor.yml
```diff
@@ -3,7 +3,7 @@ environment:
       PYTHON: "C:\\Python36"
 
     matrix:
-        - TOXENV: py27-pip8.1.1
+        - TOXENV: py27-pip8.1.1-coverage
           PIP: 8.1.1
```
atugushev: Do we really need PIP variable?

blueyed: It is used for matrix expansion, in the codecov name, and in https://github.com/blueyed/pip-tools/blob/3add05df690b4fd3350e1deb48909055897f45b1/tests/test_cli_compile.py#L19-L24.

PR #758, opened 2019-03-07
Add an option to allow enabling/disabling build isolation

### LINKED ISSUE #721: pip-compile fails on packages using poetry
`pip-compile` fails if a package uses `poetry` which is PEP 517 compliant (https://poetry.eustace.io/docs/pyproject/#poetry-and-pep-517). 

Simply doing `pip install pendulum` does not require e.g. `poetry` to be installed. However, installing it made `pip-compile` work which I'm not sure we as consumers of the package in question should need to do.

##### Environment Versions

1. Alpine 3.8
1. Python version: `3.7.2`
1. pip version:  `19.0.1`
1. pip-tools version: `3.3.1`

##### Steps to replicate
```Dockerfile
# DOES NOT WORK

FROM python:3.7.2-alpine3.8

RUN pip3 install --upgrade pip==19.0.1 setuptools pip-tools>==3.3.1

RUN echo "pendulum" > req.txt

RUN pip-compile --generate-hashes req.txt


# WORKS

FROM python:3.7.2-alpine3.8

RUN pip3 install --upgrade pip==19.0.1 setuptools pip-tools>==3.3.1

RUN echo "pendulum" > req.txt

RUN pip3 install poetry

RUN pip-compile --generate-hashes req.txt
```

atugushev:
Closes #721.

**Changelog-friendly one-liner**: Add an option to allow enabling/disabling build isolation

##### Contributor checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -534,3 +534,23 @@ def test_cert_option(MockPyPIRepository, runner, option, attr, expected):
     for call in MockPyPIRepository.call_args_list:
         pip_options = call[0][0]
         assert getattr(pip_options, attr) == expected
+
+
+@pytest.mark.parametrize('option, expected', [
+    ('--build-isolation', True),
+    ('--no-build-isolation', False),
+])
+@mock.patch('piptools.scripts.compile.PyPIRepository')
+def test_build_isolation_option(MockPyPIRepository, runner, option, expected):
+    """
+    A value of the --build-isolation/--no-build-isolation flag must be passed to the PyPIRepository.
+    """
+    with open('requirements.in', 'w'):
+        pass
+
+    runner.invoke(cli, [option])
+
+    # Ensure the build_isolation in PyPIRepository has have the expected option
+    for call in MockPyPIRepository.call_args_list:
+        build_isolation = call[0][2]
+        assert build_isolation is expected
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -551,6 +551,4 @@ def test_build_isolation_option(MockPyPIRepository, runner, option, expected):
     runner.invoke(cli, [option])
 
     # Ensure the build_isolation option in PyPIRepository has the expected value.
-    for call in MockPyPIRepository.call_args_list:
-        build_isolation = call[0][2]
-        assert build_isolation is expected
+    assert [call[0][2] for call in MockPyPIRepository.call_args_list] == [expected]
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -550,7 +550,7 @@ def test_build_isolation_option(MockPyPIRepository, runner, option, expected):
 
     runner.invoke(cli, [option])
 
-    # Ensure the build_isolation in PyPIRepository has have the expected option
+    # Ensure the build_isolation option in PyPIRepository has the expected value.
     for call in MockPyPIRepository.call_args_list:
         build_isolation = call[0][2]
         assert build_isolation is expected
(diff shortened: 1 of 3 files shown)


tests/test_cli_compile.py
```diff
@@ -534,3 +534,23 @@ def test_cert_option(MockPyPIRepository, runner, option, attr, expected):
...
+def test_build_isolation_option(MockPyPIRepository, runner, option, expected):
+    """
+    A value of the --build-isolation/--no-build-isolation flag must be passed to the PyPIRepository.
+    """
+    with open('requirements.in', 'w'):
+        pass
+
+    runner.invoke(cli, [option])
+
+    # Ensure the build_isolation in PyPIRepository has have the expected option
```
blueyed:
```suggestion
    # Ensure the build_isolation option in PyPIRepository has the expected value.
```

tests/test_cli_compile.py
```diff
@@ -534,3 +534,23 @@ def test_cert_option(MockPyPIRepository, runner, option, attr, expected):
...
+    for call in MockPyPIRepository.call_args_list:
+        build_isolation = call[0][2]
+        assert build_isolation is expected
```
blueyed: I'd prefer to assert/compare the whole list, but maybe only still the single option.  This would include the number of calls then also - currently it might just assert nothing silently.

atugushev: Sorry didn't get your point. Could you shoot some code-snippet for the instance?

tests/test_cli_compile.py
atugushev: Aw, thanks!

tests/test_cli_compile.py
blueyed: 1. `MockPyPIRepository.call_args_list` might be empty

```python
assert [call[0][2] for call in MockPyPIRepository.call_args_list] == [expected, …
```]

Comment (blueyed) [reply tests/test_cli_compile.py]: btw: you could use `[ci skip]` for the commit with the suggested changes - since it should be squashed later anyway, and would save some hours of CI.

Comment (atugushev) [reply tests/test_cli_compile.py]: Do you mean something like `MockPyPIRepository.assert_called_once_with(..., ..., expected)`?

Comment (blueyed) [reply tests/test_cli_compile.py]: Yes, or the above, where you expect it to be a non-empty list.

Comment (atugushev) [reply tests/test_cli_compile.py]: > btw: you could use [ci skip] for the commit with the suggested changes - since it should be squashed later anyway, and would save some hours of CI.

Right, okay.

Comment (atugushev) [reply tests/test_cli_compile.py]: Or we just could assert it called once (before iteration over call_args_list) to ensure the call_args_list is not empty:
```python
MockPyPIRepository.assert_called_once()
```

atugushev: In that case `test_cert_option` also should be fixed.

blueyed: > Or we just could assert it called once (before iteration over call_args_list) to ensure the call_args_list is not empty:
> 
> ```python
> MockPyPIRepository.assert_called_once()
> ```

Yes.  Although I would prefer a single assertion using a list comprehension instead.

atugushev: I was thinking, you know, one assertion looks better. Fixed the test.

blueyed: Yes, one assertion is better in my opinion, too.

`assert [call[0][2] for call in MockPyPIRepository.call_args_list] == [expected]` is just that, no? :)

Thanks for adjusting!

jkp:
Will this get merged?

atugushev:
`pip-tools v3.5.0` is released

PR #766, opened 2019-03-19
Add full test coverage for the resolver module

atugushev:
Increases test coverage up to 100% for `piptools/resolver.py`.


atugushev:
@blueyed thanks!

PR #769, opened 2019-03-23
Cli tests refactoring

atugushev:
- uses runner fixture instead context manager directly
- adds a TEST_DATA_PATH const
- removes unused tmpdir fixtures in some tests (probably copypasted)


blueyed:
> * marks to use `pip_conf` fixture, instead of using it as function param, because it's not used in the test directly

Wonder what the reason for this is?  Linting tools warning about unused args?
(I find it easier to read without the decorator)
btw: the tests could maybe even use the config file explicitly (but there appears to be no option for it with pip-compile at least).
But no strong opinion on this.

atugushev:
@blueyed 

Some time ago i read a manual, especially this section
> Sometimes test functions do not directly need access to a fixture object. For example, tests may require to operate with an empty directory as the current working directory but otherwise do not care for the concrete directory. Here is how you can use the standard tempfile and pytest fixtures to achieve it.
> https://docs.pytest.org/en/latest/fixture.html#using-fixtures-from-classes-modules-or-projects

and understood it in a way, that this is how to do when you don't need a directly access to a fixture object. But, now i've reread the manual and figured out that is not mandatory and this mark suppose (AFAICU) to use fixtures for a class or module (it's definitely convenient way).

Thank you for pointing out. I've googled and didn't found any mentions about the right way to use it (param or decorator). Maybe pytest should have a note for that case.

i've reconsidered it and gonna to change to a function param, because it's a much shorter form.

atugushev:
@blueyed thanks!

PR #770, opened 2019-03-27
Add isort and black

atugushev:
- adds `isort`
- adds `black`
- adds `checkqa` env to `tox`
- runs `checkqa` env on `python3`
- cleanup the code


piptools/resolver.py
```diff
@@ -203,24 +228,32 @@ def _resolve_one_round(self):
...
+            log.debug("Removed dependencies in this round:")
+            for removed_dependency in sorted(
+                removed, key=lambda req: key_from_req(req.req)
+            ):
+                log.debug("  removing {}".format(removed_dependency))
+            log.debug("Unsafe dependencies in this round:")
+            for unsafe_dependency in sorted(
+                unsafe, key=lambda req: key_from_req(req.req)
+            ):
+                log.debug("  remembering unsafe {}".format(unsafe_dependency))
```
rpkilby: Is this block really more readable than the way it was before?

blueyed: Black is not about opinions.. ;)

I find this change OK (apart from the clutter that changing quotes brings here), but black can certainly make things less readable (https://github.com/ambv/black/issues/601).

rpkilby: The three lines aren't that much worse on their own - I'm referring to the entire block of code. Reading the previous section, the vertical flow visually reads better, while the new code doesn't read as well visually. 

> Black is not about opinions.. ;)

Isn't this exactly the point? Black has opinions so you don't have to?

blueyed: > Isn't this exactly the point? Black has opinions so you don't have to?

Yeah, that's what I've meant - "Black is not about (your) opinions".

I agree that the flow is worse now.

We could use "# fmt: off" / "# fmt: on" markers here.

setup.cfg
```diff
@@ -7,3 +7,15 @@ license_file = LICENSE
 [tool:pytest]
 norecursedirs = .* build dist venv test_data piptools/_compat/* piptools/_vendored/*
 testpaths = tests piptools
+
+[flake8]
+# E501: line too long
+# W503: line break before binary operator
+ignore = E501, W503
```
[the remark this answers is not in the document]
atugushev: In that case:
```bash
$ flake8 piptools tests/ | wc -l
      47
```

Should we split strings manually then? BTW, even black [ignore E501](https://github.com/ambv/black/blob/master/.flake8).

tox.ini
```diff
@@ -33,9 +33,16 @@ setenv =
...
+[testenv:linters]
+basepython = python3
+deps =
+    flake8
+    black
+    isort
+commands =
+    black --check --diff .
+    isort --recursive --check-only --diff .
+    flake8 piptools tests
```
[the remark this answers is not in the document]
atugushev: Good point! Thanks.

piptools/resolver.py
atugushev: IMO, this block looks okay (i've seen much worse). Anyways, PRs are welcome to fix "ugly" formattings.

.travis.yml
```diff
@@ -51,8 +51,8 @@ jobs:
       python: "2.7"
 
     # QA checks.
-    - env: TOXENV=flake8
-      python: 2.7
+    - env: TOXENV=linters
```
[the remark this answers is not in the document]
atugushev: Like it 👍

setup.cfg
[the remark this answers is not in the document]
atugushev: So, i've fixed long lines manually

atugushev:
Travis got stuck :(

atugushev:
Finally it's passed out!

atugushev:
@blueyed are we okay with this PR?

PR #771, opened 2019-03-28
Exclude from coverage exception line

atugushev:
This could happen accidentally, no need to cover that.


atugushev:
@blueyed take a look a this trivial PR please 🙏

atugushev:
@blueyed thanks!

PR #774, opened 2019-03-29
Enable FussyFox for flake8 & isort

codingjoe:
FussyFox is a real-time code linter, to provide contributers with faster feedback
and to take load of CI systems.

##### Contributor checklist


.checks.yml
```diff
@@ -0,0 +1 @@
+- flake8
```
atugushev: What about `.fussyfox.yml`?

codingjoe: done

.travis.yml
```diff
@@ -51,9 +51,6 @@ jobs:
       python: "2.7"
 
     # QA checks.
-    - env: TOXENV=checkqa
```
atugushev: In `checkqa` we had black+isort+flake8 checking. Maybe `isort` should be also moved to fussyfox also and `checkqa` renamed to `black`.

codingjoe: Since you seem to be already on it, just do the changes that you think are best.

atugushev: @codingjoe isort checking has been failed. Investigating.

atugushev:
@codingjoe have no idea why isort is failing. On Travis CI worked like a charm. Could you take a look what's going on AWS?

rpkilby:
> Failing after **594m** — isort

Ouch. almost 10 hours of VM use.

rpkilby:
I don't think this helps the purported goal of removing load from Travis. The current `checkqa` build is part of the baseline stage. If qa fails, then this prevents the test stage from running. By moving isort/flake8 to fussyfox, this can cause tests to run unnecessarily when qa fails. 

Also, the `checkqa` build only takes 50s on Travis, while fussyfox takes 27s (with an upper limit of failing in 10 hours?). Is removing 50s of build time from Travis worth the additional complexity of relying on another CI system?

codingjoe:
> I don't think this helps the purported goal of removing load from Travis. The current `checkqa` build is part of the baseline stage. If qa fails, then this prevents the test stage from running. By moving isort/flake8 to fussyfox, this can cause tests to run unnecessarily when qa fails.

It's not so much about the speed, but how it is integrated. You don't need may clicks to realize what when wrong and what you need to fix. The point is to give contributors instant feedback. The load part is just a sideeffect.

> Also, the `checkqa` build only takes 50s on Travis, while fussyfox takes 27s (with an upper limit of failing in 10 hours?). Is removing 50s of build time from Travis worth the additional complexity of relying on another CI system?

The lambda functions don't run longer than 3min. I don't know what happened here. But i will investigate it.

codingjoe:
> @codingjoe have no idea why isort is failing. On Travis CI worked like a charm. Could you take a look what's going on AWS?

I think that is because on AWS it is static. The dependencies are not installed so isort does no know if something is 3rd party or not. I usually like to add 3rd party packages explicitly.

This might be a good thing. Not all contributors setup a virtual environment to submit changes. So they will have the same errors too. I can fix the isort config for you, if you want me to.

blueyed:
> If qa fails, then this prevents the test stage from running. By moving isort/flake8 to fussyfox, this can cause tests to run unnecessarily when qa fails.

@codingjoe 
What was meant here is that we're not running tests unnecessarily if something basic fails already - this is not the case with removing it from there.

blueyed:
@codingjoe 
btw: it would be useful to have `--version` output with runs.

rpkilby:
> It's not so much about the speed, but how it is integrated.

That's definitely fair - it is nice having the QA failures one click away. That said, I'd argue that if fussyfox is added, it should be used *in addition* to `checkqa` on Travis, per...

> If qa fails, then this prevents the test stage from running. By moving isort/flake8 to fussyfox, this can cause tests to run unnecessarily when qa fails.

blueyed:
> That said, I'd argue that if fussyfox is added, it should be used _in addition_ to `checkqa` on Travis

I agree.  And then it is questionable if it makes sense at all, given that there are two configurations then etc.  As you can see currently it might pass on Travis, but fail on FussyFox - or the other way around.

atugushev:
@rpkilby @blueyed 
> The current `checkqa` build is part of the baseline stage. If qa fails, then this prevents the test stage from running.

In that case we should also consider to add `checkqa` in `appveyour` to not waste the time if QA checks are failed.

codingjoe:
So I fixed the isort config, works now on a system without the dependencies installed.
@blueyed the `--version` idea is a good idea, also printing the command that is run, would be helpful info. I'll get to it.

atugushev:
@codingjoe 
Also i'm inclined to think that toxenv `checkqa` should not be removed from the `tox.ini`, because it's a convenient one-liner `tox -e checkqa` to see that changes are not breaking the style conventions. But now user has to install black+isort+flake8 locally and run it one by one manually.

codingjoe:
Ok, I reverted the changes, not its an addition rather than a replacement

blueyed:
Thanks!

blueyed:
@codingjoe 
Maybe you want to create a FussyFox PRs also for pytest?

blueyed:
@atugushev 
> @rpkilby @blueyed
> 
> > The current `checkqa` build is part of the baseline stage. If qa fails, then this prevents the test stage from running.
> 
> In that case we should also consider to add `checkqa` in `appveyour` to not waste the time if QA checks are failed.

Fair point - I usually like to have fast-failures with AppVeyor, since Travis would be the denominator, and that would work with this, but OTOH adding another job to AppVeyor also adds overhead/redundancy, so I'd rather leave it as-is (except for maybe enabling fast-failures if that is not the case already).

atugushev:
@blueyed 
> except for maybe enabling fast-failures if that is not the case already

AFAIK fast-failures has been already enabled for appveyour.

PR #775, opened 2019-03-29
Enable bandit security linter

codingjoe:
Enable bandit security linter and address related concernts.

**Changelog-friendly one-liner**: Add bandit security linter

##### Contributor checklist

### DIFF
--- a/.bandit
+++ b/.bandit
@@ -1,2 +1,2 @@
 [bandit]
-exclude: tests,.tox,.eggs
+exclude: tests,.tox,.eggs,.venv,.git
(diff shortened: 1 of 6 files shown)


codingjoe:
There are actually some security concerns that should be addressed. I don't see them as cirical, but they should be addressed.

Should I do that in the same PR, but different commits, or separate PRs?

atugushev:
> Should I do that in the same PR, but different commits, or separate PRs?

I think it's better to fix it in this PR.

.bandit
```diff
@@ -0,0 +1,2 @@
+[bandit]
+exclude: tests,.tox,.eggs
```
atugushev:
```suggestion
exclude: tests,.tox,.eggs,.venv,.git
```

PR #784, opened 2019-04-06
Add full test coverage for fs_str func

atugushev:
Fixes decrease coverage from the 5097833 commit.


atugushev:
@codingjoe why there are no fussyfox checks?

codingjoe:
@atugushev that is a very good point. We only listen to check check suite signal. Which should be triggered automatically. But maybe it is only for people that have the app installed. I will reach out to GitHub ask, when they send the event and when they don't. Thanks for bringing this to my attention 👍

atugushev:
@codingjoe thanks for reviewing!

PR #790, opened 2019-04-19
Add full test coverage for the cache module

atugushev:
Increases test coverage up to 100% for `piptools/cache.py`.

### DIFF
--- a/piptools/cache.py
+++ b/piptools/cache.py
@@ -121,10 +121,6 @@ class DependencyCache(object):
         self.cache[pkgname][pkgversion_and_extras] = values
         self.write_cache()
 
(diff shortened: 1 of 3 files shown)


piptools/cache.py
```diff
@@ -121,10 +121,6 @@ def __setitem__(self, ireq, values):
         self.cache[pkgname][pkgversion_and_extras] = values
         self.write_cache()
 
-    def get(self, ireq, default=None):
```
atugushev: NOTE: usage of this method has been removed here https://github.com/jazzband/pip-tools/pull/153/files#diff-4a3323f7294510dd8436d96838ca4a61L308

PR #704, opened 2019-01-02
Add support for --find-links in pip-compile input

estan:
This adds support to pip-compile for the `-f, --find-links <url>` flag in the input `requirements.in`.

Let me know if there should be a corresponding `--emit-find-links / --no-emit-find-links` as I wasn't sure.

I also took the liberty of moving down the deduplication of `repository.finder.index_urls/find_links` so that it's after the last `parse_requirements` call, otherwise dupes creep in after the deduplication.

**Changelog-friendly one-liner**: Add support for --find-links in pip-compile input

@nvie @vphilippon @davidovich

PR #793, opened 2019-04-22
Add support for --find-links in pip-compile output

atugushev:
This adds support to pip-compile for the -f, --find-links <url> flag in the input requirements.in. Rebased #704 and adjusted tests. **Thanks @estan for the contribution!**

**Changelog-friendly one-liner**: Add support for `--find-links` in `pip-compile` output.

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -280,18 +280,6 @@ def cli(
         }
         repository = LocalRequirementsRepository(existing_pins, repository)
 
-    log.debug("Using indexes:")
-    # remove duplicate index urls before processing
-    repository.finder.index_urls = list(dedup(repository.finder.index_urls))
-    for index_url in repository.finder.index_urls:
-        log.debug("  {}".format(index_url))
-
-    if repository.finder.find_links:
-        log.debug("")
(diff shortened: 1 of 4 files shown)


suutari:
FYI, I've implemented something similar in #453 over a year ago. Though with the difference that only relative directories are written out, since absolute ones rarely make sense to be stored into the txt file. If this gets merged, then #453 can be closed (or the other way around).

atugushev:
Hello @suutari,

Thank you for pointing that out! I agree there should be relative links. I'll port your code here if you don't mind.

atugushev:
@suutari 

> Though with the difference that only relative directories are written out

I've found an interesting [FIXME](https://github.com/pypa/pip/blob/master/src/pip/_internal/req/req_file.py#L241-L244):
```
# FIXME: it would be nice to keep track of the source
# of the find_links: support a find-links local path
# relative to a requirements file.
```

Maybe we should fix it on the `pip` side instead of hard coding that case in the `pip-tools`?

estan:
@atugushev I've been completely swamped and almost forgot about my PR. Thanks for taking care of this!

PR #803, opened 2019-05-02
Drop python 3.4

auvipy:
Remove py34 from:
- [X] setup
- [X] tox
- [X] appveyour 
- [X] travis


auvipy:
https://github.com/pypa/pip/releases/tag/19.1 is already on the wild :dancer:

atugushev:
> https://github.com/pypa/pip/releases/tag/19.1 is already on the wild 💃

19.1 still supports 3.4 with a warning.

auvipy:
> Hello @auvipy,
> 
> Thank you for the contribution! I would rather not to drop the support. I think we should wait at least a `pip` release (it should be right [after 19.1](https://github.com/pypa/pip/blob/f8732ac06dcb7c6af81dcdc0cc3158196b796baf/src/pip/_internal/cli/base_command.py#L141-L143)) with dropped support of the `python 3.4` and after, we could drop the support and make a release for the `pip-tools`.
> 
> But since (as i can see in the master branch) we have already removed the support in #804 and #805 (why they are not **approved**, but merged by the way, @auvipy?) my concerns may not reasonable anymore.
> 
> @vphilippon @blueyed what do you think?

IMHO, as it's not breaking anything and a very trivial valid change, these shouldn't go through the full formal process always, since practicality beats purity. And we can revert again if that creates a problem.  but I didn't merge anything with Red Button :dagger:

atugushev:
@auvipy 

> IMHO, as it's not breaking anything and a very trivial valid change, these shouldn't go through the full formal process always, since practicality beats purity. And we can revert again if that creates a problem. but I didn't merge anything with Red Button

It's not a problem. Just let's be more collaborative (even on trivial changes) :)

PS
It's been forgotten to drop support from `appveyour`, so i've fixed it.
`

auvipy:
I do agree with you buddy!

atugushev:
I've removed the current milestonoe to prepare new release.

auvipy:
that's alright

atugushev:
Since we drop 3.4 support i'd suggest to bump major version.

atugushev:
`pip-tools v4.0.0` is released.

PR #807, opened 2019-05-03
support URLs as packages

### LINKED ISSUE #272: Error with pip : pip-compile does not support URLs as packages

jcushman:
Fix #272

This PR handles URLs in requirements.in, building on work by @nim65s and @toejough.

This is similar to the previous PR #764, but I rebased @toejough's latest on top of master, updated the PyPI download url and some tests, and added support for --generate-hashes.

Note that URL requirements don't work for packages that use setuptools_scm, including pip-tools itself, but that's an inherent limitation of URL requirements and not an issue with pip-tools.

**Changelog-friendly one-liner**: support URLs as packages

##### Contributor checklist


piptools/utils.py
```diff
@@ -48,13 +48,33 @@ def make_install_requirement(name, version, extras, constraint=False):
     )
 
 
+def is_url_requirement(ireq, repository=None):
+    """
+    Finds if a requirement is a URL
+    """
+    if (
+        not ireq.link
+        or "files.pythonhosted.org" in ireq.link.url
```
atugushev: It could be valid for the link like this `https://example.com/storage/files.pythonhosted.org.zip`. Maybe it's better to check the `netloc`?

piptools/utils.py
```diff
@@ -48,13 +48,33 @@ def make_install_requirement(name, version, extras, constraint=False):
     )
 
 
+def is_url_requirement(ireq, repository=None):
+    """
+    Finds if a requirement is a URL
+    """
+    if (
```
atugushev: Could you please describe in the comment line why `"files.pythonhosted.org" in ireq.link.url` is not an URL requirement?

piptools/utils.py
```diff
@@ -48,13 +48,33 @@ def make_install_requirement(name, version, extras, constraint=False):
...
+        not ireq.link
+        or "files.pythonhosted.org" in ireq.link.url
+        or is_file_url(ireq.link)
+    ):
+        return False
+    if repository is not None and hasattr(repository, "finder"):
```
atugushev: Would be nice to have a comment here to explain that block.

tests/test_cli_compile.py
```diff
@@ -273,6 +273,35 @@ def test_locally_available_editable_package_is_not_archived_in_cache_dir(
...
+        "#egg=pytest-django"
+    )
+    with open("requirements.in", "w") as req_in:
+        req_in.write("-e " + vcs_package)
+    out = runner.invoke(cli, ["-n", "--rebuild"])
+    assert out.exit_code == 0
+    assert vcs_package in out.output
+    assert "pytest" in out.output  # dependency of pytest-django
+
+
```
atugushev: Could you add a test for a package with a link something like`https://files.pythonhosted.org/.../pip-8.1.1-py2.py3-none-any.whl`

piptools/utils.py
```diff
@@ -48,13 +48,33 @@ def make_install_requirement(name, version, extras, constraint=False):
     )
 
 
+def is_url_requirement(ireq, repository=None):
```
atugushev: Could you add tests for this func to `tests/test_utils.py`?

piptools/utils.py
```diff
@@ -48,13 +48,33 @@ def make_install_requirement(name, version, extras, constraint=False):
...
+
+
 def format_requirement(ireq, marker=None, hashes=None):
     """
     Generic formatter for pretty printing InstallRequirements to the terminal
     in a less verbose way than using its `__str__` method.
     """
     if ireq.editable:
         line = "-e {}".format(ireq.link)
+    elif is_url_requirement(ireq):
```
atugushev: Could you add a test like `test_format_requirement_url` in the `tests/test_utils.py` and cover that branch?

piptools/utils.py
```diff
@@ -99,7 +119,10 @@ def is_pinned_requirement(ireq):
     if ireq.editable:
         return False
 
-    if len(ireq.specifier._specs) != 1:
+    try:
```
atugushev: Could you add a test for this block to `tests/test_utils.py`?

blueyed:
Thanks for working on this!

Tried this quickly, and it fails when not using `#egg=foo` at the end:
```
Current constraints:
  git+https://github.com/blueyed/starlette.git@local

Finding the best candidates:
  found candidate git+https://github.com/blueyed/starlette.git@local (constraint was <any>)

None
  File "…/Vcs/pip-tools/.venv/bin/pip-compile", line 11, in <module>
    load_entry_point('pip-tools', 'console_scripts', 'pip-compile')()
  File "…/Vcs/pip-tools/.venv/lib/python3.7/site-packages/click/core.py", line 764, in __call__
  File "…/Vcs/pip-tools/.venv/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "…/Vcs/pip-tools/.venv/lib/python3.7/site-packages/click/core.py", line 956, in invoke
  File "…/Vcs/pip-tools/.venv/lib/python3.7/site-packages/click/core.py", line 555, in invoke
  File "…/Vcs/pip-tools/.venv/lib/python3.7/site-packages/click/decorators.py", line 17, in new_func
  File "…/Vcs/pip-tools/piptools/scripts/compile.py", line 415, in cli
    key_from_req(ireq.req) for ireq in constraints if not ireq.constraint
  File "…/Vcs/pip-tools/piptools/scripts/compile.py", line 415, in <setcomp>
    key_from_req(ireq.req) for ireq in constraints if not ireq.constraint
  File "…/Vcs/pip-tools/piptools/utils.py", line 30, in key_from_req
    key = req.name
AttributeError: 'NoneType' object has no attribute 'name'
```

Used `git+https://github.com/blueyed/starlette.git@local` - with `git+https://github.com/blueyed/starlette.git@local#egg=starlette` it worked, but just left it as-is - is that expected?

piptools/utils.py
```diff
@@ -48,13 +48,33 @@ def make_install_requirement(name, version, extras, constraint=False):
     )
 
 
+def is_url_requirement(ireq, repository=None):
+    """
+    Finds if a requirement is a URL
```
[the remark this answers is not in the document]
jcushman: Just replying quickly to a style issue before I get to the hard stuff. :)

Both are grammatically correct (because it's fine to pronounce it "a you are ell" or "an earl") but "[a URL](https://www.google.com/search?q="a+url")" is about five times as common as "[an URL](https://www.google.com/search?q="an+url")". Could I suggest leaving it as "a URL" to be more likely to match the usage of future PRs?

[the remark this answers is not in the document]
atugushev: Aw, okay :) Leave it as "a URL", thanks for the info.

jcushman:
Thanks for the reviews! I believe my updates address all the comments so far. I was able to greatly simplify the `is_url_requirement` function, which was the messiest part of the logic.

@blueyed:

> it fails when not using #egg=foo at the end

Good catch -- this should be fixed now.

> with git+https://github.com/blueyed/starlette.git@local#egg=starlette it worked, but just left it as-is - is that expected?

I think so, yes -- it should just copy URLs directly to requirements.txt, along with their deps. Looks like starlette doesn't have any dependencies so you'd just see the original URL.

blueyed:
> it should just copy URLs directly to requirements.txt

Shouldn't it pin it at the current commit?  (that's what I did in https://github.com/jazzband/pip-tools/pull/169 back then)

jcushman:
Update: the code I was adapting only applied to http/scm URLs and excluded file URLs, but I think file URLs actually work just fine. The latest code removes the exclusion and adds tests for file URLs. This also lets us drop the whole UnsupportedConstraint codepath because we're no longer excluding anything as unsupported.

jcushman:
> Shouldn't it pin it at the current commit? (that's what I did in #169 back then)

Oh, I see what you mean! Yes, it would be better to pin vcs URLs at the current commit.

What do you think of saving that feature for a separate pull request? The current PR is not a regression, AFAIK -- it treats `git+https://github.com/blueyed/starlette.git@local#egg=starlette` the same as `-e git+https://github.com/blueyed/starlette.git@local#egg=starlette`, which is also not pinned. In the meantime, you can work around the limitation (as I already do with `-e`) by using a commit URL instead of a tag/branch.

jcushman:
> I've noticed that pip-sync would try to sync URL package every time.

Good catch. The latest commit allows you to avoid this by providing the pinned version in the egg fragment:

```
$ pip-compile
#
# To update, run:
#
#    pip-compile
#
https://github.com/pallets/click/archive/7.0.zip#egg=click==7.0
$ pip-sync
Everything up-to-date
```

I think that's the best we can do, right? If the version isn't provided in the URL, the archive has to be downloaded and checked.

atugushev:
> Good catch. The latest commit allows you to avoid this by providing the pinned version in the egg fragment:

Great!

> I think that's the best we can do, right? If the version isn't provided in the URL, the archive has to be downloaded and checked.

You are right. Turns out that's correct behaviour. 

BTW, i've tried egg fragment with only package name:

```
$ pip-compile
#
# To update, run:
#
#    pip-compile
#
https://github.com/pallets/click/archive/7.0.zip#egg=click

$ pip-sync
Everything up-to-date
```

Shouldn't it download and check version if there is only package name in egg fragment and no version?

jcushman:
> Shouldn't it download and check version if there is only package name in egg fragment and no version?

Oops, yes it should. In fact it should download and reinstall, even if the version number matches in the download, because we don't know if the URL has changed. For example, suppose you change requirements.in from:

`https://github.com/jazzband/pip-tools/archive/master.zip#egg=example`

to

`https://github.com/jcushman/pip-tools/archive/master.zip#egg=example`

It would be surprising if pip-sync didn't change anything because I didn't update the version in my fork, right? We don't know how the installed module was specified, so we have to uninstall and reinstall on every pip-sync.

My latest commit hopefully strikes a good balance:

- If you choose to explicitly provide `#egg=example==7.0` in requirements.txt, then pip-sync believes you and only downloads and reinstalls if the version doesn't match.
- If you just provide `#egg=example`, or no name, then pip-sync downloads, uninstalls, and reinstalls.

atugushev:
> My latest commit hopefully strikes a good balance:
> 
> * If you choose to explicitly provide `#egg=example==7.0` in requirements.txt, then pip-sync believes you and only downloads and reinstalls if the version doesn't match.
> * If you just provide `#egg=example`, or no name, then pip-sync downloads, uninstalls, and reinstalls.

Awesome! Would've been nice to have tests for each scenario in `test_cli_sync.py`.

jcushman:
> Would've been nice to have tests for each scenario in test_cli_sync.py.

Those scenarios are tested via `diff()` unit tests in test_sync.py, the same as the rest of the pip-sync diff behavior. test_cli_sync.py only seems to have tests for cli wrapper flags and error messages.

What else is left to get this merged? I looked at the codecov/patch report, and I think the only lines that lost coverage are guards that are no longer hit, but that I'm wary of removing in case there are edge cases I don't know about.

atugushev:
@jcushman could you squash your commits before we merge this?

jcushman:
Squashed! `git rebase -i HEAD~19` is a personal record for me. :)

jedie:
> Anything else pending on this?

Seems that this will land in 3.7.0 and there is one issued open, see: https://github.com/jazzband/pip-tools/milestone/19 isn't it?

atugushev:
> Seems that this will land in 3.7.0 and there is one issued open, see: https://github.com/jazzband/pip-tools/milestone/19 isn't it?

We can postpone this issue to the next release.

atugushev:
I'll prepare a release within next 24 hours.

vphilippon:
@atugushev I validated pip-tools 3.7.0 release, it's on pypi.org now

Issue #806, opened 2019-05-03
Should --generate-hashes imply --allow-unsafe?


jcushman:
Some packages, such as [pytest](https://github.com/pytest-dev/pytest/blob/master/setup.py) and [Markdown](https://github.com/Python-Markdown/markdown/blob/master/setup.py), include minimum versions of setuptools in their requirements. Running `pip-compile --generate-hashes` without `--allow-unsafe` for those packages creates a `requirements.txt` file that doesn't work with `pip install`. This creates a bug that is easy to miss until deployment. Should `--allow-unsafe` be the default behavior for `--generate-hashes`?

##### Environment Versions

1. OS Type: macOS 10.14.4
1. Python version: Python 3.5.3
1. pip version: pip 19.1
1. pip-tools version: pip-compile, version 3.6.2.dev19+g8a1a8ef

##### Steps to replicate

1. New virtualenv:

```
$ pyenv virtualenv 3.5.3 piptest
$ pyenv local piptest
$ pip show setuptools
Name: setuptools
Version: 28.8.0
```

2. Require `Markdown`, which requires a newer version of setuptools:

```
$ echo "Markdown" > requirements.in
$ $ pip-compile --generate-hashes
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --generate-hashes
#
markdown==3.1 \
    --hash=sha256:fc4a6f69a656b8d858d7503bda633f4dd63c2d70cf80abdc6eafa64c4ae8c250 \
    --hash=sha256:fe463ff51e679377e3624984c829022e2cfb3be5518726b06f608a07a3aad680
```

3. Attempt pip install:

```
$ pip install -r requirements.txt
Collecting markdown==3.1 (from -r requirements.txt (line 7))
  Using cached https://files.pythonhosted.org/packages/f5/e4/d8c18f2555add57ff21bf25af36d827145896a07607486cc79a2aea641af/Markdown-3.1-py2.py3-none-any.whl
Collecting setuptools>=36 (from markdown==3.1->-r requirements.txt (line 7))
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
[...]

PR #814, opened 2019-05-12
Warn when --generate-hashes output is uninstallable

jcushman:
`pip-compile --generate-hashes` can generate a `requirements.txt` file that will be rejected by `pip install`, in at least two ways:

(1) When unsafe packages are left out (#806):

```
$ pip-compile --generate-hashes
#
# To update, run:
#
#    pip-compile --generate-hashes
#
markdown==3.1 \
    --hash=sha256:fc4a6f69a656b8d858d7503bda633f4dd63c2d70cf80abdc6eafa64c4ae8c250 \
    --hash=sha256:fe463ff51e679377e3624984c829022e2cfb3be5518726b06f608a07a3aad680
$ pip install -r requirements.txt
Collecting markdown==3.1 (from -r requirements.txt (line 7))
  Using cached https://files.pythonhosted.org/packages/f5/e4/d8c18f2555add57ff21bf25af36d827145896a07607486cc79a2aea641af/Markdown-3.1-py2.py3-none-any.whl
Collecting setuptools>=36 (from markdown==3.1->-r requirements.txt (line 7))
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    setuptools>=36 from https://files.pythonhosted.org/packages/ec/51/f45cea425fd5cb0b0380f5b0f048ebc1da5b417e48d304838c02d6288a1e/setuptools-41.0.1-py2.py3-none-any.whl#sha256=c7769ce668c7a333d84e17fe8b524b1c45e7ee9f7908ad0a73e1eda7e6a5aebf (from markdown==3.1->-r requirements.txt (line 7))
```

(2) When unhashable URL requirements are included:

```
$ pip-compile --generate-hashes
#
# To update, run:
#
#    pip-compile --generate-hashes
#
-e git+https://github.com/jazzband/pip-tools@7d86c8d3ecd1faa6be11c7ddc6b29a30ffd1dae3#egg=pip-tools
click==7.0 \
    --hash=sha256:2335065e6395b9e67ca716de5f7526736bfa6ceead690adf616d925bdc622b13 \
    --hash=sha256:5b94b49521f6456670fdb30cd82a4eca9412788a93fa6dd6df72c94d5a8ff2d7
first==2.0.2 \
    --hash=sha256:8d8e46e115ea8ac652c76123c0865e3ff18372aef6f03c22809ceefcea9dec86 \
    --hash=sha256:ff285b08c55f8c97ce4ea7012743af2495c9f1291785f163722bd36f6af6d3bf
six==1.12.0 \
    --hash=sha256:3350809f0555b11f552448330d0b52d5f24c91a322ea4a15ef22629740f3761c \
    --hash=sha256:d16a0141ec1a18405cd4ce8b4613101da75da0e9a7aec5bdd4fa804d0e0eba73
$ pip install -r requirements.txt
Obtaining pip-tools from git+https://github.com/jazzband/pip-tools@7d86c8d3ecd1faa6be11c7ddc6b29a30ffd1dae3#egg=pip-tools (from -r requirements.txt (line 7))
ERROR: The editable requirement pip-tools from git+https://github.com/jazzband/pip-tools@7d86c8d3ecd1faa6be11c7ddc6b29a30ffd1dae3#egg=pip-tools (from -r requirements.txt (line 7)) cannot be installed when requiring hashes, because there is no single file to hash.
```

This PR prints warning messages in the generated requirements.txt file for both of those situations, and also prints a warning message to stderr at the end:

```
$ pip-compile --generate-hashes
#
# To update, run:
#
#    pip-compile --generate-hashes
#
# WARNING: pip install will require the following package to be hashed.
# Consider using a hashable URL like https://github.com/jazzband/pip-tools/archive/SOMECOMMIT.zip
-e git+https://github.com/jazzband/pip-tools@7d86c8d3ecd1faa6be11c7ddc6b29a30ffd1dae3#egg=pip-tools
click==7.0 \
    --hash=sha256:2335065e6395b9e67ca716de5f7526736bfa6ceead690adf616d925bdc622b13 \
    --hash=sha256:5b94b49521f6456670fdb30cd82a4eca9412788a93fa6dd6df72c94d5a8ff2d7
first==2.0.2 \
    --hash=sha256:8d8e46e115ea8ac652c76123c0865e3ff18372aef6f03c22809ceefcea9dec86 \
    --hash=sha256:ff285b08c55f8c97ce4ea7012743af2495c9f1291785f163722bd36f6af6d3bf
markdown==3.1 \
    --hash=sha256:fc4a6f69a656b8d858d7503bda633f4dd63c2d70cf80abdc6eafa64c4ae8c250 \
    --hash=sha256:fe463ff51e679377e3624984c829022e2cfb3be5518726b06f608a07a3aad680
six==1.12.0 \
    --hash=sha256:3350809f0555b11f552448330d0b52d5f24c91a322ea4a15ef22629740f3761c \
    --hash=sha256:d16a0141ec1a18405cd4ce8b4613101da75da0e9a7aec5bdd4fa804d0e0eba73

# WARNING: The following packages were not pinned, but pip requires them to be
# pinned when the requirements file includes hashes. Consider using the --allow-unsafe flag.
# setuptools==41.0.1        # via markdown
[to stderr:]
The generated requirements file may be rejected by pip install. See # WARNING lines for details.
```

Note that the warning message text is static; this doesn't attempt to suggest a fixed URL for Github or similar VCS URLs, though that would be possible as an enhancement.

Here's an image of the output with color:

image

Tests are included for the two warning situations.

**Changelog-friendly one-liner**: Warn when pip-compile --generate-hashes output is uninstallable

##### Contributor checklist

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -14,6 +14,27 @@ from .utils import (
     key_from_ireq,
 )
 
+MESSAGE_UNHASHED_PACKAGE = comment(
+    "# WARNING: pip install will require the following package to be hashed."
+    "\n# Consider using a hashable URL like "
+    "https://github.com/jazzband/pip-tools/archive/SOMECOMMIT.zip"
+)
+
+MESSAGE_UNSAFE_PACKAGES_UNPINNED = comment(
+    "# WARNING: The following packages were not pinned, but pip requires them to be"
+    "\n# pinned when the requirements file includes hashes. "
+    "Consider using the --allow-unsafe flag."
+)
+
+MESSAGE_UNSAFE_PACKAGES = comment(
+    "# The following packages are considered to be unsafe in a requirements file:"
+)
+
+MESSAGE_UNINSTALLABLE = (
+    "The generated requirements file may be rejected by pip install. "
+    "See # WARNING lines for details."
+)
+
 
 class OutputWriter(object):
     def __init__(
@@ -104,12 +125,24 @@ class OutputWriter(object):
     def _iter_lines(
         self,
         results,
-        unsafe_requirements,
-        reverse_dependencies,
-        primary_packages,
-        markers,
-        hashes,
+        unsafe_requirements=None,
+        reverse_dependencies=None,
+        primary_packages=None,
+        markers=None,
+        hashes=None,
     ):
+        # default values
+        unsafe_requirements = unsafe_requirements or []
+        reverse_dependencies = reverse_dependencies or {}
+        primary_packages = primary_packages or []
+        markers = markers or {}
+        hashes = hashes or {}
+
+        # Check for unhashed or unpinned packages if at least one package does have
+        # hashes, which will trigger pip install's --require-hashes mode.
+        warn_uninstallable = False
+        has_hashes = hashes and any(hash for hash in hashes.values())
+
         for line in self.write_header():
             yield line
         for line in self.write_flags():
@@ -125,6 +158,9 @@ class OutputWriter(object):
         packages = sorted(packages, key=self._sort_key)
 
         for ireq in packages:
+            if has_hashes and not hashes.get(ireq):
+                yield MESSAGE_UNHASHED_PACKAGE
+                warn_uninstallable = True
             line = self._format_requirement(
                 ireq,
                 reverse_dependencies,
@@ -137,10 +173,11 @@ class OutputWriter(object):
         if unsafe_requirements:
             unsafe_requirements = sorted(unsafe_requirements, key=self._sort_key)
             yield ""
-            yield comment(
-                "# The following packages are considered "
-                "to be unsafe in a requirements file:"
-            )
+            if has_hashes and not self.allow_unsafe:
+                yield MESSAGE_UNSAFE_PACKAGES_UNPINNED
+                warn_uninstallable = True
+            else:
+                yield MESSAGE_UNSAFE_PACKAGES
 
             for ireq in unsafe_requirements:
                 req = self._format_requirement(
@@ -154,6 +191,9 @@ class OutputWriter(object):
                     yield comment("# {}".format(req))
                 else:
                     yield req
+
+        if warn_uninstallable:
+            log.warning(MESSAGE_UNINSTALLABLE)
 
     def write(
         self,
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -3,7 +3,12 @@ from pytest import fixture, mark, raises
 from piptools._compat import FormatControl
 from piptools.scripts.compile import cli
 from piptools.utils import comment
-from piptools.writer import OutputWriter
+from piptools.writer import (
+    MESSAGE_UNHASHED_PACKAGE,
+    MESSAGE_UNSAFE_PACKAGES,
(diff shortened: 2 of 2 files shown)


piptools/writer.py
```diff
@@ -104,12 +125,17 @@ def write_flags(self):
     def _iter_lines(
         self,
         results,
-        unsafe_requirements,
-        reverse_dependencies,
-        primary_packages,
-        markers,
-        hashes,
+        unsafe_requirements=[],
```
atugushev: Why not `unsafe_requirements=None`?

jcushman: `None` works for me -- updated.

(I think my intuition here is to not make something a maybe/optional type if it doesn't have to be -- it's easier to write correct code against "zero or more items in a list" than against "zero or more items in a list, or nothing." But I don't feel strongly about it.)

atugushev: I'm just saying that mutable default values should be avoided (unless it's done on purpose for some reason). BTW, the rest of defaults of arguments are still mutable.

jcushman: Oh, I see. I updated the code to set all default values within the function.

Personally I think mutable default values are fine, because it should be the baseline assumption that a function does not modify its inputs anyway. If a function does modify its inputs then it shouldn't have mutable defaults, but it should also clearly express what it does in the signature so callers know what to expect. But again, don't feel strongly about it. :)

piptools/writer.py
```diff
@@ -104,12 +125,24 @@ def write_flags(self):
...
+        unsafe_requirements = unsafe_requirements or []
+        reverse_dependencies = reverse_dependencies or {}
+        primary_packages = primary_packages or []
+        markers = markers or {}
+        hashes = hashes or {}
+
+        # Check for unhashed or unpinned packages if at least one package does have
+        # hashes, which will trigger pip install's --require-hashes mode.
+        warn_uninstallable = False
+        has_hashes = hashes and any(hash for hash in hashes.values())
```
blueyed: What does the `any()` do here really, is it necessary?

jcushman: The idea is to only warn if the file is actually going to be rejected by pip, meaning some requirements have hashes and some don't.

The `any()` means we won't warn if a file contains *only* non-hashable requirements, like this:

```
$ cat requirements.in
-e git+https://github.com/pallets/click@master#egg=click
$ pip-compile requirements.in --generate-hashes
#
# To update, run:
#
#    pip-compile --generate-hashes requirements.in
#
-e git+https://github.com/pallets/click@master#egg=click
```

`hashes` in that situation will be `{<InstallRequirement object: click from git+https://github.com/pallets/click@master#egg=click (from -r requirements.in (line 1)) editable=True>: set()}`, so we have to check the values.

It seems better not to warn in this case because people might be intentionally splitting out editable packages into separate files -- it's been [recommended by pip maintainers](https://github.com/pypa/pip/issues/4995#issuecomment-360828403) to have a separate requirements.txt for editable requirements if you're using hashes. So users might run pip-compile with the same settings on a folder full of requirements files and not need to see warnings in this case.

PR #831, opened 2019-05-30
Always obey --upgrade-package, and allow it together with --upgrade

### LINKED ISSUE #829: pip-compile --upgrade-package doesn't apply when requirements.txt doesn't exist
I have some Django libraries I'm testing. I compile my `requirements.in` to multiple output files in a script, with a `--upgrade-package` (well `-P`) to pin each Django version.

I just discovered that if I remove the `.txt` files and run my script, the latest Django is output in each. If I run the script again, the second invocation of `pip-compile` then obeys the `--upgrade-package` pins.

One PR where I had to fix this: https://github.com/adamchainz/django-perf-rec/pull/198

##### Environment Versions

1. macOS
1. Python version: Pythons 3.5-3.7
1. pip version:  19.1.1
1. pip-tools version: 3.7.0

##### Steps to replicate

```
$ echo django > requirements.in
$ python -m piptools compile --upgrade-package 'django<2.2'
#
# To update, run:
#
#    pip-compile
#
django==2.2.1
pytz==2019.1              # via django
sqlparse==0.3.0           # via django
$ python -m piptools compile --upgrade-package 'django<2.2'
#
# To update, run:
#
#    pip-compile
#
django==2.1.8
pytz==2019.1              # via django
```

##### Expected result

Django should be pinned even when the output file doesn't exist. Sometimes it can't exist, such as when it's stdout!

##### Actual result

It isn't pinned unless the file exists

### LINKED ISSUE #830: pip-compile --upgrade and --upgrade-package needn't be mutually exclusive
Related to #829. I test some Django packages, with one requirements file for each. Sometimes I want to upgrade all the requirements in my files *except* Django, which I pin for each by passing e.g. `--upgrade-package 'Django>=2.1,<2.2`.

Currently the options are mutually exclusive but I think it makes sense to allow `pip-compile --upgrade --upgrade-package 'Django>=2.1,<2.2` to mean "Upgrade all packages, except from add this constraint on Django.

##### Environment Versions

1. macOS
1. Python version: Pythons 3.5-3.7
1. pip version: 19.1.1
1. pip-tools version: 3.7.0

##### Steps to replicate

See description

##### Expected result

The pins combine

##### Actual result

Error "Only one of --upgrade or --upgrade-package can be provided as an argument"

adamchainz:
Fixes #829. Closes #830.

Allow the two options together, essentially making `--upgrade-package` a way of passing an extra constraint. Always parse `--upgrade-package`, regardless of whether the output file exists.

`--upgrade` and `--upgrade-package` are no longer mutually exclusive.
`--upgrade-package` now works even if the output file doesn't exist.

##### Contributor checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -362,36 +362,63 @@ def test_input_file_without_extension(runner):
     assert os.path.exists("requirements.txt")
 
 
-def test_upgrade_packages_option(runner):
+@pytest.mark.parametrize("existing_requirements_txt", [True, False])
+def test_upgrade_packages_option(runner, existing_requirements_txt):
     """
     piptools respects --upgrade-package/-P inline list.
     """
     with open("requirements.in", "w") as req_in:
         req_in.write("small-fake-a\nsmall-fake-b")
-    with open("requirements.txt", "w") as req_in:
-        req_in.write("small-fake-a==0.1\nsmall-fake-b==0.1")
+    if existing_requirements_txt:
+        with open("requirements.txt", "w") as req_in:
+            req_in.write("small-fake-a==0.1\nsmall-fake-b==0.1")
 
     out = runner.invoke(cli, ["-P", "small-fake-b", "-f", MINIMAL_WHEELS_PATH])
 
     assert out.exit_code == 0
-    assert "small-fake-a==0.1" in out.output
+    if existing_requirements_txt:
+        assert "small-fake-a==0.1" in out.output
     assert "small-fake-b==0.3" in out.output
 
 
-def test_upgrade_packages_version_option(runner):
+@pytest.mark.parametrize("existing_requirements_txt", [True, False])
+def test_upgrade_packages_version_option(runner, existing_requirements_txt):
     """
     piptools respects --upgrade-package/-P inline list with specified versions.
     """
     with open("requirements.in", "w") as req_in:
         req_in.write("small-fake-a\nsmall-fake-b")
-    with open("requirements.txt", "w") as req_in:
-        req_in.write("small-fake-a==0.1\nsmall-fake-b==0.1")
+    if existing_requirements_txt:
+        with open("requirements.txt", "w") as req_in:
+            req_in.write("small-fake-a==0.1\nsmall-fake-b==0.1")
 
     out = runner.invoke(cli, ["-P", "small-fake-b==0.2", "-f", MINIMAL_WHEELS_PATH])
 
     assert out.exit_code == 0
-    assert "small-fake-a==0.1" in out.output
+    if existing_requirements_txt:
+        assert "small-fake-a==0.1" in out.output
     assert "small-fake-b==0.2" in out.output
+
+
+@pytest.mark.parametrize("existing_requirements_txt", [True, False])
+def test_upgrade_packages_version_option_and_upgrade(runner, existing_requirements_txt):
+    """
+    piptools respects --upgrade-package/-P inline list with specified versions
+    whilst also doing --upgrade.
+    """
+    with open("requirements.in", "w") as req_in:
+        req_in.write("small-fake-a\nsmall-fake-b")
+    if existing_requirements_txt:
+        with open("requirements.txt", "w") as req_in:
+            req_in.write("small-fake-a==0.1\nsmall-fake-b==0.1")
+
+    out = runner.invoke(
+        cli, ["--upgrade", "-P", "small-fake-b==0.1", "-f", MINIMAL_WHEELS_PATH]
+    )
+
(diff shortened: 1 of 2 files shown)


adamchainz:
As an external contributor, I can't request a reviewer or assign to a milestone 🙊

atugushev:
Thanks for the contribution! I would suggest split it to two PR, since #830 needs to be discussed.

adamchainz:
I will split if anyone finds the proposal offensive but I think it's quite an easy win

atugushev:
> I will split if anyone finds the proposal offensive but I think it's quite an easy win

At least it would be much easier to review for maintainers.

adamchainz:
The fix for #830 is just the removal of the constraint on line 228 of compile.py. The actual behavioural fix is done by the move of code that fixes #829. If I split it will really mean branch-on-branch for #830 which isn't better is it?

atugushev:
> The fix for #830 is just the removal of the constraint on line 228 of compile.py. The actual behavioural fix is done by the move of code that fixes #829. If I split it will really mean branch-on-branch for #830 which isn't better is it?

It would be better if this two issues depended on each other. But they look like not depended and can be fixed(implemented) separately. I must emphasise *it's not mandatory*, just a simple recommendation :)

Have you researched why the options `--upgrade` and `--upgrade-package` are mutually exclusive by a deliberate design decision?

atugushev:
> It would be better if this two issues depended on each other. But they look like not depended and can be fixed(implemented) separately. I must emphasise _it's not mandatory_, just a simple recommendation :)

Oh, sorry, they should be fixed by one shot of course.

tests/test_cli_compile.py
```diff
@@ -362,38 +362,65 @@ def test_input_file_without_extension(runner):
...
+        with open("requirements.txt", "w") as req_in:
+            req_in.write("small-fake-a==0.1\nsmall-fake-b==0.1")
+
+    out = runner.invoke(
+        cli, ["--upgrade", "-P", "small-fake-b==0.1", "-f", MINIMAL_WHEELS_PATH]
+    )
+
+    assert out.exit_code == 0
+    assert "small-fake-a==0.2" in out.output
+    assert "small-fake-b==0.1" in out.output
```
atugushev: Shouldn't the `small-fake-b` be also upgraded to `0.3` version?

atugushev: I see,  that's ok since `"-P", "small-fake-b==0.1"`

tests/test_cli_compile.py
```diff
@@ -362,38 +362,65 @@ def test_input_file_without_extension(runner):
...
+    if existing_requirements_txt:
+        with open("requirements.txt", "w") as req_in:
+            req_in.write("small-fake-a==0.1\nsmall-fake-b==0.1")
 
     out = runner.invoke(cli, ["-P", "small-fake-b", "-f", MINIMAL_WHEELS_PATH])
 
     assert out.exit_code == 0
-    assert "small-fake-a==0.1" in out.output
+    if existing_requirements_txt:
+        assert "small-fake-a==0.1" in out.output
```
atugushev: Would be better to test `small-fake-a` has been upgraded to `0.2` if `not existing_requirements_txt`:
```
    if existing_requirements_txt:
        assert "small-fake-a==0.1" in out.output
    else:
        assert "small-fake-a==0.2" in out.output
```

tests/test_cli_compile.py
```diff
@@ -362,38 +362,65 @@ def test_input_file_without_extension(runner):
...
+    if existing_requirements_txt:
+        with open("requirements.txt", "w") as req_in:
+            req_in.write("small-fake-a==0.1\nsmall-fake-b==0.1")
 
     out = runner.invoke(cli, ["-P", "small-fake-b==0.2", "-f", MINIMAL_WHEELS_PATH])
 
     assert out.exit_code == 0
-    assert "small-fake-a==0.1" in out.output
+    if existing_requirements_txt:
+        assert "small-fake-a==0.1" in out.output
```
atugushev: Here also need to check `small-fake-a` version if `not existing_requirements_txt`.

adamchainz:
> Have you researched why the options --upgrade and --upgrade-package are mutually exclusive by a deliberate design decision?

It doesn't seem there was too much deliberation in #409 which added `--upgrade-package` and added the mutual exclusivity. I think it was just easier at the time given the way the code was laid out then.

adamchainz:
Thanks @atugushev ! 🎉

atugushev:
@adamchainz thanks great work! I'll prepare a release within 24h.

atugushev:
`pip-tools v3.8.0` is released

PR #844, opened 2019-07-10
Ignore deprecation warnings from pip module

atugushev:
Hides a deprecation warnings that happen in code that `piptools` has no control over. 

```
pip/_vendor/msgpack/fallback.py:222: PendingDeprecationWarning: encoding is deprecated, Use raw=False instead.
    PendingDeprecationWarning)
```

Reduces `tox` output:

- before -  https://travis-ci.org/jazzband/pip-tools/jobs/554999138
- after - https://travis-ci.org/atugushev/pip-tools/jobs/556739390


atugushev:
@auvipy thanks

PR #846, opened 2019-07-12
Add option to preserve original inline comments (using best effort)

### LINKED ISSUE #391: Include comments from input files in output file
Sometimes it's useful to have comments in input files where versions have been restricted, e.g.: `pytest<3  # Don't upgrade to Pytest 3.x because <foobar>`.
##### Steps to replicate
1. Add an inline comment to a line in an input file.
2. Run `pip-compile`.
##### Expected result

The comment should be preserved in the output file against the same line.
##### Actual result

The comment doesn't appear in the output file.
##### Specific use case

My actual use case here is compatibility with [Pyup](https://pyup.io/)'s [dependency filter](https://pyup.io/docs/filter/) functionality, which uses comments in requirements files like:

```
foo==1.2.3  # pyup: != 1.2.4  # disable 1.2.4, breaks tests
bar  # pyup: ignore
```

Currently, the best solution I can think of is to extend my `Makefile` to add these comments after running `pip-compile`.

MartinFalatic:
Resolves #391

Add `--preserve-inline-comments` option to preserve inline comments from `requirements.in` files.
This allows tooling that requires such annotations to operate properly.

**Changelog-friendly one-liner**:  Add option to preserve original inline comments (using best effort)

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -2,8 +2,10 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import os
+import re
 import sys
 import tempfile
+from collections import defaultdict
 
 from click.utils import safecall
 
@@ -161,6 +163,12 @@ DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
     "Build dependencies specified by PEP 518 must be already installed "
     "if build isolation is disabled.",
 )
+@click.option(
+    "--preserve-inline-comments",
+    is_flag=True,
+    default=False,
+    help="Preserve inline requirement comments",
+)
 def cli(
     ctx,
     verbose,
@@ -186,6 +194,7 @@ def cli(
     src_files,
     max_rounds,
     build_isolation,
+    preserve_inline_comments,
 ):
     """Compiles requirements.txt from requirements.in specs."""
     log.verbosity = verbose - quiet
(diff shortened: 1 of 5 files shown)


MartinFalatic:
Tests are WIP as there is no test that actually uses a custom requirements.in file prior to this (the existing ones only appear to be generated from setup files).

auvipy:
checkq is failing

MartinFalatic:
@auvipy I know - black and isort apparently. Will be doing more with it this weekend to resolve those.

atugushev:
Hello @MartinFalatic,

Thanks for your efforts to improve `pip-tools`. Just wondering why do we need inline comments (or any other comments) in the output file? Wouldn't enough comments in the input file?

MartinFalatic:
@atugushev Because tooling like Pyup depends on comments in the `requirements.txt` files to properly determine whether an update exists for a given package, without bringing in unwanted updates (e.g., one might want to be notified about Django 1.9 updates, but not Django 2.x updates).

This is the reason https://github.com/jazzband/pip-tools/issues/391 was opened in the first place. I'd like to make better use pip-compile as an integral part of our build and integration process, but the lack of this functionality prevents that adoption. Such functionality would make it possible to run pip-compile without losing those functional inline comments, and therefore without having to fork or hack around the lack thereof.

As this is additional functionality it shouldn't impact existing functionality in pip-tools, but instead augment it on demand.

MartinFalatic:
The current build error (`"__init__() got an unexpected keyword argument 'allow_all_prereleases'"`) only seen on the `PIP=master` builds _appears_ to be something outside of my changes.

atugushev:
@MartinFalatic 

> @atugushev Because tooling like Pyup depends on comments in the `requirements.txt` files to properly determine whether an update exists for a given package, without bringing in unwanted updates (e.g., one might want to be notified about Django 1.9 updates, but not Django 2.x updates).

> Edit: And no, the input file would not be sufficient. The input file would have the minimum dependencies, while pip-compile produces a much more comprehensive set of pinned requirements on output. Tooling like Pyup look at the totality of pinned requirements to see what packages have upgrades available - leveraging any comments - and reports on that in detail.

I see. On the other hand, how are you going to add pyup filters to external dependencies? For example if you have this `requirements.in` file:
```
$ cat requirements.in
django # pyup: ignore
six
```

And this is your `requirements.txt` file:
```
$ cat requirements.txt
django==2.2.3 # pyup: ignore
pytz==2019.1              # via django
six==1.12.0
sqlparse==0.3.0           # via django
```

As you can see you can't add `# pyup: ignore` to `pytz` (just an example), since it's not listed in `requirements.in`.

atugushev:
@MartinFalatic 

> The current build error (`"__init__() got an unexpected keyword argument 'allow_all_prereleases'"`) only seen on the `PIP=master` builds _appears_ to be something outside of my changes.

Test for `PIP=master` are allowed to fail.

MartinFalatic:
@atugushev You are correct - you normally wouldn't pin a non-primary dependency for pyup purposes until you know it's going to be a problem.

Pyup (like pip-compile) will recommend a dependency upgrade when it's available, within any constraints it is given in the comment. In our workflow we run an automated build for such recommendations, to see if the dependency causes any automated test failures (or a build failure). At that point, we know we need to pin that particular dependency down to a narrower range.

In a workflow with pip-compile involved, that means adding the problematic dependency to the primary deps in `requirements.in` with a suitable limit, e.g:

    pytz==2019.1   # pyup: <2019.6

Then, when that is propagated to `requirements.txt` using `--preserve-inline-comments`, pyup will no longer have issues with that dependency.

(Such pyup annotations are periodically checked for necessity, and may contain other commentary to that effect - it's not a best practice to keep secondary dependencies pinned down indefinitely, but it's an effective compromise when it's necessary.)

I will look into the code coverage issues next.

atugushev:
@MartinFalatic 

> In a workflow with pip-compile involved, that means adding the problematic dependency to the primary deps in `requirements.in` with a suitable limit, e.g:
> 
> ```
> pytz==2019.1   # pyup: <2019.6


Yes, in that case you have to add this package to a primary packages list. Quite acceptable though.

I found that current PR doesn't handle workflow with a `base.in` requirement file. See:

```bash
$ cat base.in
django # pyup: <2.0

$ cat requirements.in
-r base.in

$ cat dev-requirements.in
-r base.in
django-debug-toolbar
```

MartyMacGyver:
I will look into the issue with the resolver and included requirements via the `-r` specifier. (I believe it's mediated by the Resolver class in https://github.com/jazzband/pip-tools/blob/b37b91d251f7c7e590047f902e6f51e9a49d276c/piptools/resolver.py#L94 so it ought to be solvable).

MartinFalatic:
I will look into the issue with the resolver and included requirements via the `-r` specifier. (I believe it's mediated by the Resolver class in https://github.com/jazzband/pip-tools/blob/b37b91d251f7c7e590047f902e6f51e9a49d276c/piptools/resolver.py#L94 so it ought to be solvable).

PR #863, opened 2019-07-31
Add retry wrapper to codecov command in CI

atugushev:
To prevent fail job if sending report to codecov is failed for some reason (usually timeout). It's happening too frequently on AppVeyor CI.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -99,5 +99,5 @@ after_test:
   - IF NOT "x%TOXENV:-coverage=%"=="x%TOXENV%" (
       pip install codecov &&
       coverage xml &&
-      codecov --required -X gcov pycov search -f coverage.xml -n %TOXENV%-windows
+      appveyor-retry codecov --required -X gcov pycov search -f coverage.xml -n %TOXENV%-windows
     )
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ jobs:
 
 after_success:
   - travis_retry pip install codecov coveralls
-  - codecov -n "py${TRAVIS_PYTHON_VERSION}-pip${PIP}-${TRAVIS_OS_NAME}"
+  - travis_retry codecov --required -n "py${TRAVIS_PYTHON_VERSION}-pip${PIP}-${TRAVIS_OS_NAME}"
   - "COVERALLS_PARALLEL=true coveralls"
 
 notifications:


.appveyor.yml
```diff
@@ -99,5 +99,5 @@ after_test:
   - IF NOT "x%TOXENV:-coverage=%"=="x%TOXENV%" (
       pip install codecov &&
       coverage xml &&
-      codecov --required -X gcov pycov search -f coverage.xml -n %TOXENV%-windows
+      scripts\retry codecov --required -X gcov pycov search -f coverage.xml -n %TOXENV%-windows
```
blueyed: Using codecov-bash would retry automatically.

atugushev: Hm, the same as [codecov-python](https://github.com/codecov/codecov-python/blob/1d5d2881878b01b37922e97c8aedef6cf6cf5e7e/codecov/__init__.py#L731-L732).

atugushev: Looks like it's failing [here](https://github.com/codecov/codecov-python/blob/1d5d2881878b01b37922e97c8aedef6cf6cf5e7e/codecov/__init__.py#L756) and doesn't go to retry. With a script it'll retry anyways.

atugushev: But i found that codecov-bash behaves differently and retry if curl fails, maybe it's better to use it.

atugushev: @blueyed BTW has codecov-bash fixed problems with CI and connection timeout in `pytest`?

blueyed: > But i found that codecov-bash behaves differently and retry if curl fails, maybe it's better to use it.

Yes.

> BTW has codecov-bash fixed problems with CI and connection timeout in `pytest`?

That was/is related to their backend (not the uploader), and currently they have disabled flags in some places in their UI to work around this (i.e. there shouldn't be an issue with regard to using (many) flags).

atugushev: Okay we can use codecov-bash on linux in Travis, but what about windows in AppVeyor? :)

blueyed: Depends on the environment I think - with neovim we're using it there also (bash is available there).

atugushev: I just figured out that we don't need to store retry script, because appveyor has own `appveyor-retry`. Look it's working https://ci.appveyor.com/project/jazzband/pip-tools/builds/26427323/job/hnbal8gb5i7scrv6

We also can wrap codecov command to `travis_retry`.

atugushev: Note, codecov-bash won't retry if [pinging](https://github.com/codecov/codecov-bash/blob/master/codecov#L1489) was unsuccessful.

See https://github.com/codecov/codecov-bash/blob/master/codecov#L1513-L1517

atugushev: @blueyed 

Do you think we could go on with [travis|appveyor]-retry wrapper, or try to use `codecov-bash` anyways?

blueyed: It breaks with status=400, which is usually due to invalid arguments (where it does not make sense to retry).
I'd just go with codecov-bash.  It is faster to install than codecov-python also, but might of course fail to download already when their service is down.  To be safe the whole command could be wrapped in travis-retry/appveyor-retry then.

atugushev:
Thanks @blueyed!

.appveyor.yml
atugushev: BTW, It is said in readme that codecov-bash doesn't work on AppVeyour. See https://github.com/codecov/codecov-bash

blueyed: They say it is not supported.. but it works for Neovim: https://ci.appveyor.com/project/neovim/neovim/builds/26438145/job/kic2rmecfk1tv2aq?fullLog=true#L10570

PR #859, opened 2019-07-25
[WIP] tox: add -U option to install_command

atugushev:
There was an issue with PIP=latest because of cache. See #858 for details. To fix it i've added to pip upgrade option to make sure that pip package (without explicit version) upgraded to a latest version.

Also print pip version explicitly for debugging. Now you know exactly which version of pip was installed with PIP=latest. 

Closes #858.

Issue #858, opened 2019-07-25
py38 task is broken on Travis


atugushev:
```
$ tox
GLOB sdist-make: /home/travis/build/jazzband/pip-tools/setup.py
py38-piplatest-coverage create: /home/travis/build/jazzband/pip-tools/.tox/py38-piplatest-coverage
py38-piplatest-coverage installdeps: pip, mock, pytest, pytest-cov
py38-piplatest-coverage inst: /home/travis/build/jazzband/pip-tools/.tox/.tmp/package/1/pip-tools-4.0.0.zip
py38-piplatest-coverage installed: atomicwrites==1.3.0,attrs==19.1.0,Click==7.0,coverage==4.5.3,importlib-metadata==0.18,mock==3.0.5,more-itertools==7.2.0,packaging==19.0,pip-tools==4.0.0,pluggy==0.12.0,py==1.8.0,pyparsing==2.4.1.1,pytest==5.0.1,pytest-cov==2.7.1,six==1.12.0,wcwidth==0.1.7,zipp==0.5.2
py38-piplatest-coverage run-test-pre: PYTHONHASHSEED='518900656'
py38-piplatest-coverage run-test: commands[0] | pytest
ERROR: InvocationError for command /home/travis/build/jazzband/pip-tools/.tox/py38-piplatest-coverage/bin/pytest (exited with code -11)
___________________________________ summary ____________________________________
ERROR:   py38-piplatest-coverage: commands failed
The command "tox" exited with 1.
```

See full log https://travis-ci.org/jazzband/pip-tools/jobs/563357768

atugushev:
Clearing cache on Travis fixes the problem.

I've also found that PIP=latest doesn't install the latest pip. I've added to the tox command `pip --version` before `pytest` and made sure of that. See this log https://travis-ci.org/atugushev/pip-tools/jobs/563767730. Looks like tox installs last cached pip version. As a workaround i've added -U to `install_command= python -m pip install -U {opts} {packages}` and now it installs correct version.

atugushev:
Fixen by #867

atugushev:
> I've also found that PIP=latest doesn't install the latest pip. I've added to the tox command `pip --version` before `pytest` and made sure of that. See this log https://travis-ci.org/atugushev/pip-tools/jobs/563767730. Looks like tox installs last cached pip version. As a workaround i've added -U to `install_command= python -m pip install -U {opts} {packages}` and now it installs correct version.

FTR, this should be fixed in #868.

PR #866, opened 2019-08-02
Update install_command in tox to force install up-to-date dependencies

atugushev:
Force pip by `--no-cache-dir` to pull in up-to-date versions of all dependencies. See https://github.com/jazzband/pip-tools/issues/858#issuecomment-515244332.

Print explicitly pip version for debugging to know which exactly version
of pip was installed with PIP=latest.

Also had to upgrade 19.0 to 19.0.3, since pip crashes with --no-cache-dir in 19.0. See https://github.com/pypa/pip/issues/6158.

P.S.
Previously i've been considered to add `-U` flag, but it doesn't work with `pip==8.1.1` on windows py36. See https://github.com/jazzband/pip-tools/pull/859#issuecomment-515373865.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,8 +13,8 @@ environment:
           PIP: 10.0.1
         - TOXENV: py27-pip18.0
           PIP: 18.0
-        - TOXENV: py27-pip19.0
-          PIP: 19.0
+        - TOXENV: py27-pip19.0.3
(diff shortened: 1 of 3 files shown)


blueyed:
In general / on top of this: https://github.com/jazzband/pip-tools/pull/867

atugushev:
Thanks @blueyed!

PR #868, opened 2019-08-04
Install dependencies in tox with --upgrade option

atugushev:
The workaround with `--cache-dir` didn't work (#866).  See the output of `pip --version` in CI jobs:
 - [Travis-CI](https://travis-ci.org/jazzband/pip-tools/jobs/567417975#L213-L214)
 - [AppVeyor](https://ci.appveyor.com/project/jazzband/pip-tools/builds/26447187/job/14xbg8r1e3a4psyh#L42)

Had to comment off py36-pip8.1.1 and py37-pip8.1.1 dimension in AppVeyor matrix due to `python -m pip install -U pip` doesn't work because of a bug. See https://github.com/pypa/pip/issues/3964 for details.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -41,8 +41,10 @@ environment:
         - TOXENV: py35-piplatest
           PIP: latest
 
-        - TOXENV: py36-pip8.1.1
-          PIP: 8.1.1
+        # Note, pip install -U pip doesn't work on py36 windows
+        # See https://github.com/pypa/pip/issues/3964
+        # - TOXENV: py36-pip8.1.1
+        #   PIP: 8.1.1
         - TOXENV: py36-pip9.0.1
           PIP: 9.0.1
         - TOXENV: py36-pip9.0.3
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -62,8 +62,10 @@ environment:
         - TOXENV: py36-piplatest
           PIP: latest
 
-        - TOXENV: py37-pip8.1.1
-          PIP: 8.1.1
+        # Note, pip install -U pip doesn't work on py37 windows
+        # See https://github.com/pypa/pip/issues/3964
+        # - TOXENV: py37-pip8.1.1
+        #   PIP: 8.1.1
         - TOXENV: py37-pip9.0.1
           PIP: 9.0.1
         - TOXENV: py37-pip9.0.3
--- a/tox.ini
+++ b/tox.ini
@@ -32,7 +32,7 @@ setenv =
     pip19.1: PIP==19.1
 
     coverage: PYTEST_ADDOPTS=--strict --doctest-modules --cov --cov-report=term-missing {env:PYTEST_ADDOPTS:}
-install_command = python -m pip install --no-cache-dir {opts} {packages}
+install_command = python -m pip install -U {opts} {packages}
 commands =
     pip --version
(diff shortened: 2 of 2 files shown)


atugushev:
Now pip version is correct:
https://travis-ci.org/jazzband/pip-tools/jobs/567437523#L228-L229

blueyed:
btw: with https://github.com/pytest-dev/pytest-cov/pull/314 upgrading virtualenv on AppVeyor was necessary.

atugushev:
Thanks @blueyed!

blueyed:
I've just tried this with pytest-cov also, where it causes to install e.g. the latest pytest, although it is pinned in `deps`..!

blueyed:
..but that might only happen with older pip (inside of tox).

atugushev:
Damn, i've just tried to seed inside `.tox/py37-piplatest/` older version of pip (18.0), re-run `tox -e py37-piplatst` and got 18.0. I guess it still does not work if i'm not mistaken.

See:

```bash
# Start from scratch
(.venv) $ rm .tox

# It's okay, here the latest version 19.2.1 is correct
(.venv) $ tox -e py37-piplatest -- --version
...
py37-piplatest run-test: commands[0] | pip --version
pip 19.2.1 from /Users/albert/Projects/pip-tools/.tox/py37-piplatest/lib/python3.7/site-packages/pip (python 3.7)
...

# Jump to .tox/py37-piplatest virtualenv and install pip==18.0
(.venv) $ source .tox/py37-piplatest/bin/activate
(py37-piplatest) $ pip install pip==18.0 -q

# What? Still 18.0? Why it's not upgraded to 19.2.1?
(py37-piplatest) $ source .venv/bin/activate
(.venv) $ tox -e py37-piplatest -- --version
...
py37-piplatest run-test: commands[0] | pip --version
pip 18.0 from /Users/albert/Projects/pip-tools/.tox/py37-piplatest/lib/python3.7/site-packages/pip (python 3.7)
...

```

blueyed:
> ```shell
> # Jump to .tox/py37-piplatest virtualenv and install pip==18.0
> (.venv) $ source .tox/py37-piplatest/bin/activate
> (py37-piplatest) $ pip install pip==18.0 -q
> ```

JFI: `tox/py37-piplatest/bin/pip ...` can be used directly (without activating).

> Maybe tox checks whether the package installed or not? If installed, it doesn't run install_command on that package.

Could be..  although changing the version in tox.ini would usually recreate it (but likely more due to the hash of the file/config).

atugushev:
> JFI: `tox/py37-piplatest/bin/pip ...` can be used directly (without activating).

Yes, indeed. Thanks!

> Could be.. although changing the version in tox.ini would usually recreate it (but likely more due to the hash of the file/config).

I've debugged it and found that install_command for deps runs only once per setup virtualenv. On the second run the install_command won't called for deps.

Maybe we should install needed version of pip on `command` setting?

atugushev:
For example:

```diff
 commands =
+    piplatest: pip install -U pip
     pip --version
     pytest {posargs}
```

atugushev:
Oh, now it works fine!

atugushev:
> I've debugged it and found that install_command for deps runs only once per setup virtualenv. On the second run the install_command won't be called for deps.

Just a thought, in this case `-U` in `install_command` makes no sence.

PR #872, opened 2019-08-05
Filter deprecation warnings from pip._vendor module

atugushev:
Ignore `DeprecationWarning` and `PendingDeprecationWarning` from module
`pip._vendor.*` that pip-tools has no control over.

Ref: #844


atugushev:
Thanks @blueyed!

PR #877, opened 2019-08-20
[WIP] Parse extras_require when compile setup.py

### LINKED ISSUE #625: Pip-compile ignores extras_require when parsing setup.py
The documentation for python-wheel describes ways to define conditional dependencies: http://wheel.readthedocs.io/en/stable/index.html?highlight=extras_require#defining-conditional-dependencies

The most well-supported of these is to use the parameter `extras_require` to specify dependencies for specific environment markers. For example:

```
from setuptools import setup
setup(
    name='mypackage',
    extras_require={
        ':platform_python_implementation=="CPython"': ['python-cjson'],
    }
)
```

This allows us to install optimisations for CPython which are not available on other implementations of Python. This works correctly for pip/setuptools (e.g. `pip install .` or `python setup.py install`) but is ignored by pip-compile.

##### Environment Versions

1. OS Type (Linux/Ubuntu)
1. Python version: 2.7
1. pip version: 9.0.1
1. pip-tools version: 1.11.0

##### Steps to replicate

1. Create a package with a conditional dependency specified using `extras_require` in `setup.py`.
2. Run pip-compile on the package in an environment which meets the conditions, and in one that does not.

##### Expected result

The generated requirements.txt should contain the conditional packages for the environment which meets the conditions, and not otherwise.

##### Actual result

The conditional dependencies are not present in any requirements files.

atugushev:
Closes #625.

**Changelog-friendly one-liner**: Parse `extras_require` when compile `setup.py` file.

TODO

- [x] handle `extras_require={'dev': "small-fake-a==0.1"}`
- [ ] research how pip handles `extras_require` via `pip install -e .[dev]`

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -303,6 +303,18 @@ def cli(
 
                 dist = run_setup(src_file)
                 tmpfile.write("\n".join(dist.install_requires))
+
+                # Parse extras_require of the form
+                # "extra:{marker}": ["package"] to
+                # "package ; {marker}" if extra is empty.
+                for extra_and_marker, packages in dist.extras_require.items():
+                    extra, marker = extra_and_marker.split(":")
+                    # TODO: non-empty extras will be ignored, perhaps we need
+                    #  an option like --with-extra to specify one. See GH-625.
+                    if extra:
+                        continue
+                    for package in packages:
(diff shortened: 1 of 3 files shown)


atugushev:
I'll close it in favor of #908.

PR #880, opened 2019-08-25
Use skipif instead of xfail in test_filter_pip_markers

atugushev:
Since the `test_filter_pip_markers` is not gonna be passed anymore
with pip<9, so better to skip it than expect to fail.

The proper support of environment markers has been added in pip 9.0.0.


atugushev:
Thanks as always, @blueyed, for reviewing this!

PR #884, opened 2019-09-05
Fix formatting issues in README

atugushev:
Split too long lines in code block to avoid scrollbar and In favour of readability.

Also expand `-P` to `--upgrade-package`. Better to use long-name options, because it's easier to remember human-readable names.


atugushev:
@EmadMokhtar thanks for taking a look at this!

PR #893, opened 2019-09-15
Add feature request template

atugushev:
A template gladly was taken from pip.

Also, update the bug-report template.


atugushev:
@blueyed thanks for reviewing this!

PR #898, opened 2019-09-17
Install pip master in tox via commands_pre directive

atugushev:
Directive `deps=-e git+https://github.com/pypa/pip.git@master#egg=pip` in tox would install pip only once during the setup environment. The problem is that for new commits in master `tox -e py37-pipmaster` won't reinstall pip.

The workaround is the same as for `piplatest`. See https://github.com/jazzband/pip-tools/issues/858#issuecomment-515244332 for the reference.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,6 @@ skip_missing_interpreters = True
 [testenv]
 deps =
     piplatest: pip
-    pipmaster: -e git+https://github.com/pypa/pip.git@master#egg=pip
     pip8.1.1: pip==8.1.1
     pip9.0.1: pip==9.0.1
     pip9.0.3: pip==9.0.3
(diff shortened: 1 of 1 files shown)


atugushev:
Codecov bot maske me sad :( CI has not finished yet!

blueyed:
> The problem is that for new commits in master tox -e py37-pipmaster won't reinstall pip.

Not really sure if that isn't OK - i.e. you would use `-r` to recreate it.
Probably not cool to get a dep updated during working with the env, and then suddenly a pip update changed behavior.

blueyed:
`piplatest` is a bit different, since it at least is more "stable", but the same applies there (ref https://github.com/jazzband/pip-tools/pull/899).

atugushev:
> Not really sure if that isn't OK - i.e. you would use `-r` to recreate it.
> Probably not cool to get a dep updated during working with the env, and then suddenly a pip update changed behavior.

Hm, that definitely makes sense. I'll close this then.

PR #899, opened 2019-09-17
Remove redundant deps=piplatest declaration from tox

atugushev:
Because tox already installs piplatest in the `commands_pre` directive.


atugushev:
Thanks @blueyed as always!

PR #906, opened 2019-09-22
Allow failing PIP=19.3 until pip being released

atugushev:
@codingjoe looks like I've forgotten it in #864. Master is broken and new PRs are CI-red now.


atugushev:
Thanks @codingjoe!

PR #907, opened 2019-09-23
Include already-installed matching requirements in the to_install set, so that pip-sync reliably installs dependencies

### LINKED ISSUE #896: pip-sync yields different results depending on pre-installed package set
Update: The shape of this issue has changed a bit since creation, due to discoveries in this discussion, and the addition of the `--pip-args` option. I'm renaming it, and adding a new summary:

---

If `pip-sync` has to actively install a requirement, its dependencies will end up in the environment. But if the requirement is already installed when `pip-sync` runs, its dependencies not present in the `txt` will be removed, "breaking" the environment.

```console
$ echo "pytest==5.1.2" > requirements.txt
$ pip-sync /dev/null
$ pip-sync requirements.txt
$ pip freeze | wc -l
12
$ pip-sync requirements.txt
$ pip freeze | wc -l
4
```

A `txt` which omits dependencies is, by pip-tools standards, malformed, and therefore `pip-sync`'s behavior is not currently guaranteed or defined. But:

1. As @Zac-HD says [below](https://github.com/jazzband/pip-tools/issues/896#issuecomment-538205892):

> IMO "running pip-sync a second time makes no further changes" is a very important property to maintain

2. IMO a package is not "properly" installed if its dependencies aren't present, and `pip-sync` should achieve "proper" installs.

3. If a user *wants* to "break" dependencies, they can explicitly do so with `--pip-args --no-deps`

4. While I don't see an advantage to the current (inconsistent) behavior, having the deps always installed (unless `--no-deps` is specified) would enable users to use `pip-sync` without issue on `txt`s that don't come from `pip-compile`. Though not officially supported, it's a nice bonus to be able to use `pip-sync` uniformly as a complete replacement/wrapper for `pip install -r`. Now that #927 is merged, I don't know of any other obstacle to supporting "foreign" `txt`s (in practice, if not policy).

My idea of a solution is #907, which keeps already-installed packages in the to-install set, as `pip` itself will refrain from reinstalling those anyway, and will ensure its dependencies get installed if necessary.

---

<details>
  <summary>Original report:</summary>

I looked but didn't find an existing issue for this.

I would like to use `pip-sync` on `requirements.txt` files that are not generated by `pip-compile`, for arbitrary (other folks') Python projects. In fact I had been doing so, and I guess I have been getting lucky, as I only recently came to a case where this failed, and was surprised to find when re-reading `pip-tools`'s README:

> **Be careful**: pip-sync is meant to be used only with a requirements.txt generated by pip-compile

The failure I notice is that `pip-sync` will uninstall hard dependencies of packages listed in the `requirements.txt`. I don't know if there are further problems.

Until/unless there are other specifically identified problems for externally crafted `requirements.txt` files, the solution would seem to be for `pip-sync` to refrain from uninstalling dependencies of explicitly listed packages. This might be as simple as doing all the uninstalling before any of the installing.

This would enable `pip-tools` users to continue to use `pip-sync` across all Python projects they work on, even if they are not in a position to enforce use of `pip-compile` to manage those projects' `requirements.txt` generation.

#### Alternative Solutions
Alternatively, of course, users can just use `pip install -r requirements.txt`. That however does not have the advantage of stripping the environment of non-required packages. Although I guess the solution could then be as simple as `pip-sync requirements.txt && pip install -r requirements.txt`.

#### Additional context
Simple demonstration of the problem:

```bash
echo "pytest==5.1.2" > requirements.txt
pip install -r requirements.txt
pip freeze
```

```
atomicwrites==1.3.0
attrs==19.1.0
Click==7.0
importlib-metadata==0.22
more-itertools==7.2.0
packaging==19.1
pip-tools==4.1.0
pluggy==0.13.0
py==1.8.0
pyparsing==2.4.2
pytest==5.1.2
six==1.12.0
wcwidth==0.1.7
zipp==0.6.0
```
```bash
pip-sync requirements.txt
pip freeze
```

```
Click==7.0
pip-tools==4.1.0
pytest==5.1.2
six==1.12.0
```

The result is a broken installation of `pytest`.
</details>

AndydeCleyre:
Have `diff` include already-installed matching requirements in its `to_install` set.

This means that during sync, dependencies of desired packages will be (re)installed after potentially being uninstalled, regardless of environment state at time of sync, for a more consistent, predictable, and practical resulting state. Fixes #896.

If this is not desired, `--no-deps` can be provided via the new `--pip-args` option, whereby it will be passed through to `pip install`, and un-locked deps will be either removed or not installed, to match the lockfile exactly.

The following existing tests check the `to_install` set, and are herein adjusted to expect the inclusion of explicit requirements (all in `test_sync.py`):

- `test_diff_should_do_nothing`
- `test_diff_should_uninstall`
- `test_diff_should_uninstall_with_markers`
- `test_diff_leave_packaging_packages_alone`
- `test_diff_leave_piptools_alone`
- `test_diff_with_matching_url_versions`

Note: `test_diff_with_matching_url_versions`, before this change, expects pip-tools to prevent reinstallation of versioned URL requirements, but this is unnecessary as `pip` itself will do that, reporting "Requirement already satisfied . . .", and the test's comment is updated to reflect that. If this is incorrect under some circumstance, please correct me!

**Changelog-friendly one-liner**: `pip-sync` now ensures installation of dependencies of a locked requirement, even if those deps are not locked and the locked requirement is already installed; this can be disabled with `pip-sync --pip-args=--no-deps`.

##### Contributor checklist

- [X] Provided the tests for the changes. (Well, modified existing tests)

If this were to be accepted, would the appropriate merge point be the next major version, as the behavior may not be what is currently expected? I view it as a bug fix, and do not know when or why the current behavior would be desired instead, but I am only me.

@atugushev Please have a look for discussion and review. Thanks!

### DIFF
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -150,7 +150,6 @@ def diff(
     """
     requirements_lut = {diff_key_from_ireq(r): r for r in compiled_requirements}
 
-    satisfied = set()  # holds keys
     to_install = set()  # holds InstallRequirement objects
     to_uninstall = set()  # holds keys
 
@@ -159,11 +158,9 @@ def diff(
         key = key_from_req(dist)
         if key not in requirements_lut or not requirements_lut[key].match_markers():
             to_uninstall.add(key)
-        elif requirements_lut[key].specifier.contains(dist.version):
-            satisfied.add(key)
 
     for key, requirement in requirements_lut.items():
-        if key not in satisfied and requirement.match_markers():
+        if requirement.match_markers():
             to_install.add(requirement)
 
     # Make sure to not uninstall any packages that should be ignored
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -118,7 +118,7 @@ def test_diff_should_do_nothing():
     reqs = []  # no requirements
 
     to_install, to_uninstall = diff(reqs, installed)
-    assert to_install == set()
+    assert to_install == set(reqs)
     assert to_uninstall == set()
 
 
@@ -136,7 +136,7 @@ def test_diff_should_uninstall(fake_dist):
     reqs = []
 
     to_install, to_uninstall = diff(reqs, installed)
-    assert to_install == set()
+    assert to_install == set(reqs)
     assert to_uninstall == {"django"}  # no version spec when uninstalling
 
 
@@ -200,7 +200,7 @@ def test_diff_leave_packaging_packages_alone(fake_dist, from_line):
     reqs = [from_line("django==1.7")]
 
     to_install, to_uninstall = diff(reqs, installed)
-    assert to_install == set()
+    assert to_install == set(reqs)
     assert to_uninstall == {"first"}
 
 
@@ -221,7 +221,7 @@ def test_diff_leave_piptools_alone(fake_dist, from_line):
     reqs = [from_line("django==1.7")]
 
     to_install, to_uninstall = diff(reqs, installed)
-    assert to_install == set()
+    assert to_install == set(reqs)
     assert to_uninstall == {"foobar"}
 
 
@@ -242,12 +242,12 @@ def test_diff_with_editable(fake_dist, from_editable):
 
 
 def test_diff_with_matching_url_versions(fake_dist, from_line):
-    # if URL version is explicitly provided, use it to avoid reinstalling
+    # if URL version is explicitly provided, pip itself will avoid reinstalling
     installed = [fake_dist("example==1.0")]
     reqs = [from_line("file:///example.zip#egg=example==1.0")]
 
     to_install, to_uninstall = diff(reqs, installed)
-    assert to_install == set()
+    assert to_install == set(reqs)
     assert to_uninstall == set()
 
 


AndydeCleyre:
@atugushev sure: #987

But I will note again, this more complete pull request (#907) is required to achieve consistent and predictable behavior.

Current behavior is to sometimes act like `--no-deps` is provided, sometimes not. So the new request #987 provides a way to consistently achieve `--no-deps` behavior, but without specifying that option the user will still sometimes get that behavior and sometimes not.

AndydeCleyre:
Before:

```console
$ echo "pytest==5.1.2" > requirements.txt
$ pip-sync /dev/null
$ pip-sync requirements.txt
$ pip freeze | wc -l
12
$ pip-sync requirements.txt
$ pip freeze | wc -l
4
$ pip-sync --pip-args --no-deps requirements.txt
$ pip freeze | wc -l
4
$ pip-sync /dev/null
$ pip-sync --pip-args --no-deps requirements.txt
$ pip freeze | wc -l
4
```

After:

```console
$ echo "pytest==5.1.2" > requirements.txt
$ pip-sync /dev/null
$ pip-sync requirements.txt
$ pip freeze | wc -l
12
$ pip-sync requirements.txt
$ pip freeze | wc -l
12
$ pip-sync --pip-args --no-deps requirements.txt
$ pip freeze | wc -l
4
$ pip-sync /dev/null
$ pip-sync --pip-args --no-deps requirements.txt
$ pip freeze | wc -l
4
```

ssbarnea:
@AndydeCleyre @atugushev are you still interested in getting this ready for merging? It is a very old PR.

At least the PR title and descriptions needs a little bit of simplification.

AndydeCleyre:
@ssbarnea 

FWIW I still want this, but there is not agreement that this is the right thing to do. But AFAIK there are no other PRs addressing #896.

ssbarnea (approved):

ssbarnea:
@atugushev I have the impression that @webknjaz took care of this one, does it look ok now? If so, lets merge it.

webknjaz:
@ssbarnea I think I only rebased it

atugushev:
Closed based on https://github.com/jazzband/pip-tools/issues/896#issuecomment-1347488704.

PR #909, opened 2019-09-24
Ensure pip-compile --no-header <blank requirements.in> creates/overwrites requirements.txt

### LINKED ISSUE #900: pip-compile --no-header <blank requirements.in> neither creates nor overwrites requirements.txt
If `requirements.in` is empty, running `pip-compile` will create or clobber `requirements.txt`, thereby removing all requirements, but if `--no-header` is passed to `pip-compile`, `requirements.txt` will not be affected in any way, no matter its state.

#### Environment Versions

1. Linux
1. Python version: `3.7.4`
1. pip version:  `19.2.3`
1. pip-tools version: `4.1.0`

#### Steps to replicate

```bash
touch requirements.in
pip-compile --no-header
# no txt file created (unexpected)
pip-compile
# txt file created (as expected)
echo plumbum > requirements.in
pip-compile
echo > requirements.in
pip-compile --no-header
cat requirements.txt
```
full, unexpected:
```python
#
# To update, run:
#
#    pip-compile
#
plumbum==1.6.7
```

AndydeCleyre:
Before, in the case where the writer should generate a blank output document (e.g. `requirements.txt`), no new file would be created, and any existing file with that name would be left unchanged. With this change, a blank document will created and overwritten if it exists.

This is implemented by having the writer keep track of whether it yields any lines, and if not, yielding an empty string (where before it would yield nothing).

A simpler solution would be to always yield an extra empty string at the end of `_iter_lines`, but that would result in a bit of awkward space at the bottom of all the `txt`s, and if that were fixed it would involve leaking the composition logic outside of `_iter_lines`.

**Changelog-friendly one-liner**: `pip-compile --no-header <blank requirements.in>` now creates/overwrites `requirements.txt`.

Fixes #900.

@atugushev I hereby request a review. If this does not satisfy the "Requested a review from another contributor" requirement, please tell me what does. Thanks!

##### Contributor checklist

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -146,10 +146,14 @@ class OutputWriter(object):
         warn_uninstallable = False
         has_hashes = hashes and any(hash for hash in hashes.values())
 
+        yielded = False
+
         for line in self.write_header():
             yield line
+            yielded = True
         for line in self.write_flags():
             yield line
+            yielded = True
 
         unsafe_requirements = (
             {r for r in results if r.name in UNSAFE_PACKAGES}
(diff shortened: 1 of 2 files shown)


atugushev:
@AndydeCleyre could you please rebase onto master to fix red CI checks?

AndydeCleyre:
Is a simple merge of master into here acceptable? GitHub pull requests don't love history rewrites.

I may be misunderstanding your alternative suggested fix, but:

1. If that's an `else` component of the `for` loop, that empty string would be written whenever the `for` loop finishes without interruption from a `break` statement, which would be every time, as there are no `break` statements. So the simpler version of this solution would be to add the extra dry run check and blank-write before the for loop.

2. While that will work, I thought changing the behavior of `_iter_lines` would be a less fragile solution, as it doesn't create a difference between what `_iter_lines` `yield`s and what `write` produces, noting that `_iter_lines` is used directly by three existing tests.

That said, always `write`ing an empty string is simpler, and I'm happy to create a replacement pull request with that, cherry picking the commit from here which adds the test, and double-checking that other tests are not broken.

piptools/writer.py
```diff
@@ -159,6 +163,8 @@ def _iter_lines(
         packages = {r for r in results if r.name not in UNSAFE_PACKAGES}
 
         packages = sorted(packages, key=self._sort_key)
+        if packages:
+            is_empty = False
 
         for ireq in packages:
```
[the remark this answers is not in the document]
AndydeCleyre: Yes, sure, thanks. I can be reluctant to nest if it's not necessary, but nesting here does have the advantage of not checking `packages` twice when it's non-empty, and it does match the `unsafe_requirements` block.

piptools/writer.py
```diff
@@ -159,23 +163,25 @@ def _iter_lines(
         packages = {r for r in results if r.name not in UNSAFE_PACKAGES}
 
         packages = sorted(packages, key=self._sort_key)
```
atugushev: Shouldn't this line be also under `if packages`?

AndydeCleyre: Indeed it should. No point in sorting nothing.

atugushev:
Could you rebase and squash before I merge this pull request?

atugushev:
Thanks, @AndydeCleyre!

Issue #671, opened 2018-08-01
pip-sync will happilly uninstall all packages if requirements.txt is empty


takeda:
This perhaps might be argued to be a feature, since `pip-sync` supposed to uninstall anything that is not in `requirements.txt` but this is dangerous in scenarios when an earlier operation has failed and caused `requirements.txt` to be empty. I'm wondering whether this safety measure could be added, if someone needs the behavior, maybe they can specify `--force`?

##### Environment Versions

1. OS Type: N/A
1. Python version: `$ python -V`: N/A
1. pip version:  `$ pip --version`: 18.0
1. pip-tools version: `$ pip-compile --version`: 2.0.2

##### Steps to replicate

1. echo -n > test.txt
2. pip-sync test.txt
3. pip list

##### Expected result

Operation would fail

##### Actual result

All packages were uninstalled

georgek:
Seems like pip-sync is doing the right thing to me. Always check the input to your programs before you run them.

takeda:
Heh, since I wrote this PR I ended up liking this behavior and started using `pip-sync /dev/null` to cleanup the environment, before installing something else.

My concern when filling this PR was using `pip-sync` in CMS code (salt, chef, puppet etc) for the purpose of deployment in case precious step failed and produced an empty file. Maybe an option `--fail-on-empty` could be used as a safety measure.

georgek:
Why not check the previous file with another tool like `wc`? Unix philosophy: do one thing and do it well.

atugushev:
I think we can ask if a user wants to continue. For example:

```python
if file_is_empty and not click.confirm('Do you want to uninstall all packages?'):
    sys.exit(0)
```

georgek:
Personally I'm against training users that commands can't do something dangerous. They'll get a shock when `rm -rf /` doesn't ask if they are sure.

Another concern is what if I really do want to do this? We'll need another option like `--yes` for running it in a non-interactive environment like CI.

atugushev:
> Another concern is what if I really do want to do this? We'll need another option like `--yes` for running it in a non-interactive environment like CI.

That makes sense. But I don't mind honestly, for I remember how I got myself into this trap once and deleted all packages in the env.

AndydeCleyre:
Uninstalling all packages is exactly what I expect and desire to happen when I `pip-sync` against an empty file.

georgek:
[...]

PR #913, opened 2019-09-27
Add --ask option to pip-sync

georgek:
This adds an `--ask` option to pip-sync which will first show the user what would change, like `--dry-run`, and then ask to confirm before continuing with the changes. This is supposed to address issue #671 without breaking existing workflows (in particular CI pipelines) and in a way that is well-established by existing tools like nmcli (network manager) and portage (Gentoo package manager).

This change requires a minor version bump.

**Changelog-friendly one-liner**: Add `--ask` option to `pip-sync`.

##### Contributor checklist

### DIFF
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -144,7 +144,14 @@ def diff(compiled_requirements, installed_dists):
     return (to_install, to_uninstall)
 
 
-def sync(to_install, to_uninstall, verbose=False, dry_run=False, install_flags=None):
+def sync(
+    to_install,
+    to_uninstall,
+    verbose=False,
+    dry_run=False,
+    install_flags=None,
+    ask=False,
+):
     """
     Install and uninstalls the given sets of modules.
     """
@@ -157,26 +164,34 @@ def sync(to_install, to_uninstall, verbose=False, dry_run=False, install_flags=N
     if not verbose:
         pip_flags += ["-q"]
 
-    if to_uninstall:
-        if dry_run:
+    if ask:
+        dry_run = True
+
+    if dry_run:
+        if to_uninstall:
             click.echo("Would uninstall:")
             for pkg in to_uninstall:
                 click.echo("  {}".format(pkg))
-        else:
+
+        if to_install:
+            click.echo("Would install:")
+            for ireq in to_install:
+                click.echo("  {}".format(format_requirement(ireq)))
+
+    if ask and click.confirm("Would you like to proceed with these changes?"):
+        dry_run = False
+
+    if not dry_run:
+        if to_uninstall:
             check_call(  # nosec
                 [sys.executable, "-m", "pip", "uninstall", "-y"]
                 + pip_flags
                 + sorted(to_uninstall)
             )
 
-    if to_install:
-        if install_flags is None:
-            install_flags = []
-        if dry_run:
-            click.echo("Would install:")
-            for ireq in to_install:
-                click.echo("  {}".format(format_requirement(ireq)))
-        else:
+        if to_install:
+            if install_flags is None:
+                install_flags = []
             # prepare requirement lines
             req_lines = []
             for ireq in sorted(to_install, key=key_from_ireq):
--- a/tests/test_cli_sync.py
+++ b/tests/test_cli_sync.py
@@ -147,3 +147,30 @@ def test_pip_install_flags(check_call, cli_flags, expected_install_flags, runner
     assert [args[6:] for args in call_args if args[3] == "install"] == [
         expected_install_flags
     ]
+
+
+@mock.patch("piptools.sync.check_call")
+def test_sync_ask_declined(check_call, runner):
+    """
+    Make sure nothing is installed if the confirmation is declined
+    """
+    with open("requirements.txt", "w") as req_in:
+        req_in.write("small-fake-a==1.10.0")
+
+    runner.invoke(cli, ["--ask"], input="n\n")
+
+    check_call.assert_not_called()
+
+
+@mock.patch("piptools.sync.check_call")
+def test_sync_ask_accepted(check_call, runner):
+    """
+    Make sure pip is called when the confirmation is accepted (even if
+    --dry-run is given)
+    """
+    with open("requirements.txt", "w") as req_in:
+        req_in.write("small-fake-a==1.10.0")
+
+    runner.invoke(cli, ["--ask", "--dry-run"], input="y\n")
+
+    assert check_call.call_count == 2
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -403,38 +403,80 @@ def test_sync_verbose(check_call, from_line):
         assert "-q" not in check_call_args
 
 
+@pytest.mark.parametrize(
+    ("to_install", "to_uninstall", "expected_message"),
+    [
+        ({"django==1.8", "click==4.0"}, set(), "Would install:"),
+        (set(), {"django==1.8", "click==4.0"}, "Would uninstall:"),
+    ],
+)
 @mock.patch("piptools.sync.click.echo")
-def test_sync_dry_run_would_install(echo, from_line):
+def test_sync_dry_run(echo, from_line, to_install, to_uninstall, expected_message):
     """
-    Sync with --dry-run option prints what's is going to be installed.
+    Sync with --dry-run option prints what's is going to be installed/uninstalled.
     """
-    to_install = {from_line("django==1.8"), from_line("click==4.0")}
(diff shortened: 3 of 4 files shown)


georgek:
@atugushev Please could you review?

piptools/sync.py
```diff
@@ -157,26 +164,34 @@ def sync(to_install, to_uninstall, verbose=False, dry_run=False, install_flags=N
     if not verbose:
         pip_flags += ["-q"]
 
-    if to_uninstall:
-        if dry_run:
+    if ask:
+        dry_run = True
```
atugushev: I think this `if` statement should be removed in favor of checking the `ask` below.

piptools/sync.py
```diff
@@ -157,26 +164,34 @@ def sync(to_install, to_uninstall, verbose=False, dry_run=False, install_flags=N
...
+
+    if dry_run:
```
atugushev:
```suggestion
    if dry_run or ask:
```
Options `dry_run` and `ask` both prints the intentions to install or uninstall packages, so check them both.

piptools/sync.py
```diff
@@ -157,26 +164,34 @@ def sync(to_install, to_uninstall, verbose=False, dry_run=False, install_flags=N
...
             for pkg in to_uninstall:
                 click.echo("  {}".format(pkg))
-        else:
+
+        if to_install:
+            click.echo("Would install:")
+            for ireq in to_install:
+                click.echo("  {}".format(format_requirement(ireq)))
+
+    if ask and click.confirm("Would you like to proceed with these changes?"):
```
atugushev:
```suggestion
    if ask and not click.confirm("Would you like to proceed with these changes?"):
```

If asking and user did NOT confirmed the action... -> (A)

piptools/sync.py
```diff
@@ -157,26 +164,34 @@ def sync(to_install, to_uninstall, verbose=False, dry_run=False, install_flags=N
...
+        dry_run = False
```
atugushev:
```suggestion
        return 0
```
-> (A) ... then finish immediately.

tests/test_sync.py
```diff
@@ -437,6 +437,64 @@ def test_sync_dry_run_would_uninstall(echo, from_line):
...
+    sync(to_install, set(), ask=True)
+
+    expected_calls = [
+        mock.call("Would install:"),
+        mock.call("  django==1.8"),
+        mock.call("  click==4.0"),
+    ]
+    echo.assert_has_calls(expected_calls, any_order=True)
+
+    confirm.assert_called_once_with("Would you like to proceed with these changes?")
```
atugushev: Could you add here an assert to make sure that sync() won't install/uninstall anything, since the side-effect declines confirmation?
```python
check_call.assert_not_called()
```

tests/test_sync.py
```diff
@@ -437,6 +437,64 @@ def test_sync_dry_run_would_uninstall(echo, from_line):
...
+    sync(set(), to_uninstall, ask=True)
+
+    expected_calls = [
+        mock.call("Would uninstall:"),
+        mock.call("  django==1.8"),
+        mock.call("  click==4.0"),
+    ]
+    echo.assert_has_calls(expected_calls, any_order=True)
+
+    confirm.assert_called_once_with("Would you like to proceed with these changes?")
```
atugushev: Same as above.
```python
check_call.assert_not_called()
```

tests/test_sync.py
```diff
@@ -437,6 +437,64 @@ def test_sync_dry_run_would_uninstall(echo, from_line):
     echo.assert_has_calls(expected_calls, any_order=True)
 
 
+@mock.patch("piptools.sync.click.confirm")
+@mock.patch("piptools.sync.click.echo")
+def test_sync_ask_declined_would_install(echo, confirm, from_line):
```
atugushev: Tests `test_sync_ask_declined_would_install` and `test_sync_ask_declined_would_uninstall` looks almost the same. Could we use parametrization instead?

tests/test_sync.py
```diff
@@ -437,6 +437,64 @@ def test_sync_dry_run_would_uninstall(echo, from_line):
...
+    pip should be called as normal when the user confirms
+    """
+    confirm.return_value = True
+
+    sync({from_line("django==1.8")}, {from_line("click==4.0")}, ask=True)
+    assert check_call.call_count == 2
+
+    confirm.assert_called_once_with("Would you like to proceed with these changes?")
+
+
```
atugushev: These tests look great! 👍 Could you also add at least one functional test to the `tests/test_cli_sync.py` module?

piptools/sync.py
georgek: I did originally write it this way, but I thought it was slightly nicer to modify the value of `dry_run` for two reasons:
* It makes it clear when reading the code that ask is essentially dry_run, except that with ask we can disable dry_run after a prompt,
* It enables `--ask` to override `--dry-run`. With the changes you suggest, when the user provides both `--ask` and `--dry-run`, it will ask for confirmation, but will still be in dry-run mode so won't do anything after the user says yes.

tests/test_sync.py
georgek: It can be done but it's a bit awkward. Since I tried to mirror the tests `test_sync_dry_run_would_uninstall` and `test_sync_dry_run_would_install` I have also parameterised these as well.

tests/test_sync.py
georgek: Done!

tests/test_sync.py
georgek: Done!

piptools/sync.py
atugushev: > It enables `--ask` to override `--dry-run`. With the changes you suggest, when the user provides both `--ask` and `--dry-run`, it will ask for confirmation, but will still be in dry-run mode so won't do anything after the user says yes.


Yeah, I've overlooked that case. Thanks for clarifying! But the tests passed. It would be nice to have a test for that case.

georgek: Good point. I'll add tests for that case.

tests/test_cli_sync.py
```diff
@@ -147,3 +147,31 @@ def test_pip_install_flags(check_call, cli_flags, expected_install_flags, runner
...
     ]
+
+
+@mock.patch("piptools.sync.check_call")
+def test_sync_ask_declined(check_call, runner):
+    """
+    Make sure nothing is installed if the confirmation is declined
+    """
+    with open("requirements.txt", "w") as req_in:
+        req_in.write("six==1.10.0")
```
atugushev: Could you use `small-fake-a` instead of `six`? No need to use the internet.

tests/test_cli_sync.py
```diff
@@ -147,3 +147,31 @@ def test_pip_install_flags(check_call, cli_flags, expected_install_flags, runner
...
+
+    check_call.assert_not_called()
+
+
+@mock.patch("piptools.sync.check_call")
+def test_sync_ask_accepted(check_call, runner):
+    """
+    Make sure pip is called when the confirmation is accepted (even if
+    --dry-run is given)
+
```
atugushev: Empty line.

tests/test_cli_sync.py
```diff
@@ -147,3 +147,31 @@ def test_pip_install_flags(check_call, cli_flags, expected_install_flags, runner
...
+    """
+    with open("requirements.txt", "w") as req_in:
+        req_in.write("six==1.10.0")
```
atugushev: Same as https://github.com/jazzband/pip-tools/pull/913#discussion_r329366949

tests/test_sync.py
```diff
@@ -403,39 +403,81 @@ def test_sync_verbose(check_call, from_line):
...
+    [
+        ({"django==1.8", "click==4.0"}, set(), "Would install:"),
+        (set(), {"django==1.8", "click==4.0"}, "Would uninstall:"),
+    ],
+)
 @mock.patch("piptools.sync.click.echo")
-def test_sync_dry_run_would_install(echo, from_line):
+def test_sync_dry_run(echo, from_line, to_install, to_uninstall, expected_message):
     """
     Sync with --dry-run option prints what's is going to be installed.
```
atugushev: Docstring should be also updated to "is going to be installed/uninstalled"

georgek:
Thanks @atugushev!

Please let me know when I should squash/rebase etc.

atugushev:
@georgek 

> Please let me know when I should squash/rebase etc.

Yes, please, squash and rebase it.

atugushev:
Thanks, @georgek! Awesome work! 👍

Issue #911, opened 2019-09-26
Require approving pull request before merging


atugushev:
#### What's the problem this feature will solve?

 Currently any jazzband member able to merge PR without code-review. Shouldn't there be at least one approval before PR being merging?

#### Describe the solution you'd like

<img width="729" alt="image" src="https://user-images.githubusercontent.com/7377671/65697891-5fdaf200-e084-11e9-8161-c8a579ff60de.png">

#### Alternative Solutions

Leave as is.

#### Additional context

https://help.github.com/en/articles/enabling-required-reviews-for-pull-requests

jezdez:
Thank you @atugushev, that seems like a good idea for sure. The project leads for the [pip-tools project](https://jazzband.co/projects/pip-tools) are @vphilippon and @davidovich. It would be best to ask them.

atugushev:
@jezdez thanks! 👍 Didn't know that project leads have admin rights to the repo.

davidovich:
I would gladly change the setting, but I do not have the Settings tab enabled for me. @jezdez do you know why ?

jezdez:
Apologies for the confusion, project leads don't have admin permissions, but they do have the last word on deciding how their project should be led, e.g. whether code reviews are required for pull requests. @davidovich Do you want me to enable the code review requirement?

vphilippon:
I agree with enabling the code review requirement, that was actually my intention when writing the project's contribution guideline.
@jezdez You have my 👍 to enable the code review requirement.

jezdez:
Enabled with the following settings:

![Screenshot_2019-09-26 jazzband pip-tools](https://user-images.githubusercontent.com/1610/65723318-78600200-e0ae-11e9-854e-8ded4772a297.png)

atugushev:
Thanks to everyone!

atugushev:
@vphilippon @davidovich 

As a following up to this issue shouldn't we remove then this checklist item?https://github.com/jazzband/pip-tools/blob/51463fcddb75f1d96569f1d0585970eba67949d7/.github/PULL_REQUEST_TEMPLATE.md#L8 

Because contributors (not members) have no rights to request an review other than writing a message like:


> <img width="664" alt="image" src="https://user-images.githubusercontent.com/7377671/65793642-f71c7400-e16e-11e9-8c9a-9643d06a0aec.png">
> 
> or
> 
> 
[...]

PR #914, opened 2019-09-30
Remove obsolete contributor checklist item in PR template

atugushev:
See https://github.com/jazzband/pip-tools/issues/911#issuecomment-536054216


atugushev:
Thanks @blueyed for reviewing this!

PR #875, opened 2019-08-15
[WIP] GitHub actions

atugushev:

PR #916, opened 2019-10-01
Add cron job for pipmaster env to GitHub actions

### LINKED ISSUE #853: pip-compile fails with current master branch of pip
If you install any version of pip after [cad71a711787ed66](https://github.com/pypa/pip/commit/cad71a711787ed66a4e2441a2acd8e46be26a110) and try to run `pip-compile` on any file, you will get an error in the constructor for `PackageFinder`. In the linked commit, it complains about unexpected keyword argument `trusted_hosts` and in later commits it complains about various other keyword arguments. The current master ([4cf08e85c09ad213](https://github.com/pypa/pip/commit/4cf08e85c09ad2136ff7212719bc8d126490a3cd))  complains about `find_links` (see full traceback below), failing on [this line](https://github.com/jazzband/pip-tools/blob/360832c76dff540ffb0182fac5d3e198560a78a9/piptools/repositories/pypi.py#L95).

The latest behavior (failing with `find_links`) starts with in [7d08bb37a55ff7cf](https://github.com/pypa/pip/commit/7d08bb37a55ff7cfc531668270ad8037594e87aa).

I will note that considering the fact that `pip-tools` is apparently making use of `pip`'s internal (private) API, it would be a good idea to add a CI job to build against the master branch of `pip`, so you can more easily catch these issues before release.

### Steps to reproduce:

1. Create temporary directory (optional):

```bash
mkdir -p /tmp/pip_compile_repro
cd /tmp/pip_compile_repro
```

2. Install the latest pip from github:

```bash
git clone https://github.com/pypa/pip.git
cd pip
# In case this behavior changes in a more recent version of pip
git checkout 4cf08e85c09ad2136ff7212719bc8d126490a3cd
cd ..
pip install ./pip
```

3. Create a `requirements.in` file and attempt to compile it

```bash
echo "six" > requirements.in
pip-compile requirements.in
```

### Actual result

Gives a traceback:

<details>

```
$ pip-compile requirements.in
  File ".../bin/pip-compile", line 10, in <module>
    sys.exit(cli())
  File ".../lib/python3.7/site-packages/click/core.py", line 764, in __call__
  File ".../lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File ".../lib/python3.7/site-packages/click/core.py", line 956, in invoke
  File ".../lib/python3.7/site-packages/click/core.py", line 555, in invoke
  File ".../lib/python3.7/site-packages/click/decorators.py", line 17, in new_func
  File ".../lib/python3.7/site-packages/piptools/scripts/compile.py", line 256, in cli
    repository = PyPIRepository(pip_options, session, build_isolation)
  File ".../lib/python3.7/site-packages/piptools/repositories/pypi.py", line 95, in __init__
    self.finder = PackageFinder(**finder_kwargs)
TypeError: __init__() got an unexpected keyword argument 'find_links'
```

</details>

TL;DR:

```
  File ".../lib/python3.7/site-packages/piptools/repositories/pypi.py",
        line 95, in __init__
    self.finder = PackageFinder(**finder_kwargs)
TypeError: __init__() got an unexpected keyword argument 'find_links'
```

### Expected Result

```bash
$ pip-compile requirements.in 
#
# To update, run:
#
#
six==1.12.0
```

CC @cjerdonek in case he wants to help out.

atugushev:
Test pip-tools against pip master every day on the latest Ubuntu and Windows using GitHub Actions.
See an example of actions on my fork: https://github.com/atugushev/pip-tools/actions

Resolves #853.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -1,0 +1,30 @@
+name: cron
+
+on:
+  schedule:
+  # Run every day at 00:00 UTC
+  - cron: 0 0 * * *
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python-version: [2.7, 3.5, 3.6, 3.7]
+        env:
+        - TOXENV: pipmaster
+        os: [ubuntu-latest, windows-latest]
+
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          pip install tox
+      - name: Test with tox ${{ matrix.env.TOXENV }}
+        run: |
+          tox
+        env: ${{ matrix.env }}
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -22,9 +22,7 @@ jobs:
         with:
           version: ${{ matrix.python-version }}
       - name: Install dependencies
-        run: |
-          pip install tox
+        run: pip install tox
       - name: Test with tox ${{ matrix.env.TOXENV }}
-        run: |
-          tox
+        run: tox
         env: ${{ matrix.env }}
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/setup-python@v1
         with:
           version: ${{ matrix.python-version }}
-      - name: Install dependencies
+      - name: Install tox
         run: pip install tox
       - name: Test with tox ${{ matrix.env.TOXENV }}
         run: tox


.github/workflows/cron.yml
```diff
@@ -0,0 +1,30 @@
...
+        os: [ubuntu-latest, windows-latest]
+
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
```
codingjoe: Don't need the pipe for single line command.
```suggestion
        run: pip install tox
```

.github/workflows/cron.yml
```diff
@@ -0,0 +1,30 @@
...
+        - TOXENV: pipmaster
+        os: [ubuntu-latest, windows-latest]
+
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          version: ${{ matrix.python-version }}
+      - name: Install dependencies
```
codingjoe: Are you really? You are installing tox and then tox installs the dependencies.
I would just drop the name. GH will display the command, is descriptive enough I guess.

.github/workflows/cron.yml
```diff
@@ -0,0 +1,30 @@
...
+    strategy:
+      matrix:
+        python-version: [2.7, 3.5, 3.6, 3.7]
+        env:
+        - TOXENV: pipmaster
+        os: [ubuntu-latest, windows-latest]
+
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python ${{ matrix.python-version }}
```
codingjoe: Does that actually work now? I know that GitHub had some issue displaying variables in step names.

.github/workflows/cron.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/setup-python@v1
+        with:
+          version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          pip install tox
+      - name: Test with tox ${{ matrix.env.TOXENV }}
+        run: |
```
codingjoe: Same here, single line does the trick.

.github/workflows/cron.yml
```diff
@@ -0,0 +1,30 @@
+name: cron
+
+on:
+  schedule:
+  # Run every day at 00:00 UTC
+  - cron: 0 0 * * *
```
codingjoe: Why not add `push` too?

.github/workflows/cron.yml
atugushev: Yeah, they've fixed that bug.

.github/workflows/cron.yml
atugushev: Maybe a later. The reason is in https://github.com/jazzband/pip-tools/pull/875#issuecomment-537260079.

atugushev: And I think push hook should be in a separate workflow file.

.github/workflows/cron.yml
atugushev: Blindly copied from the default template. Maybe "Install tox" is good enough?

.github/workflows/cron.yml
atugushev: Resolved. Thanks!

.github/workflows/cron.yml
atugushev: Resolved. Thanks!

atugushev:
Thanks, @pradyunsg!

atugushev:
@codingjoe thanks for reviewing this as always! I've addressed some request changes in cd5a9cb and answered to the others.

.github/workflows/cron.yml
codingjoe: Sure

.github/workflows/cron.yml
codingjoe: Ok, can be done in a separate PR, but I would not do it in a separate file unless you want to execute a different suite.

.github/workflows/cron.yml
atugushev: Addressed in 9962ce3.

.github/workflows/cron.yml
atugushev: If we find a way how to make "allow_failures" for "pipmaster" except for a cron job, then we can do it in a single workflow suite.

atugushev:
Thanks, @codingjoe!

PR #922, opened 2019-10-07
Refactor several tests to reduce the time spent (issue #894)

hramezani:
Refactor several tests to reduce the time spent (issue #894)

PR #930, opened 2019-10-08
Rename test packages dir to same names as the package names.

hramezani:
https://github.com/jazzband/pip-tools/pull/922#discussion_r332154180


hramezani:
> Maybe it would be neat to move all directories of packages to `tests/test_data/packages`? What do you think?

Good idea :+1: . I did it.

atugushev:
Thanks @hramezani! Much appreciated!

PR #931, opened 2019-10-09
Fixed pip-compile --upgrade-package removes "via" annotation

### LINKED ISSUE #929: pip-compile --upgrade-package removes "via" annotation from lock file
`pip-compile --upgrade` removes source annotations (e.g. `# via jinja2`).

#### Environment Versions

1. OS Type: Linux
1. Python version: `3.6.6`
1. pip version:  `19.2.3`
1. pip-tools version: `4.1.0`

#### Steps to replicate

Prepare virtualenv:

```shell
$ pyenv install 3.6.6 py36
$ pyenv activate py36
```

Prepare lock file:

```shell
$ echo "jinja2" > requirements.in
$ pip-compile --no-header --no-index
jinja2==2.10.3
markupsafe==1.1.1         # via jinja2
$ # edit requirements.txt and change markupsafe version to 1.0.0
$ cat requirements.txt
jinja2==2.10.3
markupsafe==1.0.0         # via jinja2
```

Let's upgrade markupsafe:

```shell
$ pip-compile -P markupsafe --no-header --no-index
jinja2==2.10.3
markupsafe==1.1.1
```

Problem is that `# via jinja2` marker was removed. `--upgrade-package` shouldn't remove the annotation, as intention was merely to upgrade it. It is still required via jinja2 and not directly (it's not in `requirements.in`)

On the other hand, I understand that I now directly influenced the package version, so annotation gets removed. But still, my initial assumption was that marker would remain.

So I am not sure about this one. Is current behavior intentional?

hramezani:
Fixes #929.

**Changelog-friendly one-liner**: Fixed `pip-compile --upgrade-package` removes "via" annotation.

##### Contributor checklist


atugushev:
Hello @hramezani,

Thanks for fixing this bug and the test! Could you use [current PR template](https://raw.githubusercontent.com/jazzband/pip-tools/master/.github/PULL_REQUEST_TEMPLATE.md) if the changes are user-facing, for it might need a changelog entry? Thanks!

tests/test_cli_compile.py
```diff
@@ -839,3 +839,27 @@ def test_empty_input_file_no_header(runner, empty_input_pkg, prior_output_pkg):
 
     with open("requirements.txt", "r") as req_txt:
         assert req_txt.read().strip() == ""
+
+
+def test_pip_compile_upgrade_package(pip_conf, runner):
+    """
+    Tests pip-compile --upgrade-package shouldn't remove "via" annotation.
+    Issue #929
```
```suggestion
    See: GH-929
```
Better to use a project-specific issue marker to avoid issue trackers collision in a possible future.

tests/test_cli_compile.py
```diff
@@ -839,3 +839,27 @@ def test_empty_input_file_no_header(runner, empty_input_pkg, prior_output_pkg):
...
+        assert (
+            "small-fake-a==0.1         # via small-fake-with-deps-a"
+            in req_txt.read().strip()
+        )
+
+    runner.invoke(cli, ["-P", "small-fake-a", "--no-header", "--no-index"])
+    with open("requirements.txt", "r") as req_txt:
+        assert (
+            "small-fake-a==0.1         # via small-fake-with-deps-a"
+            in req_txt.read().strip()
```
```suggestion
            in req_txt.read().splitlines()
```
Same as above.

tests/test_cli_compile.py
```diff
@@ -839,3 +839,27 @@ def test_empty_input_file_no_header(runner, empty_input_pkg, prior_output_pkg):
 
     with open("requirements.txt", "r") as req_txt:
         assert req_txt.read().strip() == ""
+
+
+def test_pip_compile_upgrade_package(pip_conf, runner):
```
```suggestion
def test_upgrade_package_doesnt_remove_annotation(pip_conf, runner):
```

This test name could be more descriptive for my taste.

hramezani:
> Hello @hramezani,
> 
> Thanks for fixing this bug and the test! Could you use [current PR template](https://raw.githubusercontent.com/jazzband/pip-tools/master/.github/PULL_REQUEST_TEMPLATE.md) if the changes are user-facing, for it might need a changelog entry? Thanks!

@atugushev Thanks for the comment. Description edited.

tests/test_cli_compile.py
```diff
@@ -839,3 +839,27 @@ def test_empty_input_file_no_header(runner, empty_input_pkg, prior_output_pkg):
...
+    """
+    Tests pip-compile --upgrade-package shouldn't remove "via" annotation.
+    Issue #929
+    """
+    with open("requirements.in", "w") as req_in:
+        req_in.write("small-fake-with-deps-a\n")
+
+    runner.invoke(cli, ["--no-header", "--no-index"])
```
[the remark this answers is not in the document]
hramezani: I don't think so. I just copy them from #931.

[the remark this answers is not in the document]
hramezani: I don't think so. I just copy them from #931.

tests/test_cli_compile.py
```diff
@@ -839,3 +839,27 @@ def test_empty_input_file_no_header(runner, empty_input_pkg, prior_output_pkg):
...
+    with open("requirements.txt", "r") as req_txt:
+        assert (
+            "small-fake-a==0.1         # via small-fake-with-deps-a"
+            in req_txt.read().splitlines()
+        )
+
+    runner.invoke(cli, ["-P", "small-fake-a"])
+    with open("requirements.txt", "r") as req_txt:
+        assert (
+            "small-fake-a==0.1         # via small-fake-with-deps-a"
```
atugushev: Can we test here that `small-fake-a` has been also upgraded to a different version?

I guest that might take change `small-fake-with-deps-a` dependencies, so we can fix it by:
1. pin small-fake-a==0.2 in https://github.com/jazzband/pip-tools/blob/8e782ae5051ef8e195bb77e1955fed3f2c607fd5/tests/test_data/packages/small_fake_with_deps_a/setup.py#L4
2. fix `test_annotate_option` to honor that change
3. update the last assertion in `test_upgrade_package_doesnt_remove_annotation`

atugushev: Probably better to do it in a separate commit, to preserve the intention to update `small-fake-with-deps-a` in `git blame`.

hramezani: Good idea. when I do this change `small-fake-a==0.2`, the `test_annotate_option` won't fail.
I don't know why? is it something related to my machine?

atugushev: Interesting... Push the commit -- we'll figure out.

atugushev: Oh, I see, you should rebuild `tests/test_data/minimal_wheels/small_fake_with_deps_a-0.1-py2.py3-none-any.whl`.

hramezani: I did the rebuild. and upgrade `small-fake-with-deps-a` dependencies to `small-fake-a==0.2`.
Also, changed `test_annotate_option`. I did them in a separate commit.

Seems `runner.invoke(cli, ["-P", "small-fake-a"])` didn't work properly and the `small-fake-a` didn't upgraded.

Am I wrong?

tests/test_cli_compile.py
```diff
@@ -839,3 +839,27 @@ def test_empty_input_file_no_header(runner, empty_input_pkg, prior_output_pkg):
...
+    See: GH-929
+    """
+    with open("requirements.in", "w") as req_in:
+        req_in.write("small-fake-with-deps-a\n")
+
+    runner.invoke(cli)
+
+    with open("requirements.txt", "r") as req_txt:
+        assert (
+            "small-fake-a==0.2         # via small-fake-with-deps-a"
```
```suggestion
            "small-fake-a==0.1         # via small-fake-with-deps-a"
```
Should't it be downgraded before upgrade?

tests/test_cli_compile.py
```diff
@@ -839,3 +839,35 @@ def test_empty_input_file_no_header(runner, empty_input_pkg, prior_output_pkg):
...
+            in req_txt.read().splitlines()
+        )
```
atugushev: I think this `with` block should be deleted now.

tests/test_cli_compile.py
```diff
@@ -839,3 +839,35 @@ def test_empty_input_file_no_header(runner, empty_input_pkg, prior_output_pkg):
...
+
+    # Downgrade small-fake-a to 0.1
+    with open("requirements.txt", "w") as req_txt:
+        req_txt.write(
+            "small-fake-with-deps-a\n"
```
```suggestion
            "small-fake-with-deps-a==0.1\n"
```
In a real world `small-fake-with-deps-a` would be pinned.

tests/test_cli_compile.py
hramezani: Done!

tests/test_cli_compile.py
hramezani: Done!

tests/test_cli_compile.py
atugushev: I'll resolve this, since it looks like you've figured out how to fix this.

atugushev:
Thanks for your excellent work and your patience, @hramezani!

PR #940, opened 2019-10-13
Skip commands_pre for checkqa and readme env in tox.ini

atugushev:
Prevents [running](https://travis-ci.org/jazzband/pip-tools/jobs/597097879#L221) `commands_pre` on `tox -e checkqa` and `tox -e readme`.

P.S.
An alternative solution is to move everything from`commands_pre` to `commands`, but intuitively I've chosen the current one.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -45,10 +45,12 @@ passenv = CI
 basepython = python3
 skip_install = True
 deps = pre-commit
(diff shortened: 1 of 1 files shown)


atugushev:
@webknjaz thanks for reviewing this!

PR #941, opened 2019-10-13
Remove useless excludes from pre-commit config

atugushev:
"pre-commit won't check files that aren't checked in"
(c) @asottile

Thanks, Anthony!


atugushev:
@codingjoe thanks for reviewing this! 👍

PR #944, opened 2019-10-14
Fixed inline comment in pull request template.

hramezani:
**Fixed inline comment in pull request template.**  

I think `<!---  --->` doesn't work for inline comment. I replaced it with `<!-- -->`

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,6 @@
 <!--- Describe the changes here. --->
 
-**Changelog-friendly one-liner**: <!--- One-liner description here --->
+**Changelog-friendly one-liner**: <!-- One-liner description here -->
 
 ##### Contributor checklist
 


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,6 +1,6 @@
 <!--- Describe the changes here. --->
 
-**Changelog-friendly one-liner**: <!--- One-liner description here --->
+**Changelog-friendly one-liner** <!-- One-liner description here -->
```
atugushev: Thanks for fixing triple dash!

```suggestion
**Changelog-friendly one-liner**: <!-- One-liner description here -->
```

I think the colon shouldn't be removed because it splits the label and text. The template is: 

**Changelog-friendly one-liner**: My awesome feature

See an example #913.

hramezani: Thanks for the comment. Colon reverted.

atugushev:
I've cancelled the CI not to waste time.

atugushev:
Thanks, @hramezani!

PR #949, opened 2019-10-17
Show progress bar when downloading package in verbose mode

### LINKED ISSUE #832: Slow compile

##### Environment Versions

1. OS Type: Ubuntu 19.04
1. Python version: Python 3.7.3
1. pip version:  pip 19.0.3
1. pip-tools version: pip-compile, version 3.7.0

##### Steps to replicate
I have a new environment with the following `requirements.in`:
```
Cython
torch
torchvision
```
when I run `pip-compile` it just hangs.
Here is the output of `pip-compile -v`:
```
Using indexes:
  https://pypi.org/simple

                          ROUND 1                           
Current constraints:
  Cython
  torch
  torchvision

Finding the best candidates:
  found candidate cython==0.29.10 (constraint was <any>)
  found candidate torch==1.1.0 (constraint was <any>)
  found candidate torchvision==0.3.0 (constraint was <any>)

  torchvision==0.3.0        requires numpy, pillow>=4.1.1, six, torch>=1.1.0
  torch==1.1.0 not in cache, need to check index
```

##### Expected result
Compiles normally

##### Actual result
Hangs on checking index?

atugushev:
Resolves #832.

Demonstration:

[asciicast](https://asciinema.org/a/mJPk60dIndxq3Ge8oYknrZfWO)

**Changelog-friendly one-liner**: Show progress bar when downloading package in a verbose mode of `pip-compile`.

##### Contributor checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -3,6 +3,8 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import collections
 import hashlib
+import itertools
+import logging
 import os
 from contextlib import contextmanager
 from shutil import rmtree
@@ -15,11 +17,12 @@ from pip._internal.models.wheel import Wheel
 from pip._internal.req import RequirementSet
 from pip._internal.req.req_tracker import get_requirement_tracker
 from pip._internal.utils.hashes import FAVORITE_HASH
+from pip._internal.utils.logging import indent_log, setup_logging
 from pip._internal.utils.misc import normalize_path
 from pip._internal.utils.temp_dir import TempDirectory, global_tempdir_manager
 from pip._internal.utils.urls import path_to_url, url_to_path
 
-from .._compat import PIP_VERSION, TemporaryDirectory, contextlib
+from .._compat import BAR_TYPES, PIP_VERSION, TemporaryDirectory, contextlib
 from ..click import progressbar
 from ..exceptions import NoCandidateFound
 from ..logging import log
@@ -79,6 +82,8 @@ class PyPIRepository(BaseRepository):
         self._cache_dir = normalize_path(cache_dir)
         self._download_dir = fs_str(os.path.join(self._cache_dir, "pkgs"))
         self._wheel_download_dir = fs_str(os.path.join(self._cache_dir, "wheels"))
+
+        self._setup_logging()
 
     def freshen_build_caches(self):
         """
@@ -142,7 +147,7 @@ class PyPIRepository(BaseRepository):
     def resolve_reqs(self, download_dir, ireq, wheel_cache):
         with get_requirement_tracker() as req_tracker, TempDirectory(
             kind="resolver"
-        ) as temp_dir:
+        ) as temp_dir, indent_log():
             preparer = self.command.make_requirement_preparer(
                 temp_build_dir=temp_dir,
                 options=self.options,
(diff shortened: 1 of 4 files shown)


AndydeCleyre:
It looks good! I just want to note, in case this is surprising or not desirable:

If run with stderr in foreground and stdout directed to a file, the progress bar is not shown.

atugushev:
Thanks for review @AndydeCleyre! It seems `pip` print logs to `stdout`, but pip-tools to `stderr`. Then I'll pull this out from the upcoming release and figure out later what to do.

atugushev:
> It seems `pip` prints logs to `stdout`, but pip-tools to `stderr`. Then I'll pull this out from the upcoming release and figure out later what to do.

I think we could fix that by patching:
1. pip's [console logger stream](https://github.com/pypa/pip/blob/814c54f5d7e17682f9c476f5978c4d5666abb52c/src/pip/_internal/utils/logging.py#L358) to `sys.stderr`
2. `pip._internal.cli.progress_bars.DownloadProgressSpinner.file = sys.stderr`


I also discovered that hashing progress bar prints to stdout, could be fixed with `progressbar(..., file=sys.stderr)` https://github.com/jazzband/pip-tools/blob/7a52ce86884804121e8771c2f6515bfaa54da04b/piptools/repositories/pypi.py#L286

atugushev:
@AndydeCleyre 

Fixed in 5cecccf6085ef11c35e6217a829652b4496a768e. Could you take a look please?

AndydeCleyre:
LGTM :+1: 

And thanks for adding the logging comments, especially the difference between pip log level and pip-tools'.

piptools/repositories/pypi.py
```diff
@@ -331,6 +342,35 @@ def _wheel_support_index_min(self, tags=None):
...
+        console_handlers = [
+            handler for handler in logger.handlers if handler.name == "console"
+        ]
+        if console_handlers:  # pragma: no branch
+            console_handlers[0].stream = log.stream
+        else:  # pragma: no cover
+            # There is always a console handler. This warning would be a signal that
+            # this block should be removed/revisited, because of pip possibly
+            # refactored-out logging config.
+            log.warning("Couldn't find a 'console' logging handler")
```
atugushev, line 368: @AndydeCleyre 

A tiny update, I refactored this block for two reasons:
1. fix decreased code coverage
2. to be sure that we won't miss the moment when the console logger is gone, so we can remove or refactor this block.

What do you think?

AndydeCleyre: Sure, looks ok with clear comments. Adding a warning to track this is a good idea.

This may be better for coverage, but for goal 2, adding an `else` to the former `for` loop would've been a nice place for the warning, maintaining that `break` upon finding the right logger. 

Alternatively, and this may be a less readable alternative, you could make that a generator instead of a comprehension, `next` it to get the first (only) item, and put the warning under an `except StopIteration`.

atugushev: Yes, there are so many ways to do the same (hello python zen №13 😄), but I've chosen the current one deliberately as a more readable IMO. In fact, first version was with `for/else`, but this construction always confuses me.

atugushev:
Byt the way, this PR would also help with investigating such issues like #1116 by `pip-compile -v --pip-args="--verbose"`.

piptools/repositories/pypi.py
atugushev: But I don't want to hold on this, if you feel strongly that the other ways are more suitable, I'd be happy to refactor this block and go-ahead 👍

AndydeCleyre: No problem, my approval stands for your current implementation. I only felt obligated to note out loud that this loops through handlers even after finding the right one. But that's not a real problem.

Looking forward to the merge! 👍

atugushev:
Thank you so much for the feedback @AndydeCleyre!

piptools/repositories/pypi.py
atugushev: I've reconsidered my opinion about `for-else` statement, hence #1117. Thank you @AndydeCleyre.

PR #953, opened 2019-10-20
Fix compatibility with pip master

atugushev:
#### Problem

Cron job is failing, see: https://github.com/jazzband/pip-tools/commit/023d7cabe36277bf9e5a5f37ebca299d87f4e36d/checks

#### What's fixed

1. **Update import path for PackageFinder/FormatControl https://github.com/pypa/pip/pull/7144**
`PackageFinder` moved to `pip._internal.index.package_finder`
`FormatControl` moved to `pip._internal.models.format_control`
2. **Add require_hashes to Resolver._resolve_one() arguments https://github.com/pypa/pip/pull/7068**
Pass require_hashes directly to Resolver

#### Additional notes

`pragma: no cover` prevents suddenly decreasing coverage if pip master fails again. Any other suggestions on how to achieve this?

**Changelog-friendly one-liner**: Add compatibility with `pip>=19.4`

##### Contributor checklist

- ~~Provided the tests for the changes.~~


atugushev:
Thanks for reviewing this, @codingjoe! 🎉

piptools/repositories/pypi.py
```diff
@@ -226,9 +226,16 @@ def resolve_reqs(self, download_dir, ireq, wheel_cache):
...
                 resolver = PipResolver(**resolver_kwargs)
-                resolver.require_hashes = False
-                results = resolver._resolve_one(reqset, ireq)
+                require_hashes = False
+                if PIP_VERSION < (19, 4):
+                    resolver.require_hashes = require_hashes
+                    results = resolver._resolve_one(reqset, ireq)
+                else:  # pragma: no cover
+                    # TODO remove pragma after pip==19.4 being released
+                    results = resolver._resolve_one(reqset, ireq, require_hashes)
```
blueyed: Would be better to not have "pragma: no cover" there in the first place, especially since this should be covered already via the pipmaster job?!
(but even if not: it is better to clearly indicate that some code is not covered)

atugushev: Thanks for getting attention to this!

What the options do we have to avoid arbitrarily decreasing/increasing the test coverage like in this https://github.com/jazzband/pip-tools/pull/909#issuecomment-535671204 or #914 PRs? It would be annoying for users (maintainers too) to see that PR with 100% coverage has failed checks because of the other code, which is genuinely covered but temporarily is not, because of the pip master job fails at import time.

On the other hand, we do have a relative coverage check (codecov/patch), so we can rely entirely on it and don't pay attention to the absolute (codecov/project) test coverage.

atugushev: @blueyed would you like to open an issue for that?

blueyed: @atugushev 
No it's fine for now I guess - I'd rather ignore any "X" from CI and override it after checking (about coverage,e.g.via codecov).

atugushev:
FTR, it's green now https://github.com/jazzband/pip-tools/commit/4dee581f2603e21271e9c7b38830507ed749357b/checks?check_suite_id=283625731

piptools/repositories/pypi.py
atugushev: Okay, I'll remove the TODO in the following pip-master compatibility PR.

atugushev: FTR, addressed in #978 (fa2c479dd19f73fd544abbaaef624357d41b6f15)

PR #964, opened 2019-10-22
Fix --trusted-host is not copied from *.in to *.txt

### LINKED ISSUE #963: Option --trusted-host is not copied from requirements.in to requirements.txt
#### Environment Versions

1. OS Type: Mac OS
1. Python version: `Python 3.7.3`
1. pip version:  `pip 19.3.1 from /.venv/lib/python3.7/site-packages/pip (python 3.7)`
1. pip-tools version: `pip-compile, version 4.2.0`

#### Steps to replicate

1. `echo "--trusted-host exmaple.com" | pip-compile - -qo-`

#### Expected result

```bash
#
# To update, run:
#
#    pip-compile --output-file=- -
#
--trusted-host exmaple.com
```

#### Actual result

```bash
#
# To update, run:
#
#    pip-compile --output-file=- -
#
```

#### Additional context

Caught by @altendky, see https://github.com/jazzband/pip-tools/pull/824#issuecomment-544758205.
Related issues/PRs: #338, #371

atugushev:
Fixes #963.

**Changelog-friendly one-liner**: Fix `pip-compile` doesn't copy `--trusted-host` from `requirements.in` to `requirements.txt`.

##### Contributor checklist


altendky:
Oh yeah...  thanks!

piptools/utils.py
```diff
@@ -380,3 +380,13 @@ def create_install_command():
     from pip._internal.commands import create_command
 
     return create_command("install")
+
+
+def get_trusted_hosts(finder):
+    """
+    Returns a list of trusted hosts from given finder.
```
webknjaz: Looking at the Pip src, it returns an iterator:  https://github.com/pypa/pip/blob/0f095af4ba7ac1af2c27024fa539af4c3a45d002/src/pip/_internal/index/package_finder.py#L692-L695

piptools/utils.py
```diff
@@ -380,3 +380,13 @@ def create_install_command():
...
+    """
+    if PIP_VERSION < (19, 2):
+        return [host for _, host, _ in finder.secure_origins]
```
webknjaz: Based on the comment above, we need to use a generator expression:
```suggestion
        return (host for _, host, _ in finder.secure_origins)
```

piptools/utils.py
```diff
@@ -380,3 +380,13 @@ def create_install_command():
...
+
+    return finder.trusted_hosts
```
webknjaz: Alternatively
```suggestion
    return list(finder.trusted_hosts)
```
so that the return type is consistent.

piptools/utils.py
atugushev: Agreed! Nice catch. I miss mypy.

piptools/utils.py
atugushev: I'd prefer a generator.

atugushev: Addressed in 7a54db3

piptools/utils.py
atugushev: Addressed in 7a54db3

webknjaz: Almost, I imagine that you'd also want to fix the docstring to be precise.
```suggestion
    Returns an iterable of trusted hosts from a given finder.
```

atugushev: Thanks!

codingjoe:
I believe the have been more than enough reviews. So I am going to save myselthe trouble, ok?

atugushev:
@webknjaz thanks for taking the time reviewing this! All request changes seem resolved, are we okay with this or are there any other issues?

atugushev:
@webknjaz @nasirhjafri thanks for reviewing this! 🎉

Issue #926, opened 2019-10-07
Deprecation policy of supported pip versions


atugushev:
#### What's the problem this feature will solve?

Since [EOL for Python 2](https://pythonclock.org) is coming it's time to discuss about deprecation policy of supported pip versions. Currently, pip-tools supports the following pip versions:

- 8.1.1 (2016-03-17)
- 9.0.1 (2016-11-06)
- 9.0.3 (2018-03-21)
- 10.0.1 (2018-04-19)
- 18.0 (2018-07-22)
- 19.0.3 (2019-02-20)
- 19.1 (2019-04-23)
- 19.2 (2019-07-23)
- 19.3 (2019-10-14)

Fortunately, pip is under active development and refactoring. The problem is each supported pip version increases [CI dimensions](https://github.com/jazzband/pip-tools/blob/6fa224387a804913225103d1754133fd9cfd7d7c/.travis.yml#L10-L21) and [version branches in code](https://github.com/jazzband/pip-tools/blob/6fa224387a804913225103d1754133fd9cfd7d7c/piptools/repositories/pypi.py#L147-L162), which makes **hard to maintain pip-tools**.

#### Describe the solution you'd like

Possible solutions are to prepare a deprecation policy based on:

1. calendar, for example:
    * v1 deprecated on date1, will be removed on date 1.1
    * v2 deprecated on date2, will be removed on date 2.1
    * etc
1. support of N latest versions (which is, I think, the current policy except N is undefined)
1. support one minor version at a time (e.g., only pip 19.3.x), suggested by @cjerdonek in https://github.com/jazzband/pip-tools/issues/853#issuecomment-517948771
1. any other suggestions?

#### Alternative Solutions

Vendor pip.

#### Additional context

This issue follows up on the discussion started in #853.

atugushev:
@vphilippon @davidovich I've pinned this and pinged you guys to bring more attention to the important issue.

AndydeCleyre:
For extra context, I'm linking here [pip's changelog](https://github.com/pypa/pip/blob/master/NEWS.rst), and noting from it that:

- pip `19.0` is the first release that warns about Python `2.7` EOL, deprecates Python `3.4`, and supports PEP 517. IMO anything older should be deprecated ASAP.
- pip `19.2` drops support for Python `3.4`, which is past EOL now.

Notes on other projects:

Project | Release | pip | EOL
--- | --- | --- | ---
CentOS | `6` | no Python3! | 2020.11.30
[...]

PR #968, opened 2019-10-25
Cleanup ephemeral wheel cache properly

atugushev:
Attribute `self.wheel_cache` has been removed in f6de247b3f7a9e776325057a1e6cf6e48d34e7e7, thus `self.wheel_cache.cleanup()` will allways throw:

```pythin
AttributeError("'PyPIRepository' object has no attribute 'wheel_cache'")
```

and `wheel_cache.cleanup()` will never be called as supposed to be.

WheelCache.cleanup() has been introduced in `pip==10.0.0`. Try-except block has been refactored, for it's better to check the pip version explicitly (LBYL) rather than ask for forgiveness (EAFP).

**Changelog-friendly one-liner**: Fix dependency resolver to clean up the ephemeral wheel cache.

##### Contributor checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -274,10 +274,10 @@ class PyPIRepository(BaseRepository):
                         os.environ["PIP_REQ_TRACKER"] = prev_tracker
                     else:
                         del os.environ["PIP_REQ_TRACKER"]
-                try:
-                    self.wheel_cache.cleanup()
-                except AttributeError:
-                    pass
+
+                # WheelCache.cleanup() introduced in pip==10.0.0
+                if PIP_VERSION >= (10,):
+                    wheel_cache.cleanup()
         return self._dependencies_cache[ireq]
 
     def get_hashes(self, ireq):
--- a/tests/test_repository_pypi.py
+++ b/tests/test_repository_pypi.py
@@ -174,3 +174,19 @@ def test_pypirepo_calls_reqset_with_str_paths(pypi_repository, from_line):
         assert isinstance(called_with_finder, PackageFinder)
         assert called_with_ireq == ireq
         assert not pf_call_kwargs
+
+
+@pytest.mark.skipif(
+    PIP_VERSION < (10,), reason="WheelCache.cleanup() introduced in pip==10.0.0"
+)
+@mock.patch("piptools.repositories.pypi.PyPIRepository.resolve_reqs")  # to run offline
+@mock.patch("piptools.repositories.pypi.WheelCache")
+def test_wheel_cache_cleanup_called(
+    WheelCache, resolve_reqs, pypi_repository, from_line
+):
+    """
+    Test WheelCache.cleanup() called once after dependency resolution.
+    """
+    ireq = from_line("six==1.10.0")
+    pypi_repository.get_dependencies(ireq)
+    WheelCache.return_value.cleanup.assert_called_once_with()


piptools/repositories/pypi.py
```diff
@@ -284,10 +284,10 @@ def get_dependencies(self, ireq):
                         os.environ["PIP_REQ_TRACKER"] = prev_tracker
                     else:
                         del os.environ["PIP_REQ_TRACKER"]
-                try:
-                    self.wheel_cache.cleanup()
-                except AttributeError:
-                    pass
+
+                # WheelCache.cleanup() introduced in pip==10.0.0
```
webknjaz: I don't see any need to change this. In fact, following [EAFP](https://docs.python.org/3/glossary.html#term-eafp) style is more Pythonic.
The only thing I'd change is using a CM instead of a try-block.
```suggestion
                # WheelCache.cleanup() introduced in pip==10.0.0
                with contextlib.suppress(AttributeError):
                    wheel_cache.cleanup()
```

atugushev: Thanks for the suggestion! This suppressor looks pythonic and way better than try-except. I like it.

A few notes about the explicit version checking:
1. the bug with unexpected AssertionError on `self` could be caught earlier
2. when we decide to remove the support of a specific pip version (e.g., pip=={version}), we would find the code by `git grep PIP_VERSION | grep {version}`.
3. if we'd be on the multi-threaded environment, the EAFP style would be better for this particular case. 
4. also, there is an excellent explanation originally posted by @nicoddemus:  https://twitter.com/nicoddemus/status/1146196274309414912. (yeah, he's talking about import statements, but it can be extrapolated)

I'm not against the suggested solution, but I'd prefer the uniform approach, which follows the current codebase.

piptools/repositories/pypi.py
```diff
@@ -284,10 +284,10 @@ def get_dependencies(self, ireq):
...
+                if PIP_VERSION >= (10,):
```
codingjoe: Does pip-tools actually support versions lower than pip v10?

atugushev: pip-tools supports 8.1.1, 9.0.1, and 9.0.3.

piptools/repositories/pypi.py
webknjaz: How is it related to `AssertionError`? Is it a typo for `AttributeError`?

webknjaz: As for (2), grep would find it if you just add a proper comment which is necessary anyway to explain the intent behind the code.

webknjaz: (3) not sure how it's related to multi-threading at all.

webknjaz: (4) it may get quite complex with nested if-elses over time.

piptools/repositories/pypi.py
codingjoe: tenor

codingjoe: Wouldn't it make sense to simply up the required pip version? Pip will complain anyways, if it's not up to date, so most people will have it on a newer version and even if we don't if pip-tools sets the correct requirement, it will simply update pip.

codingjoe: Right, I have the feeling I am missing something fundamental 🤔

atugushev: He-he 😊 You should join the discussion #926!

piptools/repositories/pypi.py
atugushev: > How is it related to `AssertionError`? Is it a typo for `AttributeError`?

Sorry, it's a typo, I meant the AttributeError.

> As for (2), grep would find it if you just add a proper comment which is necessary anyway to explain the intent behind the code.

Fair enough!

> (3) not sure how it's related to multi-threading at all.

The explanation is here https://docs.python.org/3/glossary.html#term-lbyl.

> (4) it may get quite complex with nested if-elses over time.

That's how we pay the cost of supporting pip>=8.1.1.


---

Also, `if-statement` makes type checkers happier.

piptools/repositories/pypi.py
codingjoe: haha, yes I should :P

piptools/repositories/pypi.py
atugushev: Hey @webknjaz! Gentle ping. If you feel strongly that we should use "EAFP style" here, so I'd rather accept your commit suggestion. Don't want to waste energy on the bikeshedding. Otherwise, let's merge it as is 😊

webknjaz: Feel free to merge as-is.

atugushev: I'll mark this as resolved then.

atugushev:
ℹ️ rebased onto master and resolved conflicts.

atugushev:
@webknjaz, @codingjoe and @suutari, thanks for reviewing this! 🙏

PR #971, opened 2019-10-26
Add rudimentary mypy annotations

atugushev:
#### What's done
- updated setup.cfg
- added mypy hook to pre-commit
- added typings for the cache module

Typings added using [pytest-annotate](https://github.com/kensho-technologies/pytest-annotate).

Refs: #972.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,3 +20,10 @@ repos:
     - id: bandit
       language_version: python3
       exclude: ^tests/
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.740
+    hooks:
+    - id: mypy
+      language_version: python3
+      # To prevent "Duplicate module named 'setup'" error
+      exclude: ^tests/test_data/packages/
--- a/piptools/cache.py
+++ b/piptools/cache.py
@@ -7,16 +7,23 @@ import sys
 
 from pip._vendor.packaging.requirements import Requirement
 
+from ._compat.typing import MYPY
 from .exceptions import PipToolsError
 from .locations import CACHE_DIR
 from .utils import as_tuple, key_from_req, lookup_table
 
+if MYPY:
+    from typing import Dict, List, Optional, Set, Tuple
+    from ._compat import InstallRequirement
+
 
 class CorruptCacheError(PipToolsError):
     def __init__(self, path):
+        # type: (str) -> None
         self.path = path
 
     def __str__(self):
+        # type: () -> str
         lines = [
             "The dependency cache seems to have been corrupted.",
             "Inspect, or delete, the following file:",
@@ -26,6 +33,7 @@ class CorruptCacheError(PipToolsError):
 
 
 def read_cache_file(cache_file_path):
+    # type: (str) -> Dict[str, dict]
     with open(cache_file_path, "r") as cache_file:
         try:
             doc = json.load(cache_file)
@@ -49,7 +57,10 @@ class DependencyCache(object):
     Where X.Y indicates the Python version.
     """
 
+    _cache = None  # type: Dict[str, dict]
+
     def __init__(self, cache_dir=None):
+        # type: (Optional[str]) -> None
         if cache_dir is None:
             cache_dir = CACHE_DIR
         if not os.path.isdir(cache_dir):
@@ -58,10 +69,10 @@ class DependencyCache(object):
         cache_filename = "depcache-py{}.json".format(py_version)
 
         self._cache_file = os.path.join(cache_dir, cache_filename)
-        self._cache = None
 
     @property
     def cache(self):
+        # type: () -> Dict[str, dict]
         """
         The dictionary that is the actual in-memory cache.  This property
         lazily loads the cache from disk.
@@ -71,6 +82,7 @@ class DependencyCache(object):
         return self._cache
 
     def as_cache_key(self, ireq):
+        # type: (InstallRequirement) -> Tuple[str, str]
         """
         Given a requirement, return its cache key. This behavior is a little weird
         in order to allow backwards compatibility with cache files. For a requirement
@@ -91,6 +103,7 @@ class DependencyCache(object):
         return name, "{}{}".format(version, extras_string)
 
     def read_cache(self):
+        # type: () -> None
         """Reads the cached contents into memory."""
         if os.path.exists(self._cache_file):
             self._cache = read_cache_file(self._cache_file)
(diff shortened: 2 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -20,3 +20,10 @@ repos:
     - id: bandit
       language_version: python3
       exclude: ^tests/
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.740
+    hooks:
+    - id: mypy
+      language_version: python3
+      # To prevent "Duplicate module named 'setup'" error
```
webknjaz: I was arguing about this with Anthony a while back. Still now sure about the best way to prevent this: https://github.com/pre-commit/mirrors-mypy/issues/5.

piptools/cache.py
```diff
@@ -49,7 +57,10 @@ class DependencyCache(object):
     Where X.Y indicates the Python version.
     """
 
+    _cache = None  # type: Dict[str, dict]
```
webknjaz: Why did you move it from instance to class?

atugushev: It fixes the following complainings:

```
piptools/cache.py:82: error: Incompatible return value type (got "None", expected "Dict[str, Dict[Any, Any]]")
piptools/cache.py:109: error: Incompatible types in assignment (expression has type "Dict[str, Dict[Any, Any]]", variable has type "None")
piptools/cache.py:111: error: Incompatible types in assignment (expression has type "Dict[<nothing>, <nothing>]", variable has type "None")
piptools/cache.py:122: error: Incompatible types in assignment (expression has type "Dict[<nothing>, <nothing>]", variable has type "None")
```

.pre-commit-config.yaml
atugushev: Would you suggest this solution? https://github.com/pre-commit/mirrors-mypy/issues/5#issuecomment-515841288

webknjaz: Maybe. But then, I think, I've had problems with that too. It's up to you atm.

piptools/cache.py
webknjaz: But you completely changed the semantics with this move, which may cause unwanted side-effects. Why didn't you just add an annotation in the initializer?

atugushev: I've tried at first:

```python
self._cache = None  # type: Dict[str, dict]
````
but
```
piptools/cache.py:71: error: Incompatible types in assignment (expression has type "None", variable has type "Dict[str, Dict[Any, Any]]")
```
then 
```python
self._cache = None  # type: Optional[Dict[str, dict]]
```
but
```
piptools/cache.py:82: error: Incompatible return value type (got "Optional[Dict[str, Dict[Any, Any]]]", expected "Dict[str, Dict[Any, Any]]")
```

I've found the solution and now can't find a reference since I made changes a month ago. I should have fixed it in a separate well-described commit with the reference 😒

atugushev: > which may cause unwanted side-effects.

Yeah, now `_cache` is in `DependencyCache.__dict__`.

atugushev: Do you have any thoughts how to fix this and keep self._cache initialized in the constructor?

.pre-commit-config.yaml
atugushev: I think there's nothing wrong with this exception since packages are raw test fixtures. I'd prefer to leave it as is and move on. It's unlikely that type checking in packages could be any useful.

piptools/cache.py
atugushev: There might a design problem also.

atugushev:
```python
from typings import Optional, Dict

class Cache:
    def __init__(self):
        # type: () -> None
        self._cache = None  # type: Optional[Dict[str, dict]]

    @property
    def cache(self):
        # type: () -> Dict[str, dict]
        if self._cache is None:
            self.read_cache()
        return self._cache

    def read_cache(self):
        # type: () -> None
        self._cache = {"cache": {}}
```

```bash
$ mypy t.py
Success: no issues found in 1 source file
```

🤷‍♂

atugushev: I've refactored `read_cache` to assign explicitly `self._cache` and it works
```diff
diff --git piptools/cache.py piptools/cache.py
index 5c5fb34..24327f1 100644
--- piptools/cache.py
+++ piptools/cache.py
@@ -57,7 +57,6 @@ class DependencyCache(object):
     Where X.Y indicates the Python version.
     """

-    _cache = None  # type: Dict[str, dict]

     def __init__(self, cache_dir=None):
         # type: (Optional[str]) -> None
@@ -69,6 +68,7 @@ class DependencyCache(object):
         cache_filename = "depcache-py{}.json".format(py_version)

         self._cache_file = os.path.join(cache_dir, cache_filename)
+        self._cache = None  # type: Optional[Dict[str, dict]]

     @property
     def cache(self):
@@ -78,7 +78,7 @@ class DependencyCache(object):
         lazily loads the cache from disk.
         """
         if self._cache is None:
-            self.read_cache()
+            self._cache = self.read_cache()
         return self._cache

     def as_cache_key(self, ireq):
@@ -103,12 +103,12 @@ class DependencyCache(object):
         return name, "{}{}".format(version, extras_string)

     def read_cache(self):
-        # type: () -> None
+        # type: () -> Dict[str, dict]
         """Reads the cached contents into memory."""
         if os.path.exists(self._cache_file):
-            self._cache = read_cache_file(self._cache_file)
+            return read_cache_file(self._cache_file)
         else:
-            self._cache = {}
+            return {}

     def write_cache(self):
         # type: () -> None
```


```bash
$ mypy .
Success: no issues found in 44 source files
```

codingjoe:
@atugushev general question, why don't we use proper PEP484 type hints? They have been implemented in Python 3.5. We don't need to support Python 3.4 since it reached EoL in march this year.

If we really want to use comments for type annotations, I would recommend going with Google style doc strings.

atugushev:
@codingjoe type annotations were presented in Python 3, and they have not been backported to Python 2. The general solution is to use type comments for projects which support legacy Python. When we finally drop Python 2 support, we can convert type comments to the proper Python 2 style type annotations using [pytype](https://github.com/google/pytype) (or similar).

codingjoe:
> @codingjoe type annotations were presented in Python 3, and they have not been backported to Python 2. The general solution is to use type comments for projects which support legacy Python. When we finally drop Python 2 support, we can convert type comments to the proper Python 3 style type annotations using [pytype](https://github.com/google/pytype) (or similar).

Why don't we just drop Python 2 support? There's only two months left until EoL. Should we find a security vulnerability we can backport that patch on an older major version of pip-tools.

It's probably going to take 2 months until we removed all Python 2 specific code anyways, so by the time we release a new major version, Python 2 is finally going to be a thing of the past.

I mean, if Django, which has 100x more users, dropped Python 2 main stream support years ago, why are we having such a hard time? Just saying ;)

atugushev:
@codingjoe 

> Why don't we just drop Python 2 support?

It makes sense to drop support Python 2 when pip does. [Discussion](https://discuss.python.org/t/packaging-and-python-2/662/116) about a final date is still on, though.

image

piptools/cache.py
atugushev: Is this approach okay?

atugushev:
@codingjoe but, personally, I'm in favor of dropping Python 2 support after 2020-01-01 and add pinned pip to the latest minor version as a dependency.

codingjoe:
@atugushev is that statistic for the latest version? Also, I believe a lot of this will be package mirrors. Then again, it doesn't matter. You can still use an old version, an we can patch an old version until end of year. All I want to achieve is to save you time and effort to implement something, that we could really do a lot better in Python 3 only.

atugushev:
@codingjoe 

> @atugushev is that statistic for the latest version? Also, I believe a lot of this will be package mirrors. 

There are all the pip-tools versions.

> Then again, it doesn't matter. You can still use an old version, and we can patch an old version until end of year. All I want to achieve is to save you time and effort to implement something, that we could really do a lot better in Python 3 only.

That's why I've added the alternative solution to the #972 😉

piptools/cache.py
webknjaz: Yeah, looks good.

atugushev:
Close this due to https://github.com/jazzband/pip-tools/issues/972#issuecomment-623595869.

PR #975, opened 2019-10-30
Unpin commented out unsafe packages

### LINKED ISSUE #973: pip-compile bumps setuptools within commented-out spec
If you have a `requirements.in` which contains a package which depends on `setuptools`, for example:
```
$ cat requirements.in 
protobuf  # or google-api-core, ipython, pandas-gbq, pydata-google-auth
```
Then after running `pip-compile`, you get:
```
$ cat requirements.txt
#
# To update, run:
#
#    pip-compile
#
protobuf==3.10.0
six==1.12.0               # via protobuf

# setuptools==41.4.0        # via protobuf
```

As you can see, `pip-compile` refuses to include `setuptools` directly in a generated `requirements.txt`, instead adding a section with the comment "The following packages are considered to be unsafe in a requirements file:".
This behaviour is fine in the general (and I don't disagree with the comment or behaviour).

However because the commented out `setuptools` line includes a version number, when `setuptools` releases an update then the `requirements.txt` file ends up looking out of date.

Say you had committed a `requirements.txt` which looked like:
```
# requirements...

# setuptools==41.4.0        # via protobuf
```

Then `setuptools` releases `41.5.0` and now when you run `pip-compile` you get:
```
# requirements...

# setuptools==41.5.0        # via protobuf
```

This is inconvenient if you happen to regenerate your `requirements.txt` at some point after an upstream releases, however if you're also using `pip-compile` to check that your `requirements.txt` is in step with your `requirements.in` (see #882) in CI then it means that your build now fails for what is essentially a non-change.

I'm not sure what the right solution is here, though perhaps if there was a way for pip-compile to understand its own commented out dependency that might help?

atugushev:
Resolves #973. See the proposal in https://github.com/jazzband/pip-tools/issues/973#issuecomment-547010219.

**Changelog-friendly one-liner**: Unpin commented out unsafe packages in `requirements.txt`.

##### Contributor checklist

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -189,17 +189,18 @@ class OutputWriter(object):
                 yield MESSAGE_UNSAFE_PACKAGES
 
             for ireq in unsafe_requirements:
-                req = self._format_requirement(
-                    ireq,
-                    reverse_dependencies,
-                    primary_packages,
-                    marker=markers.get(key_from_ireq(ireq)),
-                    hashes=hashes,
-                )
+                ireq_key = key_from_ireq(ireq)
                 if not self.allow_unsafe:
-                    yield comment("# {}".format(req))
+                    yield comment("# {}".format(ireq_key))
                 else:
-                    yield req
+                    line = self._format_requirement(
+                        ireq,
+                        reverse_dependencies,
+                        primary_packages,
+                        marker=markers.get(ireq_key),
+                        hashes=hashes,
+                    )
+                    yield line
 
         # Yield even when there's no real content, so that blank files are written
         if not yielded:
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -691,8 +691,7 @@ def test_annotate_option(pip_conf, runner, option, expected):
 
 
 @pytest.mark.parametrize(
-    "option, expected",
-    [("--allow-unsafe", "\nsetuptools=="), (None, "\n# setuptools==")],
+    "option, expected", [("--allow-unsafe", "\nsetuptools=="), (None, "\n# setuptools")]
 )
 @pytest.mark.network
 def test_allow_unsafe_option(runner, option, expected):
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -129,7 +129,7 @@ def test_iter_lines__unsafe_dependencies(writer, from_line, allow_unsafe):
         "test==1.2",
         "",
         MESSAGE_UNSAFE_PACKAGES,
-        "setuptools==1.10.0" if allow_unsafe else comment("# setuptools==1.10.0"),
+        "setuptools==1.10.0" if allow_unsafe else comment("# setuptools"),
     )
     assert tuple(lines) == expected_lines
 
@@ -147,7 +147,7 @@ def test_iter_lines__unsafe_with_hashes(writer, from_line):
         "test==1.2 \\\n    --hash=FAKEHASH",
         "",
         MESSAGE_UNSAFE_PACKAGES_UNPINNED,
-        comment("# setuptools==1.10.0"),
+        comment("# setuptools"),
     )
     assert tuple(lines) == expected_lines
 


atugushev:
It turns out it was already discussed before, see #276 (comment), and fixed in #294, but has been broken later unintentionally.

tests/test_cli_compile.py
```diff
@@ -691,8 +691,7 @@ def test_annotate_option(pip_conf, runner, option, expected):
 
 
 @pytest.mark.parametrize(
-    "option, expected",
-    [("--allow-unsafe", "\nsetuptools=="), (None, "\n# setuptools==")],
+    "option, expected", [("--allow-unsafe", "\nsetuptools=="), (None, "\n# setuptools")]
```
webknjaz: I don't see any difference in this change. Looks unrelated...

piptools/writer.py
```diff
@@ -189,17 +189,18 @@ def _iter_lines(
...
-                    marker=markers.get(key_from_ireq(ireq)),
-                    hashes=hashes,
-                )
+                ireq_key = key_from_ireq(ireq)
                 if not self.allow_unsafe:
-                    yield comment("# {}".format(req))
+                    yield comment("# {}".format(ireq_key))
                 else:
-                    yield req
+                    line = self._format_requirement(
```
webknjaz: Maybe just `yield` it right away? I mean, it's expected that only lines are yielded by this generator function anyway, so this variable name doesn't add much semantical meaning to the code.

piptools/writer.py
```diff
@@ -189,17 +189,18 @@ def _iter_lines(
...
 
             for ireq in unsafe_requirements:
-                req = self._format_requirement(
-                    ireq,
-                    reverse_dependencies,
-                    primary_packages,
-                    marker=markers.get(key_from_ireq(ireq)),
-                    hashes=hashes,
-                )
+                ireq_key = key_from_ireq(ireq)
```
webknjaz: could as well generate this above as
```python
unsafe_requirements_map = ((key_from_ireq(r), r) for r in unsafe_requirements)
```
and iterate over that

piptools/writer.py
```diff
@@ -189,17 +189,18 @@ def _iter_lines(
...
                 if not self.allow_unsafe:
-                    yield comment("# {}".format(req))
+                    yield comment("# {}".format(ireq_key))
```
webknjaz: If you stick `continue` here, you could dedent the following code block
```suggestion
                    yield comment("# {}".format(ireq_key))
                    continue
```

tests/test_cli_compile.py
webknjaz: Oh, I see. The second one doesn't have `==`. It's been hard to spot. I suggest that you minimize the diff in order to make it review-friendly.

atugushev: Yeah, that's why I've split the PR into two commits.

webknjaz: Well, it's not visible in the PR diff and folks mostly don't do reviews per commit (I do sometimes but to all the time).

OTOH, you mix up two separate tasks in this PR:
1) reformatting the code
2) changing things

This makes it non-atomic, which is not good in my book. As you can see, this confuses even me. And I do a lot more reviews than most of the regular folks (compared to other types of contributions). Which opens up possibilities to sneak in a few more bugs unnoticed. Humans are error-prone and byte-by-byte comparisons are hard without any highlighting.

webknjaz: Also, these two changes are even in the same commit.

atugushev: What do you mean? This is the fix of the test.

atugushev: Yeah, it should have been separate PRs. I thought these changes would be easy to review (which is apparently not). My bad. Would you like me to split this into two PRs or we could go on?

piptools/writer.py
atugushev: But we still need an `ireq` (which uses in `self._format_requirement`) reference in the for-loop body.

piptools/writer.py
atugushev: Yeah, we can do that. I did it the same way as it's done above (see 171th line). IMO the `yield line` looks more readable than `yield function call with many arguments` and it's debugger-friendly.

piptools/writer.py
atugushev: You probably right, but I think it's not related to the issue. I'd prefer to touch these lines in a separate PR. First things first.

atugushev: And it looks subjective, is there a PEP which suggests to use:
```python
for x in z:
    if x:
        foo()
    else:
        bar()    
```

instead of

```python
for x in z:
    if x:
        foo()
        continue
    bar()    
```
? Just wondering.

tests/test_cli_compile.py
webknjaz: Yes, it would be better to avoid reformatting things in PRs that target changing the behavior.

piptools/writer.py
webknjaz: ok

piptools/writer.py
webknjaz: yes, by using `for ireq_key, ireq in unsafe_requirements_map:`

atugushev: Oh, that's tricky. I'd prefer it keep simple :)

piptools/writer.py
webknjaz: Yes, it's subjective. I don't recall this being mentioned in any PEPs.
1) It reduces the indentation which is better for readability
2) There's a similar pylint rule for `return`
   ```python
   def func():
       if cond1:
           return xxx
       else:  # pylint would emit a warning here
           return yyy
   ```
   I think it's reasonable to apply this rule to other similar structures.
3) Yes, it makes sense to do this in a separate refactoring PR to keep the diff small.

tests/test_cli_compile.py
atugushev: I must note the first commit is not about reformatting. It improves the test. The actual test was not complete enough to make changes.

webknjaz: This relates to the previous chain of comments explaining that it's bad to mix formatting and logic changes together.

atugushev: "Yes" means split or go on?

webknjaz: Please split

piptools/writer.py
webknjaz: ok

tests/test_cli_compile.py
webknjaz: P.S. these comments refer to the second commit which both (1) reformats 2 separate lines into one **and** (2) changes what's needed for the test. Hence two atomic changes in a single commit.

atugushev: We use black. Reformat is necessary, but not the intention.

webknjaz: Damn. black again...

atugushev: > Please split

okay 👍

atugushev: First one #986.

atugushev: I've rebased the branch onto the master.

tests/test_cli_compile.py
```diff
@@ -691,8 +691,7 @@ def test_annotate_option(pip_conf, runner, option, expected):
 
 
 @pytest.mark.parametrize(
-    "option, expected",
-    [("--allow-unsafe", "\nsetuptools=="), (None, "\n# setuptools==")],
```
atugushev: FTR, would be nice to pin `setuptools` here as well, but it would take to rewrite the test.

tests/test_cli_compile.py
atugushev: FTR, would be nice to pin setuptools here as well, but it would take to rewrite the test.

webknjaz: Wait, it still has reformatting and a useful change smashed in together.

atugushev: Because first, we do useful change, then black reformats it. And we can't split it, because of the linter.

webknjaz: And that's bad because the change doesn't show up highlighted in diff and it looks like you just took two lines and merged them into one because of this making the review next to impossible.

atugushev: > you just took two lines and merged them into

this is what black does, not me :)

webknjaz: you could always `# black off` to work around black's downsides

webknjaz: I mean, if this keeps happening, we need to know how to make things review-friendly. Otherwise, we'll just silently accumulate bugs.

atugushev: This is definitely what I don't want to do. Better to not use black at al then.

atugushev: It would be nice if github's diff-tool would highlight changed string (which is removed "==" in this particular case).

atugushev: > I mean, if this keeps happening, we need to know how to make things review-friendly. Otherwise, we'll just silently accumulate bugs.

Just wondering is there any discussion about black and non-friendly-review changes.  If not, let's open the discussion in black's issue-tracker for a start?

webknjaz: Yes, I usually advocate for not using black because of its problems and freedoms it steels but it's not my call to make here, I guess :)

Regarding diff. I don't think that `git diff` as well as the default `diff` tool is capable of detecting string merge *and* edit. To these tools, it just looks like removing one line and editing another.

OTOH if you look to this diff line https://github.com/jazzband/pip-tools/pull/975/files#diff-9960f0e0139a2d6000305af0c80dac37R194 you'll see a part of the string with highlighted _which is what our goal should be_.

I dream of a tool that works similar to black but only reformats things that violate PEP8 and not everything unrelated to the target change...

webknjaz: > Just wondering is there any discussion about black and non-friendly-review changes. If not, let's open the discussion in black's issue-tracker for a start?

Not that I know off. But whenever I start it, everybody starts hating me for this. So I'm not very enthusiastic about participating in yet another argument personally...

webknjaz: I've converted this into an issue: #988.

atugushev: Thanks!

atugushev: > OTOH if you look to this diff line https://github.com/jazzband/pip-tools/pull/975/files#diff-9960f0e0139a2d6000305af0c80dac37R194 you'll see a part of the string with highlighted _which is what our goal should be_.

That's what I'm talking about!

> Not that I know of. But whenever I start it, everybody starts hating me for this. So I'm not very enthusiastic about participating in yet another argument personally...

Haha! That's a rough path 😄

webknjaz: Anyway. I think that one commit should have a logic change and then autoformatting should be a separate commit.
Or just merge it if you want.

atugushev: That makes sense, but as you mentioned before:
> Well, it's not visible in the PR diff and folks mostly don't do reviews per commit (I do sometimes but to all the time).
 
And I can't merge this unless someone approves it.

webknjaz: Oh, that approval from @PeterJCLaw doesn't work..

webknjaz: > as you mentioned before

I'd like to point out that both individual commits and PRs can be atomic, they just refer to different abstraction levels.

atugushev:
@webknjaz thanks for reviewing this!

Issue #882, opened 2019-08-31
Add compliation check mechanism


PeterJCLaw:
I'd like to be able to run a command in CI to ensure that my `requirements.in` produces my `requirements.txt`, mostly as a guard against accidental edits to the latter which get undone when `pip-compile` is next run.

I'm currently using `pip-compile --dry-run` with some bodging of the output then run through a `diff` against my `requirements.txt` to enable this, however it's (predictably) not exactly reliable (sometimes I get warning text in the output which wouldn't be in the requirements file).

I can see a couple of solutions:
- make `pip-compile` output the requirements.txt to `STDOUT`, rather than `STDERR`, so that there's a stream which I can use which is sure to be free of other text
- make an actual check mode

The latter feels like the much better solution, though it seems likely to be more effort. I've no idea what the former would break though. (`pip-compile` seems not to use `STDOUT` at all?)

I'm using Python 3.5 and pip-tools 4.1.0.

atugushev:
Hello @PeterJCLaw,

Thanks for bringing this up. I'd prefer to do it in a way of "Single Responsibility Principle":

```bash
$ pip-compile --quiet && git diff --exit-code
```

If check is failed you can see the changes of `requirements.txt`.

PeterJCLaw:
That's definitely nicer than my previous approach, though it does assume that the requirements file is under version control (I agree this is likely), which feels like an odd restriction, and that no other changes have been made to the working area.

As far as I know in `pip-compile` can only operate on an output file which is the same as it's reference/lock file (I'm not sure what term you use for that). If there were a way to have a different input file than the output that might help here. (This is kinda where I was going with suggesting the change to `--dry-run`'s output). That said, I'm not actually sure that this is a good idea either (and introducing it just for this case feels a little contrived).

I can see that it would be nice to defer the comparison part to another tool, however one could argue that the `--dry-run` behaviour is already headed down this path -- why would you need `--dry-run` when you can can just run `pip-compile` and then reset the file from version control if you don't want the changes? (There's even an argument that that's preferable to `--dry-run` because you can easily diff the file)

In any case, `git diff --exit-code` works for my use-case, so I'm happy to use that for now.

atugushev:
@PeterJCLaw 

> In any case, `git diff --exit-code` works for my use-case, so I'm happy to use that for now.

I'm glad to help you. Anyways, let's leave this issue open and see whether the community is interested in the optional check mode.

Zac-HD:
[...]

PR #976, opened 2019-10-30
Add pre-commit hook for pip-compile

### LINKED ISSUE #974: pre-commit hook for pip-compile

Running `pip-compile` to ensure that `requirements.txt` is in sync with `setup.py`

A [pre-commit](https://pre-commit.com/) hook for this purpose, so users don't have to do this manually

At this point, a CI check will fail if `pip-compile` generates a modified `requirements.txt` than what is present in the source, to indicate to the user that they should do this on their end.
This is conceptually similar to running a linter like [black](https://github.com/psf/black) etc which have pre-commit hooks for this purpose, so it makes sense that `pip-compile` could have one too.

atugushev:
Resolves #974.
Refs #882.

**Changelog-friendly one-liner**: Add pre-commit hook for `pip-compile`.

##### Contributor checklist

- ~~Provided the tests for the changes.~~


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,6 @@
+- id: pip-compile
+  name: pip-compile
+  description: Automatically compile requirements.
+  entry: pip-compile
+  language: python
+  files: (requirements\.in|setup\.py)$
```
asottile: I'd suggest `pyproject.toml` and `setup.cfg` as well for completeness

and probably some docs on how to use this / when you might want to use this

atugushev: > I'd suggest `pyproject.toml` and `setup.cfg` as well for completeness

Thanks for noticing! I've missed that out. :(

> and probably some docs on how to use this / when you might want to use this

Could be enough something like https://github.com/asottile/flake8-walrus#as-a-pre-commit-hook?

atugushev: Can we borrow it? :)

asottile: sure! you may want to suggest `stages: [pre-push]` as well as `pre-commit install --hook-type pre-push` as well

asottile: lol, you can tell how often I use that feature :upside_down_face: (and I'm the author!)

atugushev: Do we really need to mention the stages? This looks good to me:

```yaml
repos:
  - repo: https://github.com/jazzband/pip-tools
    rev: 4.3.0
    hooks:
      - id: pip-compile
```

asottile: up to you, not my project :D

atugushev: I've tried this hook as a user on my repositories and it looks `pass_filenames: false` and probably `files` too need to be removed. Otherwise, pre-commit will run something like:

```console
$ pip-compile setup.py setup.cfg
```

which is totally wrong, since setup.cfg is not a requirement file.

I'd let users configure `files` and `args` manually for their special case.

asottile: if there's no good default, it might make sense to set `files: ^$` + `pass_filenames: false` and document that the consumer must configure `args: [requirements.in]` + `files: ^(...)$` (for example)

dirn:
@atugushev does the hook fail for you? I tried using it but, while I successfully get a new `requirements.txt` file, the `pre-commit` shows the hook as passing.

atugushev:
@dirn 

> does the hook fail for you? I tried using it but, while I successfully get a new `requirements.txt` file, the `pre-commit` shows the hook as passing.

I've updated the hook. Could you try the last one?

dirn:
I think I understand my problem a bit better now. The hook passes if `requirements.txt` doesn't already exist. Once it does, the hook correctly fails. I'm not sure there's much `pip-compile` can do about that.

atugushev:
@dirn Yeah, this works if `requirements.in` and `requirements.txt` are in VCS.

sfdye:
I tried in my project with the this 

<details><summary>config</summary>
<pre>
repos:
  - repo: https://github.com/atugushev/pip-tools
    rev: pre-commit-hook
    hooks:
      - id: pip-compile
</pre>
```
</details>

```
$ pre-commit run
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/xxx/.cache/pre-commit/patchxxx.
pip-compile..............................................................Passed
[INFO] Restored changes from /Users/lwan/.cache/pre-commit/patchxxx.
```

It worked from a `pip-compile` perspective, but unable to detect any files other than `requirements.(in|txt)`. (e.g. `requirements/main.txt`)

It will be nice to support any requirement file in any location.

**Comment by atugushev** [comment]
@sfdye 

Thank you for checking this out. Default pre-commit config handles only default values. If you have other than default values, you should configure `args` and `files` correspondingly. For example:
```yml
repos:
  - repo: https://github.com/atugushev/pip-tools
    rev: pre-commit-hook
    hooks:
      - id: pip-compile
        files: ^requirements/main\.(in|txt)$
        args: [requirements/main.in]
```

atugushev:
@sfdye 

Since you've had an issue with this, it seems I should definitely address this https://github.com/jazzband/pip-tools/pull/976#discussion_r342092925.

sfdye:
Yes, please. Would love to see this merged😁

atugushev:
@sfdye great! Please, review the latest changes if you get a chance.

.pre-commit-hooks.yaml
atugushev: Addressed in 5f2dea3db93cb5e313b654c0c26750195827b86f

README.rst
```diff
@@ -310,6 +310,34 @@ You can install requirements in development stage by:
...
+
+You might use ``pip-compile`` as a hook for the `pre-commit <https://github.com/pre-commit/pre-commit>`_.
+See `pre-commit docs <https://pre-commit.com/>`_ for instructions.
+Sample ``.pre-commit-config.yaml``:
+
+.. code-block:: yaml
+
+    repos:
+      - repo: https://github.com/jazzband/pip-tools
+        rev: 5.0.0
```
sfdye: Why 5.0.0?

sfdye: Never mind @atugushev I just saw this:
https://github.com/jazzband/pip-tools/milestone/32

Looks good to me then

atugushev: 👍

atugushev:
@sfdye thanks for reviewing this! 🙏

PR #977, opened 2019-10-31
Add short name version option (-V) to the CLI scripts

atugushev:
Prints version on `pip-compile -V` and `pip-sync -V` commands (like `python -V` or `pip -V`), under the hood I had to do the following changes:
* Set a name to the cli commands, so that the tests can get the correct name (not just a `cli`)
* Use `setuptools-scm` at a build time and writes explicit version to `piptools/__version__.py`
* Add `piptools.__version__` variable

**Changelog-friendly one-liner**: Add short name version option `-V` to the CLI scripts.

##### Contributor checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ venv/
 requirements.in
 requirements.txt
 .eggs/
+
+# Dynamic version file written by setuptools_scm
+/piptools/__version__.py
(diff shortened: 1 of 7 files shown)


atugushev:
Too much code for the zero-valued feature.

PR #984, opened 2019-11-09
Install fresh virtualenv on AppVeyor

### LINKED ISSUE #983: test_realistic_complex_sub_dependencies fails on Windows py27-pip8.1.1
It's failing since 69e4142d1c2ffa615ca8343b6c32a382d8e8115e. See the [log on AppVeyor](https://ci.appveyor.com/project/atugushev/pip-tools/builds/28738464/job/3n61g7vsxks9r10g).

#### Environment Versions

1. Windows
1. Python version: 2.7
1. pip version:  8.1.1
1. pip-tools version: 4.2.0

#### Steps to replicate

1. pytest -k test_realistic_complex_sub_dependencies

#### Expected result

Test passes.

#### Actual result

```
================================== FAILURES ===================================
___________________ test_realistic_complex_sub_dependencies ___________________

runner = <click.testing.CliRunner object at 0x000000000806CE80>

    @pytest.mark.network
    def test_realistic_complex_sub_dependencies(runner):
        wheels_dir = "wheels"
    
        # make a temporary wheel of a fake package
        subprocess.check_output(
            [
                "pip",
                "wheel",
                "--no-deps",
                "-w",
                wheels_dir,
                os.path.join(PACKAGES_PATH, "fake_with_deps", "."),
            ]
        )
    
        with open("requirements.in", "w") as req_in:
            req_in.write("fake_with_deps")  # require fake package
    
        out = runner.invoke(cli, ["-v", "-n", "--rebuild", "-f", wheels_dir])
    
>       assert out.exit_code == 0
E       AssertionError: assert 1 == 0
E        +  where 1 = <Result TypeError(u'Expected a pinned InstallRequirement, got unknown from htt...e38 (from pyramid==1.5.8->fake-with-deps==0.1->-r requirements.in (line 1))',)>.exit_code

C:\projects\pip-tools\tests\test_cli_compile.py:240: AssertionError
---------------------------- Captured stderr call -----------------------------
You are using pip version 8.1.1, however version 19.3.1 is available.

You should consider upgrading via the 'python -m pip install --upgrade pip' command.

------------------------------ Captured log call ------------------------------
WARNING  pip.req.req_install:req_install.py:453 Running setup.py (path:c:\users\appveyor\appdata\local\temp\1\tmpq7ozxubuild\venusian\setup.py) egg_info for package venusian produced metadata for project name unknown. Fix your #egg=venusian fragments.
```

Expanded assertion error:

```python
TypeError(
    u'Expected a pinned InstallRequirement, got unknown from '
    u'https://files.pythonhosted.org/packages/53/d9/29816f12828a5ff539e66afe270a764d07bf4bb4cefc6613e331c9d38ba9/venusian-3.0.0.tar.gz#sha256=f6842b7242b1039c0c28f6feef29016e7e7dd3caaeb476a193acf737db31ee38 '
    u'(from pyramid==1.5.8->fake-with-deps==0.1->-r requirements.in (line 1))'
)
```

#### Additional context

`Venusian` package has recently dropped Python 2.7 support, see the [changelog](https://github.com/Pylons/venusian/blob/master/CHANGES.rst#200-2019-10-04). Also, it moved all setup keywords to setup.cfg, which most likely causes the problem, see warning:

```
WARNING  pip.req.req_install:req_install.py:453 Running setup.py (path:c:\users\appveyor\appdata\local\temp\1\tmpqebvambuild\venusian\setup.py) egg_info for package venusian produced metadata for project name unknown. Fix your #egg=venusian fragments.
```

atugushev:
Fixes #983. See https://github.com/jazzband/pip-tools/issues/983#issuecomment-552114449.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -102,6 +102,7 @@ matrix:
 
 install:
     - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+    - pip install -U virtualenv
     - pip install tox
 
 build: false
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -102,7 +102,12 @@ matrix:
 
 install:
     - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-    - pip install -U virtualenv
+
+    # Temporary workaround to prevent tests failing.
+    # See GH-983 for the details.
+    # TODO remove this after tox>3.14.0 being released
+    - pip install "virtualenv>=16.0.0"
+
     - pip install tox
 
 build: false


.appveyor.yml
```diff
@@ -102,6 +102,7 @@ matrix:
 
 install:
     - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+    - pip install -U virtualenv
```
codingjoe: Why not install virtualenv and tox within one line?
I'd prefer you use `--update`, since it's more human readable.

atugushev: It's a workaround and easy to git-blame. Plus tox doesn't need to be upgraded.

Maybe we should better put there a comment about why do we really need to upgrade `virtualenv`?

atugushev: Actual, I prefer long-options too, especially in tests when we run something like `invoke(cli, ["--upgrade-package", "foo"])`, cause I always forget what does `-P` mean. 

These `pip -e -r -U` options are so famous and I type it because of a muscle memory 😅
We could normalize all options here and in the other configs in a separate PR. What do you thnk?

atugushev: Also, we can remove this workaround after tox releases a new version, where virtualenv [has bumped to >=16.0.0](https://tox.readthedocs.io/en/latest/changelog.html#vdraft-2019-10-23).

atugushev:
@codingjoe thanks for reviewing this!

PR #986, opened 2019-11-15
Refactor-out unassertive OutputWriter() tests

atugushev:
These changes make tests more assertive. Now tests check the exactly expected output, rather than a substring in the output.

Addresses https://github.com/jazzband/pip-tools/pull/975#discussion_r346959435.


atugushev:
@webknjaz, thanks! 🙏

Issue #321, opened 2016-02-17
Pass additional arguments to pip


31z4:
It would be nice to have an ability to pass some non harmful additional arguments to `pip` during `pip-sync`. E. g. `--no-cache-dir` or `--cache-dir <dir>`.

estan:
I would like to be able to pass `--upgrade` to `pip` when doing `pip-sync`, the reason is I'm unfortunately required to use `--system-site-packages` on my virtualenv, to get access to the distro installed PyQt (which is not on PyPI yet, so not in `requirements.txt`), but I still want `pip-sync` to install as much as possible into the virtualenv (and not skip stuff that is already satisfied by the system-wide `site-packages`).

estan:
Just want to check, is there still no way to do this?

tkwon:
+1 on adding options like --no-cache-dir

merwok:
A nice interface for this could be the command-line end marker: `pip-sync requirements.txt -- --pip-option-1 --pip-option-2=no`

atugushev:
That's a great idea! How to handle `--` using click API? I skimmed docs and code and didn't find the solution.

aorumbayev:
@atugushev any updates on this issue ?

atugushev:
Hello @aorumbayev,

Unfortunately, there are no updates. It's still not clear how to pass the additional arguments.

atugushev:
@merwok 

> A nice interface for this could be the command-line end marker: `pip-sync requirements.txt -- --pip-option-1 --pip-option-2=no`

What if we need to process files containing double-dash? For example:

```console
$ touch -- --requirements--.in
$ pip-compile -- --requirements--.in
```

merwok:
@atugushev Not sure I get your question.  Recognizing a single token with the exact value of `--` shouldn’t interfere with other params containing two dashes.

atugushev:
@merwok 

> Not sure I get your question. Recognizing a single token with the exact value of `--` shouldn’t interfere with other params containing two dashes.

I mean how to distinguish filenames and options after `--`? For instance:

```console
$ pip-compile -- --requirements--.in --no-cache-dir
```

merwok:
That wouldn’t be valid.  `--` means «end of processing, so all parameters for pip-compile mist appear before it:

```
$ pip-compile --some-opt file1.in --more-opt file2--weird.in -- --pip-option1 --pip-option2 arg --pip-option3
```

atugushev:
[...]

PR #987, opened 2019-11-15
Add --no-deps flag to pip-sync, which is simply passed through to pip install

AndydeCleyre:
`--no-deps` can now be provided to `pip-sync`, whereby `--no-deps` will be passed through to `pip install`, and un-locked deps of locked requirements will be either removed or not installed in the first place, in order to match the lockfile exactly, despite breaking package dependencies.

**Changelog-friendly one-liner**: `pip-sync` now accepts `--no-deps`, to match locked `txt`s exactly, ignoring any unlocked dependencies

##### Contributor checklist

### DIFF
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -28,6 +28,7 @@ DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
     help="Only show what would happen, don't change anything",
 )
 @click.option("--force", is_flag=True, help="Proceed even if conflicts are found")
+@click.option("--no-deps", is_flag=True, help="Don't install package dependencies")
 @click.option(
     "-f",
     "--find-links",
@@ -72,6 +73,7 @@ def cli(
     ask,
     dry_run,
     force,
+    no_deps,
     find_links,
     index_url,
     extra_index_url,
@@ -120,6 +122,8 @@ def cli(
     install_flags = []
     for link in find_links or []:
         install_flags.extend(["-f", link])
+    if no_deps:
+        install_flags.append("--no-deps")
     if no_index:
         install_flags.append("--no-index")
     if index_url:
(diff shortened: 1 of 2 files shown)


piptools/scripts/sync.py
```diff
@@ -28,6 +28,7 @@
     help="Only show what would happen, don't change anything",
 )
 @click.option("--force", is_flag=True, help="Proceed even if conflicts are found")
+@click.option("--no-deps", is_flag=True, help="Don't install package dependencies")
```
atugushev:
```suggestion
@click.option("--no-deps", is_flag=False, help="Don't install package dependencies")
```
`is_flag=True` might be a breaking change.

atugushev: The `is_flag` has different meaning and value is `False` by default . Sorry, this always confuses me :(

atugushev:
@AndydeCleyre 

Just wondering, why do you prefer to mark tasks this way:

```
[✓] Provided the tests for the changes.
```
instead of
```
[x] Provided the tests for the changes.
```
wich is the current Github's markdown syntax? See the [doc](https://help.github.com/en/github/managing-your-work-on-github/about-task-lists).


Thie first one we get this progress bar on PRs list:
image

The second one marks tasks correctly:
image

AndydeCleyre:
> Just wondering, why do you prefer to mark tasks this way?

Just because I never read those docs, and generally think of a check as "yes" and an `x` as "no," but thank you, now I know better.

atugushev:
@AndydeCleyre 

There is an issue #321 the implementation of which can supersede this PR. Let me know what you think of it.

AndydeCleyre:
@atugushev If `--no-deps` functionality is enabled by an implementation for #321 that's great and should be merged instead. My only interest here is that there be an acceptable interface for people who don't want #907 behavior, so we can fix #896 and its sub-issue of sync behavior inconsistency.

Inviting @ulope and @Zac-HD to comment

atugushev:
Close this in favor of #1080.

PR #1004, opened 2019-11-25
Fix pip version from 19.4 to 20.0

atugushev:
The next major pip version will be `20.0`.

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -202,7 +202,7 @@ class PyPIRepository(BaseRepository):
                 )
                 resolver_kwargs["make_install_req"] = make_install_req
 
-            if PIP_VERSION >= (19, 4):
+            if PIP_VERSION >= (20,):
                 preparer_kwargs["session"] = self.session
                 del resolver_kwargs["session"]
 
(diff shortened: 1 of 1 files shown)


atugushev:
@codingjoe I miscalculated the next major pip version 🤦‍♂

piptools/repositories/pypi.py
```diff
@@ -220,7 +220,7 @@ def resolve_reqs(self, download_dir, ireq, wheel_cache):
 
                 resolver = PipResolver(**resolver_kwargs)
                 require_hashes = False
-                if PIP_VERSION < (19, 4):
+                if PIP_VERSION < (20,):
```
codingjoe: Maybe we shouldn't use `<` but `<=` to a version number that is already released.

atugushev: Do you mean this condition?
```python
if PIP_VERSION <= (19, 3):
    results = resolver._resolve_one(reqset, ireq, require_hashes)
else:
    resolver.require_hashes = require_hashes
    results = resolver._resolve_one(reqset, ireq)
```

Sounds good, but the condition would fail for the following minor releases, e.g. (19, 3, 1).

atugushev:
@codingjoe 

> Yes, it's hard to anticipate the next release.

Actually, it's not that hard, because pip has a release [cadence](https://pip.pypa.io/en/stable/development/release-process/) of releasing whatever is on master every 3 months. Release months are January, April, July, October. It was just my mistake to choose 19.4 :/

piptools/repositories/pypi.py
codingjoe: Maybe `PIP_VERSION[:2] <= (19, 3)`?

atugushev: Yeah, that'll work! I'll try this in the following PR. Thanks! 👍

PR #1008, opened 2019-11-26
Add .venv to .gitignore

atugushev:


atugushev:
@auvipy thanks for reviewing and merging this 👍

PR #1012, opened 2019-11-27
Add CI workflow for GitHub Actions

atugushev:


atugushev:
Tracking issue for tokenless codecov reports: https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491/5

jezdez:
Remember to let me know if you ever remove Travis so we can set up the Jazzband release system via a GH action.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+      - name: Set PY
+        run: echo "::set-env name=PY::$(python -VV | sha256sum | cut -d' ' -f1)"
+      - uses: actions/cache@v1
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
```
webknjaz, line 27: Instead of setting a global env var, I'd rather use outputs:
```suggestion
      - name: Set PY
        id: py-ver
        run: echo "::set-output name=sha-256::$(python -VV | sha256sum | cut -d' ' -f1)"
      - uses: actions/cache@v1
        with:
          path: ~/.cache/pre-commit
          key: pre-commit|${{ jobs.py-ver.outputs.sha-256 }}|${{ hashFiles('.pre-commit-config.yaml') }}
```
_(https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter)_

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        toxenv:
+          - checkqa
+          - readme
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
```
webknjaz, line 21: You forgot to actually pass the arg:
```suggestion
        uses: actions/setup-python@v2
        with:
          python-version: ${{ matrix.python-version }}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+
+jobs:
+  qa:
+    name: QA
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        toxenv:
+          - checkqa
+          - readme
```
webknjaz, line 17: You use this matrix value later but didn't declare it:
```suggestion
          - readme
        python-version:
        - 3.8
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
```
webknjaz, line 6: Not sure if it'll run on tags if set up like this

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+jobs:
+  qa:
+    name: QA
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        toxenv:
+          - checkqa
+          - readme
+    steps:
```
webknjaz, line 18:
```suggestion
    env:
      TOXENV: ${{ matrix.toxenv }}
    steps:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Set PY
+        run: echo "::set-env name=PY::$(python -VV | sha256sum | cut -d' ' -f1)"
+      - uses: actions/cache@v1
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
+      - name: Install tox
+        run: pip install tox
+      - name: Test ${{ matrix.toxenv }}
+        run: tox -e ${{ matrix.toxenv }}
```
webknjaz, line 31: Better set it via env var https://github.com/jazzband/pip-tools/pull/1012/files#r419001428:
```suggestion
        run: tox
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  qa:
+    name: QA
```
webknjaz, line 11: Make it a bit more specific:
```suggestion
    name: QA / ${{ matrix.toxenv }}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+        uses: actions/setup-python@v2
+      - name: Set PY
+        run: echo "::set-env name=PY::$(python -VV | sha256sum | cut -d' ' -f1)"
+      - uses: actions/cache@v1
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
+      - name: Install tox
+        run: pip install tox
+      - name: Test ${{ matrix.toxenv }}
```
webknjaz, line 30:
```suggestion
      - name: Prepare tox testenv
        run: tox --notest -p auto --parallel-live
      - name: Test ${{ matrix.toxenv }}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+      matrix:
+        os:
+          - Ubuntu
+          - Windows
+        python-version:
+          - 2.7
+          - 3.5
+          - 3.6
+          - 3.7
+          - 3.8
```
webknjaz, line 46: Let's prioritize latest a bit:
```suggestion
          - 3.8
          - 2.7
          - 3.5
          - 3.6
          - 3.7
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: pip install tox
+      - name: Test pip ${{ matrix.pip-version }}
+        run: tox -e pip${{ matrix.pip-version }}-coverage
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1.0.6
+        with:
+          file: ./coverage.xml
+          name: ${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.pip-version }}
+
+  pypy:
```
webknjaz, line 66: I don't think this deserves a separate job setup. It could be handled in the previous matrix.

.github/workflows/ci.yml
atugushev: Yeah, there should be "tags:".

.github/workflows/ci.yml
atugushev: I'd rather remove the `python-version`, forgot to do that.

.github/workflows/ci.yml
atugushev: I didn't pas it intentionally, because I wanted to use the default which is `3.x`.

.github/workflows/ci.yml
atugushev: It's taken from official setup https://pre-commit.com/#github-actions-example

.github/workflows/ci.yml
atugushev: I intentionally didn't use env to be consistent with `cpython|pypy` jobs.

.github/workflows/ci.yml
atugushev: Yeah, it makes sense.

.github/workflows/ci.yml
webknjaz: I know, but when Anthony wrote that official setup, I think the commands were not fully documented. No need to pollute env now.

.github/workflows/ci.yml
webknjaz: I think it's cleaner to use env everywhere. Besides, you still need to declare other env vars.

.github/workflows/ci.yml
atugushev: Yes, it could but it would take too many lines to run pypy jobs only with the latest pip version and ubuntu (which is done to reduce job numbers previously, still actual though) using exclude directive.

.github/workflows/ci.yml
webknjaz: You also used it in `name:`

.github/workflows/ci.yml
atugushev: That's why I decided to put it under a separate job

.github/workflows/ci.yml
atugushev: I but it by a mistake, i'll remove it.

.github/workflows/ci.yml
webknjaz: I'm not convinced. Excludes may actually be way more flexible than you think. You shouldn't need to declare many of them.

.github/workflows/ci.yml
yuqieang19841028: []()

atugushev: Have you tried to update official docs?

.github/workflows/ci.yml
atugushev: > Besides, you still need to declare other env vars.

Which ones?

.github/workflows/ci.yml
webknjaz: Not yet, I'm currently building a state-of-the-art workflow @ https://github.com/ansible/pylibssh/blob/devel/.github/workflows/build-test-n-publish.yml that tests the dists that are going to be published rather than testing the source separately. And when I'm done with polishing it I'll go and contribute the lessons learned to other places...

.github/workflows/ci.yml
webknjaz: The ones that tox understands: https://github.com/jazzband/pip-tools/pull/1012#discussion_r419001428.

Plus, it's also the best practice to have a separate step with `tox --notest` in CI.

.github/workflows/ci.yml
atugushev: Could you send a diff with updated cpython job wich has 2 additional pypy jobs described below ?

- ubuntu + piplatest + pypy2
- ubuntu + piplatest + pypy3

Let see how it would look like.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+          - Windows
+        python-version:
+          - 2.7
+          - 3.5
+          - 3.6
+          - 3.7
+          - 3.8
+        pip-version:
+          - "20.0"
+          - "latest"
```
webknjaz, line 49: The exclusions would be
```suggestion
          - "latest"
        exclude:
        - os: Windows
          python-version: pypy
        - os: Windows
          python-version: pypy3
        - pip-version: "20.0"
          python-version: pypy
        - pip-version: "20.0"
          python-version: pypy3
```
and this'll allow to consolidate jobs

.github/workflows/ci.yml
webknjaz: Done: https://github.com/jazzband/pip-tools/pull/1012#discussion_r419004600

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  qa:
```
webknjaz, line 10: I usually put the linters matrix into a separate file, this allows having a separate Checks section, not mixing up with the actual tests.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+          python-version: ${{ matrix.python-version }}
+      - name: Install tox
+        run: pip install tox
+      - name: Test pip ${{ matrix.pip-version }}
+        run: tox -e pip${{ matrix.pip-version }}-coverage
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1.0.6
+        with:
+          file: ./coverage.xml
+          name: ${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.pip-version }}
```
webknjaz, line 64: In this context, there's also a `runner.os` which stands for `Linux`/`Windows`/`macOS` rather than your factors. Maybe use it instead?
```suggestion
          name: ${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.pip-version }}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
+      - name: Install tox
+        run: pip install tox
+      - name: Test ${{ matrix.toxenv }}
+        run: tox -e ${{ matrix.toxenv }}
+
+  cpython:
+    name: Test
```
webknjaz, line 34:
```suggestion
    name: ${{ matrix.os }}/${{ matrix.python-version }}/${{ matrix.pip-version }}
```

.github/workflows/ci.yml
atugushev: Yeah, this is much better. Thanks!

.github/workflows/ci.yml
atugushev: image

Don't you like this notation?

.github/workflows/ci.yml
webknjaz: The key thing that is not very well documented/understood is that you don't have to list all the factors in your exclusions which allows you to have fewer lines of them.

.github/workflows/ci.yml
atugushev: Could you elaborate on that? What's the benefit?

.github/workflows/ci.yml
webknjaz: Unfortunately, the problem with job names in GitHub Actions is that some places where it's rendered are very narrow so job names must be short to fit.

.github/workflows/ci.yml
atugushev: Seems reasonable.

.github/workflows/ci.yml
webknjaz: Screen Shot 2020-05-02 at 10 44 40 PM

.github/workflows/ci.yml
webknjaz: Screen Shot 2020-05-02 at 10 44 40 PM

.github/workflows/ci.yml
atugushev: But what's wrong with `Windows-3.8-20.0`? Do you suggest `windows-latest-3.8-20.0` instead?

.github/workflows/ci.yml
webknjaz: I often use emojis and short forms trying to fight this.

.github/workflows/ci.yml
atugushev: pip-tools don't run parallel jobs with tox. What's the point with preparing testenv in a separate step?

.github/workflows/ci.yml
webknjaz: https://github.com/ansible/ansible-lint/runs/638612156:
Screen Shot 2020-05-02 at 10 47 24 PM

webknjaz: When there's ~100 jobs, it's way easier when at least part of them are filtered out

.github/workflows/ci.yml
webknjaz: No, it'd be just `Windows`. But it'll use the actual runner OS, not the name of _your_ matrix factor.

webknjaz: So it'll not depend on whether you rename things in the matrix

.github/workflows/ci.yml
atugushev: Makes sense, let's move to a separate workflow then.

.github/workflows/ci.yml
atugushev: Could you attach the full screen of the page?

.github/workflows/ci.yml
atugushev: Got it, good point.

.github/workflows/ci.yml
webknjaz: It's not about parallelism. It's about separating the virtualenv creation and deps installation from the actual tests run. First of all, it's easier to consume when you look at the separate steps and you can measure the time separately. But also, there were tox versions where the preparation step produced broken envs and still attempted to run tests which was awfully painful to debug. In some other CIs you they also have separate step names for install and script stages which fail and error out differently.

.github/workflows/ci.yml
atugushev: Yeah, Include and exclude syntax are too unclear.

.github/workflows/ci.yml
webknjaz: This is a well-recognized technique nowadays: https://github.com/tox-dev/tox/issues/829#issuecomment-392290825 / https://github.com/tox-dev/tox/issues/1504#issuecomment-616083656 / https://github.com/tox-dev/azure-pipelines-template/blob/master/run-tox-env.yml#L133

atugushev: I see, good call.

.github/workflows/ci.yml
webknjaz: I understand where this is coming from but no, it's not because of narrow window size, there's no responsiveness — the width is fixed.
It's here, see for yourself: https://github.com/tox-dev/azure-pipelines-template/blob/master/run-tox-env.yml#L133 (scroll down the list).
Screen Shot 2020-05-02 at 11 02 17 PM

.github/workflows/ci.yml
atugushev: image

Isn't it specific?

.github/workflows/ci.yml
atugushev: Wrong link I guess, but I get that. Yeah, that page pretty fixed.

.github/workflows/ci.yml
atugushev: Could you elaborate on why there is `--parallel-live` option? Also, it's not clear why there is `-p auto`, since we run only a single envlist per job.

.github/workflows/ci.yml
atugushev: @asottile I wonder what do you think?

.github/workflows/ci.yml
atugushev: > I think it's cleaner to use env everywhere. Besides, you still need to declare other env vars.

Could you suggest env solution for `cpython` job?

.github/workflows/ci.yml
webknjaz: oh, just power of habit

.github/workflows/ci.yml
webknjaz: sure

.github/workflows/ci.yml
webknjaz: Also, for better or for worse, separate workflow failures won't prevent tests from running, which is rather good IMHO.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+        python-version:
+          - 2.7
+          - 3.5
+          - 3.6
+          - 3.7
+          - 3.8
+        pip-version:
+          - "20.0"
+          - "latest"
+    steps:
```
webknjaz, line 50:
```suggestion

    env:
      PY_COLORS: 1
      TOXENV: ${{ matrix.toxenv }}
      TOX_PARALLEL_NO_SPINNER: 1

    steps:
```

.github/workflows/ci.yml
atugushev: That depends on the POV, but I think it's for good too.

.github/workflows/ci.yml
webknjaz: Also injected pypy there ^

.github/workflows/ci.yml
asottile: if it works send a patch 🤷 -- I don't use the github action all that much myself

atugushev: Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Thanks! Looks good.  Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: I'll leave those params then in case we decide to run something in parallel. Thanks!  Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Thanks for the suggestion! Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Split up into ci.yml and qa.yaml in 988dfbc.

.github/workflows/ci.yml
atugushev: Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Thanks! Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Thanks!  Addressed in 988dfbc.

.github/workflows/ci.yml
atugushev: Actually I've reconsidered and put matrix.python-version = 3.x explicitly in 988dfbc. Thanks!

.github/workflows/ci.yml
atugushev: I've put just toxenv to the name in 988dfbc. Looks good IMO.

atugushev:
Hey @webknjaz. Thank you so much for such huge feedback! Much appreciated! ❤️ I've resolved all the discussions (I hope, feel free to unresolve otherwise) and I'm ready for another round of review! 🚀

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+    tags:
```
webknjaz, line 8: This is minor but you may want to also enable cron:
```suggestion
    tags:
  schedule:
  # https://crontab.guru/#53_3_*_*_*
  - cron: 53 3 * * *  # At minute 53 past 3am every day.
```

atugushev: Yeah, why not? Should QA job also be scheduled?

webknjaz: It could. But I guess linters are less important than legit tests.

webknjaz:
> Looks good. I am not sure I need tox in a GitHub actions world, but that is certainly a question of preference.

It's needed with any CI because it works like a glue between different platforms, plus it allows running tests locally exactly the same way. It's a well-established practice to use it like this.

.github/workflows/qa.yml
```diff
@@ -0,0 +1,43 @@
+name: QA
```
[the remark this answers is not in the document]
webknjaz: It was like this originally but I recommended this to be separate because linters shouldn't block tests when there's a lot of free parallel resources. Plus when it's ~100 jobs it's way easier when at least part of them are filtered out into a separate section.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: pip install tox
+      - name: Prepare test environment
+        run: tox --notest -p auto --parallel-live
+      - name: Test pip ${{ matrix.pip-version }}
+        run: tox
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1.0.6
+        with:
+          file: ./coverage.xml
+          name: ${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.pip-version }}
```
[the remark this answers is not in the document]
webknjaz: Does python package work without a token? The action can do this. Also, it's better to have an officially maintained piece of the puzzle that is designed for this ecosystem instead of reinventing the wheel.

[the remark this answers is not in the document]
codingjoe: @webknjaz hm… both versions are maintained by Codecov, so I don't see a big problem in that regard. However, you do need a token. I am just saying its fast, as in a lot faster. If you don't care waiting for CI this should be fine.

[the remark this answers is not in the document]
webknjaz: Weird. It shouldn't be slower because it's a JS action, not docker.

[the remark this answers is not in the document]
webknjaz: Oh, have you tried it after Nov 19? This is when they switched from Docker to JS.

.github/workflows/qa.yml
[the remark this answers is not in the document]
codingjoe: I understand your point, I am just not sure I share your opinion here. In any case, this is a matter of opinion or taste at the end. Therefore, I am OK either way.

.github/workflows/ci.yml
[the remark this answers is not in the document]
webknjaz: > both versions are maintained by Codecov

But maintained here I meant "and designed for use within the Actions ecosystem"

webknjaz:
@atugushev maybe on the next iteration we could make the CI actually test the dists too, not just source. And add macOS into the matrix. But right now it seems ready to go. Done is better than perfect.

.github/workflows/qa.yml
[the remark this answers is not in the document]
atugushev: @codingjoe 

> I believe this should not be a separate workflow but rather a job within a single workflow.

Personally I'm ambivalent about whether it would be in a single workflow or in a separate as long as there are not so many jobs.


> So I'd add `needs: qa` to the tests' workflow.

I'd prefer not to add `needs` as it slows down the parallel jobs since all 22 test jobs would wait at least 35 seconds until QA jobs are done.

.github/workflows/ci.yml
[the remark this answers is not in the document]
atugushev: Screenshot 2020-05-04 at 16 15 31

Four seconds to get job done - it's pretty good in my opinion.

atugushev:
> maybe on the next iteration we could make the CI actually test the dists too, not just source. And add macOS into the matrix. But right now it seems ready to go. Done is better than perfect.

What do you mean by "test the dists"?

webknjaz:
> What do you mean by "test the dists"?

When you run tox, it works with the source in the project dir and even creates some temporary dists. But they are not what you're going to publish to PyPI hence you test not exactly the same as users will get. [Also, you may be testing not what's installed for various reasons.](https://blog.ganssle.io/articles/2019/08/test-as-installed.html)
Better way would be to divide this process into stages:
```
  (1) build dists -> (2) test dists from [1] -> (3) upload *exactly* those dists from [1] that have been tested
```

atugushev:
@webknjaz 

Good suggestion! There is an issue #1134 to track.

.github/workflows/ci.yml
atugushev: Addressed in 790872f.

atugushev:
@webknjaz 

I've added MacOS to the matrix.

atugushev:
image

`pypy` test jobs are so slow 😞 By the way `pip` doesn't run tests on `pypy`, see https://github.com/pypa/pip/issues/7279#issuecomment-612911682. It's tempting to turn them off...

webknjaz:
You could add a condition to only run them on cron and push, for example

.github/workflows/cron.yml
```diff
@@ -1,28 +1,73 @@
-name: cron
+name: Cron
```
webknjaz, line 1: Oh, I didn't think you'd copy all the workflow into another file. I meant we could inject a small if-clause on the job level to filter out things.

atugushev: Could you post a suggestion?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
...
+    branches:
+      - master
+    tags:
+  schedule:
+    # Run everyday at 03:53 UTC
+    - cron: 53 3 * * *
+
+jobs:
+  test:
+    name: ${{ matrix.os }} / ${{ matrix.python-version }} / ${{ matrix.pip-version }}
```
webknjaz, line 15:
```suggestion
    if: >-
      !startsWith(matrix.python-version, 'pypy') ||
      github.event_name != 'pull_request'
```

.github/workflows/cron.yml
webknjaz: Something like this should do https://github.com/jazzband/pip-tools/pull/1012/files#r419714725

.github/workflows/ci.yml
atugushev: Hm... In that case `test` job is not run.

webknjaz: Why? It should only exclude matching jobs.

webknjaz: I may have made a mistake in the condition, though.

webknjaz: No, it looks correct

atugushev: Should the condition be wrapped with `${{ ... }}`?

webknjaz: No, that's only needed where you interpolate values.

webknjaz: https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions:
> When you use expressions in an `if` conditional, you may omit the expression syntax (`${{ }}`) because GitHub automatically evaluates the `if` conditional as an expression.

atugushev: Looks like the `matrix` parameters can't be used inside `if` at a job level:

Screenshot 2020-05-05 at 04 26 18

atugushev: See also https://github.community/t5/GitHub-Actions/Matrix-cannot-be-used-in-jobs-level-if/td-p/41048

webknjaz: That's because you have to apply my suggestion exactly like it's submitted. But you added a leading `!` there w/o quoting. This works only if you wrap the whole value with quotes or use a multiline form as I originally suggested.

webknjaz: It's not a context access problem but a YAML parsing one.

webknjaz: Ah, I see. Maybe you're right but that post is old, have you checked if it works with the proper syntax?

atugushev: Screenshot 2020-05-05 at 04 35 23

Nope, doesn't work.

webknjaz: :(

.github/workflows/cron.yml
atugushev: I'm okay with the copy approach for a while. As soon as GitHub provides an opportunity to make matrix conditionals at a job level we'll refactor it.

webknjaz: Fair enough.

webknjaz: Well, there *is* a crazy way to implement this: https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#fromjson 🤪

webknjaz:
@codingjoe do you still want to block this?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
+name: CI
+
+on:
+  pull_request:
```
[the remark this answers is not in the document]
webknjaz: No, it has to be a mapping since we have settings under some of them

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+    name: QA
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        toxenv:
+          - checkqa
+          - readme
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python ${{ matrix.python-version }}
```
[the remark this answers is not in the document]
atugushev: It's fixed in https://github.com/jazzband/pip-tools/pull/1012/commits/988dfbcc41f81e28a3fe158dc61fc4b12dea4444#diff-dea0784c97e61f16e6ca1d3926d39c62R19-R20

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+
+  pypy:
+    name: Test
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      matrix:
+        os:
+          - Ubuntu
+          # TODO uncomment after test_realistic_complex_sub_dependencies being fixed
+          # - Windpws
```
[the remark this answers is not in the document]
atugushev: Huh 😄 Fixed in https://github.com/jazzband/pip-tools/commit/988dfbcc41f81e28a3fe158dc61fc4b12dea4444#diff-dea0784c97e61f16e6ca1d3926d39c62R19-R20

.github/workflows/cron.yml
atugushev: Wow 😄

atugushev:
Seems like it's ready to go! 🚀

atugushev:
I'm so happy we've finally done this. Huge thanks to the reviewers! ❤️

.github/workflows/ci.yml
atugushev: > if it works send a patch 🤷 -- I don't use the github action all that much myself

Addressed in https://github.com/pre-commit/pre-commit.com/pull/432

PR #1013, opened 2019-11-28
Refactor test_generate_hashes_without_interfering_with_each_other.

### LINKED ISSUE #1002: Refactor test_generate_hashes_without_interfering_with_each_other
This test checks a very special case with `cffi` and `maptolib`, see 877241d14d84a97801e98f4be473e5c63de5e717. Could we leave it as is and refactor it in the following PR?

_Originally posted by @atugushev in https://github.com/jazzband/pip-tools/pull/998#discussion_r349934456_

---

I think we should describe in detail what exactly this test checks, since it's not obvious what could go wrong with `PyPIRepository.get_hashes()` and add links to #512 and #514 issues (I've spent too much time to track these issues). For example:
```python
    """
    The PyPIRepository._get_file_hash() used to call unpack_url(),
    when generating the hash. Unpacking both packages to the same directory
    will then fail. E.g. matplotlib-2.0.2.tar.gz has a directory named LICENSE,
    but many other packages have a file named LICENSE.

    See GH-512 and GH-544.
    """"
```

and assert the result of `get_hashes()`.

hramezani:
Fixes #1002.
Thanks @atugushev for description.

**Changelog-friendly one-liner**: Refactor `test_generate_hashes_without_interfering_with_each_other`.

##### Contributor checklist

### DIFF
--- a/tests/test_repository_pypi.py
+++ b/tests/test_repository_pypi.py
@@ -58,9 +58,7 @@ def test_generate_hashes_all_platforms(from_line, pypi_repository):
 
 
 @pytest.mark.network
-def test_generate_hashes_without_interfering_with_each_other(
-    from_line, pypi_repository
-):
+def test_get_file_hash_without_interfering_with_each_other(from_line, pypi_repository):
     """
     The PyPIRepository._get_file_hash() used to call unpack_url(),
     when generating the hash. Unpacking both packages to the same directory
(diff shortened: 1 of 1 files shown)


hramezani:
@atugushev , Should I use `with pypi_repository.allow_all_wheels()` for this test as well?

hramezani:
@atugushev Thanks for your suggestion.

tests/test_repository_pypi.py
```diff
@@ -61,8 +61,35 @@ def test_generate_hashes_all_platforms(from_line, pypi_repository):
 def test_generate_hashes_without_interfering_with_each_other(
```
atugushev: Probably we should also rename the test name.

hramezani: what about `test_get_file_hash_without_interfering_with_each_other`?

atugushev: yeah, looks good 👍

PR #1022, opened 2019-12-24
Add --cache-dir option to pip-compile

richafrank:
Ensure tests pass without dependencies cached from previous tests. 

Running tests individually, I was seeing intermittent failures depending on whether the cache was populated or not, specifically `test_stdin` and `test_filter_pip_markers`. Also I refactored a test to use an existing fixture.

**Changelog-friendly one-liner**: Add `--cache-dir` option to `pip-compile`.

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -176,10 +176,10 @@ pip_defaults = install_command.parser.get_default_values()
 )
 @click.option(
     "--cache-dir",
-    help="Specify a directory to cache dependency information (defaults to {})".format(
-        CACHE_DIR
-    ),
+    help="Store the cache data in DIRECTORY.",
+    default=CACHE_DIR,
     envvar="PIP_TOOLS_CACHE_DIR",
+    show_default=True,
     show_envvar=True,
     type=click.Path(file_okay=False, writable=True),
 )
@@ -359,9 +359,6 @@ def cli(
         log.debug("Configuration:")
         for find_link in dedup(repository.finder.find_links):
             log.debug("  -f {}".format(find_link))
-
-    if cache_dir is None:
-        cache_dir = CACHE_DIR
 
     try:
         resolver = Resolver(
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -9,7 +9,9 @@ from click.utils import safecall
 
 from .. import click
 from .._compat import install_req_from_line, parse_requirements
+from ..cache import DependencyCache
 from ..exceptions import PipToolsError
+from ..locations import CACHE_DIR
 from ..logging import log
 from ..repositories import LocalRequirementsRepository, PyPIRepository
 from ..resolver import Resolver
@@ -172,6 +174,15 @@ pip_defaults = install_command.parser.get_default_values()
     default=True,
     help="Add the find-links option to generated file",
 )
+@click.option(
+    "--cache-dir",
+    help="Specify a directory to cache dependency information (defaults to {})".format(
+        CACHE_DIR
+    ),
+    envvar="PIP_TOOLS_CACHE_DIR",
+    show_envvar=True,
+    type=click.Path(file_okay=False, writable=True),
+)
 def cli(
     ctx,
     verbose,
@@ -198,6 +209,7 @@ def cli(
     max_rounds,
     build_isolation,
     emit_find_links,
+    cache_dir,
 ):
     """Compiles requirements.txt from requirements.in specs."""
     log.verbosity = verbose - quiet
@@ -348,11 +360,15 @@ def cli(
         for find_link in dedup(repository.finder.find_links):
             log.debug("  -f {}".format(find_link))
 
+    if cache_dir is None:
+        cache_dir = CACHE_DIR
+
     try:
         resolver = Resolver(
             constraints,
             repository,
             prereleases=repository.finder.allow_all_prereleases or pre,
+            cache=DependencyCache(cache_dir),
             clear_caches=rebuild,
             allow_unsafe=allow_unsafe,
         )
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -272,7 +272,9 @@ def cli(
         for host in trusted_host:
             pip_args.extend(["--trusted-host", host])
 
-    repository = PyPIRepository(pip_args, build_isolation=build_isolation)
+    repository = PyPIRepository(
+        pip_args, build_isolation=build_isolation, cache_dir=cache_dir
+    )
 
     # Parse all constraints coming from --upgrade-package/-P
     upgrade_reqs_gen = (install_req_from_line(pkg) for pkg in upgrade_packages)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -124,8 +124,11 @@ def repository():
 
 
 @fixture
-def pypi_repository():
-    return PyPIRepository(["--index-url", PyPIRepository.DEFAULT_INDEX_URL])
+def pypi_repository(tmpdir):
+    return PyPIRepository(
+        ["--index-url", PyPIRepository.DEFAULT_INDEX_URL],
+        cache_dir=str(tmpdir / "pypi-repo"),
+    )
 
 
 @fixture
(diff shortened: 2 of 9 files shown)


richafrank:
Thanks @atugushev . I can definitely make it more explicit!

Since this is manifesting in the cli tests, my recommendation would be to add a `--cache-dir` cli arg to `pip-compile` that falls back to an env var, as you suggested, followed by the current `CACHE_DIR`, and then change the parameter to `Resolver` from a cache instance to the cache dir to pass through. The only consumers of the `cache` parameter right now are the tests, which only configure the `cache_dir` anyway. For isolation in the tests then, we could have the `runner` fixture provide a `CliRunner` that uses a `tmpdir` for `--cache-dir`. 

Adding a cli arg in addition to the env var would mean more transparency and give more flexibility, allowing tests to run in parallel in the same process even. But that's more to change, so curious your thoughts.

richafrank:
(I updated this branch with my suggestion from above, if that sounds good.)

richafrank:
Unfortunately, as long as we have the default cache directory, one could easily add a test that relies on it unintentionally. That would be a good reason to add back the autouse `clear_cache` fixture as well.

piptools/scripts/compile.py
```diff
@@ -354,6 +363,7 @@ def cli(
             repository,
             prereleases=repository.finder.allow_all_prereleases or pre,
             clear_caches=rebuild,
+            cache_dir=cache_dir,
```
atugushev:
```suggestion
            cache=DependencyCache(cache_dir),
```
I'd prefer this way, the benefit is that we can use different cache backends (mocked in tests for example)

tests/conftest.py
```diff
@@ -129,21 +128,21 @@ def pypi_repository():
...
+    return str(tmpdir / "dep-cache")
 
 
 @fixture
-def resolver(depcache, repository):
+def resolver(depcache_dir, repository):
     # TODO: It'd be nicer if Resolver instance could be set up and then
     #       use .resolve(...) on the specset, instead of passing it to
     #       the constructor like this (it's not reusable)
-    return partial(Resolver, repository=repository, cache=depcache)
```
atugushev: Based on the above comment this file can be reverted.

tests/test_cli_compile.py
```diff
@@ -1,42 +1,62 @@
...
 from .utils import invoke
 
 from piptools._compat.pip_compat import PIP_VERSION, path_to_url
+from piptools.cache import DependencyCache
 from piptools.scripts.compile import cli
 
 
-def test_default_pip_conf_read(pip_with_index_conf, runner):
+@pytest.fixture
+def invoke_cli(runner, depcache_dir):
```
atugushev: instead of `invoke_cli` we can use the following fixture:
```python
@pytest.fixture(autouse=True)
def temp_dep_cache(tmpdir, monkeypatch):
    monkeypatch.setenv('PIP_TOOLS_CACHE_DIR', str(tmpdir / 'cache'))
```

so we don't have to touch all the below. lines. What do you think?

tests/test_cli_compile.py
```diff
@@ -572,38 +594,42 @@ def test_default_index_url(pip_with_index_conf):
...
 
-def test_stdin(runner):
+def test_stdin(invoke_cli, from_line, depcache_dir):
     """
     Test compile requirements from STDIN.
     """
-    out = runner.invoke(
-        cli, ["-", "--output-file", "requirements.txt", "-n"], input="small-fake-a==0.1"
+    dep_cache = DependencyCache(depcache_dir)
+    dep_cache[from_line("small-fake-a==0.1")] = []
```
atugushev: Better to use `pip_conf`fixture. This allows you to install test wheels.

piptools/scripts/compile.py
```diff
@@ -172,6 +173,13 @@
     default=True,
     help="Add the find-links option to generated file",
 )
+@click.option(
+    "--cache-dir",
+    default=lambda: os.environ.get("PIP_TOOLS_CACHE_DIR", CACHE_DIR),
+    show_default="PIP_TOOLS_CACHE_DIR env var or {}".format(CACHE_DIR),
+    help="Specify a directory to cache dependency information.",
+    type=click.Path(file_okay=False, writable=True),
+)
```
atugushev: How do you like this option?

```python
@click.option(
    "--cache-dir",
    help="Specify a directory to cache dependency information (defaults to {})".format(CACHE_DIR),
    envvar="PIP_TOOLS_CACHE_DIR",
    show_envvar=True,
    type=click.Path(file_okay=False, writable=True),
)
```

and below:

```python
if cache_dir is None:
    cache_dir = CACHE_DIR
resolver = Resolver(
   ...
   cache=DependencyCache(cache_dir)
   ...
)
```

richafrank:
Thanks @atugushev . Sounds like I went a bit too explicit! Your suggestions look good to me.

tests/test_cli_compile.py
richafrank: Ah, I hadn't understood the purpose of that fixture. That's great.

tests/test_cli_compile.py
```diff
@@ -592,7 +596,7 @@ def test_not_specified_input_file(runner):
     assert out.exit_code == 2
 
 
-def test_stdin(runner):
+def test_stdin(pip_conf, runner, from_line):
```
```suggestion
def test_stdin(pip_conf, runner):
```

`from_line` could be deleted since it's not used in this test anymore

tests/test_cli_compile.py
```diff
@@ -519,7 +523,7 @@ def test_generate_hashes_verbose(pip_conf, runner):
 
 
 @pytest.mark.skipif(PIP_VERSION < (9,), reason="needs pip 9 or greater")
-def test_filter_pip_markers(runner):
+def test_filter_pip_markers(pip_conf, runner, from_line):
```
```suggestion
def test_filter_pip_markers(pip_conf, runner):
```
`from_line` could be deleted since it's not used in this test anymore

tests/test_cli_compile.py
```diff
@@ -69,26 +73,26 @@ def test_command_line_setuptools_read(pip_conf, runner):
...
+    package.write(
+        dedent(
+            """\
+            from setuptools import setup
+            setup(install_requires=[])
+            """
         )
-        package.close()
+    )
+    package.close()
```
atugushev: Thanks for refactoring this test! Just one note: could you also refactor-out `package = open(...)` thing? For example:

```python
with open(...) as package:
   ...
```

richafrank:
👍 on the cleanup

piptools/scripts/compile.py
```diff
@@ -348,11 +360,15 @@ def cli(
         for find_link in dedup(repository.finder.find_links):
             log.debug("  -f {}".format(find_link))
 
+    if cache_dir is None:
+        cache_dir = CACHE_DIR
```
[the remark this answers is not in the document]
atugushev: I think we can remove this branch in favor of:
```python
@click.option(
    "--cache-dir",
    help="Store the cache data in DIRECTORY",
    default=CACHE_DIR,
    envvar="PIP_TOOLS_CACHE_DIR",
    show_default=True,
    show_envvar=True,
    type=click.Path(file_okay=False, writable=True),
)
```

piptools/scripts/compile.py
```diff
@@ -172,6 +174,15 @@
     default=True,
     help="Add the find-links option to generated file",
 )
+@click.option(
+    "--cache-dir",
+    help="Specify a directory to cache dependency information (defaults to {})".format(
```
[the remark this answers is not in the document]
atugushev: See also https://github.com/jazzband/pip-tools/pull/1022#discussion_r361790087.

piptools/scripts/compile.py
```diff
@@ -198,6 +209,7 @@ def cli(
     max_rounds,
     build_isolation,
     emit_find_links,
+    cache_dir,
```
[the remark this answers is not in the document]
atugushev: It should affect `test_locally_available_editable_package_is_not_archived_in_cache_dir`.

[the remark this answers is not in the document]
richafrank: Good catch. I should have looked into `PyPIRepository` when that import needed updating.

piptools/scripts/compile.py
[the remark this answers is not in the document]
richafrank: Ah, nice - click does it all!

atugushev:
@richafrank thanks for the contribution!

richafrank:
Thanks @atugushev !

atugushev:
`pip-tools v4.4.0` is released!

PR #1026, opened 2019-12-30
Use str.format instead of printf-style formatting

atugushev:
The official Python 3 documentation doesn't recommend printf-style string formatting.
See https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -159,8 +159,8 @@ class Resolver(object):
                 raise RuntimeError(
                     "No stable configuration of concrete packages "
                     "could be found for the given constraints after "
-                    "%d rounds of resolving.\n"
-                    "This is likely a bug." % max_rounds
+                    "{max_rounds} rounds of resolving.\n"
+                    "This is likely a bug.".format(max_rounds=max_rounds)
                 )
 
             log.debug("")


duoi:
I appreciate the idea, but this doesn't particularly resolve the spirit of the issue that the docs you linked raise. The appropriate fix here (to follow the recommendation of the docs) is to have named substitution, i.e

```
"{max_rounds} rounds of resolving.\n"
"This is likely a bug.".format(
    max_rounds=max_rounds
)
```

atugushev:
@duoi both variants look good to me :) Could you point to the recommendation to use `{name}` instead of `{}`?

vphilippon:
Without having documented recommendation, I do have a preference for named arguments with `str.format`. I feel it's generally easier to figure out what goes where. It doesn't make much difference here, so it's up to you at this point.

So I guess that means: recommendation source = me 😄

atugushev:
I presume It's subjective, that's why there're no PEP or guides. Usually, I prefer `{}` if there is a single param in `format(foo)` (mostly to keep strings thin), as well as the codebase though:

```console
$ git grep '{}' | wc -l
      72
```

vs

```console
$ git grep  '{\w\+}' | wc -l
      11
```

I don't want to waste energy on subjective things, so I'll just fix the code. Anyways, later there will be `f""` strings and there will be `f"...{max_rounds}..."`, so your recommendations would help :)

PR #1029, opened 2020-01-09
Fix a bug where pip-compile would keep outdated options

### LINKED ISSUE #1018: duplicate --find-links
I've managed to get 2 `--find-links` in my requirements.txt when I altered the single value in requirements.in . The first is the one before the change and the second is after.

#### Environment Versions

1. OS Type: MacOS 10.13
1. Python version: 3.7.5
1. pip version:  19.3.1
1. pip-tools version: 4.3.0

#### Steps to replicate

1. add `--find-links file:./local_src` in requirements.in
1. `pip-compile requirements.in`
1. change entry in 1 to `--find-links ./local_src`
1. `pip-compile requirements.in`

#### Expected result
(I would actually expect the result to remain a relative path, but that's #204)
```
--find-links /Users/me/myrepo/./local_src
```

#### Actual result

```
--find-links file:./local_src
--find-links /Users/me/myrepo/./local_src
```

atugushev:
Fixes #1018.

Consider a test case:

```console
$ cat requirements.in
--find-links foo

$ cat requirements.txt
--find-links bar
```
The bug was that pip-compile would keep both options from requirements.in and requirements.txt:

```console
$ pip-compile --no-header
--find-links foo
--find-links bar
```

It should keep the options from requirements.in file:

```console
$ pip-compile --no-header
--find-links foo
```

The fix is to pass a temporary finder, session, and options instances to `parse_requirements()` when parsing an existing requirements.txt.

**Changelog-friendly one-liner**: Fix a bug where `pip-compile` would keep outdated options from `requirements.txt`.

##### Contributor checklist


atugushev:
@vphilippon thanks for reviewing this!

PR #1031, opened 2020-01-17
pip-compile -P PKGSPEC ignores pkgs not already parsed as constraints

### LINKED ISSUE #759: Pip-compile upgrades a package even it doesn't exist in a requrements.*
Pip-compile should raise an error while trying to upgrade a package which doesn't exist in a requirements.in.

##### Environment Versions

1. OS Type: any
1. Python version: any
1. pip version:  any
1. pip-tools version: any

##### Steps to replicate

1. echo "click" > requirements.in
2. pip-compile
3. pip-compile -P django

##### Expected result

Raise a error/warning or skip non-existing package.

##### Actual result

```
#
# To update, run:
#
#    pip-compile -P pytz
#
click==7.0
django==1.11.20
pytz==2018.9
```

AndydeCleyre:
Fixes #759

`pip-compile -P <pkgspec>`

When adding requested-to-upgrade pkgspecs (`ireq`s) to the `constraints` list, omit those whose key is not present among the `primary_packages` set.

**Changelog-friendly one-liner**: `pip-compile --upgrade-package` silently ignores those passed packages not already required according to the *.in and *.txt files.

##### Contributor checklist


AndydeCleyre:
Thanks!

I didn't think about user-managed versions of requirements not entered into the `.in` file. Some options:

a) alter the resolver to handle an additional constraint variant which limits only if needed
b) run the resolver twice in the case that `upgrade_install_reqs` isn't empty, the second time without those constraints added
c) only support `-P` for reqs.in entries

atugushev:
@AndydeCleyre 

d) this also can be fixed by:

```python
if not upgrade and os.path.exists(output_file.name):
    ...
    existing_pins_to_upgrade = {
        key_from_req(ireq.req): ireq
        for ireq in ireqs
        if is_pinned_requirement(ireq)
        and key_from_req(ireq.req) in upgrade_install_reqs
    }
...
constraints.extend(
    ireq for key, ireq in upgrade_install_reqs.items()
    if key in primary_packages or key in existing_pins_to_upgrade
)
```

AndydeCleyre:
Alright, so after your last comment I implemented your fix only slightly modified by populating `existing_pins_to_upgrade` and `existing_pins` during a single iteration through `ireqs`.

But it drew my attention to the fact that I wasn't yet familiar with `LocalRequirementsRepository`, and then I thought the distinction between pinned and to-upgrade was unnecessary, so that a better solution would be to stop separating them, and just make sure to update the existing pins with the to-upgrade reqs (and never directly add `upgrade_install_reqs` to the `constraints`:

```python
existing_pins = {
    key_from_req(ireq.req): ireq
    for ireq in ireqs
    if is_pinned_requirement(ireq)
}
existing_pins.update(upgrade_install_reqs)
```

At first I thought it was working, but then realized that it breaks when using `-P` with specifiers other than `==`, as `<`, `<=`, etc., aren't "pinned." And converting one of those to a true pin isn't so simple, as it entails using the resolver and accessing the real repository.

So I've replaced this PR content with the modified version of your last-suggested solution.

I tried to add a little extra to `test_upgrade_packages_version_option` to cover [this](https://github.com/jazzband/pip-tools/pull/1031#pullrequestreview-344925098), but I don't sufficiently understand the way the test_data fake packages with deps are handled to get it right. If the fix looks good, can you add a test for that case?

I still keep in my head as a fallback for "correct" (if inefficient) outcome: in the case of upgrades, resolve/pin/compile once with all constraints, and again without the to-upgrade ones.

AndydeCleyre:
Ah, the star unpacking isn't py2 compatible, I'll update shortly.

atugushev:
@AndydeCleyre 

Thanks for the updates! I've skimmed the PR but review it properly soon.

Here is the PR https://github.com/AndydeCleyre/pip-tools/pull/1 with tests. Also, I've added a test for downgrade deps.

tests/test_cli_compile.py
```diff
@@ -370,6 +370,22 @@ def test_upgrade_packages_option(pip_conf, runner):
...
+def test_upgrade_packages_option_irrelevant(pip_conf, runner):
+    """
+    piptools ignores --upgrade-package/-P items not already constrained.
+    """
+    with open("requirements.in", "w") as req_in:
+        req_in.write("small-fake-a")
+    with open("requirements.txt", "w") as req_in:
+        req_in.write("small-fake-a==0.1")
+
+    out = runner.invoke(cli, ["-P", "small-fake-b"])
```
```suggestion
    out = runner.invoke(cli, ["--upgrade-package", "small-fake-b"])
```

minor: long-opts are more readable

tests/test_cli_compile.py
```diff
@@ -370,6 +370,22 @@ def test_upgrade_packages_option(pip_conf, runner):
...
+
+    assert out.exit_code == 0
+    assert "small-fake-a==0.1" in out.stderr
```
```suggestion
    assert "small-fake-a==0.1" in out.stderr.splitlines()
```

minor: `"small-fake-a==0.1" in out.stderr` could be clashed with `small-fake-a==0.1.1`. Works with `--no-annotate`.

piptools/scripts/compile.py
```diff
@@ -293,13 +295,15 @@ def cli(
...
-        }
+        # constraints, and separately gather pins to be upgraded
+        existing_pins = {}
+        for ireq in filter(is_pinned_requirement, ireqs):
+            key = key_from_req(ireq.req)
+            (
+                existing_pins_to_upgrade
+                if key in upgrade_install_reqs
+                else existing_pins
+            )[key] = ireq
```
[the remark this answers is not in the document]
AndydeCleyre: Yup, changing it, thanks.

piptools/scripts/compile.py
```diff
@@ -282,6 +282,8 @@ def cli(
         key_from_req(install_req.req): install_req for install_req in upgrade_reqs_gen
     }
 
+    existing_pins_to_upgrade = {}
```
atugushev:
```suggestion
    existing_pins_to_upgrade = set()
```

Since the values ​​aren't used a `set ()` could be more suitable. What do you think?

AndydeCleyre:
You're totally right about it naturally being a set rather than a dict, that was just a holdover from trying to more tightly combine the population of `existing_pins` and `existing_pins_to_upgrade`, which we scrapped anyway.

Thank you for your continued work on this project, and ensuring I don't muck it up.

atugushev:
@AndydeCleyre thanks!

atugushev:
`pip-tools v4.4.0` is released!

PR #1032, opened 2020-01-18
Add full test coverage for the exceptions module

atugushev:


atugushev:
@vphilippon thanks for reviewing this! I've fixed conflicts and rebased onto master.

Issue #908, opened 2019-09-24
Refactor parsing setup.py in favour of pip install -e .


atugushev:
#### What's the problem this feature will solve?

The current implementation of `pip-compile setup.py` doesn't handle the following use-cases and errors:

1) doesn't support extra requirements, see #625
2) doesn't support environment markers, see #1139
5) doesn't support declarative config `setup.cfg`
6) doesn't support modern `pyproject.toml` config
3) suppresses an error message when setup.py parse failed, see #572 #684
4) raises an unexpected error during parsing a valid setup.py, see #511 #695

The solution based on `-e .` became "de facto" workaround, so why can't we handle it by pip-compile itself?

#### Describe the solution you'd like

Instead of:
```python
from distutils.core import run_setup
dist = run_setup(src_file)
tmpfile.write("\n".join(dist.install_requires))
```

get dependencies using something like (consider bash onliner as a demonstration):

```bash
$ echo "-e ." | pip-compile - -qo- | sed '/^-e / d' > tmpfile
```

For example:
```bash
$ cat setup.py
from setuptools import setup
setup(
    name="mypackage",
    version="0.1",
    install_requires=['django']
)

$ echo "-e ." | pip-compile - -qo- | sed '/^-e / d'
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --output-file=- -
#
django==2.2.5
pytz==2019.2              # via django
sqlparse==0.3.0           # via django
```

#### Alternative Solutions

Use [pip's internal api](https://github.com/pypa/pip/blob/b96ee36ba9c5d7e79592eab52c653ae9a37dbcb8/src/pip/_internal/legacy_resolve.py#L364-L367) to get dependencies from `setup.py`:

```python
...
abstract_dist = pip._internal.legacy_resolver.Resolver._get_abstract_dist_for(req_to_install)

# Parse and return dependencies
dist = abstract_dist.get_pkg_resources_distribution()
...
```

But i found it challenging to maintain, especially for old pip versions.

#### Additional context

N/A

**UPDATE 2020-05-16**: The downside is that `-e .` solution loses the environment markers.

barrywhart:
I believe there is another issue where it currently doesn't handle `extras_require`. Could this solution be extended to handle that as well? I think it could, for example instead of running `pip-compile` with `-e .`, it'd run it with (for example), `-e .[extra1,extra2]`.

atugushev:
@barrywhart 

[...]

PR #1042, opened 2020-01-24
Parse environment markers from setup.py install_requires

iamhatesz:
**Changelog-friendly one-liner**: Added scanning of `extras_require` in `setup.py` for additional dependencies which might come from using environment markers in `install_requires`.

##### Contributor checklist


iamhatesz:
Hi @atugushev 

I am not sure if I understand #908 correctly. You propose the support of `extras_require` by passing `--extras` flag to `pip-compile` explicitly. This is fine when we have manually defined extras like `dev`, but I am afraid this could be a problem, when these extras are dynamically generated e.g. by specifying environment markers. In this case, I don't know what value should I use when calling `pip-compile` - maybe `:"linux" in sys_platform`, maybe `:platform_system == "Linux"`, too many cases.

atugushev:
> I am not sure if I understand #908 correctly. You propose the support of `extras_require` by passing `--extras` flag to `pip-compile` explicitly. This is fine when we have manually defined extras like `dev`, but I am afraid this could be a problem, when these extras are dynamically generated e.g. by specifying environment markers. In this case, I don't know what value should I use when calling `pip-compile` - maybe `:"linux" in sys_platform`, maybe `:platform_system == "Linux"`, too many cases.

The original propose is to use `pip. install -e .` to parse `setup.py`.  To parse deps via `pip install -e .[my-extra]` (e.g., extra markers) proposed  `--extra=my-extra` option. You are right, it won't work for environment markers, but it makes no sense since, for example, it's difficult to build Windows dependencies on a Linux platform.

iamhatesz:
Hm, I don't consider parsing Windows-specific dependencies when being on Linux. I am fine with compiling dependencies twice: once on a Windows machine, and the second time on a Linux machine. 

In my project, I have a few dependencies that can be satisfied only on Linux. Thus in my `setup.py` I use environment markers, so installing my package from `pip` works well. However, with current `pip-tools` I can't successfully compile my `setup.py` to `requirements.txt`, because it doesn't parse `extra_requires`. The solution with `--extras` flag is a workaround, but in this case I have to know what value I have to provide there. And for some cases, I can't be sure that, because some other packages could use environment markers as well, and I don't know to what exact extras key they will be transformed.

atugushev:
@iamhatesz 

> If `pip install -e .` parses extras which come from environment markers (satisfying only these appropriate for current platform), then this is exactly what I'd like to achieve.

Yes, it works for the environment markers. You can try it how it works by:

```
$ echo "-e ." | pip-compile - -qo- | sed '/^-e / d'
```

iamhatesz:
@atugushev cool, just tested, seems like a solution for me. What's the status of #908 then? Can I help you with the implementation?

atugushev:
@iamhatesz 

> cool, just tested, seems like a solution for me. What's the status of #908 then? 

It's still under discussion, you might join there and express your thoughts.

> Can I help you with the implementation?

Sure, any contributions are highly welcomed and encouraged 😉

tests/test_cli_compile.py
```diff
@@ -58,6 +58,36 @@ def test_command_line_setuptools_read(pip_conf, runner):
...
 
+def test_command_line_setuptools_with_env_markers_read(pip_conf, runner):
+    package = open("setup.py", "w")
+    package.write(
+        dedent(
+            """\
+            from setuptools import setup
+            setup(
+                install_requires=[
+                    'small-fake-a==0.1 ; "linux" in sys_platform',
```
webknjaz, line 69: I'd add some corner cases. Like empty or whitespace-only string after `;`, also multiple conditions

anthrotype:
Hi. Any updates on this? I'd love to see support for environment markers integrated in `pip-compile setup.py`. Thank you

atugushev:
I'm going to close this in favor of #1311. Thanks for the PR anyway! 🙏🏼

PR #1075, opened 2020-02-25
Strip line number annotations

### LINKED ISSUE #1074: Remove "(line X)" annotations from output

I love the new source annotations in 4.5.0 from #1058. But the line number annotations cause unnecessary diff churn. For example on a project I'm working on, I added a couple requirements that are alphabetically early, and it means *every* requirement afterwards has a new line number:

Screenshot 2020-02-25 at 15 22 19

Adding a single requirement with no dependencies can become a 100 line diff 😢

Equally sorting one's `requirements.in` would make diff noise just from changing the line numbers.
 
Remove the line numbers from the output.

#### Alternative Solutions

A flag? I prefer sensible defaults though.

adamchainz:
Fixes #1074.

**Changelog-friendly one-liner**: Strip line number annotations such as "(line 12)" from file requirements, to prevent diff noise when modifying input requirement files.

##### Contributor checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -662,7 +662,7 @@ def test_multiple_input_files_without_output_file(runner):
         (
             "--annotate",
             "small-fake-a==0.1         "
-            "# via -c constraints.txt (line 1), small-fake-with-deps\n",
+            "# via -c constraints.txt, small-fake-with-deps\n",
         ),
         ("--no-annotate", "small-fake-a==0.1\n"),
     ],
--- a/README.rst
+++ b/README.rst
@@ -114,7 +114,7 @@ Now, run ``pip-compile requirements.in``:
     #    pip-compile requirements.in
     #
     asgiref==3.2.3            # via django
-    django==3.0.3             # via -r requirements.in (line 1)
+    django==3.0.3             # via -r requirements.in
     pytz==2019.3              # via django
     sqlparse==0.3.0           # via django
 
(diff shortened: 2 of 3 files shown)


tests/test_cli_compile.py
```diff
@@ -662,7 +662,7 @@ def test_multiple_input_files_without_output_file(runner):
         (
             "--annotate",
             "small-fake-a==0.1         "
-            "# via -c constraints.txt (line 1), small-fake-with-deps\n",
+            "# via -c constraints.txt, small-fake-with-deps\n",
```
graingert: It's possible to have duplicate constraints in one file, in that case the line number would be useful for disambiguation

adamchainz: I mean, how useful? Wouldn't it be better to warn/error about duplicate constraints in a file, since they can always be combined into one line with commas?

jmbowman: I'd be fine with something that just flagged that there are duplicates from a file, not necessarily saying which lines.  Also, another drawback I've discovered with the new output is that it makes merge conflicts much more likely; almost any two distinct changes to the input requirements or constraints will change many of the line numbers to different values.

graingert: Ok maybe it's best to leave duplicates for another PR. The line number diffnoise is too much

adamchainz: Yeah I'm merging more requirements changes today and the diffnoise is slowing down review quite a bit 😱

Made #1077 for warning on duplicated constraints.

adamchainz:
Not a problem, thanks for the maintenance!

atugushev:
`pip-tools v4.5.1` is released.

PR #824, opened 2019-05-23
Support options in requirements.txt in pip-sync

atugushev:
Adds support of the following options in `requirements.txt` in `pip-sync`:

```
--index-url
--extra-index-url
--no-index
--find-links
--no-binary
--only-binary
--trusted-host
```

Closes #637 
Closes #638
Inspired by #703 #464.

**Changelog-friendly one-liner**: Support `requirements.txt` options in pip-sync

##### Contributor checklist

- [X] Provided the tests for the changes.
- [x] Requested a review from another contributor.
- [X] Gave a clear one-line description in the PR (that the maintainers can add to CHANGELOG.md on release).
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #1080, opened 2020-02-28
Pass arbitrary args to 'pip' from pip-compile and pip-sync, with '--pip-args'

### LINKED ISSUE #321: Pass additional arguments to pip

AndydeCleyre:
~Technically, pass all args after the second standalone '--' arg, whether or not they're consecutive.~

Fixes #321

```console
$ pip-sync <pip-sync arg/flag>... --pip-args '<pip install arg/flag>...'
$ pip-sync <pip-sync arg/flag>... --pip-args '<pip install arg/flag>...'
```

**Changelog-friendly one-liners**: 

- ~~Any flags or arguments following two standalone double-dashes (' -- -- ') at the end of pip-compile or pip-sync's argument list are passed on to pip~~
- pip-compile output headers are now more accurate when ' -- ' is used to escape filenames
- pip-compile and pip-sync now pass anything provided to the new --pip-args option on to pip

##### Contributor checklist

### DIFF
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import itertools
 import os
+import shlex
 import sys
 
 from pip._internal.commands import create_command
@@ -73,6 +74,7 @@ DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
     "the private key and the certificate in PEM format.",
 )
 @click.argument("src_files", required=False, type=click.Path(exists=True), nargs=-1)
+@click.option("--pip-args", help="Arguments to pass directly to pip install.")
 def cli(
     ask,
     dry_run,
@@ -87,6 +89,7 @@ def cli(
     cert,
     client_cert,
     src_files,
+    pip_args,
 ):
     """Synchronize virtual environment with requirements.txt."""
     if not src_files:
@@ -139,7 +142,7 @@ def cli(
         user_only=user_only,
         cert=cert,
         client_cert=client_cert,
-    )
+    ) + shlex.split(pip_args or "")
     sys.exit(
         sync.sync(
             to_install,
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -328,6 +328,10 @@ def get_compile_command(click_ctx):
         # Collect variadic args separately, they will be added
         # at the end of the command later
         if option.nargs < 0:
+            # These will necessarily be src_files
+            # Re-add click-stripped '--' if any start with '-'
+            if any(val.startswith("-") and val != "-" for val in value):
+                right_args.append("--")
             right_args.extend([shlex_quote(force_text(val)) for val in value])
             continue
 
@@ -366,10 +370,20 @@ def get_compile_command(click_ctx):
                 left_args.append(shlex_quote(arg))
             # Append to args the option with a value
             else:
-                left_args.append(
-                    "{option}={value}".format(
-                        option=option_long_name, value=shlex_quote(force_text(val))
+                if option.name == "pip_args":
+                    # shlex_quote would produce functional but noisily quoted results,
+                    # e.g. --pip-args='--cache-dir='"'"'/tmp/with spaces'"'"''
+                    # Instead, we try to get more legible quoting via repr:
+                    left_args.append(
+                        "{option}={value}".format(
+                            option=option_long_name, value=repr(fs_str(force_text(val)))
+                        )
                     )
-                )
+                else:
+                    left_args.append(
+                        "{option}={value}".format(
+                            option=option_long_name, value=shlex_quote(force_text(val))
+                        )
+                    )
 
     return " ".join(["pip-compile"] + sorted(left_args) + sorted(right_args))
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -730,7 +730,7 @@ def test_allow_unsafe_option(pip_conf, monkeypatch, runner, option, expected):
 @mock.patch("piptools.scripts.compile.parse_requirements")
 def test_cert_option(parse_requirements, runner, option, attr, expected):
     """
-    The options --cert and --client-crt have to be passed to the PyPIRepository.
+    The options --cert and --client-cert have to be passed to the PyPIRepository.
     """
     with open("requirements.in", "w"):
         pass
@@ -757,6 +757,20 @@ def test_build_isolation_option(parse_requirements, runner, option, expected):
 
     # Ensure the options in parse_requirements has the expected build_isolation option
     assert parse_requirements.call_args.kwargs["options"].build_isolation is expected
+
+
+@mock.patch("piptools.scripts.compile.PyPIRepository")
+def test_forwarded_args(PyPIRepository, runner):
+    """
+    Test the forwarded cli args (--pip-args 'arg...') are passed to the pip command.
+    """
+    with open("requirements.in", "w"):
+        pass
+
+    cli_args = ("--no-annotate", "--generate-hashes")
+    pip_args = ("--no-color", "--isolated", "--disable-pip-version-check")
+    runner.invoke(cli, cli_args + ("--pip-args", " ".join(pip_args)))
+    assert set(pip_args).issubset(set(PyPIRepository.call_args.args[0]))
 
 
 @pytest.mark.parametrize(
--- a/tests/test_cli_sync.py
+++ b/tests/test_cli_sync.py
@@ -111,31 +111,52 @@ def test_merge_error(runner):
 
 
 @pytest.mark.parametrize(
-    "install_flags",
+    ("cli_flags", "expected_install_flags"),
     [
-        ["--find-links", "./libs1", "--find-links", "./libs2"],
-        ["--no-index"],
-        ["--index-url", "https://example.com"],
-        ["--extra-index-url", "https://foo", "--extra-index-url", "https://bar"],
-        ["--trusted-host", "foo", "--trusted-host", "bar"],
-        ["--user"],
-        ["--cert", "foo.crt"],
-        ["--client-cert", "foo.pem"],
+        (
+            ["--find-links", "./libs1", "--find-links", "./libs2"],
+            ["--find-links", "./libs1", "--find-links", "./libs2"],
+        ),
+        (["--no-index"], ["--no-index"]),
+        (
+            ["--index-url", "https://example.com"],
+            ["--index-url", "https://example.com"],
+        ),
+        (
+            ["--extra-index-url", "https://foo", "--extra-index-url", "https://bar"],
+            ["--extra-index-url", "https://foo", "--extra-index-url", "https://bar"],
+        ),
+        (
+            ["--trusted-host", "foo", "--trusted-host", "bar"],
+            ["--trusted-host", "foo", "--trusted-host", "bar"],
+        ),
+        (["--user"], ["--user"]),
+        (["--cert", "foo.crt"], ["--cert", "foo.crt"]),
+        (["--client-cert", "foo.pem"], ["--client-cert", "foo.pem"]),
+        (
+            ["--pip-args", "--no-cache-dir --no-deps --no-warn-script-location"],
+            ["--no-cache-dir", "--no-deps", "--no-warn-script-location"],
+        ),
+        (["--pip-args='--cache-dir=/tmp'"], ["--cache-dir=/tmp"]),
+        (
+            ["--pip-args=\"--cache-dir='/tmp/cache dir with spaces/'\""],
+            ["--cache-dir='/tmp/cache dir with spaces/'"],
+        ),
     ],
 )
 @mock.patch("piptools.sync.check_call")
-def test_pip_install_flags(check_call, install_flags, runner):
+def test_pip_install_flags(check_call, cli_flags, expected_install_flags, runner):
     """
     Test the cli flags have to be passed to the pip install command.
     """
     with open("requirements.txt", "w") as req_in:
         req_in.write("six==1.10.0")
 
-    runner.invoke(cli, install_flags)
+    runner.invoke(cli, cli_flags)
 
     call_args = [call[0][0] for call in check_call.call_args_list]
     called_install_options = [args[6:] for args in call_args if args[3] == "install"]
-    assert called_install_options == [install_flags], "Called args: {}".format(
+    assert called_install_options == [expected_install_flags], "Called args: {}".format(
         call_args
     )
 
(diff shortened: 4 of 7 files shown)


AndydeCleyre:
Process question: Am I supposed to check the clear-oneliner box myself, or leave that to a reviewer's judgment?

atugushev:
By the way, we could do the same way on `pip-compile` to pass additional args to `pip_args`.

piptools/scripts/sync.py
```diff
@@ -84,6 +84,15 @@ def cli(
     src_files,
 ):
     """Synchronize virtual environment with requirements.txt."""
+    try:
+        fwd_args_idx = src_files.index("--")
+    except ValueError:
+        install_flags = []
+    else:
+        install_flags = list(src_files[fwd_args_idx + 1 :])  # noqa
```
atugushev: Could you update in a separate PR `setup.cfg` to ignore `E203`? See https://github.com/psf/black#slices.

atugushev:
@AndydeCleyre 

> Process question: Am I supposed to check the clear-oneliner box myself, or leave that to a reviewer's judgment?

Both are fine.

piptools/scripts/sync.py
AndydeCleyre: Sure, see #1081. I'm guessing that one doesn't need tests or changelog entries.

AndydeCleyre: Accordingly, I've removed the one-off noqa marker, so the tests will fail until the #1081 is merged.

atugushev: Merged.

atugushev:
> By the way, we could do the same way on `pip-compile` to pass additional args to `pip_args`.

@AndydeCleyre what do you think?

AndydeCleyre:
Oh, you mean in this same PR? 

OK, I can get to it within 12 hours

AndydeCleyre:
FYI I've started the pip-compile work locally, but there are still a few wrinkles to iron out before I push. I'll have to resume work a little later.

AndydeCleyre:
I don't know if this is within scope of this PR, but these changes have exposed to me and expanded already existing problems with the header generation omitting ` -- ` from the argument list.

```console
$ echo six > --requirements.in
$ pip-compile -- --requirements.in
#
# To update, run:
#
#    pip-compile --requirements.in
#
six==1.14.0               # via -r --requirements.in
$ pip-compile -- --requirements.in -- --disable-pip-version-check
#
# To update, run:
#
#    pip-compile -- --disable-pip-version-check --requirements.in
#
six==1.14.0               # via -r --requirements.in
```
Those commands in the headers are not valid and will fail. The first will read `--requirements.in` as an illegal option, and the second will read `--disable-pip-version-check` as a nonexistent path.

AndydeCleyre:
Tomorrow I will have a go at improving `get_compile_command` for inclusion in this PR.

Note to self:

```python
click_ctx.params == {'src_files': ('--requirements.in', '--', '--disable-pip-version-check'),...}
```

AndydeCleyre:
OK, so `get_compile_command` intentionally omits:
- `arguments that don't change build behaviour like '--verbose'`
- `one-off arguments like '--upgrade'`
- `values that are already default`

and also does `expanding options short to long`

What do you think of the following judgments regarding "forwarded" `pip` args?

1. We shouldn't bother checking if values are default or not; those details are too far outside this project.
2. Similarly, we shouldn't bother doing any expansion for them.
3. Again, we shouldn't be concerned with whether or not they change build behavior. Basically, I think we should be "hands off" as much as possible for these, not trying to be too clever.
4. The big question for me: should they all be treated as one-offs, and omitted? In favor of that:
    - simpler implementation (no extra sorting, and we can ignore 1, 2, and 3)
    - if users are regularly using that feature as standard project practice, they can always provide `CUSTOM_COMPILE_COMMAND`

AndydeCleyre:
I've updated `get_compile_command` and a corresponding test to be more accurate when encountering one or two uses of ` -- ` in the pip-compile argument list.

Arguments that are forwarded to pip are not transformed or sorted or omitted, but left as-is as much as possible.

Please let me know what you think, and if you want these two commits squashed.

AndydeCleyre:
Ack, this isn't quite right yet, in it's handling of ` - ` to signify stdin/out. Stay tuned.

AndydeCleyre:
I guess the real question regarding the last comment is if we want to change behavior for

```console
$ pip-compile --output-file=- -- -
```

to treat `-` as an actual filename, whereas now any `src_file` with that exact name is taken to signify stdin rather than a file. I think it's fine to keep as-is.

AndydeCleyre:
Looks like this will need a careful rebase now that c4dcaf706ffa223a140613e79603700e0033758c is pushed. I see that pip-compile's `install_flags` is now [built differently](https://github.com/jazzband/pip-tools/commit/c4dcaf706ffa223a140613e79603700e0033758c#diff-b0bb5ed53e50b127ae3b78c8f1a6e518L122). 

Would it make sense to add a kwarg to `_compose_install_flags` like `extra_flags`, passing any ` -- -- ` "forwarded" ones as a list/tuple?

Or would it be preferable to simply extend its return value after it's called, outside that function?

AndydeCleyre:
I'm looking at / merging this c4dcaf706ffa223a140613e79603700e0033758c . . . Is it just me, or are `get_trusted_hosts` and `create_install_command` ghosts?

I don't think that was ready for master, maybe I should hold off on merging it into this PR for now.

atugushev:
@AndydeCleyre  see https://github.com/jazzband/pip-tools/pull/824#issuecomment-594512120 for details. I'll fix this shortly.

atugushev:
@AndydeCleyre fixed

AndydeCleyre:
Alright, after merging in those changes last night, I did a proper rebase onto master today.  Please have a look. I've included all the concerns I raised above, below.

Review Pointers
===

`tests/test_cli_sync.py:test_pip_install_flags`
---
#824 changed some things about this test, maybe making it neater, but AFAICT also making it insufficiently flexible for the cases I needed in this PR, so those were partially reverted.

The issue is just that the flags passed to `pip-sync` are not identical to those which should be passed to `pip install` (namely, the `--` instances). Someone who has a better idea than me of the intent and benefits behind those test changes, please pay attention.

`piptools/scripts/sync.py:_compose_install_flags`
---
#824 introduced `_compose_install_flags`. As-is, it does not accept any "extra" or "right" args to append to the more deliberately handled flags. It may or may not be preferable to add this functionality to that function. I chose to leave the function alone and instead append the new forwarded args to the list returned by that function.

stdin marker `-`
---
```console
$ pip-compile --output-file=- -- -
```
What *should* the behavior be for this command? As is, ` -- ` signifies that the following arguments are *args*, not options. `-` is hard-coded to signify "stdin" rather than a file. It is conceivable that someone would want or expect the ` -- ` to instead signify that the following `-` is a literal *filename*. Of course no sane person would use a file with that name, and they should be able to get by with `./-`, but I'd like to make sure this decision is hashed out.

`piptools/utils.py:get_compile_command` and `pip-compile --output-file=- -`
---
This is closely related to the previous pointer and will be affected by the corresponding consensus, whatever that turns out be.

Before this PR, a command like:
```console
$ pip-compile --output-file=- -
```
would generate the following as documentation in the output header:
```bash
#    pip-compile --output-file=- -
```

The current state of this PR changes that part of the header to:
```bash
#    pip-compile --output-file=- -- -
```
With the current treatment of `-` as stdin (never a literal filename), this is functionally equivalent and also correct. It is of course possible to add a bit more logic to `get_compile_command` to strip out the excess `-- `. IMO, if we are to keep our current treatment of `-- -`, doing so is unwarranted. But we should do it if either you disagree, or we change our interpretation of `-- -` to handle `-` as a filename.

`piptools/utils.py:get_compile_command` and sorting, omission, & expansion
---
I chose to include the forwarded args and flags in the `get_compile_command` output, without applying any of those operations to them (sorting, omission, expansion). I view forwarded args after ` -- -- ` with a "take your stinking parsers off me, you damn dirty pip-tools" attitude, as much as possible. I think there's also a decent case for omitting them altogether, but including them is my preference.

atugushev:
Just a thought, what if we add a new option `--pip-args` (same as `pipx`) instead of `-- --`. For example:

```console
$ pip-compile --pip-args="--no-cache-dir"
$ pip-compile --pip-args="--no-cache-dir --no-deps --force-reinstall"
```

AndydeCleyre:
I'm not eager to implement it as a single quoted arg, as then we'll have to be careful to unquote and requote the inner arguments appropriately. Maybe that's not really an issue and things will "just work," but it's my initial concern.

I will reimplement it that way if it's preferred. Would love to get some feedback from more users looking forward to this feature.

tim-mitchell:
I don't mind either way. But if you go with the --pip-args option then no escaping should be done.  That should be the users responsibility.
AFAICT 
`--pip-args="--some-arg=\"path with spaces\""` will pass 
`--some-arg="path with spaces"` to pip (on windows at least).

AndydeCleyre:
Thanks everyone, I'll start re-implementing this as a single, quoted arg soon. Time is a little limited for now, especially during the week, but optimistically this PR will ready for another review next week.

In the meantime, you are welcome to mention here all the crazy examples you might use with this feature (with spaces and single and double quotes and whatever) to help me test properly.

AndydeCleyre:
For posterity, the "double-dash islands" implementation now lives at my [bugfix/321-doubledash-islands](https://github.com/AndydeCleyre/pip-tools/tree/bugfix/321-doubledash-islands) branch.

I've just pushed a re-implementation to this branch, using `--pip-args`, and relying on `shlex` to do proper arg splitting.

Please review and test!

tests/test_cli_sync.py
```diff
@@ -111,31 +111,47 @@ def test_merge_error(runner):
...
+            ["--trusted-host", "foo", "--trusted-host", "bar"],
+            ["--trusted-host", "foo", "--trusted-host", "bar"],
+        ),
+        (["--user"], ["--user"]),
+        (["--cert", "foo.crt"], ["--cert", "foo.crt"]),
+        (["--client-cert", "foo.pem"], ["--client-cert", "foo.pem"]),
+        (
+            ["--pip-args", "--no-cache-dir --no-deps --no-warn-script-location"],
+            ["--no-cache-dir", "--no-deps", "--no-warn-script-location"],
+        ),
```
atugushev, line 139: Could you add a few cases:

```
$ pip-sync --pip-args="--cache-dir=/tmp"
$ pip-sync --pip-args="--cache-dir='/tmp/cashe dir with spaces/'"
```

AndydeCleyre: Sure, added those cases, with the equals-style syntax.

atugushev:
```console
$ echo | pip-compile - -qo- --pip-args="--cache-dir='/tmp/with spaces'"
#
# To update, run:
#
#    pip-compile --output-file=- --pip-args='--cache-dir='"'"'/tmp/with spaces'"'"'' -
#
#
```
I've checked a header, it looks weird.

AndydeCleyre:
Thanks, I'll check it out.

AndydeCleyre:
> I've checked a header, it looks weird.

This may just be the price of using `--pip-args='<str>'` here instead of letting the shell handle it, [as feared](https://github.com/jazzband/pip-tools/pull/1080#issuecomment-601311368).

I think mitigating the redundant quotes would be functionally fragile and hacky.

AndydeCleyre:
Maybe I'll dig into `click` tonight and see if something in there can help.

AndydeCleyre:
@atugushev Please take a look at the (currently unsquashed) new commit, which special-cases `pip_args` within `get_compile_command` to just pass its value through `repr` instead of `shlex_quote`.

It's probably a fragile idea. Proper quoting can be hell, which is why I prefer the first implementation, to avoid dealing with it. But this hack is only used for the comment, so its perfection isn't as critical as it would be if it were affecting actual `pip` calls. It _might_ be good enough.

So I'm thinking either:

1. you can turn this new commit's idea into something saner
2. decide this hack is good enough, and I'll squash it
3. revert to the ` -- -- ` implementation; I'd backport a couple test cases and rebase that preserved branch onto master and bulldoze this one
4. live with the ugly redundant yet reliably functional quoting exhibited without this new commit

tests/test_utils.py
```diff
@@ -265,12 +265,25 @@ def test_force_text(value, expected_text):
...
+            "pip-compile --pip-args='--disable-pip-version-check --isolated'",
+        ),
     ],
 )
 def test_get_compile_command(tmpdir_cwd, cli_args, expected_command):
     """
     Test general scenarios for the get_compile_command function.
     """
+    with open("--requirements.in", "w"):
+        pass
```
atugushev, line 286: This block could be moved out to a separate test.

tests/test_cli_compile.py
```diff
@@ -759,6 +759,20 @@ def test_build_isolation_option(parse_requirements, runner, option, expected):
...
 
+@mock.patch("piptools.scripts.compile.PyPIRepository")
+def test_forwarded_args(PyPIRepository, runner):
+    """
+    Test the forwarded cli args (--pip-args 'arg...') are passed to the pip command.
+    """
+    with open("requirements.in", "w"):
+        pass
+
+    pipc_args = ("--no-annotate", "--generate-hashes")
```
atugushev, line 770:
```suggestion
    cli_args = ("--no-annotate", "--generate-hashes")
```
A nitpick. Feel free to disregard it :)

piptools/utils.py
```diff
@@ -366,10 +370,17 @@ def get_compile_command(click_ctx):
...
             else:
-                left_args.append(
-                    "{option}={value}".format(
-                        option=option_long_name, value=shlex_quote(force_text(val))
+                if option.name == "pip_args":
+                    left_args.append(
+                        "{option}={value}".format(
+                            option=option_long_name, value=repr(fs_str(force_text(val)))
+                        )
+                    )
```
atugushev, line 378: Could you add a comment about why `pip_args` handled in that way?

AndydeCleyre: Will push soon:
```python
if option.name == "pip_args":
    # shlex_quote would produce functional but noisily quoted results,
    # e.g. --pip-args='--cache-dir='"'"'/tmp/with spaces'"'"''
    # Instead, we try to get more legible quoting via repr:
```

tests/test_cli_compile.py
AndydeCleyre: :+1:

piptools/utils.py
atugushev: Neat!

tests/test_utils.py
AndydeCleyre: Sure, next commit will grant that its own test.

AndydeCleyre:
*splat*

atugushev:
Thank you, @AndydeCleyre, for the great work and patience 🎉 🎉 🎉

tim-mitchell:
Thanks @AndydeCleyre and @atugushev

PR #1055, opened 2020-02-03
Pin pip-tools to pip>=20.0

atugushev:
Refs #926.

**Progress**

- [X] remove `pip<20.0` compatible code
- [X] remove obsolete tests
- [X] remove `do_import()` from `pip_compat` module
- [X] refactor codebase to import objects directly from pip
- [X] remove `pip<20.0` envs from tox.ini
- [X] remove `pip<=20.0` jobs from travis
- [X] remove `pip<=20.0` jobs from appveyor
- [X] pin `pip>=20` in `setup.py`

**Changelog-friendly one-liner**: Pin `pip-tools` to `pip>=20.0`.

##### Contributor checklist

- ~~Provided the tests for the changes.~~
- [X] Gave a clear one-line description in the PR (that the maintainers can add to CHANGELOG.md on release).
- [X] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #1090, opened 2020-03-22
Remove build_isolation param from PyPIRepository constructor

atugushev:
Pass `--build-isolation/--no-build-isolation` option to the `pip_args` of `PyPIRepository` constructor.

Addresses https://github.com/jazzband/pip-tools/pull/1055#discussion_r377464546.


atugushev:
Thanks, @AndydeCleyre!

PR #1091, opened 2020-03-23
Use click API to test the output from click.echo()

atugushev:
The rationale is to use fewer mocks if possible.

### DIFF
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -173,12 +173,12 @@ def sync(
     if dry_run:
         if to_uninstall:
             click.echo("Would uninstall:")
-            for pkg in to_uninstall:
+            for pkg in sorted(to_uninstall):
                 click.echo("  {}".format(pkg))
 
         if to_install:
             click.echo("Would install:")
-            for ireq in to_install:
+            for ireq in sorted(to_install, key=key_from_ireq):
                 click.echo("  {}".format(format_requirement(ireq)))
 
     if ask and click.confirm("Would you like to proceed with these changes?"):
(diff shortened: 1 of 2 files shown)


AndydeCleyre:
I like the rationale, and appreciate any reduction in the decorator stacks. Testing is still one of the more opaque parts of the codebase for me.

The changes look good, but can you clarify a bit of the `click` business with bytes vs unicode for me? The hosted doc links for `isolated_filesystem` are currently broken ( https://click.palletsprojects.com/en/7.x/search/?q=isolated_filesystem ).

I see at `click`'s [testing:CliRunner](https://click.palletsprojects.com/en/7.x/api/#testing):

```
 property stdout
    The standard output as unicode string.
 stdout_bytes = None
    The standard output as bytes.
```

But I'm guessing from the content of this PR that the `stdout.getvalue()` from `isolated_filesystem` is always guaranteed to be `bytes`?

atugushev:
@AndydeCleyre 

> But I'm guessing from the content of this PR that the `stdout.getvalue()` from `isolated_filesystem` is always guaranteed to be `bytes`?

Regarding to [the code](https://github.com/pallets/click/blob/97de2c97275f52adc5c6ef8bb8aa993ce997fe7e/src/click/testing.py#L236) `isolation()` yields bytes:

```python
yield (bytes_output, not self.mix_stderr and bytes_error)
```

AndydeCleyre:
Thanks @atugushev, I see it's using `io.BytesIO` consistently. In that case, looks good, except: any reason why you `decode` for (just) 2/3 of them?

atugushev:
@AndydeCleyre 

> any reason why you `decode` for (just) 2/3 of them?

That's my loss, good catch! Addressed in 428d5e0.

atugushev:
@AndydeCleyre thank you for reviewing this! 🙏

PR #1093, opened 2020-03-28
Compile editables that are both primary reqs and constraints

richafrank:
If an editable package is both a primary requirement and a constraint, it was being considered as solely a constraint, and so it and its dependencies were excluded from pip-compile output.

Instead of combining the editable requirements, we were returning the first one, and the two are sorted the same up to `"... -c constraints.in"` and `"... -r requirements.in"`.

**Changelog-friendly one-liner**: Ensure editables that are both primary reqs and constraints appear in pip-compile output

##### Contributor checklist

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -60,7 +60,8 @@ def combine_install_requirements(ireqs):
     combined_ireq = copy.deepcopy(source_ireqs[0])
     for ireq in source_ireqs[1:]:
         # NOTE we may be losing some info on dropped reqs here
-        combined_ireq.req.specifier &= ireq.req.specifier
+        if combined_ireq.req is not None and ireq.req is not None:
+            combined_ireq.req.specifier &= ireq.req.specifier
         combined_ireq.constraint &= ireq.constraint
         # Return a sorted, de-duped tuple of extras
         combined_ireq.extras = tuple(
@@ -220,13 +221,6 @@ class Resolver(object):
 
         """
         for _, ireqs in full_groupby(constraints, key=key_from_ireq):
-            ireqs = list(ireqs)
-            editable_ireq = next((ireq for ireq in ireqs if ireq.editable), None)
-            if editable_ireq:
-                # ignore all the other specs: the editable one is the one that counts
-                yield editable_ireq
-                continue
-
             yield combine_install_requirements(ireqs)
 
     def _resolve_one_round(self):
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -219,6 +219,29 @@ def test_editable_package(pip_conf, runner):
     assert "small-fake-a==0.1" in out.stderr
 
 
+@pytest.mark.parametrize(("req_editable",), [(True,), (False,)])
+def test_editable_package_in_constraints(pip_conf, runner, req_editable):
+    """
+    piptools can compile an editable that appears in both primary requirements
+    and constraints
+    """
+    fake_package_dir = os.path.join(PACKAGES_PATH, "small_fake_with_deps")
+    fake_package_dir = path_to_url(fake_package_dir)
+
+    with open("constraints.txt", "w") as constraints_in:
+        constraints_in.write("-e " + fake_package_dir)
+
+    with open("requirements.in", "w") as req_in:
+        prefix = "-e " if req_editable else ""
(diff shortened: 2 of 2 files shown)


piptools/resolver.py
```diff
@@ -219,14 +219,16 @@ def _group_constraints(self, constraints):
...
+                # Within each group, sort by editables first, and within that,
+                # non-constraints before constraints (False sorts before True):
+                key=lambda x: (key_from_ireq(x), not x.editable, x.constraint),
+            ),
+            key=key_from_ireq,
+        ):
             ireqs = list(ireqs)
-            editable_ireq = next((ireq for ireq in ireqs if ireq.editable), None)
-            if editable_ireq:
-                # ignore all the other specs: the editable one is the one that counts
```
richafrank: I couldn't find enough context to know what depended on returning just the first editable, if a reviewer could comment. This would also help determine whether the tie-breaking I added above is necessary or not.

atugushev:
Hello @richafrank!

> If an editable package is both a primary requirement and a constraint, it was being considered as solely a constraint, and so it and its dependencies were excluded from pip-compile output.

It looks like a bug. Could you please provide also reproducible steps to test?

richafrank:
Hi @atugushev !

(I thought the unit test might serve as those steps. It fails on master.) 

Put an editable in your requirements.in and the same one in a constraints via a second .in file. I can spell that out more precisely if that'd be helpful!

atugushev:
Usually, editables don't suppose to be in constraints file, but it looks like this bug related to layered workflow. Am I right?

piptools/resolver.py
```diff
@@ -219,14 +219,16 @@ def _group_constraints(self, constraints):
...
-        for _, ireqs in full_groupby(constraints, key=key_from_ireq):
+        for _, ireqs in groupby(
+            sorted(
+                constraints,
+                # Within each group, sort by editables first, and within that,
+                # non-constraints before constraints (False sorts before True):
+                key=lambda x: (key_from_ireq(x), not x.editable, x.constraint),
+            ),
+            key=key_from_ireq,
+        ):
```
atugushev: AFAIK `sort` and `groupby` by different key functions could be a potential bug.

atugushev: AFAIK `sort` and `groupby` by different key functions could be a potential bug.

atugushev: > The operation of groupby() is similar to the uniq filter in Unix. It generates a break or new group every time the value of the key function changes (which is why it is usually necessary to have sorted the data using the same key function). That behavior differs from SQL’s GROUP BY which aggregates common elements regardless of their input order.

https://docs.python.org/3/library/itertools.html#itertools.groupby

richafrank:
> Usually, editables don't suppose to be in constraints file, but it looks like this bug related to layered workflow. Am I right?

Yea, I'm generating a lockfile from a base set of requirements. Then I'm using that lockfile as constraints when generating a lockfile for a superset of the base requirements, to keep them in sync. The base requirements include an editable package, so it appears in the lockfile. Thoughts on this workflow?

piptools/resolver.py
richafrank: I believe in this case, because the sort key is the group key plus extra tie-breakers, it should satisfy the needs of the `groupby`. The `groupby` cares about comparing the group key between elements. In this way, we can sort within the groups when we sort everything. Alternatively, we could sort each group again after the sort+groupby. 

Ultimately, we should decide if this sort order is even necessary - I added it only for backwards compatibility with the comment I removed about yielding just the editable. Now it yields a copy of the editable with updates from others in the group. But we could implement that however, if it's desirable.

atugushev: In fact, we may don't touch these lines and everything would work. The **order** of the sort doesn't matter actually, so these changes can be reverted.

atugushev: The point is that the fact of *sort* is matters, not order.

richafrank:
Thanks for the review @atugushev . I'm certainly on board with your suggestions - I'm glad we don't need all that ordering compat.

piptools/resolver.py
```diff
@@ -219,14 +219,16 @@ def _group_constraints(self, constraints):
...
             ireqs = list(ireqs)
```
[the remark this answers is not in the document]
richafrank: Ah, good find.

atugushev:
Thanks, @richafrank!

PR #1094, opened 2020-03-28
Suggest "pip install --upgrade" using module invocation in README

atugushev:
Suggest `python -m pip install --upgrade` rather than `pip install --upgrade` since it's the only way to upgrade `pip` on Windows, otherwise you'll get a PermissionError during upgrade `pip.exe`.


atugushev:
Thanks @hramezani!

PR #1101, opened 2020-04-02
Add a test for uncovered branch in the writer module

atugushev:
Now it's 100% coverage in the writer module.

### DIFF
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -159,6 +159,28 @@ def test_iter_lines__hash_missing(writer, from_line):
     assert tuple(lines) == expected_lines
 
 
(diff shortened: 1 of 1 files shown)


atugushev:
@AndydeCleyre awesome catch! 🎉 That's so tricky. I'll add a test for that case then.

AndydeCleyre:
The test looks good :+1: . Is it still the plan to change the `has_hashes` tracking at all, or is this PR just for the test now?

atugushev:
@AndydeCleyre 

Title updated.

atugushev:
@AndydeCleyre 

Thank you for the analysis and for reviewing this!

PR #1109, opened 2020-04-19
Get hashes from PyPI JSON API

### LINKED ISSUE #543: use sha256 hashes from warehouse
Generating hashes is too slow, use the hashes from warehouse.

eg: click the SHA256 buttons: https://pypi.org/project/gitchart/#files

atugushev:
Resolves #543.

##### Before
```console
$ time echo web3 | pip-compile - -qo- --generate-hashes > /dev/null

real	1m32.754s
user	0m10.959s
sys	0m3.083s
```

##### After
```console
$ time echo web3 | pip-compile - -qo- --generate-hashes > /dev/null

real	0m4.824s
user	0m2.603s
sys	0m0.141s
```

**Changelog-friendly one-liner**: Get hashes from PyPI JSON API.

##### Contributor checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -9,7 +9,7 @@ from shutil import rmtree
 
 from pip._internal.cache import WheelCache
 from pip._internal.commands import create_command
-from pip._internal.models.index import PyPI
+from pip._internal.models.index import PackageIndex, PyPI
 from pip._internal.models.link import Link
 from pip._internal.models.wheel import Wheel
 from pip._internal.req import RequirementSet
@@ -18,12 +18,14 @@ from pip._internal.utils.hashes import FAVORITE_HASH
 from pip._internal.utils.misc import normalize_path
 from pip._internal.utils.temp_dir import TempDirectory, global_tempdir_manager
 from pip._internal.utils.urls import path_to_url, url_to_path
+from pip._vendor.requests import RequestException
 
 from .._compat import PIP_VERSION, TemporaryDirectory, contextlib
 from ..click import progressbar
 from ..exceptions import NoCandidateFound
 from ..logging import log
 from ..utils import (
+    as_tuple,
     fs_str,
     is_pinned_requirement,
     is_url_requirement,
@@ -227,6 +229,47 @@ class PyPIRepository(BaseRepository):
 
         return self._dependencies_cache[ireq]
 
+    def _get_project(self, ireq):
+        """
+        Return a dict of a project info from PyPI JSON API for a given
+        InstallRequirement. Return None on HTTP/JSON error or if a package
+        is not found on PyPI server.
+
+        API reference: https://warehouse.readthedocs.io/api-reference/json/
+        """
+        package_indexes = (
+            PackageIndex(url=index_url, file_storage_domain="")
+            for index_url in self.finder.search_scope.index_urls
+        )
+        for package_index in package_indexes:
+            url = "{url}/{name}/json".format(url=package_index.pypi_url, name=ireq.name)
+            try:
+                response = self.session.get(url)
+            except RequestException as e:
+                log.debug(
+                    "Fetch package info from PyPI failed: {url}: {e}".format(
+                        url=url, e=e
+                    )
+                )
+                continue
+
+            # Skip this PyPI server, because there is no package
+            # or JSON API might be not supported
+            if response.status_code == 404:
+                continue
+
+            try:
+                data = response.json()
+            except ValueError as e:
+                log.debug(
+                    "Cannot parse JSON response from PyPI: {url}: {e}".format(
+                        url=url, e=e
+                    )
+                )
+                continue
+            return data
+        return None
+
     def get_hashes(self, ireq):
         """
         Given an InstallRequirement, return a set of hashes that represent all
@@ -257,6 +300,50 @@ class PyPIRepository(BaseRepository):
         if not is_pinned_requirement(ireq):
             raise TypeError("Expected pinned requirement, got {}".format(ireq))
 
+        log.debug("{}".format(ireq.name))
+
+        with log.indentation():
+            hashes = self._get_hashes_from_pypi(ireq)
+            if hashes is None:
+                log.log("Couldn't get hashes from PyPI, fallback to hashing files")
+                return self._get_hashes_from_files(ireq)
+
+        return hashes
+
+    def _get_hashes_from_pypi(self, ireq):
+        """
+        Return a set of hashes from PyPI JSON API for a given InstallRequirement.
+        Return None if fetching data is failed or missing digests.
+        """
+        project = self._get_project(ireq)
+        if project is None:
+            return None
+
+        _, version, _ = as_tuple(ireq)
+
+        try:
+            release_files = project["releases"][version]
+        except KeyError:
+            log.debug("Missing release files on PyPI")
+            return None
+
+        try:
+            hashes = {
+                "{algo}:{digest}".format(
+                    algo=FAVORITE_HASH, digest=file_["digests"][FAVORITE_HASH]
+                )
+                for file_ in release_files
+            }
+        except KeyError:
+            log.debug("Missing digests of release files on PyPI")
+            return None
+
+        return hashes
+
+    def _get_hashes_from_files(self, ireq):
+        """
+        Return a set of hashes for all release files of a given InstallRequirement.
+        """
         # We need to get all of the candidates that match our current version
         # pin, these will represent all of the files that could possibly
         # satisfy this constraint.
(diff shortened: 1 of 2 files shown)


piptools/repositories/pypi.py
```diff
@@ -227,6 +229,23 @@ def get_dependencies(self, ireq):
...
         return self._dependencies_cache[ireq]
 
+    def _get_project(self, ireq):
+        """
+        Given an `InstallRequirement`, return a dict of a project info
+        from PyPI JSON API. Return None on HTTP or JSON parsing error.
+
+        See https://warehouse.readthedocs.io/api-reference/json/
+        """
+        url = "{url}/{name}/json".format(url=PyPI.pypi_url, name=ireq.name)
```
atugushev, line 239: A thought: probably, we should configure a `PyPI.pypi_url`.

atugushev: Yeah, this part must be improved when a user has an extra URLs, or a private index url, or ignores package index by `--no-index`.

atugushev: @graingert any thoughts on this?

graingert: I'm not sure about this, we could try to see if the index chosen supports the PyPI JSON API, or use hashes from PyPI.org for files with the same filename, or ignore it when using a non PyPI.org index

or make it a cli flag?

graingert:
it might be worth always getting the hashes now rather than skipping them if the hashes are already in the requirements.txt (this behaviour causes me problems, eg when new wheels are uploaded to an old version I don't get the hash of the new wheel

atugushev:
We could ignore current hashes using `--rebuild`. What do you think?

atugushev:
But if think it deserves a discussion on a separate issue.

piptools/repositories/pypi.py
atugushev: Addressed in ed6d8158d3aa0a161c2ec4403677dcac4515f087.

This handles all configured index URLs and also handles disabled package index via `--no-index`.

piptools/repositories/pypi.py
```diff
@@ -236,15 +236,23 @@ def _get_project(self, ireq):
...
-            return None
-        return data
+        package_indexes = [
+            PackageIndex(url=index_url, file_storage_domain="")
+            for index_url in self.finder.search_scope.index_urls
+        ]
+        for package_index in package_indexes:
+            url = "{url}/{name}/json".format(url=package_index.pypi_url, name=ireq.name)
+            try:
+                data = self.session.get(url).json()
```
graingert, line 246: we should probably remember which URLs support the PyPI JSON api

atugushev: How would we know that JSON API of an extra index URL is supported?

graingert: if this URL doesn't 404

atugushev: If it raises 404 then it falls back to the files hash checking.

atugushev: Perhaps 404 should be handled explicitly

graingert: but there's no point re-requesting the API from indexes that we've already tried

atugushev: Good call. But what if https://custom.pypi-server.com/pypi/my-package/json raise 404 because there is no `my-package` on the server, but it supports JSON API?

graingert: when you fall back to calculating hashes directly you'll know it was because the index doesn't support the api

atugushev: Sorry, I might be missing your point, but I'll try to describe mine in detail. 

Using this heuristic, we can't be sure that JSON API is not supported. Consider the following case:

- package1 from https://pypi1-server/pypi/package1/json
- package2 from https://pypi2-server/pypi/package2/json
- package3 from https://pypi1-server/pypi/package1/json

Suppose we run the following command:
```console
$ pip-compile --index-url=https://pypi1-server/ --extra-index-url=https://pypi2-server/
```

Step 1. When pip-tools processes `package1` everything is okay because it hits https://pypi1-server/pypi/package1/json which responses 200. 

Step2. After,  pip-tools compiles `package2` and hit the first index URL, which is https://pypi1-server/pypi/package2/json. But this URL raises 404 because the package is on pypi2-server. So it has to go to the second index URL, which is https://pypi2-server/pypi/package2/json, which responses 200.

Step3. Then pip-tools processes `package3 `the same as in *step 1* with the first index URL.


> How would we know that JSON API of a custom index URL is supported?
> If this URL doesn't 404.

Considering the above statement the test case will be failed on *step 3* because `pypi1-server` will be marked as "no supported JSON API" on *step 2* because pip-tools couldn't find package2 on `pypi1-server` and it responded 404.

Am I missing something?

graingert: if the simple index and the JSON API 404s then you don't know
if the simple index 200s and the JSON API 404s then you know the API is unsupported

atugushev: I see your point. Yes, that would work. Assuming that suggestion is to hit less the PyPI servers now we have to hit the servers twice on 404. So what's the point?

graingert:
You have to hit it twice to download the hashes from indexes that don't
support the JSON API

piptools/repositories/pypi.py
atugushev: > You have to hit it twice to download the hashes from indexes that don't
support the JSON API

Sorry, I don't understand what are you trying to achieve by that. Could you elaborate?

atugushev:
Okay, it's ready for review now.

AndydeCleyre:
I may be missing something in `pip`'s internal vendored `requests.session` business, but AFAIU there's no `timeout` getting set during the `self.session.get(url)` call, which could be trouble.

If so, I notice there's one other case like that in `pypi.py` that predates this PR (from 627dbaf9).

I'll take a closer look at this whole thing tonight, but it looks really nice, especially that timed example you've provided.

Thanks!

atugushev:
@AndydeCleyre 

It respects pip's `options.timeout`, which is default 15 seconds.

piptools/repositories/pypi.py
```diff
@@ -257,6 +300,52 @@ def get_hashes(self, ireq):
         if not is_pinned_requirement(ireq):
             raise TypeError("Expected pinned requirement, got {}".format(ireq))
 
+        log.debug("  {}".format(ireq.name))
+
+        hashes = self._get_hashes_from_pypi(ireq)
+        if hashes is None:
+            # TODO: implement indentation context manager to avoid hard coded
+            #       spaces prefixes for debug messages.
+            #       See pip._internal.utils.logging.indent_log() for a reference.
```
atugushev, line 309: Addressed in #1110.

AndydeCleyre:
Great! This is a much appreciated improvement.

If you like, minor English edits in the docstrings: 

- "by a given InstallRequirement" -> "for a given..." (x2)
- "from every release files" -> "of all release files" or "for all release files"

atugushev:
Thanks @AndydeCleyre for reviewing this! Your suggestions addressed in a03f3fd17dafd0e8477f11785f03099e224f222c.

atugushev:
Rebased onto the master and squashed commits.

atugushev:
@AndydeCleyre @auvipy @graingert 

Thanks for reviewing this! 🙏

piptools/repositories/pypi.py
atugushev: If you think this should be fixed anyway please feel free to open an issue/PR. Thank you for the help!

graingert:
@atugushev the speedup here is going to be incredible! :heart: :heart: :heart: :heart: :tada: 

I was just rehashing a requirements that uses numpy pandas and lxml and thinking longingly of this PR

atugushev:
`pip-tools v5.1.0` is released. Thanks @graingert!

graingert:
@atugushev fyi this PR means you now get .egg hashes in the requirements.txt, eg:
```
 toml==0.10.0 \
     --hash=sha256:229f81c57791a41d65e399fc06bf0848bab550a9dfd5ed66df18ce5f05e73d5c \
     --hash=sha256:235682dd292d5899d361a811df37e04a8828a5b1da3115886b73cf81ebc9100e \
+    --hash=sha256:f1db651f9657708513243e61e6cc67d101a39bad662eaa9b5546f789338e07a3 \
```

atugushev:
Erm, I wonder should we skip egg hashes?

graingert:
@atugushev what mechanism skipped them before?

atugushev:
pip doesn't support installation from eggs, so it's just picked wheels and source dists.

atugushev:
So we could skip `"packagetype": "bdist_egg"` digests, since they aren't useful.

graingert:
@atugushev https://github.com/jazzband/pip-tools/issues/1121

PR #1115, opened 2020-04-21
[WIP] Regression: Fix editable and non-editable requirement duplicate

vphilippon:
WIP, input welcome. The first commit contains a test to reproduce.
This is a regression from 4.5.1

Given a requirement file like:
```
-e /path/to/small_fake_a
small_fake_with_unpinned_deps  # Has dependency on small_fake_a.
```

The output will be:
```
-e file://path/to/small_fake_a
small_fake_a==0.2  # Hey, this is already covered by the editable install above!
small_fake_with_unpinned_deps==0.1
<other dependencies>
```

#1093 introduced the regression, although the effect was hard to spot.

The cause:
- Passing an editable `ireq` to pip resolver (our `_iter_dependency`)has the side effect of setting the `ireq.req` on it, effectively giving it a package name and specifier.
Having `ireq.req` set on it for subsequent cycle (most of the time (always?), editables come from the requirements file, in round 1) cause `_group_constraints` (with `full_groupby`) to group the editable and non-editable requirements for a same package together.
- The change in #1093 made `combine_install_requirements` handle all `ireq`, including editable ones. But `combine_install_requirements` returns a deep-copy of the `ireq`, which means we loose the side effect described above, thus loosing the grouping, thus causing the duplicate in the output.

Removing the deepcopy at https://github.com/jazzband/pip-tools/blob/master/piptools/resolver.py#L60 fixes the issue, but could cause other issues, notably with the diff use to determine if the round is stable, so that's not a solution.

The general solution seems to be:
- Find a way to always have `ireq.req` to have the package name.
  Maybe by resolving links (not their dependencies) as a pre-step before doing anything else.
- Implement a proper logic to combine non-editable and editable requirements together in `combine_install_requirements` rather than returning the first editable as before the #1093.

I'm not yet sure how to do either of those at the moment. Any help/input appreciated.

**Changelog-friendly one-liner**: Fix duplicate editable and non-editable requirement duplicate.

##### Contributor checklist

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -56,6 +56,11 @@ def combine_install_requirements(ireqs):
     for ireq in ireqs:
         source_ireqs.extend(getattr(ireq, "_source_ireqs", [ireq]))
 
+    # TODO Find a way (and logic) to:
+    # - Have the ireq.req be set on editables so they end up grouped with non-editable
+    #   (previously a pip resolver side effect on don-deep-copied editable ireq).
+    # - Properly combine the editable and non-editable requirement for the same package.
+
     # deepcopy the accumulator so as to not modify the inputs
     combined_ireq = copy.deepcopy(source_ireqs[0])
     for ireq in source_ireqs[1:]:
@@ -67,6 +72,7 @@ def combine_install_requirements(ireqs):
         combined_ireq.extras = tuple(
             sorted(set(tuple(combined_ireq.extras) + tuple(ireq.extras)))
         )
+        # TODO propably "merge" editable attibutes here.
 
     # InstallRequirements objects are assumed to come from only one source, and
     # so they support only a single comes_from entry. This function breaks this
@@ -221,6 +227,7 @@ class Resolver(object):
 
         """
         for _, ireqs in full_groupby(constraints, key=key_from_ireq):
+            # TODO full_groupby has to group ediatble and non-editable ireqs somehow.
             yield combine_install_requirements(ireqs)
 
     def _resolve_one_round(self):
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -219,6 +219,27 @@ def test_editable_package(pip_conf, runner):
     assert "small-fake-a==0.1" in out.stderr
 
 
+def test_editable_package_without_non_editable_dupplicate(pip_conf, runner):
+    """
+    piptools keeps editable requirement,
+    without also adding a dupplicate "non-editable" requirement variation
+    """
+    fake_package_dir = os.path.join(PACKAGES_PATH, "small_fake_a")
+    fake_package_dir = path_to_url(fake_package_dir)
+    with open("requirements.in", "w") as req_in:
+        req_in.write(
+            "-e " + fake_package_dir +  # require editable fake package
+            "\nsmall_fake_with_unpinned_deps"  # This one also requires small_fake_a
+        )
+
+    out = runner.invoke(cli, ["-n"])
+
+    assert out.exit_code == 0
+    assert fake_package_dir in out.stderr
+    # Shouldn't include a non-editable small-fake-a==<version>.
+    assert "small-fake-a==" not in out.stderr
+
+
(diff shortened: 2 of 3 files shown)


richafrank:
Thanks for the heads up @vphilippon . I knew there had to be *some* reason for that previous behavior! Will give it some thought...

richafrank:
> The general solution seems to be:
    - Find a way to always have ireq.req to have the package name. Maybe by resolving links (not their dependencies) as a pre-step before doing anything else.
    - Implement a proper logic to combine non-editable and editable requirements together in combine_install_requirements rather than returning the first editable as before the #1093.

If the issue with the current combination logic is that we don't get the naming side effect, then seems like the naming pre-step alone would solve that, and we don't need to touch the combination logic. Definitely let me know if I'm missing something.

richafrank:
I ran with that idea in https://github.com/vphilippon/pip-tools/pull/1 and got the tests passing, but I'm not sure about the TODO I added there...

vphilippon:
@richafrank The note about adapting the combination logic is that currently, we don't "merge" the `ireq.link` and `ireq.editable` attribute and such.
I think that might be necessary, but I could be wrong. Or maybe not for that specific issue, but I forsee other issues with the change.

Ex 1: You have something like
```
-e git:<some_repo_url>/myproject@master
-c some_file.txt

# some_file.txt:
-e git:<some_repo_url>/myproject@another_branch
```
The "combination" logic has to "merge" the branch specification, or in some way detect that we have an incompatible requirement by requestion 2 different branches (the same way having `==0.1,==0.2` would be detected as an issue). Although, that that I write this, if we actually have the `ireq.req.specifier` set (like the previous side effect did), then some later part of the resolver would detect the incompatibility.

Ex 2 (my original example case):
```
-e /path/to/small_fake_a
small_fake_with_unpinned_deps  # Has dependency on small_fake_a.
```
We have to make sure that in all cases, the first `ireq` is the editable one, or else we loose the editable information.
I'm also trying to think if there exist a case where and editable requirement could appear at a later stage, or in some way not be the first `ireq` we handle.

If we can rule that out, and are sure that an editable install will always be the first one and serve as the "source" for the `combined_req`, they we are ok.
A part of me feels like not relying on that condition, and handle that in a generic way, but that might be my perfectionism speking :P .

richafrank:
> We have to make sure that in all cases, the first ireq is the editable one, or else we loose the editable information.
I'm also trying to think if there exist a case where and editable requirement could appear at a later stage, or in some way not be the first ireq we handle.

> If we can rule that out, and are sure that an editable install will always be the first one and serve as the "source" for the combined_req, they we are ok.

We could have `_group_constraints` sort the editable ireq first - I had actually done that [originally](https://github.com/jazzband/pip-tools/pull/1093#discussion_r399687223), but we removed it when we didn't think it was necessary. Would it be the case then that each time we combine them, we'll take the first editable's information, even if it appeared at a later stage?

> A part of me feels like not relying on that condition, and handle that in a generic way, but that might be my perfectionism speking :P .

As you said, we could instead "merge" the `editable` and `link` attributes, though I'm not sure if there's other state that would need to come along with them...

richafrank:
Thinking about other cases to handle, I added a failing test to my [PR](https://github.com/vphilippon/pip-tools/pull/1/commits/e48467cac499b473912cdc045c33c8cdf40d752b) with:

```
# reqs.in
-c constraints.txt
small_fake_with_unpinned_deps"  # requires small_fake_a

# constraints.txt
-e /path/to/small_fake_a
```

My attempted fix for that causes`test_editable_package_in_constraints[True]` to fail. Ultimately, I keep having to work around:
1. our reliance on mutating ireqs (e.g. their req/name) that are cached in a dict `PyPIRepository._dependencies_cache` by identity, i.e. the copies from `combine_install_requirements` don't hash the same;
1. the pip resolver doesn't return any results for dependencies if `req_to_install` is a constraint or already `prepared`.

atugushev:
FTR, I pulled out this PR from the milestone to do a release.

richafrank:
I just opened https://github.com/jazzband/pip-tools/pull/1132 - take a look...

PR #1117, opened 2020-04-25
Refactor to for-else block

atugushev:
I've [reconsidered](https://www.youtube.com/watch?v=8P7lXLXR_3c) my opinion about `for-else` statement.

Addresses https://github.com/jazzband/pip-tools/pull/949#discussion_r412622027


atugushev:
@AndydeCleyre thanks!

Issue #811, opened 2019-05-10
pip-compile 3.7.0 no longer honors --no-index


jbergknoff-rival:
pip-compile no longer honors the --no-index argument, so the index URL (containing repository credentials) is present in the header of the generated file. This is a regression between pip-tools 3.6.1 and 3.7.0. My guess is it's related to this change: https://github.com/jazzband/pip-tools/pull/800.

##### Environment Versions

1. OS Type: **Linux**
1. Python version: `$ python -V`: **3.6.8**
1. pip version:  `$ pip --version`: **19.0.3**
1. pip-tools version: `$ pip-compile --version`: **3.7.0**

##### Steps to replicate

1. `pip-compile --no-index --output-file requirements/runtime.txt requirements/runtime.in`
2. Look at the header in requirements/runtime.txt

##### Expected result

The header of requirements/runtime.txt shouldn't have an index URL containing repository credentials

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --no-index --output-file requirements/runtime.txt requirements/runtime.in
#
```

##### Actual result

The index URL is present.

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile --index-url=... --no-index --output-file=requirements/runtime.txt requirements/runtime.in
#
```

atugushev:
Hello @jbergknoff-rival,

Thanks for the report! Where did you get the `index-url`? Is it configured in `pip.conf`? Can't reproduce locally though.

jbergknoff-rival:
Hi, it's set as an environment variable `PIP_INDEX_URL`.

**3.7.0**

```
$ docker run -it --rm python:3.6.8-alpine3.9 sh
/ # pip install pip-tools==3.7.0
Collecting pip-tools==3.7.0
  Downloading https://files.pythonhosted.org/packages/96/43/34412d316bdbf1cd9c9a0e487138b40db4e1e11212cee2d46440b6b49b08/pip_tools-3.7.0-py2.py3-none-any.whl
Collecting click>=6 (from pip-tools==3.7.0)
  Downloading https://files.pythonhosted.org/packages/fa/37/45185cb5abbc30d7257104c434fe0b07e5a195a6847506c074527aa599ec/Click-7.0-py2.py3-none-any.whl (81kB)
    100% |████████████████████████████████| 81kB 3.2MB/s 
Collecting six (from pip-tools==3.7.0)
[...]

PR #1130, opened 2020-05-02
Deprecate --index/--no-index in pip-compile

### LINKED ISSUE #373: --index/--no-index flag means different things in pip-compile and pip-sync
`--no-index` is an existing `pip` internal flag. It does the same thing on `pip-sync` (ignore default pypi url), and something different on `pip-compile` (don't annotate index url into `requirements.txt`).

**Proposal: Change `--index/--no-index` in `pip-compile` to `--annotate-index/--no-annotate-index`.**

Achieves:
- Consistency (between `pip-sync` and `pip-compile`)
- Consistency (with `pip --no-index`)
- Consistency (with other flag `--annotate/--no-annotate`)

atugushev:
To avoid ambiguity of `pip-sync --no-index`.

Resolves #373.

**Changelog-friendly one-liner**: Deprecate `--index/--no-index` in favor of `--emit-index-url/--no-emit-index-url` options in `pip-compile`.

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -7,6 +7,7 @@ import sys
 import tempfile
 import warnings
 
+from click import Command
 from click.utils import safecall
 from pip._internal.commands import create_command
 from pip._internal.req.constructors import install_req_from_line
@@ -30,7 +31,30 @@ install_command = create_command("install")
 pip_defaults = install_command.parser.get_default_values()
 
 
-@click.command()
+class CompileCommand(Command):
+    def __init__(self, *awrgs, **kwargs):
+        super(CompileCommand, self).__init__(*awrgs, **kwargs)
+        self._os_args = None
+
+    def parse_args(self, ctx, args):
+        """
+        Override `parse_args` to store args since click doesn't provide
+        this information in the context.
+        """
+        self._os_args = set(args)
+        return super(CompileCommand, self).parse_args(ctx, args)
+
+    def has_arg(self, arg_name):
+        """
+        Detect whether given arg name is present in the argument part of `sys.argv`.
+        """
+        command_options = {option.name: option for option in self.params}
+        option = command_options[arg_name]
+        args = set(option.opts + option.secondary_opts)
+        return bool(self._os_args & args)
+
+
+@click.command(cls=CompileCommand)
 @click.version_option()
 @click.pass_context
 @click.option("-v", "--verbose", count=True, help="Show more output")
@@ -95,7 +119,7 @@ pip_defaults = install_command.parser.get_default_values()
 @click.option(
     "--index/--no-index",
     is_flag=True,
-    default=None,
+    default=True,
     help="DEPRECATED: Add index URL to generated file",
 )
 @click.option(
@@ -183,7 +207,7 @@ pip_defaults = install_command.parser.get_default_values()
 @click.option(
     "--emit-index-url/--no-emit-index-url",
     is_flag=True,
-    default=None,
+    default=True,
     help="Add index URL to generated file",
 )
 def cli(
@@ -254,20 +278,18 @@ def cli(
         # Close the file at the end of the context execution
         ctx.call_on_close(safecall(output_file.close_intelligently))
 
-    if index is not None and emit_index_url is not None:
+    if cli.has_arg("index") and cli.has_arg("emit_index_url"):
         raise click.BadParameter(
             "--index/--no-index and --emit-index-url/--no-emit-index-url "
             "are mutual exclusive."
         )
-    elif index is not None:
+    elif cli.has_arg("index"):
         warnings.warn(
             "--index and --no-index are deprecated and will be removed "
             "in future versions. Use --emit-index-url/--no-emit-index-url instead.",
             category=FutureWarning,
         )
         emit_index_url = index
-    elif emit_index_url is None:
-        emit_index_url = True
 
     ###
     # Setup
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -5,6 +5,7 @@ import os
 import shlex
 import sys
 import tempfile
+import warnings
 
 from click.utils import safecall
 from pip._internal.commands import create_command
@@ -94,8 +95,8 @@ pip_defaults = install_command.parser.get_default_values()
 @click.option(
     "--index/--no-index",
     is_flag=True,
-    default=True,
-    help="Add index URL to generated file",
+    default=None,
+    help="DEPRECATED: Add index URL to generated file",
 )
 @click.option(
     "--emit-trusted-host/--no-emit-trusted-host",
@@ -179,6 +180,12 @@ pip_defaults = install_command.parser.get_default_values()
     type=click.Path(file_okay=False, writable=True),
 )
 @click.option("--pip-args", help="Arguments to pass directly to the pip command.")
+@click.option(
+    "--emit-index-url/--no-emit-index-url",
+    is_flag=True,
+    default=True,
+    help="Add index URL to generated file",
+)
 def cli(
     ctx,
     verbose,
@@ -207,6 +214,7 @@ def cli(
     emit_find_links,
     cache_dir,
     pip_args,
+    emit_index_url,
 ):
     """Compiles requirements.txt from requirements.in specs."""
     log.verbosity = verbose - quiet
@@ -245,6 +253,14 @@ def cli(
 
         # Close the file at the end of the context execution
         ctx.call_on_close(safecall(output_file.close_intelligently))
+
+    if index is not None:
+        warnings.warn(
+            "--index and --no-index are deprecated and will be removed "
+            "in future versions. Use --emit-index-url/--no-emit-index-url instead.",
+            category=FutureWarning,
+        )
+        emit_index_url = index
 
     ###
     # Setup
@@ -410,7 +426,7 @@ def cli(
         click_ctx=ctx,
         dry_run=dry_run,
         emit_header=header,
-        emit_index=index,
+        emit_index_url=emit_index_url,
         emit_trusted_host=emit_trusted_host,
         annotate=annotate,
         generate_hashes=generate_hashes,
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -31,22 +31,19 @@ install_command = create_command("install")
 pip_defaults = install_command.parser.get_default_values()
 
 
-class CompileCommand(Command):
-    def __init__(self, *awrgs, **kwargs):
-        super(CompileCommand, self).__init__(*awrgs, **kwargs)
-        self._os_args = None
+class BaseCommand(Command):
+    _os_args = None
 
     def parse_args(self, ctx, args):
         """
-        Override `parse_args` to store args since click doesn't provide
-        this information in the context.
+        Override base `parse_args` to store the argument part of `sys.argv`.
         """
         self._os_args = set(args)
-        return super(CompileCommand, self).parse_args(ctx, args)
+        return super(BaseCommand, self).parse_args(ctx, args)
 
     def has_arg(self, arg_name):
         """
-        Detect whether given arg name is present in the argument part of `sys.argv`.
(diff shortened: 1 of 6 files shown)


atugushev:
Hello @auvipy. Could you please do merging `master` after the review is done and right before merging PR to the master to avoid redundant noise. 🙏

atugushev:
@auvipy and I'd prefer to rebase (not to merge) to keep clean the commits history.

atugushev:
I've just thought about the case when `--index` and `--no-emit-index-url` conflict to each other. Shouldn't we throw the error?

atugushev:
@hugovk @webknjaz 

As a follow up of the [twitter thread](https://twitter.com/metalikus/status/1255968363051810816) I wonder what do you think about the deprecation technique in this PR?

hugovk:
Looks about right!

In our project, we decided not to use `PendingDeprecationWarning` first, and go straight to `DeprecationWarning` to increase visibility from the start.

piptools/scripts/compile.py
```diff
@@ -246,6 +255,14 @@ def cli(
         # Close the file at the end of the context execution
         ctx.call_on_close(safecall(output_file.close_intelligently))
 
+    if index is not None:
+        warnings.warn(
+            "--index and --no-index are deprecated and will be removed "
+            "in future versions. Use --emit-index-url/--no-emit-index-url instead.",
+            category=PipToolsPendingDeprecationWarning,
```
webknjaz, line 262: Always set stacklevel because you want to point to the end-user's code where they should stop using it, not here.
```suggestion
            category=PipToolsPendingDeprecationWarning,
            stacklevel=2,
```

piptools/warnings.py
```diff
@@ -0,0 +1,29 @@
...
+class PipToolsBaseWarning:
+    pass
+
+
+class PipToolsPendingDeprecationWarning(PipToolsBaseWarning, PendingDeprecationWarning):
+    """
+    Deprecates things in future versions.
+    """
+
+    pass
```
webknjaz, line 16: This is redundant, a docstring is already a no-op node in the class.
```suggestion
```

piptools/warnings.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class PipToolsDeprecationWarning(PipToolsBaseWarning, DeprecationWarning):
+    """
+    Deprecates things in the next or a specific version.
+    """
+
+    pass
```
webknjaz, line 24:
```suggestion
```

piptools/warnings.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+# Setup if warning filters are not configured
+if not sys.warnoptions:
+    warnings.simplefilter("once", category=PipToolsBaseWarning)
```
webknjaz, line 29: I'm not a fan of import-time logic execution. You should instead add a function like `setup_warnings()` and call it first thing in `main()`. This way you're more in control of how and when the code is invoked.

tests/test_utils.py
```diff
@@ -327,7 +327,7 @@ def test_get_compile_command_sort_args(tmpdir_cwd):
         pass
 
     args = [
-        "--no-index",
```
webknjaz, line 330: I think that during the deprecation period, the old flag should still be tested and behave like the new one.

piptools/warnings.py
atugushev: TBH I had the same thoughts and that was the first implementation with the same naming :D But `pip-tools` doesn't have `main()` or such. In this case, `setup_warnings()` should be called in `piptools.scripts.compile.cli()` and `piptools.scripts.sync.cli()` separately. A few more options in my mind:
- decorator like `@setup_warnings` for `cli()` functions (which is not bad idea)
- refactor cli scripts to have single entry point `main()` and call there `setup_warnings()` (would touch many lines I guess)
- anything more?

With the current implementation main idea is if one day there will be nothing to deprecated for a while, thus there will not be any imports of `piptools.warnings` module (except tests), so warnings filters won't be touched. What do you think?

tests/test_utils.py
atugushev: It's tested in `test_index_option`. The `test_get_compile_command_sort_args ` is not about `--index`'s behaviour actually, it tests sorting args in `get_compile_command()`.

piptools/scripts/compile.py
atugushev: AFAIK it's relevant for libraries, not for cli apps, see the message otherwise with `stacklevel=2`:

```console
$ pip-compile --extra-index-url=https://test.pypi.org/simple/ --no-emit-index-url --index
/Users/albert/Projects/pip-tools/.venv/lib/python3.8/site-packages/click/decorators.py:21: PipToolsPendingDeprecationWarning: --index and --no-index are deprecated and will be removed in future versions. Use --emit-index-url/--no-emit-index-url instead.
#
# To update, run:
#
#    pip-compile --extra-index-url=https://test.pypi.org/simple/ --index --no-emit-index-url
#
--extra-index-url https://test.pypi.org/simple/

six==1.14.0               # via -r requirements.in
```

Current;y message is:

```
pip-compile --extra-index-url=https://test.pypi.org/simple/ --no-emit-index-url --index
/Users/albert/Projects/pip-tools/piptools/scripts/compile.py:259: PipToolsPendingDeprecationWarning: --index and --no-index are deprecated and will be removed in future versions. Use --emit-index-url/--no-emit-index-url instead.
  warnings.warn(
```

piptools/warnings.py
atugushev: Good call! Thanks.

piptools/warnings.py
atugushev: Good call! Thanks.

piptools/warnings.py
webknjaz: A decorator sounds good. I waw thinking about it too.

piptools/scripts/compile.py
webknjaz: Well, maybe you could detect somehow CLI vs import. Can you guarantee that nobody is using the APIs directly via imports?

atugushev: I think nobody uses `piptools.scripts.compile.cli` and calls it because it won't work. AFAIK The only way is to invoke `pip-compile` or `python -m piptools compile` or use click's `CliRunner()`.

piptools/warnings.py
atugushev: Okay, let's go ahead with the decorator then.

atugushev: Addressed in 5f8aa2f.

piptools/warnings.py
atugushev: Addressed in 5f8aa2f.

piptools/warnings.py
atugushev: Addressed in 5f8aa2f.

piptools/scripts/compile.py
atugushev: Kindly ping. What do you think?

webknjaz: I'm not concerned too much. Ignore it. After all, done is better than perfect :)

atugushev: Thanks. Nice rule of thumb! 👍

atugushev:
@AndydeCleyre 

I wonder what would you think before we merge this?

> I've just thought about the case when `--index` and `--no-emit-index-url` conflict to each other. Shouldn't we throw the error or prefer `--no-emit-index-url`?

Especially for this case.

AndydeCleyre:
Nice work! Three notes/questions:

1.
> I've just thought about the case when `--index` and `--no-emit-index-url` conflict to each other. Shouldn't we throw the error or prefer `--no-emit-index-url`?

I say either throw an error, or prefer the *right-most* flag (not necessarily the not-deprecated one).

2. Slightly off topic: I don't like the click-style `--flag/--no-flag` options in general; is there any way for a user to know which is the default? It doesn't seem to be indicated in the `--help` output.

3. I think we only need one type of warning so far, and should change the name and base for it.

Why bother with a `PipToolsPendingDeprecationWarning`? You may be using the term "deprecation" to mean something closer to "unsupported." The docstring for `PipToolsDeprecationWarning` says:

> Deprecates things in the next or a specific version.

but aren't we trying to warn about things that are *currently* deprecated, to be unsupported/removed in the future?

The Python docs describe their own [`PendingDeprecationWarning`](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning) as:

> Base class for warnings about features which are obsolete and expected to be deprecated in the future, but are not deprecated at the moment.

> This class is rarely used as emitting a warning about a possible upcoming deprecation is unusual, and `DeprecationWarning` is preferred for already active deprecations.

Then the description of their `DeprecationWarning` makes me think that isn't the best analogue either:

> Base class for warnings about deprecated features when those warnings are intended for other Python developers.

We are really targeting our *users* here, although the line between "users" and "Python developers" is not so obvious for a project like this.

I think what we're going for is a much better fit with Python's [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning):

> Base class for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

Note that this is not ignored by default as the other two are.

Seeing this at the [warnings docs](https://docs.python.org/3/library/warnings.html#warning-categories) made it clear that the conceptual classifications have changed pretty recently, making `FutureWarning` decidedly more appropriate here:

> Changed in version 3.7: Previously `DeprecationWarning` and `FutureWarning` were distinguished based on whether a feature was being removed entirely or changing its behaviour. They are now distinguished based on their intended audience and the way they’re handled by the default warnings filters.

atugushev:
@AndydeCleyre 

Thanks for the analysis as always and sorry for delayed answer. Regarding to the docs:

> FutureWarning — Base category for warnings about deprecated features when those warnings are intended **for end users of applications** that are written in Python.
 
it seems `FutureWarning ` is our candidate since it's intended to warn end-users of the application and it's visible default so there is no need in `setup_warning()`.

atugushev:
@AndydeCleyre 

I came up with the solution acae7bf8236ad7266438693ba49814cdc2081ba6. What do you think?

atugushev:
@AndydeCleyre kindly ping

piptools/scripts/compile.py
```diff
@@ -246,6 +278,19 @@ def cli(
         # Close the file at the end of the context execution
         ctx.call_on_close(safecall(output_file.close_intelligently))
 
+    if cli.has_arg("index") and cli.has_arg("emit_index_url"):
+        raise click.BadParameter(
+            "--index/--no-index and --emit-index-url/--no-emit-index-url "
+            "are mutual exclusive."
```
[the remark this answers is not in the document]
atugushev: Addressed in f7e312.

tests/test_cli_compile.py
```diff
@@ -152,20 +153,64 @@ def test_trusted_host(pip_conf, runner):
...
         pass
-    out = runner.invoke(cli, ["-v", "--no-emit-find-links"])
-    assert "--find-links" not in out.stderr
+
+    out = runner.invoke(cli, options)
+
+    assert out.exit_code == 2
+    assert (
+        "--index/--no-index and --emit-index-url/--no-emit-index-url "
+        "are mutual exclusive"
```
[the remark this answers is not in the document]
atugushev: Addressed in f7e3121ab1df69d888f1d5e112ba04bcbd326946.

piptools/scripts/compile.py
```diff
@@ -29,7 +31,30 @@
 pip_defaults = install_command.parser.get_default_values()
 
 
-@click.command()
+class CompileCommand(Command):
+    def __init__(self, *awrgs, **kwargs):
+        super(CompileCommand, self).__init__(*awrgs, **kwargs)
```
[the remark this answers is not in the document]
atugushev: Aw, nice catch! Fixed in 6320c81.

atugushev:
@AndydeCleyre 

Thanks! I believe `cli.has_arg(...)` would help to refactor and simplify `get_compile_command`, and also to fix #811 where we need to track os args, so this change is definitely useful!

PR #1132, opened 2020-05-03
Fix grouping of requirements to handle editables better

richafrank:
Fixes https://github.com/jazzband/pip-tools/pull/1115 , or at least the test cases added there and in https://github.com/vphilippon/pip-tools/pull/1

- Ask repository for dependencies early for the side effect of assigning each nameless ireq a name. This means the correct ireqs will be grouped together.
- Sort editables first within grouping, so that we start with all their attributes.
- Update the dependency cache when we make a copy of an ireq. Otherwise, the pip resolver will tell us that the ireq has no dependencies because it was already prepared when we got its name. This was done because we hash ireqs in the cache based on identity. Alternatively, we could change that notion of identity, but that felt less safe, trying to determine the correct hash function.

**Changelog-friendly one-liner**:  Fix grouping of editables and non-editables

##### Contributor checklist

### DIFF
--- a/piptools/repositories/base.py
+++ b/piptools/repositories/base.py
@@ -44,3 +44,11 @@ class BaseRepository(object):
         """
         Monkey patches pip.Wheel to allow wheels from all platforms and Python versions.
         """
+
+    def copy_ireq_dependencies(self, source, dest):
+        """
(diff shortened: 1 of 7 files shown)


andersk:
This last commit introduced a regression when using the same zip file with different subdirectories (#1155).

PR #1147, opened 2020-05-16
Fix broken cron CI job

atugushev:
See https://github.com/jazzband/pip-tools/actions/runs/106207484

See fix demonstration with a02385f in https://github.com/jazzband/pip-tools/actions/runs/106447309


webknjaz:
>           # TODO: fix test_realistic_complex_sub_dependencies test on Windows

btw this could be an xfailing test instead

atugushev:
@webknjaz good point! Here is the tracking issue #1148.

atugushev:
Thanks, @webknjaz!

PR #1119, opened 2020-04-25
Fix a bug where pip-compile would ignore some of versions if prefer binary distributions

atugushev:
Fixes #1118.

Relevant lines of code: https://github.com/jazzband/pip-tools/blob/9865035944f4a4d2c9cf3dc6ba9c77a98e9fc989/piptools/repositories/pypi.py#L125-L127

The parameter `unique=True` was the root of the issue and the cause of losing installation candidates. See examples of `lookup_table()` results below.

#### Before

```python
candidates_by_version = {
    ...
    <Version('0.2.2')>: <InstallationCandidate('entrypoints', <Version('0.2.2')>, <Link https://files.pythonhosted.org/packages/f8/ad/0e77a853c745a15981ab51fa9a0cb4eca7a7a007b4c1970106ee6ba01e0c/entrypoints-0.2.2-py2.py3-none-any.whl#sha256=0a0685962ee5ac303f470acbb659f0f97aef5b9deb6b85d059691c706ef6e45e (from https://pypi.org/simple/entrypoints/) (requires-python:>=2.7)>)>,
    <Version('0.2.3')>: <InstallationCandidate('entrypoints', <Version('0.2.3')>, <Link https://files.pythonhosted.org/packages/27/e8/607697e6ab8a961fc0b141a97ea4ce72cd9c9e264adeb0669f6d194aa626/entrypoints-0.2.3.tar.gz#sha256=d2d587dde06f99545fb13a383d2cd336a8ff1f359c5839ce3a64c917d10c029f (from https://pypi.org/simple/entrypoints/) (requires-python:>=2.7)>)>,
    <Version('0.3')>: <InstallationCandidate('entrypoints', <Version('0.3')>, <Link https://files.pythonhosted.org/packages/b4/ef/063484f1f9ba3081e920ec9972c96664e2edb9fdc3d8669b0e3b8fc0ad7c/entrypoints-0.3.tar.gz#sha256=c70dd71abe5a8c85e55e12c19bd91ccfeec11a6e99044204511f9ed547d48451 (from https://pypi.org/simple/entrypoints/) (requires-python:>=2.7)>)>
}
```

Note `entrypoints-0.3.tar.gz` in `<Version('0.3')>` .

#### After

```python

candidates_by_version = {
    <Version('0.2.2')>: {
[...]

PR #1149, opened 2020-05-19
Avoid URL requirement name collisions

### LINKED ISSUE #1145: Multiple entries in requirements.in with the same file cause collision
When a requirements.in contains multiple entries with the same file name, there is a collision when they are downloaded. e.g.

requirements.in contents:
`https://github.com/cloudify-cosmo/cloudify-common/archive/master.zip#egg=cloudify-common[dispatcher]`
`https://github.com/cloudify-cosmo/cloudify-cli/archive/master.zip#egg=cloudify`

This results in a warning that the cloudify 5.1.0.dev1 does not provide the extra 'dispatcher',
followed by an error that no version could be found for cloudify-common==5.1.0.dev1

#### Environment Versions

1. Ubuntu 18.04
1. Python version: Python 2.7.17  # Though by the behaviour I expect this to still cause problems when we switch this to Python 3.
1. pip version:  pip 20.1
1. pip-tools version: pip-compile, version 5.1.2

#### Steps to replicate

1. Create an empty dir, and cd to it
2. Make a requirements.in file in this dir, containing:
`https://github.com/cloudify-cosmo/cloudify-common/archive/master.zip#egg=cloudify-common[dispatcher]`
`https://github.com/cloudify-cosmo/cloudify-cli/archive/master.zip#egg=cloudify`
3. Run `pip-compile`

#### Expected result

requirements.txt is produced.

#### Actual result

Errors abound (see above for specific errors).

geokala:
Fixes #1145.

This avoids cache collisions when using URLs which both end with the same filename where a later dependency depends on an earlier one, e.g.
`file:///tmp/mydep1/master.zip`
`file:///tmp/mydep2/master.zip`
(where mydep2 depends on mydep1)

**Changelog-friendly one-liner**: Avoid name collisions with URL based requirements.

##### Contributor checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -215,7 +215,7 @@ class PyPIRepository(BaseRepository):
                 # using git-checkout-index, which gets rid of the .git dir.
                 download_dir = None
             else:
-                download_dir = self._download_dir
+                download_dir = self._get_download_path(ireq)
                 if not os.path.isdir(download_dir):
                     os.makedirs(download_dir)
             if not os.path.isdir(self._wheel_download_dir):
@@ -288,6 +288,20 @@ class PyPIRepository(BaseRepository):
             return data
         return None
 
+    def _get_download_path(self, ireq):
+        """
+        Determine the download dir location in a way which avoids name
+        collisions.
+        """
+        if ireq.link:
+            salt = hashlib.sha224(ireq.link.url_without_fragment.encode()).hexdigest()
+            # Nest directories to avoid running out of top level dirs on some FS
+            # (see pypi _get_cache_path_parts, which inspired this)
+            salt = [salt[:2], salt[2:4], salt[4:6], salt[6:]]
+            return os.path.join(self._download_dir, *salt)
+        else:
+            return self._download_dir
+
     def get_hashes(self, ireq):
         """
         Given an InstallRequirement, return a set of hashes that represent all
@@ -308,7 +322,7 @@ class PyPIRepository(BaseRepository):
                 # Directly hash URL requirements.
                 # URL requirements may have been previously downloaded and cached
                 # locally by self.resolve_reqs()
-                cached_path = os.path.join(self._download_dir, link.filename)
+                cached_path = os.path.join(self._get_download_path(ireq), link.filename)
                 if os.path.exists(cached_path):
                     cached_link = Link(path_to_url(cached_path))
                 else:
--- a/tests/test_repository_pypi.py
+++ b/tests/test_repository_pypi.py
@@ -347,3 +347,49 @@ def test_get_project__handles_404(from_line, tmpdir, monkeypatch, pypi_repositor
 
     actual_data = pypi_repository._get_project(ireq)
     assert actual_data is None
+
+
+def test_name_collision(from_line, pypi_repository, make_package, make_sdist, tmpdir):
+    """
+    Test to ensure we don't fail if there are multiple URL-based requirements
+    ending with the same filename where later ones depend on earlier, e.g.
+    https://git.example.com/requirement1/master.zip#egg=req_package_1
+    https://git.example.com/requirement2/master.zip#egg=req_package_2
+    In this case, if req_package_2 depends on req_package_1 we don't want to
+    fail due to issues such as caching the requirement based on filename.
+    """
+    packages = {
+        "test_package_1": make_package("test_package_1", version="0.1"),
+        "test_package_2": make_package(
+            "test_package_2", version="0.1", install_requires=["test-package-1"]
+        ),
+    }
+
+    for pkg_name, pkg in packages.items():
+        pkg_path = os.path.join(tmpdir, pkg_name)
+        os.mkdir(pkg_path)
+
+        make_sdist(pkg, pkg_path, "--formats=zip")
+
+        os.rename(
+            os.path.join(pkg_path, "{}-{}.zip".format(pkg_name, "0.1")),
+            os.path.join(pkg_path, "master.zip"),
+        )
+
+    name_collision_1 = (
+        "file://"
+        + str(os.path.join(tmpdir, "test_package_1", "master.zip"))
+        + "#egg=test_package_1"
+    )
+    ireq = from_line(name_collision_1)
+    pypi_repository.get_dependencies(ireq)
+
+    name_collision_2 = (
+        "file://"
+        + str(os.path.join(tmpdir, "test_package_2", "master.zip"))
+        + "#egg=test_package_2"
+    )
+    ireq = from_line(name_collision_2)
+    reqs = pypi_repository.get_dependencies(ireq)
+    assert len(reqs) == 1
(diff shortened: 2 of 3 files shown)


tests/test_repository_pypi.py
```diff
@@ -347,3 +347,25 @@ def mock_get(*args, **kwargs):
...
     actual_data = pypi_repository._get_project(ireq)
     assert actual_data is None
+
+
+def test_name_collision(from_line, pypi_repository):
+    packages_base = os.path.join(os.path.dirname(__file__), "test_data", "packages")
+
+    name_collision_1 = (
+        "file://"
+        + str(os.path.join(packages_base, "fake_name_collision_1", "master.zip"))
```
atugushev, line 357: #1119 introducing new fixtures to create test packages rather than add binary archives to the VCS. I hope soon we refactor `test_data` in favor of functional fixtures to make tests more clear since it's difficult to maintain and review binary files. What do you think @vphilippon @AndydeCleyre?

geokala:
Re: Cache directory, good point. I've factored out the logic to a _get_download_path function and used it in both places it appears to be needed.
I've not introduced a specific test for this as I believe it constitutes internal implementation detail rather than external behaviour that should be asserted upon, but let me know if a test is required.

tests/test_repository_pypi.py
geokala: Agreed that this would be nicer if it weren't a couple of zips. Let me know if there's a better way to do it that fits with your general approach.

I explicitly wanted to avoid:
1. External dependencies.
2. Shelling out (or similar) to create the zips.

piptools/repositories/pypi.py
```diff
@@ -287,6 +287,17 @@ def _get_project(self, ireq):
             return data
         return None
 
+    def _get_download_path(self, ireq):
+        """
+        Determine the download dir location in a way which avoids name
+        collisions.
+        """
+        if ireq.link:
+            salt = hashlib.sha256(ireq.link.url_without_fragment.encode()).hexdigest()
```
atugushev, line 296: I've researched pip codebase and have found this [function](https://github.com/pypa/pip/blob/9999f0ec098f9dc4687eecd118f52a3233e64e8c/src/pip/_internal/cache.py#L89-L120). It probably makes sense to do the same for caching files. What do you think?

geokala: Sure, that looks like it'd work nicely. Only issue would be that it's depending on another pip internal function, so increases the risk of breakage every time they update.
If you're comfortable with that then I can switch it to use that this evening.

Otherwise, I can grab the nesting part of it to give the same behaviour and avoid that dependency.

What's your preference?

atugushev: I would grab the nesting part.

tests/test_repository_pypi.py
atugushev: #1119 is merged, so you can use the fixtures `make_package` and `make_sdist` now.

piptools/repositories/pypi.py
geokala: OK, that'll be in the next push. Just looking at the test now.

tests/test_repository_pypi.py
geokala: OK, I've shifted to using those, and confirmed the test will fail without the new code.

atugushev: Looks good!

AndydeCleyre:
Can we get a nice explanation of the problem being solved, as the docstring for the test?

geokala:
@AndydeCleyre Sure, I've just added that- hopefully it's not too verbose?

geokala:
Not sure why CI is failing at the moment, will take a look when it deigns to give me logs.

atugushev:
@geokala 

> Not sure why CI is failing at the moment, will take a look when it deigns to give me logs.

I've fixed it in https://github.com/geokala/pip-tools/pull/1. You could merge this if it works.

atugushev:
@geokala 

FYI, also fixed test `test_name_collision` due to changes in https://github.com/geokala/pip-tools/pull/1/commits/2c4c5719bc2be19ce8fa61a39df6e78ab3191ce8.

geokala:
@atugushev Awesome, thanks for that- makes sense, so I've merged it into my PR.

geokala:
Looks like CI is failing on 2.7 pip latest (I think?):
  File "/tmp/easy_install-P_Otzu/jaraco.windows-5.0.0/temp/easy_install-B_XSyo/path.py-12.4.0/temp/easy_install-kjgCqJ/more-itertools-8.3.0/more_itertools/__init__.py", line 1, in <module>
  File "/tmp/easy_install-P_Otzu/jaraco.windows-5.0.0/temp/easy_install-B_XSyo/path.py-12.4.0/temp/easy_install-kjgCqJ/more-itertools-8.3.0/more_itertools/more.py", line 482
    yield from iterable
             ^
SyntaxError: invalid syntax

=================================== log end ====================================
ERROR: FAIL could not package project - v = InvocationError(u'/opt/hostedtoolcache/Python/2.7.18/x64/bin/python setup.py sdist --formats=zip --dist-dir /home/runner/work/pip-tools/pip-tools/.tox/dist', 1)

Has pip already stopped supporting 2.7 for make_sdist or something?

geokala:
A local run of `tox --notest -p auto --parallel-live` doesn't give me that error, though of course my environment doesn't match travis.

geokala:
So yeah, looks like more-itertools is being used by easy_install but doesn't support python2.7 any more- I'm guessing I already had an older version of easy_install installed so didn't run into this problem.
Presumably this means that something needs pinning- easy_install or more-itertools- and presumably it wants pinning in tox??

atugushev:
Unfortunately, recent `setuptools-scm-4.0.0` broke our CI. Working on a fix.

atugushev:
@geokala 

Please rebase onto `master` it should be fixed now.

geokala:
Rebased. Fingers crossed.

piptools/repositories/pypi.py
atugushev: Could you update `_get_download_path` in accordance with this [function](https://github.com/pypa/pip/blob/9999f0ec098f9dc4687eecd118f52a3233e64e8c/src/pip/_internal/cache.py#L89-L120)?

geokala: Can do- but what's the objective?
I can do a straight copy of that function, which will not break on changes to that function's name, but will break if there are changes to the vendored packages that pip uses. Otherwise, what's currently in this PR is a partial copy which should address the concerns but may cause slight cache bloat.
Is the preference here to reduce cache bloat but at the possible expense of stability? If so, I could change it to call that function directly?

atugushev: The intention is to kill two birds with one stone. But, I don't want to block this improvement due to mysterious bugs and rather approve this as is. Thanks for your contribution!

atugushev:
@geokala 

Great first-time contribution 🚀 Thanks!

geokala:
No worries, and thanks for your help!

PR #1151, opened 2020-05-25
Fix py27 failing jobs

atugushev:
Refs https://github.com/pypa/setuptools_scm/issues/442

1. Upgrade `virtualenv` fixed [issue](https://github.com/jazzband/pip-tools/runs/705017713?check_suite_focus=true#step:5:155) with `path.py`

```
File "/tmp/easy_install-KNyzCx/jaraco.windows-5.0.0/temp/easy_install-m7aKTl/path.py-12.4.0/temp/easy_install-LMhEDY/more-itertools-8.3.0/more_itertools/more.py", line 482
    yield from iterable
             ^
SyntaxError: invalid syntax
```

2. Released `setuptools-4.1.0` fixed [issue](https://ci.appveyor.com/project/jazzband/pip-tools/builds/33084292/job/7u6m2qjorh2xkyx3#L85)  with:

```python
File "c:\users\vssadm~1\appdata\local\temp\pip-install-yxxmaf\setuptools-scm\src\setuptools_scm\git.py", line 9, in <module>
    from os.path import samefile
ImportError: cannot import name samefile
```

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,7 +35,7 @@ matrix:
 
 install:
     - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-    - pip install tox
+    - python -m pip install -U tox virtualenv
 
 build: false
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,8 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install tox
-        run: pip install tox
+      - name: Install test dependencies
+        run: python -m pip install -U tox virtualenv
       - name: Prepare test environment
         run: tox --notest -p auto --parallel-live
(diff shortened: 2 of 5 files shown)


atugushev:
Oof, keep `py27` alive till Jan 2021 would be tough... 😞

AndydeCleyre:
I'm all for upgrading past these problems. Only 1 and a half questions:

- is this this adding `virtualenv` to py3 workflows when it wasn't there before?
  - if so, is that necessary?

atugushev:
@AndydeCleyre 

It's not necessary for py3, but as we seen before (#984) it's better to install fresh `virtualenv`. Just in case.

AndydeCleyre:
If `virtualenv` is not used at all by our py3 workflow, I don't think we should install it in our py3 workflow.

atugushev:
@AndydeCleyre 

`tox` uses `virtualenv` as a dependency.

atugushev:
@AndydeCleyre 

Thank you so much! 🙏

PR #1169, opened 2020-06-28
Add py3.9 to CI workflow

atugushev:
Uses https://github.com/deadsnakes/action to install python 3.9.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
         os:
           - Ubuntu
           - Windows
-          - MacOS
+          - macOS
         python-version:
           - 3.8
           - 2.7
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -29,16 +29,29 @@ jobs:
         pip-version:
           - "latest"
           - "20.0"
+        include:
+          - os: Ubuntu
```
webknjaz, line 33: How about
```suggestion
          - os: Ubuntu-20.04
```
(latest is 16.04 atm)

.github/workflows/ci.yml
```diff
@@ -29,16 +29,29 @@ jobs:
...
+
     env:
       PY_COLORS: 1
       TOXENV: pip${{ matrix.pip-version }}-coverage
       TOX_PARALLEL_NO_SPINNER: 1
     steps:
       - uses: actions/checkout@master
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python ${{ matrix.python-version }} from GitHub
+        if: "!endsWith(matrix.python-version, '-dev')"
```
webknjaz, line 44: my preference is to avoid extra quotes but it's up to you
```suggestion
        if: >-
          !endsWith(matrix.python-version, '-dev')
```

atugushev:
@webknjaz 

Thanks for reviewing this! I'll try to address "Ubuntu-20.04" suggestion in a follow-up PR.

PR #1172, opened 2020-07-03
Update sync to return non-zero exit code on dry run

### LINKED ISSUE #1166: Feature: pip-sync --dry-run should return a non-zero exit code if changes were to occur
I'm looking to add a pre-commit hook to check the environment is up to date

```
-   repo: local
    hooks:
    -   id: pip-sync
        name: pip-sync check
        entry: pip-sync --dry-run
        language: system
        always_run: true
        pass_filenames: false
```

```
$ pip-sync --dry-run
Would install:
  numpy==1.18.5
$ $?
2
```

#### Alternative Solutions
various | awk stuff

#### Additional context
NA

francisbrito:
Resolves #1166

**Description**
Update sync function and CLI script so that they return non-zero exit codes if they're `dry-run` option is given.

**Changelog-friendly one-liner**: Update sync to return non-zero exit code on dry run

##### Contributor checklist

### DIFF
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -183,7 +183,7 @@ def sync(
             for ireq in sorted(to_install, key=key_from_ireq):
                 click.echo("  {}".format(format_requirement(ireq)))
 
-        exit_code = 2
+        exit_code = 1
 
     if ask and click.confirm("Would you like to proceed with these changes?"):
         dry_run = False
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -158,10 +158,12 @@ def sync(
     """
     Install and uninstalls the given sets of modules.
     """
+    exit_code = 0
+
     if not to_uninstall and not to_install:
         if verbose:
             click.echo("Everything up-to-date")
-        return 0
+        return exit_code
 
     pip_flags = []
     if not verbose:
@@ -181,8 +183,11 @@ def sync(
             for ireq in sorted(to_install, key=key_from_ireq):
                 click.echo("  {}".format(format_requirement(ireq)))
 
+        exit_code = 2
+
     if ask and click.confirm("Would you like to proceed with these changes?"):
         dry_run = False
+        exit_code = 0
 
     if not dry_run:
         if to_uninstall:
@@ -215,4 +220,4 @@ def sync(
             finally:
                 os.unlink(tmp_req_file.name)
 
-    return 0
+    return exit_code
--- a/tests/test_cli_sync.py
+++ b/tests/test_cli_sync.py
@@ -233,8 +233,7 @@ def test_sync_ask_accepted(check_call, runner):
     assert check_call.call_count == 2
 
 
-@mock.patch("piptools.sync.check_call")
-def test_sync_dry_run_returns_non_zero_exit_code(check_call, runner):
+def test_sync_dry_run_returns_non_zero_exit_code(runner):
     """
     Make sure non-zero exit code is returned when --dry-run is given.
     """
--- a/tests/test_cli_sync.py
+++ b/tests/test_cli_sync.py
@@ -120,7 +120,7 @@ def test_merge_error(req_lines, should_raise, runner):
         assert out.exit_code == 2
         assert "Incompatible requirements found" in out.stderr
     else:
-        assert out.exit_code == 2
+        assert out.exit_code == 1
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


francisbrito:
Notes & suggestions:

* In order to improve legibility and avoid making mistakes, I'd suggest that exit codes be abstracted into module-level variables. This way, tests and script code can be refactored to use said variables instead of relying on "magic numbers", e.g:

```python
EXIT_CODE_SUCCESS = 0
EXIT_CODE_ERROR = 2
``` 

* The exit code to be returned when `--dry-run` is provided is `2`. This is in line with what was requested in #1166. If it makes more sense to return an alternative exit code for this case in particular. Please let me know so.

tests/test_cli_sync.py
```diff
@@ -231,3 +231,16 @@ def test_sync_ask_accepted(check_call, runner):
     runner.invoke(cli, ["--ask", "--dry-run"], input="y\n")
 
     assert check_call.call_count == 2
+
+
+@mock.patch("piptools.sync.check_call")
+def test_sync_dry_run_returns_non_zero_exit_code(check_call, runner):
```
atugushev, line 237:
```suggestion
def test_sync_dry_run_returns_non_zero_exit_code(runner):
```

I wonder, do we actually need to patch `check_call`?

francisbrito: Oh, no, actually it is not required as the `sync` won't call `check_call` if ran in dry mode. My bad.

piptools/sync.py
```diff
@@ -181,8 +183,11 @@ def sync(
             for ireq in sorted(to_install, key=key_from_ireq):
                 click.echo("  {}".format(format_requirement(ireq)))
 
+        exit_code = 2
```
atugushev, line 186:
```suggestion
        exit_code = 1
```

code `1` would be more appropriate to not clash with code `2`, which is used by click in the [usage error context](https://github.com/pallets/click/blob/123dd717439d8620d8d6be5574d2c9f007952326/click/exceptions.py#L51).

tests/test_cli_sync.py
```diff
@@ -231,3 +231,15 @@ def test_sync_ask_accepted(check_call, runner):
...
+def test_sync_dry_run_returns_non_zero_exit_code(runner):
+    """
+    Make sure non-zero exit code is returned when --dry-run is given.
+    """
+    with open("requirements.txt", "w") as req_in:
+        req_in.write("small-fake-a==1.10.0")
+
+    out = runner.invoke(cli, ["--dry-run"])
+
+    assert out.exit_code == 2
```
atugushev, line 245:
```suggestion
    assert out.exit_code == 1
```

code `1` would be more appropriate to not clash with code `2`, which is used by click in the [usage error context](https://github.com/pallets/click/blob/123dd717439d8620d8d6be5574d2c9f007952326/click/exceptions.py#L51).

atugushev:
Hey @francisbrito!

Thanks for your contribution! I took the liberty to address review comments and push changes to your branch since I'm going to release it soon. Feel free to fix "magic numbers" in a follow-up PR.

tests/test_cli_sync.py
```diff
@@ -120,7 +120,7 @@ def test_merge_error(req_lines, should_raise, runner):
         assert out.exit_code == 2
         assert "Incompatible requirements found" in out.stderr
     else:
-        assert out.exit_code == 0
+        assert out.exit_code == 2
```
atugushev, line 123:
```suggestion
        assert out.exit_code == 1
```

atugushev:
Thanks @francisbrito for your first contribution! Good job!

francisbrito:
@atugushev You're welcome!

(PS: Sorry for taking so long to reply, had a busy past week)

pmi123:
I know I am late to the game here. However, I don't see the reason for always returning the error code 1 when `--dry-run` is called. As a comparison, `rsync --dry-run` returns 0 if no errors are detected, and only returns an error code if there is an actual error (e.g. invalid path). 

IMO, the purpose of a `--dry-run` option is to perform a test to see if the actual command and parameters will work. As long as there are no errors thrown, the `--dry-run` should return 0, just as if one had called `pip-sync` or `rsync` without `--dry-run`. Calling `--dry-run` is not an error condition in and of itself. 

Always returning an error code just because `--dry-run` was called eliminates the ability to test whether the `pip-sync` command will work as expected in a script. How does a script know if the error is from the actual command and parameters (e.g. invalid requirements.txt file) versus the error is just the default return value?

Mark

PR #1175, opened 2020-07-09
Do not search for best_match for -c constraints that are not (yet) requirements

### LINKED ISSUE #1178: pip-compile fails on unavailable irrelevant constraint
pip-compile tries to find a match for all constraints, event those that are not a requirement and come from a constraint file.

As a result if a constraints file defines a constraint on a package that is unavailable (be it a typo, a private lib on a currently unavailable repository, or a deprecated version no longer available) pip-compile won't work, even for requirements.in that will not need that package.

#### Steps to replicate
requirements.in:
```
-c constraints.txt
```

constraints.txt:
```
invalid-dependency~=0.0
```

#### Expected result

```
#
# To update, run:
#
#
```

#### Actual result
```
Could not find a version that matches invalid-dependency~=0.0 (from -c constraints.txt (line 1))
No versions found
```

This should be fixed by #1175

clslgrnc:
Fixes #1178.

In my organization we use a mono-repository with a huge constraints file for security purposes (via the -c option).
Additionally we have a few internal libs with a lot of versions.

As a result pip-compile is extremely slow as it tries to find a match for all constraints in our constraints file, at each round.
This is not necessary until these constraints become requirements.

The debug line should probably be changed, as an example we have:
```
found candidate urllib3>1.25.7 (constraint was >1.25.7)
```
Maybe I should change it to:
```
unused constraint urllib3>1.25.7
```
or something like that.

**Changelog-friendly one-liner**: Do not find_best_match for constraints (as opposed to requirements)

##### Contributor checklist

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -323,6 +323,10 @@ class Resolver(object):
             # NOTE: it's much quicker to immediately return instead of
             # hitting the index server
             best_match = ireq
+        elif ireq.constraint:
+            # NOTE: This is not a requirement (yet) and does not need
+            # to be resolved
+            best_match = ireq
         else:
             best_match = self.repository.find_best_match(
                 ireq, prereleases=self.prereleases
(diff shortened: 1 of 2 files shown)


clslgrnc:
Added test fails without the patch

atugushev:
`pip-tools v5.3.0` is released.

Issue #1135, opened 2020-05-04
get sha256 hash from simple endpoint


graingert:
#### What's the problem this feature will solve?
when using devpi or other non- pypi.org servers the hashing falls back to downloading the asset and hashing it locally

#### Describe the solution you'd like
use the sha256 hash from the /simple/ endpoint

#### Alternative Solutions
https://github.com/devpi/devpi/issues/801#issuecomment-623510074

#### Additional context
/cc @fschulze
https://github.com/jazzband/pip-tools/pull/1109
view-source on: https://m.devpi.net/root/pypi/+simple/devpi-server/
and view-source on: https://pypi.org/simple/devpi-server/

graingert:
Another option that would be standardized across HTTP hosts
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Want-Digest

graingert:
Artifactory now supports sha256 in the simple api

PR #1184, opened 2020-07-20
default to --no-reuse-hashes

graingert:
A breaking change for v6

**Changelog-friendly one-liner**: default to --no-reuse-hashes

##### Contributor checklist

Now that hashing is way faster than it once was, I think the slight speed improvement from caching no longer outweighs the danger of trusting local/accidental changes to the .txt pins

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -186,7 +186,7 @@ class BaseCommand(Command):
 @click.option(
     "--reuse-hashes/--no-reuse-hashes",
     is_flag=True,
-    default=True,
+    default=False,
     help=(
         "Improve the speed of --generate-hashes by reusing the hashes from an "
         "existing output file."


jdufresne:
> Now that hashing is way faster than it once was, I think the slight speed improvement from caching no longer outweighs the danger of trusting local/accidental changes to the .txt pins

If this is the case, would it make sense to deprecate/remove the feature entirely?

graingert:
@jdufresne I'm in favour of that

graingert:
@jdufresne https://github.com/jazzband/pip-tools/issues/1135 would probably need doing first though

atugushev:
FTR: I've removed this PR from `6.0.0`. I'm going to prepare a release soon and it's unclear to me whether there should be changed the default value for `--no-reuse-hashes` or should be completely removed. Seems like this PR could be postponed for now.

plannigan:
[Artifactory doesn't support the `/json` endpoint](https://www.jfrog.com/jira/browse/RTFACT-16101), so users in a corporate environment that don't directly pull from PyPI would be required to always download each file to be hashed. So removing the capability to reuse hashes would increase the time it takes to compile a lock file. This is part of the reason I implemented single hash (#1406).

graingert:
@plannigan https://github.com/jazzband/pip-tools/issues/1135 we could pull the hashes from the simple API

plannigan:
Good to know. If sequence was try `/json`/try simple/download file then that would resolve my concerns.

graingert:
Probably not worth doing the /json endpoint now as the simple api will already have been downloaded with the hashes in

neykov:
There are widely used Python projects that rely extensively on `--find-links` / `--extra-index-url` for distribution that don't provide checksums in their index files. For example:
* [pytorch](https://pytorch.org/get-started/previous-versions/) - https://download.pytorch.org/whl/torch/
* [pytorch-scatter](https://github.com/rusty1s/pytorch_scatter) - https://data.pyg.org/whl/torch-1.12.0+cpu.html

pytorch is a ~2GB download. Until these projects catch up generating the hashes will be too expensive (slow and bandwidth intensive) if `--reuse-hashes` is dropped.

PR #1191, opened 2020-07-29
Drop pip-20.0 support

atugushev:
Due to `pip-tools`' [policy](https://github.com/jazzband/pip-tools/issues/926) to support only the last two minor pip versions, it's time to drop the `pip-20.0`.

- [x] pin `pip >= 20.1` in `setup.py`
- [x] remove `20.0` compatible code
- [x] remove `20.0` from CI 
- [x] update "Versions and compatibility" in `README`

**Changelog-friendly one-liner**: Drop `pip-20.0` support.

##### Contributor checklist


atugushev:
@AndydeCleyre 

> * Should we remove `BAR_TYPES` from `_compat/{pip_compat,__init__}.py` and import it directly in `pypi.py`?

Yeah, thanks for noticing!




> * Do you think it would be worth the trouble to template some of these files, for more automated pip-version updates? Specifically: `.appveyor.yml`, `ci.yml`, `.travis.yml`, `setup.cfg`, `tox.ini`

Would be nice to automate this. 

> * Do we have special reasons to stick with `setup.{cfg,py}` for now, rather than moving to `pyproject.toml`?

I'd prefer to stick to `setup.cfg` since pip doesn't support install editables with `pyproject.toml`.

atugushev:
Thank you @AndydeCleyre for reviewing this! Much appreciated!

PR #1222, opened 2020-11-15
Complete testing and documentation of Python 3.9 support

jdufresne:
Document and add support for Python 3.9.

##### Contributor checklist

---

Testing was added in 7503ad75d849b96c6e29afeefbc2d735681854ae but is now
included in all CI matrices and is documented as trove classifier.

Python 3.9 was released November 15, 2020.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,17 +21,18 @@ jobs:
           - Windows
           - macOS
         python-version:
-          - 3.8
+          - 3.9
           - 2.7
           - 3.5
           - 3.6
           - 3.7
+          - 3.8
         pip-version:
           - "latest"
           - "previous"
         include:
           - os: Ubuntu
-            python-version: 3.9-dev
+            python-version: 3.10-dev
             pip-version: latest
 
     env:
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -16,11 +16,12 @@ jobs:
           - Windows
           - MacOS
         python-version:
-          - 3.8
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -21,18 +21,15 @@ jobs:
...
           - 3.5
           - 3.6
           - 3.7
+          - 3.8
         pip-version:
           - "latest"
           - "20.1"
-        include:
-          - os: Ubuntu
-            python-version: 3.9-dev
```
atugushev, line 34: Maybe test `3.10-dev` now? Deadsnakes supports it.

jdufresne: Done!

atugushev (approved):
Thanks 👍 I think it would be good to remove py35 (due to EOL) in a following up PR.

jdufresne:
> I think it would be good to remove py35 (due to EOL) in a following up PR.

Would be happy to.

What is the pip-tools policy on supported Pythons? 

If we're using EOL as a guideline, then we should begin work towards dropping Python 2. Something I'm also happy to contribute to.  See https://devguide.python.org/#status-of-python-branches

Or, on the other hand, are we trying to match the pip support? If so, they still support Python 3.5.

Or, is the policy drop EOL Pythons except 2.7?

I looked through the docs and wasn't able to find anything explicit. I could add something to <https://github.com/jazzband/pip-tools#versions-and-compatibility> once I understand the policy.

My preference is to follow Python EOL branches, but I recognize that may too fast for some projects.

PR #1231, opened 2020-11-20
Release 5.4.0

atugushev:


atugushev:
@vphilippon where we should announce the `pip-20.0` drop line? Currently, it's under "features".

vphilippon:
I'd suggest putting it under a `Compatibility Changes:` section, as the first section.

jdufresne:
> Drop `pip-20.0` support

The project says it follows semantic versioning, so shouldn't this necessitate a bumping the major version number?

vphilippon:
Hm, I re-read the discussions around pip version support, and it seems this fell into a crack. We haven't clearly established if a change in dependency was a breaking change.

I guess it depends how we see this:
- Changing the requirement (dependency) on the version of pip required by pip-tools: Not a breaking change (ex: `requests` changing its dependency on `urllib3` isn't a breaking change that required a major version bump.
vs
- Removing the feature/functionality that is the support of a given version of `pip`: Removal of a feature, backward-incompatible, that's a breaking changes that would require a major version bump.

Note that I simply described the exact same change in 2 different manners. So it boils down to "What's our contract" and "What are our users expecting.

Since we've decided to specify the version of `pip` to use in `pip-tools` `install_require`, I'd go for the 1st option, the dependency change, thus it's not a breaking change. That seems to be the line we are taking now.

If a user does `python -m pip install -U pip-tools`, and then uses `pip-tools` as before, then every feature of `pip-tools` that used to work still works now.

atugushev:
@vphilippon 

> I'd go for the 1st option

Yeah, I remember we agreed on 1st option. How about this?

> Features:
> - ...
>
> Bug Fixes:
> - ...
> 
> Dependencies:
> - Bump minimum `pip` version to `>= 20.1`.

AndydeCleyre (approved):
Thanks for getting this out! Good call on holding back the new resolver changes for this important release.

atugushev:
Thanks @vphilippon and @AndydeCleyre!

PR #1233, opened 2020-11-21
Enable Python warnings during tests and fix all in the code

jdufresne:
When running the test suite with Python warnings enabled, several
warnings are emitted. They look like:

    tests/test_utils.py::test_format_requirement
      /usr/lib64/python3.9/tempfile.py:817: ResourceWarning: Implicitly cleaning up <TemporaryDirectory '/tmp/tmpezxehep1source'>
        _warnings.warn(warn_message, ResourceWarning)

To solve, make Repository objects a context manager. Upon entering the
with statement, the build an source caches are created and upon exit
they are cleaned up. This ensures they only live for the duration they
are required. This could theoretically be used creating and destroying
any necessary resources. Tests were adjusted.

    tests/test_repository_pypi.py::test_pypirepo_build_dir_is_str
      .../pip/_vendor/requests/adapters.py:59: ResourceWarning: unclosed file <_io.FileIO name='/tmp/pytest-of-jon/pytest-70/test_open_local_or_remote_file5/foo.txt' mode='rb' closefd=True>
        super(BaseAdapter, self).__init__()

To solve, test_open_local_or_remote_file__remote_file() was adjusted to
close its file after use.

    tests/test_cli_compile.py::test_redacted_urls_in_verbose_output[--find-links]
      .../pip-tools/piptools/scripts/compile.py:306: FutureWarning: --index and --no-index are deprecated and will be removed in future versions. Use --emit-index-url/--no-emit-index-url instead.
        warnings.warn(

To solve, use --no-emit-index-url in
test_redacted_urls_in_verbose_output().

There remains one warning left, but this is out of the control of
pip-tools. An upstream PR has been filed:
https://github.com/pypa/pip/pull/9156

Enabling the warnings in test runs will help catch them earlier and make
debugging/fixing easier.

##### Contributor checklist

### DIFF
--- a/piptools/repositories/base.py
+++ b/piptools/repositories/base.py
@@ -12,6 +12,8 @@ class BaseRepository(object):
     def clear_caches(self):
         """Should clear any caches used by the implementation."""
 
+    @abstractmethod
+    @contextmanager
     def freshen_build_caches(self):
         """Should start with fresh build/source caches."""
 
--- a/piptools/repositories/local.py
+++ b/piptools/repositories/local.py
@@ -56,8 +56,10 @@ class LocalRequirementsRepository(BaseRepository):
     def clear_caches(self):
         self.repository.clear_caches()
 
+    @contextmanager
     def freshen_build_caches(self):
-        self.repository.freshen_build_caches()
+        with self.repository.freshen_build_caches():
+            yield
 
     def find_best_match(self, ireq, prereleases=None):
         key = key_from_ireq(ireq)
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -87,7 +87,8 @@ class PyPIRepository(BaseRepository):
         self._dependencies_cache = {}
 
         # Setup file paths
-        self.freshen_build_caches()
+        self._build_dir = None
+        self._source_dir = None
         self._cache_dir = normalize_path(cache_dir)
         self._download_dir = fs_str(os.path.join(self._cache_dir, "pkgs"))
         if PIP_VERSION[:2] <= (20, 2):
@@ -95,6 +96,7 @@ class PyPIRepository(BaseRepository):
 
         self._setup_logging()
 
+    @contextmanager
     def freshen_build_caches(self):
         """
         Start with fresh build/source caches.  Will remove any old build
@@ -102,14 +104,21 @@ class PyPIRepository(BaseRepository):
         """
         self._build_dir = TemporaryDirectory(fs_str("build"))
         self._source_dir = TemporaryDirectory(fs_str("source"))
+        try:
+            yield
+        finally:
+            self._build_dir.cleanup()
+            self._build_dir = None
+            self._source_dir.cleanup()
+            self._source_dir = None
 
     @property
     def build_dir(self):
-        return self._build_dir.name
+        return self._build_dir.name if self._build_dir else None
 
     @property
     def source_dir(self):
-        return self._source_dir.name
+        return self._source_dir.name if self._source_dir else None
 
     def clear_caches(self):
         rmtree(self._download_dir, ignore_errors=True)
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -170,7 +170,13 @@ class Resolver(object):
 
             log.debug("")
             log.debug(magenta("{:^60}".format("ROUND {}".format(current_round))))
-            has_changed, best_matches = self._resolve_one_round()
+            # If a package version (foo==2.0) was built in a previous round,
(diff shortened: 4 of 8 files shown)


piptools/resolver.py
```diff
@@ -170,7 +170,8 @@ def resolve(self, max_rounds=10):
 
             log.debug("")
             log.debug(magenta("{:^60}".format("ROUND {}".format(current_round))))
-            has_changed, best_matches = self._resolve_one_round()
+            with self.repository:
```
atugushev, line 173:
```suggestion
            # If a package version (foo==2.0) was built in a previous round,
            # and in this round a different version of foo needs to be built
            # (i.e. foo==1.0), the directory will exist already, which will
            # cause a pip build failure.  The trick is to start with a new
            # build cache dir for every round, so this can never happen.
            with self.repository.freshen_build_caches():
```

I would keep the comment — it explains the intention. And I think it makes sense to call an explicit context manager in favor of readability (minor).

jdufresne: Thanks for the review. I made this change in the latest revision.

piptools/repositories/base.py
```diff
@@ -12,8 +12,10 @@ class BaseRepository(object):
     def clear_caches(self):
         """Should clear any caches used by the implementation."""
 
+    @contextmanager
     def freshen_build_caches(self):
         """Should start with fresh build/source caches."""
+        yield
```
atugushev, line 18:
```suggestion
    @abstractmethod
    @contextmanager
    def freshen_build_caches(self):
        """Should start with fresh build/source caches."""
```

piptools/repositories/local.py
```diff
@@ -56,8 +56,11 @@ def DEFAULT_INDEX_URL(self):
     def clear_caches(self):
         self.repository.clear_caches()
 
+    @contextmanager
     def freshen_build_caches(self):
-        self.repository.freshen_build_caches()
+        with super(LocalRequirementsRepository, self).freshen_build_caches():
+            with self.repository.freshen_build_caches():
+                yield
```
atugushev, line 63:
```suggestion
        with self.repository.freshen_build_caches():
            yield
```

piptools/repositories/pypi.py
```diff
@@ -87,29 +87,39 @@ def __init__(self, pip_args, cache_dir):
...
+        with super(PyPIRepository, self).freshen_build_caches():
+            self._build_dir = TemporaryDirectory(fs_str("build"))
+            self._source_dir = TemporaryDirectory(fs_str("source"))
+            try:
+                yield
+            finally:
+                self._build_dir.cleanup()
+                self._build_dir = None
+                self._source_dir.cleanup()
+                self._source_dir = None
```
atugushev, line 114:
```suggestion
        self._build_dir = TemporaryDirectory(fs_str("build"))
        self._source_dir = TemporaryDirectory(fs_str("source"))
        try:
            yield
        finally:
            self._build_dir.cleanup()
            self._build_dir = None
            self._source_dir.cleanup()
            self._source_dir = None
```

atugushev (commented):
Since `freshen_build_caches` does nothing interesting but `yield` it can be declared as an abstract method (same as `allow_all_wheels`). What do you think?

jdufresne:
> it can be declared as an abstract method

Can you explain a bit how that helps us, what the benefit is, or what it works towards solving?

piptools/repositories/local.py
jdufresne: Calling the parent implementation allows us to easily make changes to the base class and be confident all children will using it without additional intervention. If we were to add behavior to the base class's method, it is all too easy to forget one of the child implementation that must change.

atugushev: Yeah, I understand your point which I think is pretty reasonable. Although following the YAGNI principle I'd rather have an abstract method instead of a base one (which does nothing useful even for a while). Once we need to add a base logic (which I slightly doubt) we could update children's methods and use the base functionality. However, If you feel strongly that we need the base method then feel free to ignore the suggestion - I'd like to not block the PR any longer 👍

> Can you explain a bit how that helps us, what the benefit is, or what it works towards solving?

The main benefit is less code and indentations.

jdufresne:
I went ahead with the abstract method approach. The reduction in repetition in indentation is nice.

One downside, this required adding an empty method to `FakeRepository` to fulfill the ABC requirements. This will be true of any new repository that doesn't require build caches. In practice, this probably won't be an issue going forward, just something to mention w/r/t the trade offs.

Thanks for the reviews!

atugushev (approved):
Thanks!

PR #1236, opened 2020-11-22
Remove unnecessary flake8 configuration option

jdufresne:
flake8 ignores these directories by default and piptools/_compat/* has
no violations anyway.


atugushev (approved):
LGTM 👍

Issue #1156, opened 2020-06-03
`pip-compile` only primary requirements defined in `requirements.in`


cabhishek:
#### What's the problem this feature will solve?

We have a requirement where our project only really needs to pin the primary requirements defined in the `requirements. in` file. And right now, there is no way to do this explicitly. So we have to generate the `requirements.txt` file and strip out all non-primary dependencies, which is quite hacky. 
 
#### Describe the solution you'd like

`pip-compile --primary-requirements-only --output-file requirements.txt --upgrade requirements.in`. 

#### Alternative Solutions

None available today in `pip-compile.` But maybe use `pip.` 

#### Additional context

None

atugushev:
Hello @cabhishek,

Thanks for the issue! IMO, this case is too specific and could be solved by:

```console
$ pip-compile -qo- | grep "via -r requirements.in"
```

But, I'd be happy to hear other opinions.

cabhishek:
@atugushev Thanks for the suggestion, and we are doing something similar. But I guess our use case is specific since we don't even need the annotations  `# "via -r requirements.in..."`. So we cannot use `--no-annotate` flag if we are relying on annotation for primary dependency filtering.  

But overall I think its quite a simple use case, just pin dependencies from `requirement.in` but not recursively.

AndydeCleyre:
A few extra notes:

- You can add a comment-stripping step in the pipeline to @atugushev's example:
```console
$ pip-compile -qUo- requirements.in | grep "via -r requirements.in" | grep -o '.*  ' > requirements.txt
```
- You can do it with one fewer pipe with a fancy tool like `pcregrep`, `pcre2grep`, or `rg` (ripgrep):
```console
$ pip-compile -qUo- requirements.in | pcregrep -o1 '(.*) # via -r requirements.in' > requirements.txt
$ pip-compile -qUo- requirements.in | pcre2grep -o1 '(.*) # via -r requirements.in' > requirements.txt
$ pip-compile -qUo- requirements.in | rg -r '$1' '(.*) # via -r requirements.in' > requirements.txt
```
- **Be warned** that in this project's current state, running `pip-sync` without options against such a resulting `requirements.txt` will behave inconsistently. Let's say we've got two primary requirements, `click` and `boto3`:
```console
$ pip-sync -q requirements.txt
$ pip freeze | wc -l
10
$ pip-sync -q requirements.txt
$ pip freeze | wc -l
4
$ pip-sync -q requirements.txt
[...]

PR #1237, opened 2020-11-22
Improve formatting of long "via" annotations

### LINKED ISSUE #1036: via lines often result in git conflict
currently when adding and removing packages in multiple github PRs the `# via` line will conflict

put each `# via` on a new line:
```
six==1.12.0 \
    --hash=sha256:3350809f0555b11f552448330d0b52d5f24c91a322ea4a15ef22629740f3761c \
    --hash=sha256:d16a0141ec1a18405cd4ce8b4613101da75da0e9a7aec5bdd4fa804d0e0eba73 \
    # via automat, bcrypt, cryptography, ecdsa, format-cef, grpcio, patchy, pathlib, prompt-toolkit, protobuf, pyhamcrest
```
instead use:
```
six==1.12.0 \
    --hash=sha256:3350809f0555b11f552448330d0b52d5f24c91a322ea4a15ef22629740f3761c \
    --hash=sha256:d16a0141ec1a18405cd4ce8b4613101da75da0e9a7aec5bdd4fa804d0e0eba73 \
    # via automat
    #   , bcrypt
    #   , cryptography
    #   , ecdsa
    #   , format-cef
    #   , grpcio
    #   , patchy
    #   , pathlib
    #   , prompt-toolkit
    #   , protobuf
    #   , pyhamcrest
```

jdufresne:
When there is more than one "via" annotation, they are now formatted one
per line. This prevents very long lines in requirements.txt output. The
very long lines often scroll beyond the width of an editor, making them
difficult to read. Worse, when these annotations change, diffs are hard
to read as the actual difference is hidden far off to the right. By
placing one annotation per line, diffs will be much more obvious. This
follows the philosophy of many automatic formatters such as Black.

To simplify the formatting and to make it more consistent, annotations
are now always on the line after the requirement, whether hashes were
generated or not.

Fixes #1036

##### Contributor checklist


setup.cfg
```diff
@@ -62,9 +62,8 @@ markers =
     network: mark tests that require internet access
 
 [flake8]
-max-line-length = 88
 exclude = build/*, dist/*, pip_tools.egg-info/*, piptools/_compat/*, .tox/*, .venv/*, .git/*, .eggs/*
-extend-ignore = E203  # E203 conflicts with PEP8; see https://github.com/psf/black#slices
+extend-ignore = E501,E203  # E203 conflicts with PEP8; see https://github.com/psf/black#slices
```
graingert, line 66: you can set the max-line-length to whatever line you created, no need to set it to infinite

tests/test_cli_compile.py
```diff
@@ -691,6 +698,90 @@ def test_generate_hashes_verbose(pip_conf, runner):
...
+            # via -r requirements.in
+        django==2.2 \\
+            --hash=sha256:7c3543e4fb070d14e10926189a7fcf42ba919263b7473dceaefce34d54e8a119 \\
+            --hash=sha256:a2814bffd1f007805b19194eb0b9a331933b82bd5da1c3ba3d7b7ba16e06dc4b
+            # via
+            #   -r requirements.in
+            #   django-debug-toolbar
+            #   django-storages
+            #   django-taggit
+            #   djangorestframework
```
graingert, line 764: nice :+1:

setup.cfg
jdufresne: Good, point. I changed it to 100 rather than dropping it.

atugushev:
Looks nice! I wonder what @vphilippon would think?

vphilippon:
I like that!
I might even prefer it with comas, like:
```
            # via
            #   -r requirements.in,
            #   django-debug-toolbar,
            #   django-storages,
            #   django-taggit,
            #   djangorestframework
```
(`',\n'.join(...)`)
I feel like it helps me spot it as an ongoing list quickly.
But that's really a preference, not a blocker or game changer.

Beside that, 👍 from me!

jdufresne:
My preference is for no comma. IMO, it doesn't add much to the formatting and I think the no-comma result looks clean and readable enought without it.

> I wrote the coma suggestion without actually thinking about the potential risk of adding some git diff conflict with that. I'll let you be the judge of that :)

This is also a concern for me. When the last dependency in the list changes, now two lines will change instead of one. this causes unnecessary diff noise. So another reason I prefer no comma.

I don't consider the above too big a deal though, so if it is considered a blocker, let me know and I'll change it.

graingert:
the version with commas:

```
    # via automat
    #   , bcrypt
    #   , cryptography
    #   , ecdsa
    #   , format-cef
    #   , grpcio
    #   , patchy
    #   , pathlib
    #   , prompt-toolkit
    #   , protobuf
    #   , pyhamcrest
```

graingert:
looking at the test output again, I really like how it looks, and I think I prefer comma-less

atugushev:
@vphilippon thanks for your feedback!

I like also "no comma" annotations. Less diff and less symbols.

atugushev (approved):
LGTM 👍

vphilippon (approved):
👍 
Agreed on the coma-less format then, I'm a-ok with that!

jdufresne:
Thanks for the feedback and reviews all!

scholarsmate:
FYI, this new formatting breaks the [requirements-txt-fixer pre-commit hook](https://github.com/pre-commit/pre-commit-hooks#requirements-txt-fixer).

adamchainz:
@jdufresne Just upgraded and love this change, thanks!

rsalmaso:
This is a breaking change... And broke my workflow :disappointed:

webknjaz:
> This is a breaking change... And broke my workflow

Sounds like a reference to
[XKCD Workflow comic #1172](https://xkcd.com/1172/)

setup.cfg
```diff
@@ -63,7 +63,7 @@ markers =
     network: mark tests that require internet access
 
 [flake8]
-max-line-length = 88
+max-line-length = 100
```
webknjaz, line 66: this looks like a bad standard :(

webknjaz:
> FYI, this new formatting breaks the [requirements-txt-fixer pre-commit hook](https://github.com/pre-commit/pre-commit-hooks#requirements-txt-fixer).

Sounds like this should've been a CLI flag and not an opinionated reformatting

bendikro:
> > FYI, this new formatting breaks the [requirements-txt-fixer pre-commit hook](https://github.com/pre-commit/pre-commit-hooks#requirements-txt-fixer).
> 
> Sounds like this should've been a CLI flag and not an opinionated reformatting

Obviously. There are always good reasons to make changes, but the very good reasons _not_ to make changes are often ignored. Such as you will probably be breaking something.

adamchainz:
> > FYI, this new formatting breaks the [requirements-txt-fixer pre-commit hook](https://github.com/pre-commit/pre-commit-hooks#requirements-txt-fixer).
> 
> Sounds like this should've been a CLI flag and not an opinionated reformatting

This pre-commit hook was already broken if you used `--generate-hashes` which forced multi-line. IMO the hook is at fault since it uses very simple parsing of requirements.txt.

webknjaz:
> This pre-commit hook was already broken if you used `--generate-hashes` which forced multi-line. IMO the hook is at fault since it uses very simple parsing of requirements.txt.

I don't use the hook but still would prefer a flag.

dannyshaw:
I'll just note here we're having to roll back to 5.4.0 because the change in formatting means that [dependabot](dependabot.com) (also built into github) apparently uses an older version of pip-tools to compile `requirements.txt` and therefore the PRs that dependabot produces now cause conflicts with those that we manually produce.

I hope that's clear.

feelepxyz:
👋  I work on Dependabot at GitHub and wondering how best to deal with this change. It seems best if Dependabot detected the current formatting and switched the used version of pip-tools based on this to maintain the same formatting to prevent merge conflicts.

Thoughts on adding a flag or some way of telling pip-compile which format to use? Or even better detect the current formatting if there is an existing output file and use that?

rsalmaso:
I've made https://github.com/jazzband/pip-tools/pull/1297 to move this improvement under an explicit flag. Feel free to comment there.

jdufresne:
The rendered formatting of `requirements.txt` is not a stable API. It is expected to be correct an informative, but wasn't built specifically to be consumed by other tools. So I disagree with the assertion that this is a breaking change. I recognize some tools were treating it as such, but I think it is up to them to handle those nuances.

> I'll just note here we're having to roll back to 5.4.0 ...

As with any other dependency, I recommend pinning `pip-tools` within your project. In some projects, I've handled this using `bootstrap-requirements.in`/`bootstrap-requirements.txt` and compiled it using pip-tools itself.

>  I work on Dependabot at GitHub and wondering how best to deal with this change.

I think the best move would be fore dependabot to upgrade its internal version of pip-tools so it is using the latest format. All new PRs opened by dependabot will then use this new format.

merwok:
But the requirements format was invented by pip, and surely tools that create requirements file shouldn’t break compatibility with pip.  pip-sync is great for deployments and tox, but sometimes you want or need to use pip install.

Can the original ticket be reopened, or a new one created, to add an option to control the formatting of `via` lines?

graingert:
> shouldn’t break compatibility with pip

this change doesn't do that - it alters a comment which is ignored by pip

merwok:
Ah my apology!  I read this part:

> The rendered formatting of requirements.txt is not a stable API. It [...] wasn't built specifically to be consumed by other tools.

and interpreted wrong.

So the issue is not functional but cosmetic — I didn’t like getting a 100% diff when upgrading pip-tools, so an option would be great.

cabhishek:
Not depending on the formatting of `requirements.txt` sounds fair, but right now, there is no way to only get the top-level dependencies in `pip-tools`, so it was recommended to us to parse the pip-compile output and infer the information. So this change breaks that recommendation. You can refer to the feature request we made [here](https://github.com/jazzband/pip-tools/issues/1156).

AndydeCleyre:
This makes awful use of my limited vertical screen real estate, and my more generous horizontal screen real estate. It's terribly inconvenient to page around to get all the same information I used to get without paging, and no longer ever all visible at once.

I looked but didn't find an existing open issue to regain the practical use of space from before this merge, but please let me know if there is one already.

AndydeCleyre:
OK, I've made #1306 to track the request.

belm0:
> The rendered formatting of requirements.txt is not a stable API. It is expected to be correct and informative, but wasn't built specifically to be consumed by other tools. So I disagree with the assertion that this is a breaking change.

Given that it's normal to commit requirements.txt files into a project repo (that's how you pin, after all), it's more than a matter of breaking tools trying to consume the file.  Suddenly we have a large diff in our project's requirements.txt since the file is entirely rewritten, and we'd like to decide when to incur that diff (e.g. by way of a friendly command line option for this formatting change) rather than have the timing be imposed upon us.

> As with any other dependency, I recommend pinning pip-tools within your project. 

Except, your project may need Python 3.9 support or other features which are orthogonal to this formatting change.

PR #1240, opened 2020-11-22
Remove six dependency in favor pip's vendored six

jdufresne:
Allows removing a dependency of pip-tools. The package is already
guaranteed to exist within pip, so might as well use it.

As more and more projects are dropping Python 2 support, they too are
dropping six as a dependency. As more projects drop six, this will
eventually remove the need for projects to pull in six as a transient
dependency.

When pip eventually drops Python 2 support, it will also drop six. When
that day comes, pip-tools can also drop Python 2 support, so the
eventual lack of six should not be an issue.

##### Contributor checklist

### DIFF
--- a/piptools/_compat/__init__.py
+++ b/piptools/_compat/__init__.py
@@ -2,7 +2,7 @@
 # flake8: noqa
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import six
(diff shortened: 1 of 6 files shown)


atugushev (approved):
That makes sense! Thanks 👍

PR #1243, opened 2020-11-25
Remove support for EOL Python 3.5 and 2.7

jdufresne:
Python 3.5 went EOL 2020-09-30.

For a list of supported Python versions, see:
https://devguide.python.org/#status-of-python-branches

For a list of EOL Python versions and their dates, see:
https://devguide.python.org/devcycle/#end-of-life-branches

##### Contributor checklist


jdufresne:
I marked this for the 6.0.0 milestone, but I'm not sure if that is considered correct.

This change was suggested in #1222.

I'll also copy over my questions from that PR, so the thread can continue here:


What is the pip-tools policy on supported Pythons? 

If we're using EOL as a guideline, then we should begin work towards dropping Python 2. Something I'm happy to contribute to.  See https://devguide.python.org/#status-of-python-branches

Or, on the other hand, are we trying to match the pip support? If so, they still support Python 3.5.

Or, is the policy to drop EOL Pythons except 2.7?

I looked through the docs and wasn't able to find anything explicit. I could add something to <https://github.com/jazzband/pip-tools#versions-and-compatibility> once I understand the policy.

My preference is to follow Python EOL schedule, but I recognize that may be too fast for some projects that use pip-tools.

atugushev:
> I marked this for the 6.0.0 milestone, but I'm not sure if that is considered correct.

That's correct. See [4.0.0](https://github.com/jazzband/pip-tools/releases/tag/4.0.0) release for example.

> What is the pip-tools policy on supported Pythons?

Usually `pip-tools` follows `pip` drop guideline. BTW `pip==20.1` drops `py3.5`, see https://github.com/pypa/pip/pull/8778.

> I looked through the docs and wasn't able to find anything explicit. I could add something to https://github.com/jazzband/pip-tools#versions-and-compatibility once I understand the policy.

Cool! That would be awesome!

hugovk:
> Usually `pip-tools` follows `pip` drop guideline. BTW `pip==20.1` drops `py3.5`, see [pypa/pip#8778](https://github.com/pypa/pip/pull/8778).

Typo: it's actually `pip==21.0` which will drop 3.5, in January 2021, and the same version will also drop 2.7 (https://github.com/pypa/pip/pull/8087 / https://github.com/pypa/pip/issues/6148).

jdufresne:
@atugushev If we're going make a new major release with backward incompatible changes, WDYT about taking this opportunity to drop Python 2?

Python 2 has now been EOL for nearly a year.

Python 2 users could continue using old version of pip-tools until they're ready to upgrade their project. We could even cut one more minor release so Python 2 users have one more chance to receive updated features (such as the new pip compatibility).

Removing it would reduce compatibility code, CI resources, and other maintenance concerns. As well, we could begin to take advantage of newer Python features and patterns. pytest, click, and other transient dependencies have removed or are removing Python 2 support.

Here are the PyPI statistics for the last three months. It shows Python 2 downloads before 20% and this percentage has been falling.

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 3.7            |  29.78% |        800,411 |
| 3.8            |  22.67% |        609,187 |
| 2.7            |  19.96% |        536,354 |
| 3.6            |  18.39% |        494,225 |
| 3.5            |   7.15% |        192,166 |
| 3.9            |   1.96% |         52,628 |
| 3.4            |   0.09% |          2,371 |
| 3.10           |   0.01% |            382 |
| 3.3            |   0.00% |             31 |
| 2.6            |   0.00% |              7 |
| Total          |         |      2,687,762 |

I'm am willing to do the necessary work. Cheers.

atugushev:
@jdufresne 

> If we're going make a new major release with backward incompatible changes, WDYT about taking this opportunity to drop Python 2?

Since `pip==21.0` drops `py27` we should drop it also.

jdufresne:
Great to hear! The latest revision of this PR drops Python 2.7 as well.

setup.cfg
```diff
@@ -26,7 +24,7 @@ classifiers =
     Topic :: System :: Systems Administration
 
 [options]
-python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
+python_requires = >=3.5
```
hugovk, line 27:
```suggestion
python_requires = >=3.6
```

hugovk:
`test_sync.py` has a couple of asserts with 2.7 checks:

```python
def test_diff_should_install_with_markers(from_line):
    installed = []
    reqs = [from_line("subprocess32==3.2.7 ; python_version=='2.7'")]

    to_install, to_uninstall = diff(reqs, installed)
    assert {str(x.req) for x in to_install} == (
        {"subprocess32==3.2.7"} if sys.version.startswith("2.7") else set()
    )
    assert to_uninstall == set()


def test_diff_should_uninstall_with_markers(fake_dist, from_line):
    installed = [fake_dist("subprocess32==3.2.7")]
    reqs = [from_line("subprocess32==3.2.7 ; python_version=='2.7'")]

    to_install, to_uninstall = diff(reqs, installed)
    assert to_install == set()
    assert to_uninstall == (
        set() if sys.version.startswith("2.7") else {"subprocess32"}
    )
```

Also pyupgrade is [intentionally timid](https://github.com/asottile/pyupgrade#f-strings) in upgrading to f-strings, https://github.com/ikamensh/flynt (optionally with `-a`) is good at converting more.

jdufresne:
Thanks for the review @hugovk I applied your suggestions in the latest revision.

atugushev (approved):
Awesome!

PR #1245, opened 2020-11-26
Minor fixes to stay in line with Jazzband CI best practices

jezdez:
This basically makes sure to first setup Python and then fetch the pip cache dir using `pip cache dir`. Also minor renaming for the sake of consistency.

##### Contributor checklist


hugovk (approved):

.github/workflows/release.yml
```diff
@@ -11,26 +11,31 @@ on:
...
+      - name: Get pip cache dir
+        id: pip-cache
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: release-${{ hashFiles('**/setup.py') }}
```
jezdez, line 35: @hugovk Should we add a `v1` here as well?

hugovk: Could do, it's also trivial to add it later when actually needing to invalidate the old cache to generate a new one.

---

Have you heard about https://github.com/asottile/all-repos? Very good for making lots of (small) changes to many repos. https://jugmac00.github.io/blog/how-to-create-hundreds-of-pull-requests-with-a-single-command/ is a good intro.

hugovk, line 35:
```suggestion
          key: release-${{ hashFiles('**/release.yml') }}
```

Another thing: we want to hash the filename of whatever defines the dependencies installed by pip. This could be `setup.py`, `setup.cfg`, `requirements.txt`, or a combination.

In this case, the stuff pip installs is defined in this file.

The idea is, when a dependency is added or removed, that file changes, and we get a new cache.

---

A bit more info about `actions/cache`:

First it looks for a cache matching the `key`. If found, great, it downloads it.

If not, it then it looks for any _partially_ matching `restore-keys`. (This is a list, even though we probably only need a single entry.)

So if no `key` was found (e.g. when we've added a new dependency), it'll download the old cache because of the partial `restore-keys` match. We then install our new stuff into the pip cache, and it'll upload that for use next time.

Docs: https://github.com/actions/cache#creating-a-cache-key

jdufresne:
Is configuring the cache worth it? Maybe it is or maybe it isn't, I'm haven't measured to know. If it doesn't offer much in performance gains, we could just drop it. For releasing pip-tools, my hunch is there won't be any noticeable difference and we may be better off with a simpler configuration.

hugovk:
Yeah, GitHub Actions is pretty snappy to downloading stuff directly from PyPI.

Performance-wise, it's probably not going to make much difference, unless needing to build stuff from source when wheels aren't available.

I generally add the cache by default anyway, if only just to try and help fractionally reduce [PyPI's gigantic (albeit donated) hosting cost](https://twitter.com/GaelVaroquaux/status/1236357401911209985).

hugovk:
https://github.com/actions/cache says:

> Caches that are not accessed within the last week will also be evicted.

And as we're discussing the release workflow here, usually releases cadence is greater than a week* so it's of less value here, especially if you want to simplify the config.

\* Unless you're [Hypothesis](https://pypi.org/project/hypothesis/#history) :)

jdufresne:
> Caches that are not accessed within the last week will also be evicted.

Interesting. Thanks for digging that up.

This makes me think we're better off without it for the release workflow.

.github/workflows/release.yml
jezdez: Yeah, I've seen all-repos, think that'll be excellent to update the workflows in the future, but the actual migration from travis.yml to GHA workflow wasn't something I was able to figure out with it.

jezdez: I've removed the caching bits for now, given the small time window in which it would be useful for a release workflow.

jezdez:
> > Caches that are not accessed within the last week will also be evicted.
> 
> Interesting. Thanks for digging that up.
> 
> This makes me think we're better off without it for the release workflow.

Agreed, the time window where caching would be useful for the release workflow is too small to worry about it (or make a dent in the PyPI usage). I've removed the bit from the file.

jdufresne (approved):

PR #1252, opened 2020-11-28
Remove filterwarnings configuration

jdufresne:
##### Contributor checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,8 +57,6 @@ norecursedirs = .* build dist venv test_data piptools/_compat/*
 testpaths = tests piptools
 filterwarnings =
     always
-    ignore::PendingDeprecationWarning:pip\._vendor.+
-    ignore::DeprecationWarning:pip\._vendor.+
 markers =
     network: mark tests that require internet access
 


atugushev:
> No longer appear during tests so these filters can safely be removed.

I've tried it locally and got these warnings:

```
================================== warnings summary ==================================
.venv/src/pip/src/pip/_vendor/packaging/version.py:127: 72 warnings
tests/test_cli_compile.py: 5049 warnings
tests/test_cli_sync.py: 936 warnings
tests/test_repository_local.py: 108 warnings
tests/test_repository_pypi.py: 756 warnings
tests/test_resolver.py: 20 warnings
tests/test_top_level_editable.py: 81 warnings
  /Users/albert/Projects/pip-tools/.venv/src/pip/src/pip/_vendor/packaging/version.py:127: DeprecationWarning: Creating a LegacyVersion has been deprecated and will be removed in the next major release
    warnings.warn(

tests/test_resolver.py: 22 warnings
  /Users/albert/Projects/pip-tools/.venv/src/pip/src/pip/_vendor/packaging/specifiers.py:283: DeprecationWarning: Creating a LegacyVersion has been deprecated and will be removed in the next major release
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

Since `pip-tools` has no control over what uses pip internally those warnings were disabled intentionally to avoid distraction. However, `pip-tools` uses vendored packages too:

```console
$ git grep pip._vendor | wc -l
      13
```

This makes me think that we should not filter those warnings to avoid missing a valuable warning 🤔

jdufresne:
Okay, I can reproduce the warnings locally. Not sure what I was looking at before.

> This makes me think that we should not filter those warnings to avoid missing a valuable warning thinking

IMO, this is the right thing to do. We don't want to ignore upstream warnings if they could point towards required action. Shall we go ahead?

atugushev (approved):
> Okay, I can reproduce the warnings locally. Not sure what I was looking at before.
> 
> > This makes me think that we should not filter those warnings to avoid missing a valuable warning thinking
> 
> IMO, this is the right thing to do. We don't want to ignore upstream warnings if they could point towards required action. Shall we go ahead?

Yeah, let's merge it then. Thanks! 👍

jdufresne:
The warning listed above looks like the upstream issue: https://github.com/pypa/setuptools/issues/2466

PR #1253, opened 2020-11-28
Avoid spurious output during tests

jdufresne:
It is often convenient to use the pytest option "-s" (shortcut for
--capture=no) to view one's own debugging print() output. When there is
already lots of spurious output, it produces lots of noise and it may be
difficult to view the intended debugging output. By avoiding unnecessary
output, it is easier to find.

Tests that have intentional output now assert that output. For example,
the output of the sync command is now asserted. In addition to the
advantage above, this creates a more robust test suite as the expected
behavior is now more explicit, precise, and better covered.

##### Contributor checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,9 @@
 import json
 import os
+import subprocess
 import sys
 from contextlib import contextmanager
 from functools import partial
-from subprocess import check_call
 from textwrap import dedent
 
 import pytest
@@ -259,6 +259,9 @@ def make_package(tmp_path):
                     setup(
                         name={name!r},
                         version={version!r},
+                        author="pip-tools",
+                        author_email="pip-tools@localhost",
+                        url="https://github.com/jazzband/pip-tools",
                         install_requires={install_requires_str},
                     )
                     """.format(
@@ -268,6 +271,12 @@ def make_package(tmp_path):
                     )
                 )
             )
+
+        # Create a README to avoid setuptools warnings.
+        readme_file = str(package_dir / "README")
+        with open(readme_file, "w"):
+            pass
+
         return package_dir
 
     return _make_package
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -259,6 +259,9 @@ def _make_package(name, version="0.1", install_requires=None):
                     setup(
                         name={name!r},
                         version={version!r},
+                        author="pip-tools",
+                        author_email="pip-tools@mailinator.com",
```
atugushev, line 263:
```suggestion
                        author_email="pip-tools@localhost",
```

Maybe localhost to avoid ads?

tests/conftest.py
```diff
@@ -281,9 +290,12 @@ def run_setup_file():
...
         setup_file = str(package_dir_path / "setup.py")
-        return check_call(
-            (sys.executable, setup_file) + args, cwd=str(package_dir_path)
-        )  # nosec
+        with open(os.devnull, "w") as fp:
+            return subprocess.check_call(
+                (sys.executable, setup_file) + args,
+                cwd=str(package_dir_path),
+                stdout=fp,
+            )  # nosec
```
atugushev, line 298:
```suggestion
            return subprocess.check_call(
                (sys.executable, setup_file) + args,
                cwd=str(package_dir_path),
                stdout=subprocess.DEVNULL,
            )  # nosec
```

Maybe [subprocess.DEVNULL](https://docs.python.org/3/library/subprocess.html#subprocess.DEVNULL)?

tests/test_sync.py
```diff
@@ -458,6 +459,17 @@ def test_sync_ask_accepted(check_call, confirm, from_line, dry_run):
 
     confirm.assert_called_once_with("Would you like to proceed with these changes?")
```
atugushev, line 460: I wonder could it be refactored to check the message in `std{err,out}` instead?

tests/conftest.py
jdufresne: Unfortunately, that is not available on Python 2 :cry:. If we start dropping it now, I'll change this code otherwise I'll fix it in PR #1243.

tests/test_sync.py
jdufresne: Good idea and thanks for the review. I have implemented this in the latest revision.

tests/conftest.py
atugushev: Aw, right... I forgot about py2

atugushev (approved):
Thanks! 👍

PR #1254, opened 2020-11-29
Remove now-unnecessary PIP_REQ_TRACKER cleanup

jdufresne:
Since pip 20.0, this is cleaned up automatically by pip. See upstream
commits:

https://github.com/pypa/pip/commit/4a97d91a39ea94c4ef285bc870d5872aa283be8b
get_requirement_tracker() is introduced which uses a context manager to
set and reset the PIP_REQ_TRACKER environment variable.

https://github.com/pypa/pip/commit/b953ef4d18b3a86e19213685c6572732b32a034b
The PIP_REQ_TRACKER is no longer set directly in favor of the context
manage mentioned above.

##### Contributor checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -247,17 +247,9 @@ class PyPIRepository(BaseRepository):
 
             with global_tempdir_manager():
                 wheel_cache = WheelCache(self._cache_dir, self.options.format_control)
-                prev_tracker = os.environ.get("PIP_REQ_TRACKER")
-                try:
-                    self._dependencies_cache[ireq] = self.resolve_reqs(
-                        download_dir, ireq, wheel_cache
-                    )
-                finally:
-                    if "PIP_REQ_TRACKER" in os.environ:
-                        if prev_tracker:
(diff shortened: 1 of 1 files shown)


atugushev (commented):
It was reverted back again https://github.com/pypa/pip/commit/c1817d8b148c5f2f0a4b689399e014ad47ad9056

jdufresne:
You're right. The work was continued in the commits:

https://github.com/pypa/pip/commit/4a97d91a39ea94c4ef285bc870d5872aa283be8b
`get_requirement_tracker()` is introduced which uses a context manager to set and reset the PIP_REQ_TRACKER environment variable.

https://github.com/pypa/pip/commit/b953ef4d18b3a86e19213685c6572732b32a034b
The `PIP_REQ_TRACKER` is no longer set directly in favor of the context manage mentioned above.

Both of these are included in pip 20.0. So I think this is still the right change, just the original reasoning was wrong. I've updated the commit message.

atugushev (approved):
> Both of these are included in pip 20.0. So I think this is still the right change, just the original reasoning was wrong. I've updated the commit message.


Aw, great then! Thanks for the analysis! 👍

PR #1255, opened 2020-11-29
Use context manager to restore environment variables in resolve()

jdufresne:
In the event of an unexpected exception, the system's environment
variable PIP_EXISTS_ACTION should still be restored to its previous
state.

This context manager already exists in the upstream pip library. Can
reuse it to avoid duplication.

Unlike the previous implementation, if the PIP_EXISTS_ACTION environment
variable existed prior to executing pip-tools, that state will be
restored.

##### Contributor checklist

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -2,11 +2,11 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import copy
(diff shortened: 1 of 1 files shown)


atugushev (approved):
Nice! 👍

PR #1261, opened 2020-12-04
Make some long hashes in tests slightly more readable

jdufresne:
The expected stdout is easier to read when the lines aren't broken up
mid-line. Expand the allowed line length just enough to make these long
lines look more like the real output.

##### Contributor checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,7 +61,7 @@ markers =
     network: mark tests that require internet access
 
 [flake8]
-max-line-length = 100
+max-line-length = 88
 extend-ignore = E203  # E203 conflicts with PEP8; see https://github.com/psf/black#slices
 
 # flake8-pytest-style
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -414,11 +414,7 @@ def test_editable_package_in_constraints(pip_conf, runner, req_editable):
 
 @pytest.mark.network
 def test_editable_package_vcs(runner):
-    vcs_package = (
-        "git+git://github.com/jazzband/pip-tools@"
-        "f97e62ecb0d9b70965c8eff952c001d8e2722e94"
-        "#egg=pip-tools"
-    )
+    vcs_package = "git+git://github.com/jazzband/pip-tools@f97e62ecb0d9b70965c8eff952c001d8e2722e94#egg=pip-tools"  # noqa: E501
     with open("requirements.in", "w") as req_in:
         req_in.write("-e " + vcs_package)
     out = runner.invoke(cli, ["-n", "--rebuild"])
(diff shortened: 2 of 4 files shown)


setup.cfg
```diff
@@ -61,7 +61,7 @@ markers =
     network: mark tests that require internet access
 
 [flake8]
-max-line-length = 100
+max-line-length = 110
```
atugushev, line 64: I do like how the code looks now. One more thing, maybe it's better to keep `max-line-length=88` and ignore `E203` only in tests? For example:

```ini
[flake8]
per-file-ignores =
    # To make long strings wth hashes in tests more readable
    test/*: E203
```

What do you think?

jdufresne: Hmm. My preference would be to use a single flake8 configuration rather than try to manage a subset per file. IMO, it is simpler and more universal.

Black will continue to wrap lines at 88 characters, this affects long strings only, of which there are only a few. So this will not encourage more long lines in new code.

FWIW, Django and other projects use a value of 119: https://github.com/django/django/blob/d8dfff2ab0edf7a1ca5255eccf45c447b2f9d57e/setup.cfg#L63

atugushev: > Hmm. My preference would be to use a single flake8 configuration rather than try to manage a subset per file. IMO, it is simpler and more universal.

I'm not sure that I understood it correctly. The suggestion is to add `per-file-ignores` to `setup.cfg` to `flake8` section, which is the single flake8 configuration.

> So this will not encourage more long lines in new code.

At least this could happen:
```python
log.debug(
    "A very-very-very-very-very-very-very-very-very-very-very-very-very log debug message" # 110-th column
)
```

jdufresne: > I'm not sure that I understood it correctly. The suggestion is to add per-file-ignores to setup.cfg to flake8 section, which is the single flake8 configuration.

Yeah, I understand. But my preference would to not introduce `per-file-ignores`. I think it would be better to keep a single, uniform configuration across the project.

> At least this could happen:

I think that example would be caught during review. Also, outside these test hashes, I think it is rare for a string to contain such a long unbroken word.

atugushev: > Yeah, I understand. But my preference would to not introduce `per-file-ignores`. I think it would be better to keep a single, uniform configuration across the project.

Agreed, it makes sense. At the same time, I prefer a uniform configuration too, which is why there was 88 on flake8 and black configuration.

> I think that example would be caught during a review. 

Right, but robots do this job better than humans 😄 

> Also, outside these test hashes, I think it is rare for a string to contain such a long unbroken word.

Yes, which is why I suggested ignoring `E501` for tests. Or we could add `flake8: ignore E501` next to those few long strings.

atugushev: The latter I like even more.

jdufresne: I've done this in the latest revision.

FWIW, I'm not thrilled with the result. It extends the lines even farther to the right and requires repetitive additions to these lines throughout the code. IMO, we're better off extending `max-line-length` to a ~110-119 to allow for such lines in tests. But if you prefer this `noqa` style, we can go ahead.

PR #1262, opened 2020-12-04
Remove bandit lint tool

jdufresne:
I'm not convinced this has demonstrated its benefit to the code base.
When introduced, it changed one assert usage to a raise -- that's it --
which doesn't seem all that interesting of a change. Beyond that, this
tool requires many useless "nosec" comments for each suprocess call,
which I find distracting.

The tool itself adds overhead -- even if small -- to all contributor's
test runs and CI. IMO, we're better off without it.

##### Contributor checklist

- [ ] Provided the tests for the changes. (n/a)

### DIFF
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -2,7 +2,7 @@ import collections
 import os
 import sys
 import tempfile
-from subprocess import run  # nosec
+from subprocess import run
 from typing import Deque, Dict, Iterable, List, Optional, Set, Tuple, ValuesView
 
 import click
@@ -202,7 +202,7 @@ def sync(
 
     if not dry_run:
         if to_uninstall:
-            run(  # nosec
+            run(
                 [
                     sys.executable,
                     "-m",
@@ -230,7 +230,7 @@ def sync(
             tmp_req_file.close()
 
             try:
-                run(  # nosec
+                run(
                     [
                         sys.executable,
                         "-m",
--- a/.bandit
+++ b/.bandit
@@ -1,3 +1,0 @@
-[bandit]
-exclude: tests,.tox,.eggs,.venv,.git
-skips: B101
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,12 +26,6 @@ repos:
         # Avoid error: Duplicate module named 'setup'
         # https://github.com/python/mypy/issues/4008
         exclude: ^tests/test_data/
-  - repo: https://github.com/PyCQA/bandit
-    rev: 1.7.0
-    hooks:
-      - id: bandit
-        args: [--ini, .bandit]
-        exclude: ^tests/
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v2.2.1
     hooks:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -284,7 +284,7 @@ def run_setup_file():
             cwd=str(package_dir_path),
             stdout=subprocess.DEVNULL,
             check=True,
-        )  # nosec
+        )
 
     return _run_setup_file
 


piptools/sync.py
```diff
@@ -2,7 +2,7 @@
 import os
 import sys
 import tempfile
-from subprocess import check_call  # nosec
```
atugushev, line 5: > many useless "nosec" comments for each subrocess call

There must be an option to add `check_call` to `allow-list`.

atugushev (commented):
I'd prefer to have extra eyes, but I'd say that I'm `-0` on this change. I wonder what @codingjoe would think?

jdufresne:
In practice, since the introduction, do you feel this has caught a single legitimate security issue? If so, do you have any details?

Any thoughts on why more Python projects on using this tool?

codingjoe:
Hi there, that is an interesting one.
I presume we all agree that security is important and needs to be taken seriously. As I understand the argument, we are questioning if bandit contributes to this goal in a meaningful way. The hypothesis, it only adds a slight overhead, without any measurable outcome.

I would like to propose another way to look that this. I would suggest adding a single comment that alerts both author and reviewer to potential security implications. Following that thought, I do not agree that this tools adds overhead, but shines a spotlight on code that needs additional attention to provide another layer of security.

Moreover, I believe that the argument can not only include the past, but must consider the future. Just because we didn't have any security issues in the past doesn't mean we don't have any in the future. We must also consider, that bandit has prevented potential security risks in the past without our knowledge.

In conclusion, I love to delete useless code, just like the next developer, but in this particular case I must disagree. `bandit` pulls its weight and rather sacrifice a contributors time over security.

piptools/sync.py
webknjaz: I think the comments must remain where they are: the check is not about whether the import is dangerous, it's rather about how the function is called and if any user input is passed there unsanitized. Such things must be checked on a case-by-case basis. You cannot just globally declare it safe and hope for the better.
The only case when it's fine to ignore it is tests: https://github.com/ansible/pylibssh/blob/afabcea/.flake8#L89.
By the way, I like [wemake-python-styleguide](https://wemake-python-stylegui.de) which is a collection of flake8 linters pulled in together. It uses flake8-bandit that is quite handy because you get to use flake8 ignore machinery when necessary and it becomes a part of the linting suite where it belongs. I recommend switching to it.

webknjaz (changes requested):
I agree with @codingjoe's sentiment. Security should be practice continuously, not only when you hit a CVE.
Moreover, this project lacks the effort: there's no security policy set up at https://github.com/jazzband/pip-tools/security (meaning that the users cannot send their findings to the maintainers in secret), and no of the code-scanning GHA workflows GH offers are enabled either.
This is why I'm :-1: on removing the only security-related thing that still exists in the project.

jdufresne:
Yes, we all agree with the goal, but this tool strikes me more as theater rather than something of substance.

Any reasonable code review should take security issues into consideration whether bandit is in use or isn't.

piptools/sync.py
jdufresne: > I think the comments must remain where they are: the check is not about whether the import is dangerous, it's rather about how the function is called and if any user input is passed there unsanitized. Such things must be checked on a case-by-case basis. You cannot just globally declare it safe and hope for the better.

I agree that subprocess has the potential for misue, but that isn't the case with pip-tools. These comments just add noise and don't fulfill the implication of a real security audit.

webknjaz:
@jdufresne bandit is a linter. Just like any other linter, it's an assistive technology. It should help reviewers and implementers not oversee rookie mistakes. The reviewer should pay attention to whether the ignores are fair, linter-wise. Of course, they need to also take into account other things but w/o automations like this, it adds more mental overhead to the process.

piptools/sync.py
webknjaz: Comments could have extended explanation, yes. But the presence of them adds more context and information that actual humans checked this. If you do global ignores, there's no trace of whether the simplest problems are even checked.

jdufresne:
Yes, I understand what the tool is and its intent. I still think it isn't a very useful for this project as I don't find it to be effective at achieving its goals and I don't agree that it reduces mental overhead. I think it increases it. I had to learn what "nosec" means. I had to investigate what the tool is all about. :shrug:

Since its introduction, it has created two concrete changes:

- Changed `assert` statements into explicit exceptions. This is a horizontal move and is really an opinion on style more so than anything security related. Yes, I'm aware `assert` can be optimized away.

- Forced every usage of `subprocess` to include a noisy `# nosec` comment. This I consider theater and an annoyance. This isn't making anyone rethink the use of subprocess. If someone were to analyze the project's usage of they would see that there isn't anything alarming about its use.

That's it. Reading through the rest of its rules, I don't find them interesting for the purposes of this project.

piptools/sync.py
jdufresne: When I read these comments, I don't think these comments communicate anything useful. They just come across as terse noise. Expanding the comment seems entirely unnecessary to me. The use of subprocess looks pretty straightforward, logical, and obvious to me.

webknjaz:
@jdufresne this sounds a lot like "since it didn't help me, there's no way it would help anybody else". I don't think everybody who faced a legitimate linter failure and fixed it would be posting about the success on an issue tracker.

> * Forced every usage of `subprocess` to include a noisy `# nosec` comment. This I consider theater and an annoyance. This isn't making anyone rethink the use of subprocess.

This again is an instance of "thinking annoys me therefore everybody must turn their brains off". This I consider theater and an annoyance. It really does make other people rethink the use of subprocess, just not you. And this is fine. One of the conclusions could be that any use of subprocess should be put in a separate module/subpackage and abstracted away from the rest of the code. This separate place would need to be watched more closely and probably implement some safeguards. It would be the only place having the "annoying" ignore comments but would have to have really comprehensive comments describing the rationale. And for the rest of the project, it would simply be forbidden to use subprocess directly. This way of separation of concerns can be a great positive impact of the bandit tool if used wisely.

> If someone were to analyze the project's usage of they would see that there isn't anything alarming about its use.

And that is why those comments are in place. Too bad that they don't have the motivation written alongside, though.

piptools/sync.py
webknjaz: It's not the use of subprocess that needs to be expanded but the reasoning behind the ignore comments.

jdufresne:
> "since it didn't help me, there's no way it would help anybody else"

...

> "thinking annoys me therefore everybody must turn their brains off"

No, this does not describe my position or how I think. With all due respect, I find the tone of this last comment a bit insulting.

---

I presented this PR, as a frequent contributor, in good faith, in an attempt to improve the tooling around the project and ease contributions for myself and others. I understand you disagree and that is fine, but these accusations like "thinking annoys me" and "therefore everybody must turn their brains off" aren't helpful or welcome.

codingjoe:
@jdufresne your contribution is much appreciated, regardless of my opinion on this matter.

I believe an argument has been made and there isn't much to add, that furthers the cause without risking being hurtful.
At the end, this seems to boil down to a difference in values and people will come to a different conclusion even if presented with the same evidence. Therefore, I would advise we leave it like that for now.

@jdufresne it is up to you to withdraw you pull-request, but my advice: It's probably not worth the effort and will make you unhappy, which I don't wish on anyone.

webknjaz:
@jdufresne I apologize for making you feel that way. I guess what I was trying to communicate is that there's a useful impact of bandit that may be perceived subjectively depending on personal preferences but nevertheless, there's a way of making it less annoying (and I personally think that it's a good practice) by quarantining any potentially dangerous invocations in a separate location and disabling rules that everyone agrees on as misleading. This would also improve the structural aspect of the app (and its maintainability).

P.S. Calling things theatrical and diminishing their value is far from being constructive and it is what felt insulting to me in the first place. I think that's why overreacted. Apologies again.

ssbarnea:
Closing as this is going nowhere and clutters the list of open PRs. Usually changes that are likely to be controversial are better to be discussed with other cores before being proposed. I made similar mistakes myself and I know the feeling of investing time and not being able to merge my patch due to failure to gather support.

To everyone involved, be cool and don't assume bad intentions, sometimes the language/culture barrier can be an extra challenge we have to face.

PR #1263, opened 2020-12-04
Remove unused .fussyfox.yml

jdufresne:
At some point, FussyFox was turned off. It no longer runs on PRs. So
this file goes unused. As it is unused, it only adds clutter to the top
level directory which is already full of other configuration files.

Regardless, I'm not sure this is needed any longer. After the move to
GitHub Actions, CI is much faster and the lint targets run in parallel
to the tests. I think whatever benefit it provided is now builtin.

##### Contributor checklist

- [ ] Provided the tests for the changes. (n/a)


atugushev (commented):
It seems "Fussy Fox" hasn't been worked in a while. @codingjoe are you okay with that change?

codingjoe:
It's not working because it's not installed. So, we can either remove it or ask a roady to install it. Since I am affiliated with the tool I will yield my opinion ;)

webknjaz:
@jazzband/roadies :point_up:

jdufresne:
What benefit would installing it achieve that isn't already there through GHA? I'm not familiar enough with the tool to know offhand and the webpage doesn't make it obvious to me.

jezdez:
Hi all, sorry for the lack of communication regarding Fussy Fox. It was turned off since I want to reduce the number of 3rd party dependencies following the issues with Travis CI.

@codingjoe I know you've spent a non-trivial amount of time on Fussy Fox, so I want to be clear that this decision is not based on my experience with Fussy Fox, which was great, but rather a strategic with regard to long-term maintenance of Jazzband project setups. Thank you for all your work, it is much appreciated.

Generally speaking GHA is pretty good and there is little risk just going all-in since we either could pay GitHub to get more workers or host the workers ourselves if we continue to grow like right now.

So my current half-baked plan is to start recommending using precommit for linting eventually (as was requested a couple of times). That would allow using the same "lint.yml" GHA config for the majority of projects and just run the [precommit action](https://github.com/pre-commit/action) effectively.

I'm also currently trialing https://pre-commit.ci/ for the Jazzband website code, but that is a bit orthogonal to all that and could again introduce an unwanted dependency to a 3rd party.

codingjoe:
> What benefit would installing it achieve that isn't already there through GHA? I'm not familiar enough with the tool to know offhand and the webpage doesn't make it obvious to me.

It's faster. That's it. No added benefits.

codingjoe:
> Hi all, sorry for the lack of communication regarding Fussy Fox. It was turned off since I want to reduce the number of 3rd party dependencies following the issues with Travis CI.
> 
> @codingjoe I know you've spent a non-trivial amount of time on Fussy Fox, so I want to be clear that this decision is not based on my experience with Fussy Fox, which was great, but rather a strategic one with regard to long-term maintenance of Jazzband project setups. Thank you for all your work, it is much appreciated.
> 
> Generally speaking GHA is pretty good and there is little risk just going all-in since we either could pay GitHub to get more workers or host the workers ourselves if we continue to grow like right now.
> 
> So my current half-baked plan is to start recommending using precommit for linting eventually (as was requested a couple of times). That would allow using the same "lint.yml" GHA config for the majority of projects and just run the [precommit action](https://github.com/pre-commit/action) effectively.
> 
> I'm also currently trialing https://pre-commit.ci/ for the Jazzband website code, but that is a bit orthogonal to all that and could again introduce an unwanted dependency to a 3rd party.

@jezdez I spend a non-trivial amount of time on many things, that I maybe shouldn't 😄

PR #1264, opened 2020-12-04
Use more descriptive exceptions than AssertionError

jdufresne:
Better match Python conventions and the stdlib by using TypeError and
ValueError rather than AssertionError.

For example, in CPython:

    $ python -c 'int(object())'
      File "<string>", line 1, in <module>
    TypeError: int() argument must be a string, a bytes-like object or a number, not 'object'

##### Contributor checklist


atugushev (approved):
LGTM 👍

PR #1266, opened 2020-12-05
Remove unnecessary system calls via os.path.exists()/os.path.isdir()

jdufresne:
Rather than checking a file exists, then reading it, simply read it and
handle the potential ENOENT OS error. This reduces the number of system
calls from:

    1. check file exists
    2. read file

To:

    1. read file

Reducing system calls is an optimization.

More importantly, this also avoids race conditions that were previously
not handled. If the file was deleted after the existence check but
before reading it, an unhandled FileNotFound error would be raised.

This approach better follows Python pattern "Easier to ask for
forgiveness than permission" rather than the "Look before you leap"
approach. See: https://docs.python.org/3/glossary.html#term-eafp

The same pattern applies to os.makedirs(). Add a compatibility shim to
allow using the exist_ok parameter on Python 2.

##### Contributor checklist

### DIFF
--- a/piptools/_compat/__init__.py
+++ b/piptools/_compat/__init__.py
@@ -2,6 +2,9 @@
 # flake8: noqa
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import errno
+import os
+
 from pip._vendor import six
 
 from .pip_compat import PIP_VERSION, parse_requirements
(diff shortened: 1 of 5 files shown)


atugushev (approved):
Awesome! Thanks 👍

PR #1270, opened 2020-12-06
Remove unreachable code in PyPIRepository.get_hashes()

jdufresne:
get_hashes() is only called after the dependencies have been resolved.
After they have been resolved, requirements are either links or pinned
requirements.

If the requirement is a link, the "if ireq.link" block is executed and
the function returns early. If the requirement is not a link, then it
has been resolved to an exact version. Either way, the exception will
never be raised.

This line is untested and so frequently appears as uncovered during
coverage analysis, erroneously marking the PR with a big red ❌ when the
PR is otherwise green.

##### Contributor checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -339,9 +339,6 @@ class PyPIRepository(BaseRepository):
                     cached_link = link
                 return {self._get_file_hash(cached_link)}
 
-        if not is_pinned_requirement(ireq):
-            raise TypeError("Expected pinned requirement, got {}".format(ireq))
-
         log.debug(ireq.name)
 
         with log.indentation():


atugushev (commented):
I'd rather add a test and keep the guard.

webknjaz:
+1 for adding a test instead.

jdufresne:
The branch represent an impossible state that can't be reached under any pip-tools use. The test itself would be contrived and artificial.

webknjaz:
Then it should be turned into an assertion.

jdufresne:
The bandit tool forbids `asser`. IMO, a misguided rule, but one that exists none the less. As long as we continue to use bandit, we can't use it `assert`.

webknjaz:
Can't we disable the `assert` rule in bandit? I'm pretty sure it's possible if it's integrated via `flake8-bandit` at least.

jdufresne:
I'm not interested in fussing with bandit. As you know, I don't find the tool useful or constructive. To me, it is security theater, results in noisy comments, and adds mental overhead to contributors to understand its usage.

My suggestion:

1. We either continue handling this code block as an exception, treat it as a defensive check only, and add the contrived test.
2. We decide we don't need this defensive check and remove it.

I went with option 2 when opening this PR, but if there is reason to believe this will catch real bugs I'll go ahead with option 1.

webknjaz:
It looks to me that the defensive check was added there as a contract verification and not a user input check because it's too deep in the control flow. I suppose the user input would be checked early in an outer layer of the software.

jdufresne:
Yes, that is my interpretation too and that explains why the code is not executed in real life runs.

jdufresne:
Closing in favor of #1272 which tests the guard.

Issue #972, opened 2019-10-26
Type annotations


atugushev:
#### What's the problem this feature will solve?

It would be nice to have type annotations throughout the codebase. This issue is a general tracking issue of how we would use mypy and typings.

#### Describe the solution you'd like

- use [pytest-annotate](https://github.com/kensho-technologies/pytest-annotate) to auto add typings (or something similar)
- add typings incrementally, one PR per module (e.g., 21b713018d2dba6efef0e3d4e71facac1c4f0b38) to help reviewers

#### Alternative Solutions

Wait until we drop Python 2 support and add Python 3 style type annotations.

#### Additional context

Docs:

- https://www.python.org/dev/peps/pep-0484/
- https://realpython.com/python-type-checking/

Useful tools:

- https://github.com/dropbox/pyannotate
- https://github.com/kensho-technologies/pytest-annotate
 
##### Maintenanse
- [X] add mypy checking to pre-commit config (#1275)
- [X] add `[mypy]` section to setup.cfg (#1275)

##### Add typings to the following files

- [x] piptools/_compat/pip_compat.py (#1315)
- [x] piptools/cache.py (#1315, #1316)
- [x] piptools/exceptions.py (#1315)
- [x] piptools/logging.py (#1315)
- [x] piptools/repositories/base.py (#1315)
- [x] piptools/repositories/local.py (#1341)
- [x] piptools/repositories/pypi.py (#1341)
- [x] piptools/resolver.py (#1316)
- [x] piptools/scripts/compile.py (#1322)
- [x] piptools/scripts/sync.py (#1331)
- [x] piptools/sync.py (#1331)
- [x] piptools/utils.py
- [x] piptools/writer.py (#1310)

##### Final steps

This list must be done after the above lists.

- [ ] add `disallow_untyped_defs = True` to setup.cfg (#1350)

atugushev:
Hey, @jazzband/pip-tools. Please, join the discussion. Have you any other thoughts or helpful tools we could use to integrate type annotations?

webknjaz:
Could probably use https://github.com/Instagram/MonkeyType to collect types from runtime.

P.S. I think, it's better to use a separate `mypy.ini` rather than flooding `setup.cfg` with things that aren't related to packaging.

codingjoe:
[...]

PR #1275, opened 2020-12-15
Use mypy for static type checking

jdufresne:
Type checking helps give confidence that APIs are being used correctly,
especially during large refactoring (e.g. eventually dropping Python 2
support).

This initial pass only includes type comments that were necessary to
make mypy pass. Future pull requests will add types for the rest of the
project once the workflow is established.

pip has started including type annotation in a subset of its code. When
major changes to pip are released, the type checking will help identify
API mismatches and areas to adjust.

mypy is configured to be as strict as possible without introducing
errors. The configuration is a subset of the --strict CLI argument.

Refs #972

##### Contributor checklist

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -16,6 +16,7 @@ jobs:
       matrix:
         toxenv:
           - checkqa
+          - mypy
           - readme
         python-version:
           - "3.x"
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -16,7 +16,6 @@ jobs:
       matrix:
         toxenv:
           - checkqa
-          - mypy
           - readme
         python-version:
           - "3.x"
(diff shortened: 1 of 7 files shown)


jdufresne:
The failure is being addressed in typeshed here: https://github.com/python/typeshed/pull/4825

atugushev:
See also #972.

jdufresne:
The failure would also be addressed by dropping Python 2 support, see #1243. If that lands, I'll rebase this work.

jdufresne:
Changes:

- Rebased
- Changed type comments to type annotations
- Dropped import guard; the typing module is available on all support Pythons

tox.ini
```diff
@@ -33,6 +34,11 @@ deps = pre-commit
 commands_pre =
 commands = pre-commit run --all-files --show-diff-on-failure
 
+[testenv:mypy]
```
atugushev, line 37: We can add mypy to `pre-commit` config instead to be able to run it on `checkqa`.

.github/workflows/qa.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       matrix:
         toxenv:
           - checkqa
+          - mypy
```
atugushev, line 19: There is no need int additional dimension if we add mypy to `pre-commit`.

jdufresne:
No problem. Moved to pre-commit.

atugushev (approved):
👍

PR #1276, opened 2020-12-16
Add `--split_wheels` option to pip-compile that generates wheels and non-wheels in separate files

violalyu:
- Updated unittest for test_cli_compile
- Include requirements with `_wheels` suffix for default lookup
- Added `utils.format_suffixed_path`, updated method for split wheel outputing to files

Added `--split_wheels` option to pip-compile that generates wheels and non-wheels in separate files

**motivation**:
We are using pip-tools to update our virtual environments and currently it is slow to do that because there are a number of non-wheels that get installed; if we have separate requirement files we'll be able to handle these separately and improve the overall performance. 

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -21,11 +21,13 @@ from ..locations import CACHE_DIR
 from ..logging import log
 from ..repositories import LocalRequirementsRepository, PyPIRepository
 from ..resolver import Resolver
-from ..utils import UNSAFE_PACKAGES, dedup, is_pinned_requirement, key_from_ireq
+from ..utils import (UNSAFE_PACKAGES, dedup, format_suffixed_path,
+                     is_pinned_requirement, key_from_ireq)
 from ..writer import OutputWriter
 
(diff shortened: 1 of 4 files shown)


jdufresne:
What is the motivation for this change? Can you provide a bit of background about the use case?

violalyu:
> Thanks for opening a pull request!
> 
> What is the motivation for this change? Can you provide a bit of background about the use case?

Thanks @jdufresne ! I've updated in the PR description in the `motivation` section as well:

"We are using pip-tools to update our virtual environments and currently it is slow to do that because there are a number of non-wheels that get installed; if we have separate requirement files we'll be able to handle these separately and improve the overall performance. "

ssbarnea:
Closing this because I do not think it will gather enough support from others (at least it did not in 6 months).

I am personally a little bit against it because it adds extra complexity for something that seems to be a very edge case. Who is going to maintain this extra feature in longer term.

If we would find enough supporters across people that contributed to the project, I would not fight it.

PR #1288, opened 2021-01-01
Remove language restrictions in pre-commit config

atugushev:
Since we have dropped py2, there is no need for `language_version: python3` restriction


jdufresne (approved):

jdufresne:
If you'd like, could do the same in `tox.ini`.

```
tox.ini
30   basepython = python3
```

atugushev:
Thanks @jdufresne!

PR #1289, opened 2021-01-01
Use prettier pre-commit hook for md, json, and yaml formatting

jdufresne:


.pre-commit-config.yaml
```diff
@@ -2,24 +2,28 @@ repos:
...
-    - id: bandit
-      language_version: python3
-      exclude: ^tests/
+      - id: bandit
+        language_version: python3
+        exclude: ^tests/
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.2.1
+    hooks:
+      - id: prettier
```
atugushev, line 29: Is `prettier` requires `node` to be installed in the OS? If so it might be not convenient for python contributors, because it would add additional steps to set up the workspace. Could we pick any other YAML/MD/JSON formatters from the official [pre-commit hooks page](https://pre-commit.com/hooks.html) for the Python ecosystem? I've tried the following below.

YAML:
- https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
- https://github.com/macisamuele/language-formatters-pre-commit-hooks (pretty-format-yaml)

Markdown:
- https://github.com/executablebooks/mdformat

JSON:
- https://github.com/pre-commit/pre-commit-hooks (pretty-format-json)

... and the results were pretty decent. What do you think?

atugushev: See also #1291

jdufresne: Yeah, prettier does require node. However, both node and prettier are de facto standard and common tools. I personally don't think it will be a large hardship to use. One option to further reduce contributor hurdles would be to use https://pre-commit.ci/. It automatically updates PR with the pre-commit changes so users don't need to run hooks locally. If you agree with that, I'm happy to go ahead with that change.

Overall, I still prefer using prettier. It is a more established and popular tool than the ones listed. There are already lots of developers familiar with it. It also combines many tools down to one.

atugushev: Yeah, `pre-commit.ci` would be a win-win. Let's go ahead with `prettier` then, its results look much prettier than the formatters' above.

atugushev (approved):

.pre-commit-config.yaml
```diff
@@ -2,24 +2,28 @@ repos:
...
-      language_version: python3
+      - id: black
+        args: [--target-version=py36]
+        language_version: python3
   - repo: https://github.com/PyCQA/isort
     rev: 5.6.4
     hooks:
-    - id: isort
-      language_version: python3
+      - id: isort
```
webknjaz, line 11: Extra sequence indentation looks weird

jdufresne: The style is quite conventional and common. Beyond being the standard formatting of prettier, it is also the formatting used by the popular Ansible project. For example: https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html

PR #1291, opened 2021-01-03
Add YAML formatter to pre-commit config

atugushev:


.github/workflows/cron.yml
```diff
@@ -3,7 +3,7 @@ name: Cron
 on:
   schedule:
     # Run every day at 00:00 UTC
```
jdufresne, line 5: This comment looks over indented.

jdufresne (commented):
I prefer the formatting produced by prettier. I think the extra indentation is more conventional and more readable.

atugushev:
Superseded by #1289.

PR #1295, opened 2021-01-03
Remove unnecessary piptools/click.py compat shim

jdufresne:
Unnecessary since dropping Python 2 support.

The attribute `click.disable_unicode_literals_warning` only affects
Python 2. On Python 3, it is a nop. For details, see:

- https://click.palletsprojects.com/en/7.x/python3/?highlight=disable_unicode_literals_warning#unicode-literals
- https://github.com/pallets/click/blob/7.1.2/src/click/_unicodefun.py#L32-L33

This removes the need to have a click compatibility shim. Now import the
package directly.

##### Contributor checklist

### DIFF
--- a/piptools/__init__.py
+++ b/piptools/__init__.py
@@ -1,6 +1,6 @@
 import locale
 
-from piptools.click import secho
+from click import secho
 
 # Needed for locale.getpreferredencoding(False) to work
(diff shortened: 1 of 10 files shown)


atugushev (approved):
Nifty-nifty! 👍🏼

PR #1301, opened 2021-01-08
Pipfile support

stefansjs:
Hey folks, I was playing with adding pipfile support to the pip-compile command locally and it looked like it worked pretty well (at least for my use cases). I thought the changes might be straight-forward enough to warrant at least sharing and see if upstream is interested in consuming those changes. I'm sure there's good reasons not to support Pipfiles right now that I just can't think of, so don't worry about being too harsh if you just don't like the idea at all.

What do y'all think?

**Changelog-friendly one-liner**: Support for reading Pipfile

##### Contributor checklist


stefansjs:
Understood. After living with this fork for a few months, it doesn't really provide a great quality of life improvement. I fully agree with closing the PR until Pipfile gets better definition and adoption.

PR #1315, opened 2021-02-06
Add typing to several areas where straightforward

jdufresne:
Refs: #972

### DIFF
--- a/piptools/exceptions.py
+++ b/piptools/exceptions.py
@@ -1,3 +1,8 @@
+from typing import Iterable
+
+from pip._internal.index.package_finder import PackageFinder
+from pip._internal.models.candidate import InstallationCandidate
+from pip._internal.req import InstallRequirement
 from pip._internal.utils.misc import redact_auth_from_url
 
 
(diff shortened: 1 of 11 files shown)


piptools/exceptions.py
```diff
@@ -6,12 +11,17 @@ class PipToolsError(Exception):
 
 
 class NoCandidateFound(PipToolsError):
-    def __init__(self, ireq, candidates_tried, finder):
+    def __init__(
+        self,
+        ireq: InstallRequirement,
+        candidates_tried: List[InstallationCandidate],
```
atugushev, line 17:
```suggestion
        candidates_tried: Iterable[InstallationCandidate],
```

Could be relaxed to pass any iterables

atugushev (approved):
Only a minor comment below. Otherwise, LGTM 👍🏼

piptools/exceptions.py
jdufresne: Thanks! I applied this suggestion in the latest revision.

PR #1316, opened 2021-02-07
Add type hints for the resolver module

atugushev:
Refs: #972 

Additionally refactored `lookup_table()`:
 - removed unused functionality
 - made function type safe
 - moved doctests to `tests/test_utils.py` and add more tests


piptools/resolver.py
```diff
@@ -25,27 +30,31 @@ class RequirementSummary:
...
-            and self.extras == other.extras
-        )
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, self.__class__):
+            return (
+                self.key == other.key
+                and self.specifier == other.specifier
+                and self.extras == other.extras
+            )
+        return NotImplemented
```
atugushev, line 46: FYI: this line left uncovered, so I've added it to `exclude_lines` to the coverage report config in a619d0dd2f34897b37ed48efff8b5a6aaa967d7e

webknjaz: Why not add tests instead?

webknjaz, line 46: Also, I believe it's better to have an inversed guard-expression as this reduces the number of nested lines
```suggestion
        if not isinstance(other, self.__class__):
            return NotImplemented

        return (
            self.key == other.key
            and self.specifier == other.specifier
            and self.extras == other.extras
        )
```

piptools/utils.py
```diff
@@ -152,89 +169,33 @@ def flat_map(fn, collection):
...
-        if key is None:
 
-            def keyval(v):
-                return v
+    def keyval(v: Union[_VT, Tuple[_KT, _VT]]) -> Tuple[_KT, _VT]:
+        if isinstance(v, tuple):
+            return v[0], v[1]
 
-        else:
+        assert key is not None, "key function must be specified"
```
webknjaz, line 184: This is a serious misuse of the `assert` statement. It must only check things that should always be true in runtime but here it seems to be used as an input validation. Moreover, it's an input validation of another function (`lookup_table()`).
`asserts` are stripped off in the optimized mode (except for the pytest runtime because it patches the bytecode so this would be different from the real runtime).
Another problem is that it's a check that (when enabled) is made multiple times, on each invocation in the loop. It should instead be done in the `lookup_table()` function itself.
Besides, this sort of checks should be done with an `if`-statement and a condition, followed by raising an appropriate exception explicitly. Raising an `AssertionError` is not a good API for this. An appropriate error to raise would be a `TypeError` or a `ValueError`.
```suggestion
```

piptools/utils.py
```diff
@@ -152,89 +169,33 @@ def flat_map(fn, collection):
...
-
-    >>> assert lookup_table(
-    ...     ['foo', 'bar', 'baz', 'qux', 'quux'],
-    ...     keyval=lambda s: (s[0], s[1:])) == {
-    ...     'b': {'ar', 'az'},
-    ...     'f': {'oo'},
-    ...     'q': {'uux', 'ux'}
-    ... }
-
     """
```
webknjaz, line 178: Here's how to improve this user-input check:
```suggestion
    """
    if key is None and values and any(not isinstance(v, tuple) for v in values):
        raise ValueError("The `key` function must be specified when the `values` are not empty.")

```

tests/test_utils.py
```diff
@@ -346,3 +347,26 @@ def test_get_compile_command_sort_args(tmpdir_cwd):
...
+        ),
+    ),
+)
+def test_lookup_table(values, key):
+    expected = {"b": {"bar", "baz"}, "f": {"foo"}, "q": {"quux", "qux"}}
+    assert lookup_table(values, key) == expected
+
+
+def test_lookup_table_requires_key():
+    with pytest.raises(AssertionError, match="^key function must be specified$"):
```
webknjaz, line 371: :exploding_head: Please don't use `AssertionError` outside of the test files as an API, it's hugely unreliable and you're not testing what you think you do.
Also, it's a good tone/convention to use r-strings for regexes.

webknjaz (changes requested):
Most of the change is good but I've added a few comments to places that are rather problematic.

.coveragerc
```diff
@@ -6,3 +6,20 @@ omit =
 
 [report]
 include = piptools/*, tests/*
+
+# Regexes for lines to exclude from consideration
+exclude_lines =
```
jdufresne, line 11: I don't think we should add this configuration. IMO, it is okay to understand that coverage isn't 100% and doing so paints a more accurate picture about what the coverage numbers mean. Plus, in this case, I think we could add a test.

.pre-commit-config.yaml
```diff
@@ -20,7 +20,7 @@ repos:
       additional_dependencies:
       - flake8-pytest-style
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.790
+    rev: v0.800
```
jdufresne, line 23: We can update pre-commit as orthogonal work. Might as well get all dependencies at once #1317

piptools/repositories/base.py
```diff
@@ -12,24 +15,26 @@ def freshen_build_caches(self):
         """Should start with fresh build/source caches."""
 
     @abstractmethod
-    def find_best_match(self, ireq):
+    def find_best_match(
+        self, ireq: InstallRequirement, prereleases: Optional[bool]
```
jdufresne, line 19: Could we default `preeleases` to `False` rather than making it optional?

piptools/resolver.py
```diff
@@ -100,34 +109,36 @@ def combine_install_requirements(repository, ireqs):
...
     def __init__(
         self,
-        constraints,
-        repository,
-        cache,
-        prereleases=False,
-        clear_caches=False,
-        allow_unsafe=False,
-    ):
+        constraints: List[InstallRequirement],
```
jdufresne, line 112: Can relax the `List` as the line coerces to a `Set` anyway.

```suggestion
        constraints: Iterable[InstallRequirement],
```

piptools/resolver.py
```diff
@@ -100,34 +109,36 @@ def combine_install_requirements(repository, ireqs):
...
+        repository: BaseRepository,
+        cache: DependencyCache,
+        prereleases: Optional[bool] = False,
```
jdufresne, line 115: Can we drop Optional?

```suggestion
        prereleases: bool = False,
```

piptools/resolver.py
```diff
@@ -389,7 +404,9 @@ def _iter_dependencies(self, ireq):
                 dependency_string, constraint=ireq.constraint, comes_from=ireq
             )
 
-    def reverse_dependencies(self, ireqs):
+    def reverse_dependencies(
+        self, ireqs: Set[InstallRequirement]
```
jdufresne, line 408: Only used for iteraration.

```suggestion
        self, ireqs: Iterable[InstallRequirement]
```

jdufresne (commented):
Nice work! Glad to see the typing continue.

piptools/utils.py
atugushev: Good point, thanks. Addressed in 1c125852462698c1d40e50f796f81c73f3411f2b.

piptools/utils.py
atugushev: Addressed in 1c125852462698c1d40e50f796f81c73f3411f2b.

tests/test_utils.py
atugushev: Okay, thanks. Addressed in 1c125852462698c1d40e50f796f81c73f3411f2b.

piptools/resolver.py
atugushev: Okay, let's add one.

atugushev: Yeah, looks better now. Thanks!

piptools/repositories/base.py
atugushev: Usually, prereleases goes to [SpecifierSet.filter()](https://github.com/pypa/pip/blob/bbf8466088655d22cd46b286c8f0b8150754c1d9/src/pip/_vendor/packaging/specifiers.py#L816) where it's optional for a special case. Do you still think we should drop `Optional`?

piptools/resolver.py
atugushev: Indeed. Thanks, addressed in 386c8e07c054176758ed65a937c9fbc9dad0c1d2.

piptools/resolver.py
atugushev: Right, thanks! Addressed in 386c8e07c054176758ed65a937c9fbc9dad0c1d2.

piptools/repositories/base.py
jdufresne: I see what you're saying. Good point. Lets match the upstream signature, so keep it optional. Thanks for explaining.

piptools/utils.py
atugushev: Note, that `assert key is not None` guard is still there to satisfy `mypy`. Any ideas to improve this?

piptools/resolver.py
atugushev: Added in cabb979.

atugushev:
@jdufresne @webknjaz thanks for helping me with this PR! 🙏🏼 I've addressed all suggestions and I think it's ready for a second review round.

piptools/utils.py
webknjaz: `typing.cast()`?

tests/test_resolver.py
```diff
@@ -441,3 +441,12 @@ def test_RequirementSummary_hash_equality(from_line, left_hand, right_hand, expe
...
     assert (hash(lh_summary) == hash(rh_summary)) is expected
+
+
+def test_requirement_summary_with_other_objects(from_line):
+    """
+    RequirementSummary should not be equal to any other object
+    """
+    requirement_summary = RequirementSummary(from_line("test_package==1.2.3"))
+    other_object = object()
+    assert (requirement_summary == other_object) is False
```
webknjaz, line 452: Maybe?
```suggestion
    assert not (requirement_summary == other_object)
```

webknjaz: Or even
```suggestion
    assert requirement_summary != other_object
```
(since there's no `__ne__()`, it should all back to `not requirement_summary.__eq__(other_object)`) and so it'll call `__eq__()` anyway.

tests/test_utils.py
```diff
@@ -346,3 +347,61 @@ def test_get_compile_command_sort_args(tmpdir_cwd):
...
             "requirements.in setup.py"
         )
+
+
+@pytest.mark.parametrize(
+    ("values", "key"),
+    (
+        pytest.param(
+            ("foo", "bar", "baz", "qux", "quux"),
+            lambda s: s[0],
```
webknjaz, line 357: I think it's more elegant to do
```suggestion
            operator.itemgetter(0),
```

tests/test_utils.py
```diff
@@ -346,3 +347,61 @@ def test_get_compile_command_sort_args(tmpdir_cwd):
...
+        pytest.param(("foo", "bar", "baz"), id="values are not tuples"),
+        pytest.param((("f", "foo"), "b"), id="one of the values is not a tuple"),
+    ),
+)
+def test_lookup_table_requires_key(values):
+    with pytest.raises(
+        ValueError,
+        match=r"^The `key` function must be specified when the `values` are not empty\.$",
+    ):
+        assert lookup_table(values)
```
webknjaz, line 396: When you expect it to raise an error, `assert` will never happen.
```suggestion
        lookup_table(values)
```

piptools/utils.py
atugushev: Could work, but doesn't look like an improvement. Maybe leave as is?

webknjaz: ok

tests/test_resolver.py
atugushev: Hah, copy-pasted from the above test. I think `requirement_summary != other_object` should be fine. Addressed in 080f85.

tests/test_utils.py
atugushev: Yeah, `operator` looks better. Thanks! Addressed in 080f85.

tests/test_utils.py
atugushev: Nice catch! Thanks. Fixed in 080f85.

webknjaz (approved):

jdufresne (approved):

atugushev:
@jdufresne @webknjaz thanks for reviewing! 🙏🏼

PR #1318, opened 2021-02-07
Enable mypy option no_implicit_reexport

jdufresne:
This option is part of the set of "strict" mypy options, bringing the
project closer to full strict checking.

https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport

> By default, imported values to a module are treated as exported and
> mypy allows other modules to import them. This flag changes the
> behavior to not re-export unless the item is imported using from-as or
> is included in `__all__`.

A side benefit of using `__all__` is that it allows removing the flake8
exceptions from the code base. Now `__init__.py` files are checked too.

Refs: #972


atugushev (approved):
Oh, that's useful! Thanks 🙏🏼

PR #1322, opened 2021-02-09
Add typings for the scripts.compile module

atugushev:
Refs: #972


piptools/scripts/compile.py
```diff
@@ -383,8 +390,8 @@ def cli(
                 from distutils.core import run_setup
 
                 dist = run_setup(src_file)
-                tmpfile.write("\n".join(dist.install_requires))
-                comes_from = f"{dist.get_name()} ({src_file})"
+                tmpfile.write("\n".join(dist.install_requires))  # type: ignore
```
webknjaz, line 393: Instead of `# type: ignore`, it's better to use `# type: ignore[error-code1,error-code2]` etc

atugushev: Yeah, that's a good point. It's WIP and I thought I could find some appropriate type to cast.

atugushev: Thank. Addressed in 7faaded935908b74e22483554e008c9486a03bb6.

piptools/utils.py
```diff
@@ -315,7 +315,7 @@ def get_compile_command(click_ctx: click.Context) -> str:
             else:
                 if isinstance(val, str) and is_url(val):
                     val = redact_auth_from_url(val)
-                if option.name == "pip_args":
+                if option.name == "pip_args_str":
```
webknjaz, line 318: this doesn't look related to typing

piptools/scripts/compile.py
```diff
@@ -216,44 +216,46 @@ def has_arg(self, arg_name):
...
+    output_file: Optional[LazyFile],
+    allow_unsafe: bool,
+    generate_hashes: bool,
+    reuse_hashes: bool,
+    src_files: Tuple[str],
+    max_rounds: int,
+    build_isolation: bool,
+    emit_find_links: bool,
+    cache_dir: str,
+    pip_args_str: Optional[str],
```
webknjaz, line 256: Why did you rename it?

webknjaz (dismissed):
Looks like you're changing the CLI API which is unrelated and is a breaking change.

piptools/scripts/compile.py
atugushev: Note, the CLI option is untouched , only changed variable name. See option decorator.

webknjaz: There's a lot of changes and renaming variables on top of everything else is hard to follow. It's usually better to make separate patches that are easier to digest for reviewers.

atugushev: The intention was to make diff less and make PR easier for reviewers.

atugushev: Would you like me to revert the variable name back? I'm okay with both variants.

piptools/utils.py
atugushev: Related to https://github.com/jazzband/pip-tools/pull/1322#discussion_r584272705

tests/conftest.py
```diff
@@ -90,6 +90,18 @@ def copy_ireq_dependencies(self, source, dest):
...
+    def options(self):
+        """Not used"""
+
+    @property
+    def session(self):
+        """Not used"""
+
+    @property
+    def finder(self):
+        """Not used"""
```
jdufresne, line 103: I think we might as well type these methods too.

While at first it might seem silly to bother typing tests, it:

1. Helps show we're using the pip-tools API correctly and in a way representative to how it should be used in production.
2. Helps show we're using third party APIs correctly.

I think it is fine to be less strict about types in tests, but when it is easy like this, might as well.

atugushev: Seems reasonable. Addressed in 5f43774eacb849e4483fd2d67f5389af1c1fa0ca.

atugushev:
Gentle ping @webknjaz, @jdufresne 👋🏼  All suggestions seem resolved. Is there anything I can do to complete the review?

webknjaz:
I don't have time to review this now but I skimmed through it quickly and I don't have additional comments. That said, I'll let someone else hit approve.

atugushev:
@webknjaz thanks for helping nonetheless 🙏🏼

jdufresne (commented):
:+1: Nice work! Lots of typing :slightly_smiling_face:

atugushev:
@jdufresne you were going to press approve button, but something went wrong? 😃

jdufresne (approved):

jdufresne:
Must have misclicked :laughing:

atugushev:
Thanks for reviewing this @webknjaz and @jdufresne 🙏🏼

PR #1323, opened 2021-02-13
Consider specifiers for equality operator to pin a dependency in make_install_requirement

### LINKED ISSUE #1114: Possible bug: problems resolving pytorch (cuda).
This may very well be PyTorch issue due to the way they distribute packages, but I will describe it here too.
The problem is that `pip-compile` doesn't stick to the way how package version is descried in `.in` file.
I am trying to install the default version of PyTorch (which comes compiled against CUDA 10.1).

#### Environment Versions

1. OS Type: Windows 10
1. Python version: 3.7.6
1. pip version:  20.0.2 
1. pip-tools version: 4.5.1

#### Steps to replicate

1. Make `requirements.in` file:
    ```
    torch===1.4.0
    torchvision===0.5.0
    ```
    Note usage of [arbitrary equality](https://www.python.org/dev/peps/pep-0440/#arbitrary-equality). I know that this is discouraged by pip. This is way I think the issue may be more related to PyTorch itself.
2. Compile it `pip-compile -f https://download.pytorch.org/whl/torch_stable.html -o requirements.txt requirements.in`

#### Expected result

Expected `requirements.txt` with `torch===1.4.0` (note triple `=`).

#### Actual result

Actual result it:
```
torch==1.4.0              # via -r req.in, torchvision
torchvision===0.5.0       # via -r req.in
```
Note double `=` on torch and triple `=` on torchvision.

If one installs the packages via `pip-sync requirements.txt`, then:
```
Looking in links: https://download.pytorch.org/whl/torch_stable.html
Collecting numpy==1.18.3
  Using cached numpy-1.18.3-cp37-cp37m-win_amd64.whl (12.8 MB)
Collecting pillow==7.1.1
  Using cached Pillow-7.1.1-cp37-cp37m-win_amd64.whl (2.0 MB)
Collecting torch==1.4.0
  Using cached https://download.pytorch.org/whl/cu92/torch-1.4.0%2Bcu92-cp37-cp37m-win_amd64.whl (641.8 MB)
Collecting torchvision===0.5.0
  Using cached torchvision-0.5.0-cp37-cp37m-win_amd64.whl (1.2 MB)
```

See that `cu92` version of `torch` is selected.

However, if `requirements.txt` would contain `torch===1.4.0`, then the results of `pip-sync` would be:
```
Looking in links: https://download.pytorch.org/whl/torch_stable.html
Collecting torch===1.4.0
  Using cached https://download.pytorch.org/whl/cu101/torch-1.4.0-cp37-cp37m-win_amd64.whl (796.8 MB)
Collecting torchvision===0.5.0
  Using cached torchvision-0.5.0-cp37-cp37m-win_amd64.whl (1.2 MB)
Successfully installed torch-1.4.0 torchvision-0.5.0
```

IceTDrinker:
Resolves #1114

Haven't written the test yet, as I would like to know if the spirit of the change suits the authors first
Tried to run `tox -e checkqa` as proposed in contribution guidelines but had a failure with github auth

- Changed function signature to take ireq directly rather than passing its
members one by one
- Changed type hint for version as I was getting a pip Version rather than
a string, maybe there are some parts of the code calling with string though
- Can confirm it generates a pinned torch with === in my use case as
expected

**Changelog-friendly one-liner**: Prefer === over == when generating requirements.txt if a dependency was pinned with ===

##### Contributor checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -1555,3 +1555,27 @@ def test_duplicate_reqs_combined(
     assert out.exit_code == 0, out
     assert str(test_package_2) in out.stderr
     assert "test-package-1==0.1" in out.stderr
+
+
+@pytest.mark.network
+def test_triple_equal_pinned_depency_is_used(runner):
+    """
+    Test that pip-compile properly emits the pinned requirement with ===
+    torchvision 0.8.2 requires torch==1.7.1 which can resolve to versions with
+    patches (e.g. torch 1.7.1+cu110), we want torch===1.7.1 without patches
+    """
+    with open("requirements.in", "w") as reqs_in:
+        reqs_in.write("torch===1.7.1\n")
+        reqs_in.write("torchvision===0.8.2\n")
+
+    out = runner.invoke(
+        cli,
+        [
+            "--find-links",
+            "https://download.pytorch.org/whl/torch_stable.html",
+        ],
+    )
+
+    assert out.exit_code == 0, out
+    assert "torch===1.7.1" in out.stderr
+    assert "torchvision===0.8.2" in out.stderr
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -1557,7 +1557,9 @@ def test_duplicate_reqs_combined(
     assert "test-package-1==0.1" in out.stderr
 
 
-def test_triple_equal_pinned_depency_is_used(runner, make_package, make_wheel, tmpdir):
+def test_triple_equal_pinned_dependency_is_used(
+    runner, make_package, make_wheel, tmpdir
+):
     """
     Test that pip-compile properly emits the pinned requirement with ===
     torchvision 0.8.2 requires torch==1.7.1 which can resolve to versions with
@@ -1576,10 +1578,10 @@ def test_triple_equal_pinned_depency_is_used(runner, make_package, make_wheel, t
 
     with open("requirements.in", "w") as reqs_in:
         reqs_in.write("test-package-1===1.7.1\n")
-        reqs_in.write("test-package-2===0.8.2\n")
+        reqs_in.write("test-package-2==0.8.2\n")
 
     out = runner.invoke(cli, ["--find-links", str(dists_dir)])
 
     assert out.exit_code == 0, out
     assert "test-package-1===1.7.1" in out.stderr
-    assert "test-package-2===0.8.2" in out.stderr
+    assert "test-package-2==0.8.2" in out.stderr
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -1557,8 +1557,44 @@ def test_duplicate_reqs_combined(
     assert "test-package-1==0.1" in out.stderr
 
 
+@pytest.mark.parametrize(
+    ("pkg2_install_requires", "req_in_content", "out_expected_content"),
+    (
+        pytest.param(
+            "",
+            ["test-package-1===0.1.0\n"],
+            ["test-package-1===0.1.0"],
+        ),
+        pytest.param(
+            "",
+            ["test-package-1==0.1.0\n"],
+            ["test-package-1==0.1.0"],
+        ),
+        pytest.param(
+            "test-package-1==0.1.0",
+            ["test-package-1===0.1.0\n", "test-package-2==0.1.0\n"],
+            ["test-package-1===0.1.0", "test-package-2==0.1.0"],
+        ),
+        pytest.param(
+            "test-package-1==0.1.0",
+            ["test-package-1===0.1.0\n", "test-package-2===0.1.0\n"],
+            ["test-package-1===0.1.0", "test-package-2===0.1.0"],
+        ),
+        pytest.param(
+            "test-package-1==0.1.0",
+            ["test-package-2===0.1.0\n"],
+            ["test-package-1==0.1.0", "test-package-2===0.1.0"],
+        ),
+    ),
+)
 def test_triple_equal_pinned_dependency_is_used(
-    runner, make_package, make_wheel, tmpdir
+    runner,
+    make_package,
+    make_wheel,
+    tmpdir,
+    pkg2_install_requires,
+    req_in_content,
+    out_expected_content,
 ):
     """
     Test that pip-compile properly emits the pinned requirement with ===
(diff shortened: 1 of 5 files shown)


IceTDrinker:
Solved the tox auth issue, will fix the issues on my end and understand what happens with the failing test cases

IceTDrinker:
Feedback welcome, squash is ready locally (removed the pytest network marker that's not required by my test anymore)

Cheers

tests/test_cli_compile.py
```diff
@@ -1534,3 +1534,32 @@ def test_duplicate_reqs_combined(
...
+    make_wheel(test_package_1, dists_dir)
+
+    test_package_2 = make_package(
+        "test_package_2", version="0.8.2", install_requires=["test-package-1==1.7.1"]
+    )
+    make_wheel(test_package_2, dists_dir)
+
+    with open("requirements.in", "w") as reqs_in:
+        reqs_in.write("test-package-1===1.7.1\n")
+        reqs_in.write("test-package-2===0.8.2\n")
```
atugushev, line 1559: Could you delete this line with `test-package-2` pin? Would be nice to test, that the resolver keeps dependency's pin operator?

IceTDrinker: You mean pinning test-package-2 with == 0.8.2 and check that it is kept ?

atugushev: Yeah, left only one pin in `requirements.in`:

```
# requirements.in
test-package-1===1.7.1
```

IceTDrinker: well that defeats the purpose of this particular test :) the issue was that the requirement from test_package_2 which pinned test_package_1 with == was taken instead of the more stringent requirement for test_package_1 with ===, plus test package 1 here does not depend on test_package_2

I can add additional test cases though if you want so that we can verify combinations of pinnings ?

Something like that for example :

```python
    # Case 1 ===
    with open("requirements.in", "w") as reqs_in:
        reqs_in.write("test-package-1===1.7.1\n")

    out = runner.invoke(cli, ["--find-links", str(dists_dir)])

    assert "test-package-1===1.7.1" in out.stderr

    # Case 2 ==
    with open("requirements.in", "w") as reqs_in:
        reqs_in.write("test-package-1==1.7.1\n")

    out = runner.invoke(cli, ["--find-links", str(dists_dir)])

    assert "test-package-1==1.7.1" in out.stderr

    # Case 3 pinned by user with === but pinned by package with ==, prefer ===
    with open("requirements.in", "w") as reqs_in:
        reqs_in.write("test-package-1===1.7.1\n")
        reqs_in.write("test-package-2==0.8.2\n")

    out = runner.invoke(cli, ["--find-links", str(dists_dir)])

    assert "test-package-1===1.7.1" in out.stderr
    assert "test-package-2==0.8.2" in out.stderr

    # Case 4 pinned by user with === but pinned by package with ==, prefer ===
    # Different pin for test_package_2
    with open("requirements.in", "w") as reqs_in:
        reqs_in.write("test-package-1===1.7.1\n")
        reqs_in.write("test-package-2===0.8.2\n")

    out = runner.invoke(cli, ["--find-links", str(dists_dir)])

    assert "test-package-1===1.7.1" in out.stderr
    assert "test-package-2===0.8.2" in out.stderr
```

atugushev: This set of test cases come to my mind:

First round where dependency is `test-package2==0.1`:
```
setup(name='test-package-1', version='0.1', install_requires=['test-package2==0.1'])
```

1. Test direct dependency keeps `===`
```
# requirements.in
test-package-1===0.1
```

2. Test sub-dependency keeps `===`
```
# requirements.in
test-package-1
test-package-2===0.1
```

Second round where dependency is `test-package2===0.1`:

```
setup(name='test-package-1', version='0.1', install_requires=['test-package2===0.1'])
```

1. Test sub-dependency keeps `===`
```
# requirements.in
test-package-1
```

2. Test sub-dependency keeps `==`
```
test-package1==0.1
```

Well, that pretty close to what you've provided, but slightly different on `setup.py` files 👍🏼 Could you also split each test case into a parametrized test?

IceTDrinker: Latest commit tests what you propose except the last case as pip cannot handle package where an install_requires string has a dependency pinned with === apparently, there is a parse error happening in pip._vendor.pyparsing. Is this material for a bug report for pip or should dependencies not be pinned with === in packages ?

IceTDrinker: Ok for some reason the bug is not happening anymore, I'll try to handle the sub-dependency case which does not preserve a package dependency specified with ===

atugushev: > pip does not support === for package dependencies at all...

Oh, interesting. Does it? I've tried it locally though:

```python
# setup.py
from setuptools import setup
setup(name='foo', version='0.1', install_requires=['six===1.10.0'])
```

```console
$ pip install -e .
Obtaining file:///private/var/folders/x4/88j2h7255_qd_td8wsn642br0000gp/T/tmp.95T5QMTv
Collecting six===1.10.0
  Using cached six-1.10.0-py2.py3-none-any.whl (10 kB)
Installing collected packages: six, foo
  Running setup.py develop for foo
Successfully installed foo six-1.10.0
WARNING: You are using pip version 20.1; however, version 21.0.1 is available.
```

IceTDrinker: Yep, "funnily" install -e must use a different route, it worked for a project I'm working on. Try installing a wheel generated from that setup.py, that's where I was out of luck

atugushev: > Yep, "funnily" install -e must use a different route, it worked for a project I'm working on. Try installing a wheel generated from that setup.py, that's where I was out of luck

I see. I've caught this error with wheels:

```
pip._vendor.pkg_resources.RequirementParseError: Parse error at "'(===1.10'": Expected stringEnd
```

> is it worth opening an issue on the pip repo at this point?

Yeah, It'll be good to know what would pip-devs think about it.

tests/test_cli_compile.py
```diff
@@ -1534,3 +1534,69 @@ def test_duplicate_reqs_combined(
...
+
+
+@pytest.mark.parametrize(
+    ("pkg2_install_requires", "req_in_content", "out_expected_content"),
+    (
+        pytest.param(
+            "",
+            ["test-package-1===0.1.0\n"],
+            ["test-package-1===0.1.0"],
+        ),
```
atugushev, line 1546:
```suggestion
        pytest.param(
            "",
            ["test-package-1===0.1.0\n"],
            ["test-package-1===0.1.0"],
            id="{what do we test here}"
        ),
```

It would be useful to have a descriptive ID on each test.

tests/test_cli_compile.py
IceTDrinker: Issue opened here: https://github.com/pypa/pip/issues/9650

I'll add ids to tests

tests/test_cli_compile.py
IceTDrinker: Done

atugushev (approved):
LGTM 👍🏼 Thanks, @IceTDrinker, for your patience and great work!

IceTDrinker:
> LGTM 👍🏼 Thanks, @IceTDrinker, for your patience and great work!

Do you need a squashed commit for all this ?

atugushev:
> Do you need a squashed commit for all this ?

I'll use the "Squash and merge" button, no worries 👍🏼

IceTDrinker:
Thanks for the reviews

Cheers

atugushev:
pip-tools `v6.0.0` is released 🚀

PR #756, opened 2019-03-06
Add full test coverage for the writer module

atugushev:
Increases test coverage up to 100% for `piptools/writer.py`.

PR #1325, opened 2021-02-15
Write out default index when it is provided as`--extra-index-url`

fahrradflucht:
If the default index is provided as `--extra-index-url` it can not safely be dropped from the written output since this changes behavior when the output is synced.

While jazzband/pip-tools#756 Explicitly introduced tests for the opposite, this seams to have been a mistake since otherwise the following lines comment about making an exception would be nonsensical.

**Changelog-friendly one-liner**:  Write out default index when it is provided as `--extra-index-url`.

##### Contributor checklist

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -103,7 +103,7 @@ class OutputWriter:
     def write_index_options(self) -> Iterator[str]:
         if self.emit_index_url:
             for index, index_url in enumerate(dedup(self.index_urls)):
-                if index_url.rstrip("/") == self.default_index_url:
+                if index == 0 and index_url.rstrip("/") == self.default_index_url:
                     continue
                 flag = "--index-url" if index == 0 else "--extra-index-url"
                 yield f"{flag} {index_url}"
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -276,7 +276,7 @@ def test_write_format_controls(writer):
                 "--extra-index-url https://index-url2.com",
             ],
         ),
-        # Ignore URLs equal to the default index-url
+        # Not ignore URLs equal to the default index-url
         # (note: the previous case is exception)
         (
             [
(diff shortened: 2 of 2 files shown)


fahrradflucht:
@atugushev Since you wrote that test, did I interpret that right?

piptools/writer.py
```diff
@@ -103,7 +103,7 @@ def write_header(self) -> Iterator[str]:
     def write_index_options(self) -> Iterator[str]:
         if self.emit_index_url:
             for index, index_url in enumerate(dedup(self.index_urls)):
-                if index_url.rstrip("/") == self.default_index_url:
+                if index == 0 and index_url.rstrip("/") == self.default_index_url:
```
atugushev, line 106: Oh, nice catch! Have you checked by the way how pip combines `--index-url` and `--extra-index-urls`? We should keep in sync with that logic.

fahrradflucht: Pip just [defaults `--index-url` to PyPI](https://github.com/pypa/pip/blob/751cbb72f5f7c3f0b98e2c1ff50fb61abde41acb/src/pip/_internal/cli/cmdoptions.py#L324) and [joins both index and extra index urls](https://github.com/pypa/pip/blob/751cbb72f5f7c3f0b98e2c1ff50fb61abde41acb/src/pip/_internal/req/req_file.py#L244). It then resolves dependencies from the index in the list order.

This is also how I hit this. If you set your private index as index-url and then add pypi as extra-index-url to make sure the private index is checked first, the requirements.txt pip-tools produces currently is not installable because it doesn't contain the extra-url option for pypi and so all public packages aren't found. If you add it manually to the file everything works as expected.

atugushev: Thanks for the analysis!

atugushev (approved):
:+1:

atugushev:
@fahrradflucht thanks for the contribution! 🎉

atugushev:
pip-tools `v6.0.0` is released 🚀

Issue #1087, opened 2020-03-12
Support for installation via pipx


georgek:
#### What's the problem this feature will solve?

`pipx` is a new tool which supports "global" installation of user tools. `pip-tools` seems like an ideal candidate for such a global tool. However, currently if you install `pip-tools` that way then anything that it installs (via `pip-sync`) goes into the `pipx` installation venv.

#### Describe the solution you'd like

I'd like to be able to install `pip-tools` once via `pipx` and then use this to sync my active virtualenv.

#### Alternative Solutions

The current alternative is to install `pip-tools` in each virtualenv but the downside is it then breaks `pip freeze` and requires updating for each virtualenv you might have on your system (`pip install -U pip-tools`).

georgek:
I will try to tackle this myself soon, but I was wondering if someone already familiar with the `pip` interface could comment on what's going to need to change and possible stumbling blocks with doing this. Is it necessarily a huge change?

atugushev:
Hello @georgek,

Thanks for bringing this up! TBH I have no experience with `pipx`, but I've heard pretty much positive feedback about this project, so I think it would be cool to support it.

Internally pip-sync installs packages via:

```python
subprocess.check_call([sys.executable, '-m', 'pip', 'install', '...'])
```

and it works pretty well for many cases. At first glance, we could configure the executable path that would resolve the issue. Any other ideas?

/cc @cs01 what do you think?

atugushev:
@georgek 

> but the downside is it then breaks `pip freeze`

Could you elaborate on why `pip-tools` would break `pip freeze`?

georgek:
@atugushev 
> Could you elaborate on why `pip-tools` would break `pip freeze`?

It doesn't break it as such, but the problem is if you add tools like pip-tools to your virtualenvs then it will be included in `pip freeze` output which probably isn't expected. I see `pip-tools` in a similar class to tools like `tox`. They are developer tools and not part of a project runtime environment.

It looks like finding python via `/usr/bin/env` as is normal with scripts would do the right thing, but I haven't tested it yet. That's also platform specific so needs some care. Probably need to read the virtualenv docs to see about cross platform ways to detect a currently active virtualenv.

One reason I thought about this is because poetry supports (and recommends) installation via `pipx`. I think it's the way to go for tools like this.

atugushev:
> One reason I thought about this is because poetry supports (and recommends) installation via `pipx`. I think it's the way to go for tools like this.

[...]

PR #1333, opened 2021-02-27
Support custom python-executable

MaratFM:
**Problem**: build pipelines and infrastructure at big companies often targets multiple independent python projects. Currently, in order to use pip-sync, pip-tools have to be installed in every environment. When pip-compile does not have such restriction.

**Solution**: add the ability to pass custom python executable.  When the option provided, list of installed packages will be collected from the target environment and all install/uninstall commands will be executed inside that environment. 

This PR implements a proof concept for collecting early feedback on the idea and implementation. If concept and task will be accepted, sufficient test coverage will be added and PR re-submitted.

Thank you!

**Added "--python-executable" option for pip-sync command. It allows targeting environments other than current**: 

##### Contributor checklist


MaratFM:
@atugushev could you please comment if this PR makes sense and maintainers agree to add such an option

atugushev:
Hello @MaratFM,

Tganks for pinging! I believe that this functionality would help for such issues as #1087. So i'm :+1: on this.

piptools/scripts/sync.py
```diff
@@ -55,6 +55,10 @@
     is_flag=True,
     help="Ignore package index (only looking at --find-links URLs instead)",
 )
+@click.option(
+    "--python-executable",
```
atugushev, line 59: We could use `type=click.Path(exists=True)` instead of this check below:

```python
    if python_executable is not None and not os.path.exists(python_executable):
         msg = "Python executable {} not found"
         log.error(msg.format(python_executable))
         sys.exit(2)
```

MaratFM: Great idea, thanks!

piptools/scripts/sync.py
```diff
@@ -55,6 +55,11 @@
...
```
atugushev, line 59: Subjective: maybe `--python` option (like in `virtualenv`)?

MaratFM: No problem changing. Interesting, which name is the most commonly used. I saw `--python-executable` in mypy
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-executable

MaratFM: Overall, it seems like with `--python` there are more ways for misinterpretation, so I am leaning towards more explicit `--python-executable` but the final call is yours

piptools/scripts/sync.py
```diff
@@ -125,7 +131,13 @@ def cli(
         log.error(str(e))
         sys.exit(2)
 
-    installed_dists = get_installed_distributions(skip=[], user_only=user_only)
+    paths = None
+    if python_executable:
+        paths = get_sys_path_for_python_executable(python_executable)
```
atugushev, line 136: I would rewrite this to:

```suggestion
    paths = None if python_executable is None else get_sys_path_for_python_executable(python_executable) 
```

or vice versa

piptools/scripts/sync.py
atugushev: > Overall, it seems like with `--python` there are more ways for misinterpretation, so I am leaning towards more explicit `--python-executable`

Oh, that's interesting. Let's wait for other opinions.

atugushev (commented):
Generally, it looks good. One thing to consider: could we improve `--python-executable` to pass binary name somehow? For example: `pip-sync --python-executable=python3`.

MaratFM:
> Generally, it looks good. One thing to consider: could we improve `--python-executable` to pass binary name somehow? For example: `pip-sync --python-executable=python3`.

Great idea, I can try using `shutil.which` for resolving
https://docs.python.org/3/library/shutil.html?highlight=.which#shutil.which

Looks like it works well for aliases and full paths, so we can always use it for resolving:
```py
In [1]: import shutil

In [2]: import sys

In [3]: shutil.which('python')
Out[3]: 'D:\\Envs\\test\\python.EXE'

In [4]: shutil.which(sys.executable)
Out[4]: 'd:\\envs\\test\\python.exe'
```

MaratFM:
@atugushev please take a look when you will have a chance. I added support for binary executable name (like `pip-sync --python-executable=python3`). In addition, while testing locally, I found that when we use a custom python executable it is no guaranteed that pip version is compatible. So I added an explicit check for it.

MaratFM:
@atugushev gentle ping

MaratFM:
@webknjaz @jdufresne could you please take a look if you have time 🙂

piptools/scripts/sync.py
```diff
@@ -9,15 +10,21 @@
 from pip._internal.commands.install import InstallCommand
 from pip._internal.index.package_finder import PackageFinder
 from pip._internal.utils.misc import get_installed_distributions
+from pip._vendor.packaging.version import Version
```
webknjaz, line 13: Have you checked if `packaging` is available directly, maybe as a transitive dep?

piptools/scripts/sync.py
```diff
@@ -9,15 +10,21 @@
...
 from ..repositories import PyPIRepository
-from ..utils import flat_map
+from ..utils import (
+    flat_map,
+    get_pip_version_for_python_executable,
+    get_sys_path_for_python_executable,
+)
 
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
+MIN_PIP_VERSION = Version("20.3")  # Should be in sync with version in setup.cfg
```
webknjaz, line 27: Instead of hardcoding such things, it's possible to parse it from the metadata via `importlib_metadata` or `pkg_util` or smth like that.

webknjaz: You may find some inspiration here https://github.com/cherrypy/cherrypy/blob/e00d162/docs/conf.py#L32-L57

piptools/scripts/sync.py
MaratFM: It is available, but because other parts of code are using imports from `pip._vendor`, mixing with direct imports will break `isinstance` like checks. Not sure about the original motivation to use  `pip._vendor` imports, most probably it is to make it more stable and not rely on libraries except pip

piptools/scripts/sync.py
MaratFM: thanks for the suggestion, updated to use `get_distribution` and compare the version with the specifier

MaratFM:
@webknjaz @atugushev could you please take a look

MaratFM:
@jdufresne could you please help with the review?

jdufresne:
Sorry, I will not have the time to review this. Please carry on with the other reviewers.

piptools/scripts/sync.py
```diff
@@ -108,6 +119,26 @@ def cli(
...
+        # Ensure that target python executable has the right version of pip installed
+        pip_version = get_pip_version_for_python_executable(python_executable)
+        required_pip_specification = get_required_pip_specification()
+        if not required_pip_specification.contains(pip_version):
+            msg = (
+                "Target python executable '{}' has pip version {} installed. "
+                "Version {} is expected."
+            )
+            log.error(
+                msg.format(python_executable, pip_version, required_pip_specification)
```
webknjaz, line 138: FWIW it's recommended (although controversial sometimes) to let the logging module take care of rendering things. And this is why it's best to pass a separate string with percent-placeholders as the first arg and pass the vars separately, instead of pre-rendering that string.
This also allows different tooling to aggregate log entries coming from the same place but with different args.

piptools/scripts/sync.py
```diff
@@ -108,6 +119,26 @@ def cli(
...
+            )
+            sys.exit(2)
```
webknjaz, line 140: This block looks like it could belong in a dedicated function.

tests/test_cli_sync.py
```diff
@@ -242,3 +244,116 @@ def test_sync_dry_run_returns_non_zero_exit_code(runner):
...
+def test_python_executable_option(
+    run,
+    get_installed_distributions,
+    get_sys_path_for_python_executable,
+    get_pip_version_for_python_executable,
+    runner,
+    fake_dist,
+):
+    """
+    Make sure sync command can run with `--python-executable` option
```
webknjaz, line 262: PEP 257 requires sentences (with a period at the end)
```suggestion
    Make sure sync command can run with `--python-executable` option.
```

webknjaz:
Same here. I hope @atugushev will be able to get this to the finish line.

piptools/scripts/sync.py
MaratFM: Thanks, I updated logging calls. 

BTW, I was just following the pattern used in the file, it could be a separate discussion if all log statements should be aligned and switch to logging formatting.

piptools/scripts/sync.py
MaratFM: thanks, extracted _validate_python_executable method

tests/test_cli_sync.py
```diff
@@ -242,3 +244,116 @@ def test_sync_dry_run_returns_non_zero_exit_code(runner):
     out = runner.invoke(cli, ["--dry-run"])
 
     assert out.exit_code == 1
+
+
+@mock.patch("piptools.scripts.sync.get_pip_version_for_python_executable")
+@mock.patch("piptools.scripts.sync.get_sys_path_for_python_executable")
+@mock.patch("piptools.scripts.sync.get_installed_distributions")
+@mock.patch("piptools.sync.run")
```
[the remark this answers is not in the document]
MaratFM: sounds good, will keep aligned with the rest of the code

tests/test_cli_sync.py
MaratFM: updated here and in other places I found in my code

atugushev (commented):
I've caught a trace when a given python executable is not found:

```
$ pip-sync -a --python-executable=python3.9
  File "/Users/albert/Projects/pip-tools/.venv/bin/pip-sync", line 11, in <module>
    load_entry_point('pip-tools', 'console_scripts', 'pip-sync')()
  File "/Users/albert/Projects/pip-tools/.venv/lib/python3.8/site-packages/click/core.py", line 764, in __call__
  File "/Users/albert/Projects/pip-tools/.venv/lib/python3.8/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/albert/Projects/pip-tools/.venv/lib/python3.8/site-packages/click/core.py", line 956, in invoke
  File "/Users/albert/Projects/pip-tools/.venv/lib/python3.8/site-packages/click/core.py", line 555, in invoke
  File "/Users/albert/Projects/pip-tools/piptools/scripts/sync.py", line 126, in cli
    _validate_python_executable(python_executable)
  File "/Users/albert/Projects/pip-tools/piptools/scripts/sync.py", line 200, in _validate_python_executable
    log.error(msg, python_executable, pip_version, required_pip_specification)
  File "/Users/albert/Projects/pip-tools/piptools/logging.py", line 40, in error
    self.log(message, *args, **kwargs)
  File "/Users/albert/Projects/pip-tools/piptools/logging.py", line 24, in log
    click.secho(prefix + message, *args, **kwargs)
TypeError: secho() got multiple values for argument 'err'
```

atugushev:
> I reverted logging message passing, was confused by the recommendation to delegate message formatting to the logging system ([#1333 (comment)](https://github.com/jazzband/pip-tools/pull/1333#discussion_r647266617)). Seems like pip-tools uses a custom logging wrapper that has incompatible method contracts.

Thanks @MaratFM! Let's address that in a following up PR.

tests/test_cli_sync.py
```diff
@@ -242,3 +244,116 @@ def test_sync_dry_run_returns_non_zero_exit_code(runner):
     out = runner.invoke(cli, ["--dry-run"])
 
     assert out.exit_code == 1
+
+
+@mock.patch("piptools.scripts.sync.get_pip_version_for_python_executable")
+@mock.patch("piptools.scripts.sync.get_sys_path_for_python_executable")
+@mock.patch("piptools.scripts.sync.get_installed_distributions")
```
atugushev, line 251: There are too much of monkeypatches. Ideally, there should be only `@mock.patch("piptools.sync.run")`.

piptools/scripts/sync.py
atugushev: I'll resolve this. Let it be `--python-executable`.

piptools/utils.py
```diff
@@ -358,3 +363,39 @@ def get_compile_command(click_ctx: click.Context) -> str:
...
 
     return " ".join(["pip-compile", *sorted(left_args), *sorted(right_args)])
+
+
+def get_required_pip_specification() -> SpecifierSet:
+    """
+    Returns pip version specifier requested by current pip-tools installation.
+    """
+    project_dist = get_distribution("pip-tools")
+    requirement = next((r for r in project_dist.requires() if r.name == "pip"), None)
```
atugushev, line 373: It seems the `None` branch left uncovered. Options:
1. add a test
2. add pragma marker to skip this branch

tests/test_utils.py
```diff
@@ -423,3 +428,15 @@ def test_drop_extras(from_line, given, expected):
...
+
+def test_get_pip_version_for_python_executable():
+    result = get_pip_version_for_python_executable(sys.executable)
+    assert Version(pip.__version__) == result
+
+
+def test_get_sys_path_for_python_executable():
+    result = get_sys_path_for_python_executable(sys.executable)
+    # not testing for equality, because pytest adds extra paths into current sys.path
+    for path in result:
```
atugushev, line 441: `assert result` should be checked to make sure that below assert would be reached.

MaratFM: good point, added an assert

piptools/utils.py
MaratFM: added pragma marker

tests/test_cli_sync.py
MaratFM: refactored

piptools/utils.py
```diff
@@ -358,3 +363,41 @@ def get_compile_command(click_ctx: click.Context) -> str:
...
+
+
+def get_required_pip_specification() -> SpecifierSet:
+    """
+    Returns pip version specifier requested by current pip-tools installation.
+    """
+    project_dist = get_distribution("pip-tools")
+    requirement = next(
+        (r for r in project_dist.requires() if r.name == "pip"), None
+    )  # pragma: no cover
```
atugushev, line 375:
```suggestion
    )  # pragma: no branch
```

There is a [no branch](https://coverage.readthedocs.io/en/coverage-4.3.3/branch.html#structurally-partial-branches) pragma.

MaratFM:
@atugushev seems like with "# pragma: no branch" it stills shows partial coverage, so I reverted back to "# pragma: no cover"

piptools/utils.py
atugushev, line 375:
```suggestion
    requirement = next(  # pragma: no branch
        (r for r in project_dist.requires() if r.name == "pip"), None
    )
```

This should work.

MaratFM:
@atugushev sorry for changing tests back and forth, I could not run tests for all these platforms locally and have to wait when someone will trigger the GitHub actions pipeline

atugushev (approved):
Thanks @MaratFM for your patience. Awesome work! 👍🏻

PR #1342, opened 2021-03-08
Simplify freshen_build_caches using contextlib.ExitStack

jdufresne:
Beyond simplifying the implementation, in the event that the 2nd
tempfile.TemporaryDirectory call fails, the first will still be cleaned
up.

##### Contributor checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -4,7 +4,7 @@ import itertools
 import logging
 import os
 import tempfile
-from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


atugushev:
I've grepped `source_dir` and `build_dir` and didn't find any usages except on tests. Would be nice to check whether it leads to hidden bugs or we can completely remove those properties.

jdufresne:
Even simpler! As best I can tell, looks like these have been unused since 50ecc4e8fd76d3540d2f6db68683ee6292af562a.

Removed them in #1344

PR #1345, opened 2021-03-09
Prefer dict over OrderedDict

jdufresne:
Starting with Python 3.6, the builtin dict retains its order. Using
OrderedDict is no longer necessary.

##### Contributor checklist

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -207,7 +207,7 @@ def dedup(iterable: Iterable[_T]) -> Iterable[_T]:
     """Deduplicate an iterable object like iter(set(iterable)) but
     order-preserved.
     """
-    return iter(collections.OrderedDict.fromkeys(iterable))
+    return iter(dict.fromkeys(iterable))
 
 
 def get_hashes_from_ireq(ireq: InstallRequirement) -> Set[str]:


atugushev:
Dictionary order is guaranteed since python 3.7:

https://docs.python.org/3/library/stdtypes.html
> Changed in version 3.7: Dictionary order is guaranteed to be insertion order.

or 3.6?

webknjaz:
@atugushev de-jure, the guarantee is for 3.7+ (language-wise) but de-facto CPython 3.6 dicts do preserve the order (which may be different for other runtimes like PyPy, MicroPython etc).

jdufresne:
Exactly right. Here is some more details from the 3.6 release notes:

https://docs.python.org/dev/whatsnew/3.6.html#new-dict-implementation

> The order-preserving aspect of this new implementation is considered an implementation detail and should not be relied upon (this may change in the future, ...

PyPy shares this behavior and is in fact the inspiration for CPython's change. So we can rely on the same behavior there.

atugushev (approved):
👍🏼

webknjaz:
> PyPy shares this behavior and is in fact the inspiration for CPython's change. So we can rely on the same behavior there.

I think it's okay for as long as we don't support things like MicroPython (this one couldn't implement the same data structure because there was not enough memory on microcontrollers IIRC, dunno if it's been addressed by now).

PR #1349, opened 2021-03-13
Release 6.0.0

atugushev:


CHANGELOG.md
```diff
@@ -1,3 +1,38 @@
+# 6.0.0 (2020-03-12)
+
+Backwards Incompatible Changes:
+
+- Remove support for EOL Python 3.5 and 2.7
+  ([#1243](https://github.com/jazzband/pip-tools/pull/1243)). Thanks @jdufresne
+- Remove long deprecated cache dir `~/.pip-tools`
+  ([#1225](https://github.com/jazzband/pip-tools/pull/1225)). Thanks @jdufresne
```
atugushev, line 8: @jdufresne what do you think, should we add this to the changelog?

jdufresne: Probably doesn't matter too much either way. Worse case scenario, the first pip-tools command takes a bit longer and users don't think about it after that. 

No strong preference here. Up to you.

jdufresne (approved):

CHANGELOG.md
atugushev: Ok, since it's not a user-facing change I'll remove it.

atugushev:
Thanks @jdufresne 🙏🏼

PR #1356, opened 2021-03-18
PEP-517 support

### LINKED ISSUE #1047: Support setup.py-less setups
Due to PEP 518, projects can now have just a setup.cfg and pyproject.toml file. Pip-compile only supports setup.py when using setuptools, though.

Pip-compile can use setup.cfg files to get at dependencies, like `pip-compile -U setup.cfg`.

#### Alternative Solutions
I suppose I can copy-paste the dependencies into their own requirements.in? That would be duplication though :(

orsinium:
**Changelog-friendly one-liner**: Support for pyproject.toml as input dependency file (PEP-517).

If pyproject.toml is explicitly specified as input file, pip-compile will use it. Example: 

```bash
pip-compile --output-file requirements.txt pyproject.toml
```

Included a simple test for [flit](https://github.com/takluyver/flit)

Closes #1047

##### Contributor checklist


piptools/scripts/compile.py
```diff
@@ -24,6 +24,7 @@
 
 DEFAULT_REQUIREMENTS_FILE = "requirements.in"
 DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
+METADATA_FILENAMES = frozenset({"setup.py", "pyproject.toml"})
```
atugushev, line 27: Should we also include `setup.cfg` in this list?

orsinium: Great catch! I thought, `setup.py` is required even if you use `setup.cfg` but docs say it is indeed optional:

> If compatibility with legacy builds (i.e. those not using the PEP 517 build API) is desired, a setup.py file containing a setup() function call is still required even if your configuration resides in setup.cfg.

I've updated the line and added a test for setup.cfg as well 👍🏿 

BTW, tox fails if there is no `setup.py`, I'm not the only one who didn't know that 🙃

tests/test_cli_compile.py
```diff
@@ -1656,3 +1656,29 @@ def test_triple_equal_pinned_dependency_is_used(
...
+    assert "small-fake-a==0.1" in out.stderr
+    assert "small-fake-b==0.2" in out.stderr
+    assert "small-fake-c" not in out.stderr
+    assert "small-fake-d" not in out.stderr
+    assert "small-fake-e" not in out.stderr
+    assert "small-fake-f" not in out.stderr
+
+
+@pytest.mark.network
+def test_setup_cfg(runner):
```
atugushev, line 1675: Would be nice to have a separate package with `setup.cfg` and without `pyproject.toml` file to make sure that `pep517` parses only `setup.cfg`.

tests/test_data/packages/flit_small_with_deps/module/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""Sample module
+"""
+
+__version__ = "0.1.1"  # pragma: no cover
```
atugushev, line 4: We try to avoid creating fixture packages in `test/test_data` and encourage to create packages dynamically in the test body or pytest fixtures instead (see [make_package](https://github.com/jazzband/pip-tools/blob/77aa19b63c9e2aad410e6f5654c19f78b8101ddf/tests/conftest.py#L242)). 

Ideally, it would be nice to hive a fixture like `make_flit_package(name, version, install_requires, extras, ...)` to avoid boilerplate.

tests/test_cli_compile.py
```diff
@@ -1656,3 +1656,29 @@ def test_triple_equal_pinned_dependency_is_used(
     assert out.exit_code == 0, out
     for line in out_expected_content:
         assert line in out.stderr
+
+
+@pytest.mark.network
+def test_flit(runner):
```
atugushev, line 1662: The body of both tests looks similar. Seems like we could combine `test_setup_cfg` and `test_flit` via `pytest.mark.parametrize`.

atugushev (commented):
Thanks for addressing the suggestion and adding a test 🙏🏻 A few comments below.

tests/test_data/packages/flit_small_with_deps/module/__init__.py
atugushev: But if you have no time or energy to spend on that, we could leave it as is and refactor later. I don't mind 👍🏻

tests/test_cli_compile.py
orsinium: It is separate, just the directory name is similar. Flit is in `flit_small_with_deps` and setup.cfg is in `setup_cfg_small_with_deps`.

tests/test_data/packages/flit_small_with_deps/module/__init__.py
orsinium: It's part of metadata, flit requires it. Or do you mean, generate any metadata (including `pyproject.toml`) as well? It's hard to tell which way is preferred now, some packages use `make_package`, some use `PACKAGES_PATH`. I'm not sure what would be better here. Having pyproject.toml as a separate TOML file gives a small benefit of syntax highlighting, formatting all that nice stuff 🤔

orsinium: Fixed.

tests/test_cli_compile.py
orsinium: Great catch! I parametrized tests, dropped auxiliary files, added a test case for poetry.

tests/test_cli_compile.py
```diff
@@ -1656,3 +1656,118 @@ def test_triple_equal_pinned_dependency_is_used(
...
+                    small-fake-b==0.2
+
+                [options.extras_require]
+                dev =
+                    small-fake-c==0.3
+                    small-fake-d==0.4
+                test =
+                    small-fake-e==0.5
+                    small-fake-f==0.6
+            """,
```
atugushev, line 1687:
```suggestion
            dedent(
                """\
                [metadata]
                name = sample_lib
                author = Vincent Driessen
                author_email = me@nvie.com

                [options]
                packages = find:
                install_requires =
                    small-fake-a==0.1
                    small-fake-b==0.2

                [options.extras_require]
                dev =
                    small-fake-c==0.3
                    small-fake-d==0.4
                test =
                    small-fake-e==0.5
                    small-fake-f==0.6
                """
            ),
```

dedent could help to reduce extra indentations

atugushev (commented):
Nice! One last thing. Could `dedent` be used alongside triple-quote strings? That helps to reduce extra indentations.

tests/test_cli_compile.py
orsinium: Sorry, what do you mean? I can drop one level of indentation from the multiline string without moving `dedent`, Python is liberal about it. It's up to personal preferences, I've chosen to add the indentation to match how we indent the code inside braces. See https://github.com/wemake-services/wemake-python-styleguide/issues/1025 for more context. However, if you prefer another style, I can change it, not a big deal.

atugushev: > I can drop one level of indentation from the multiline string without moving dedent, Python is liberal about it.

Yeah, but the result files will be indented too wide. In `pip-tools` we use `dedent` in that cases. See references:
- [1](https://github.com/jazzband/pip-tools/blob/37946f212248e932d5f7022f7fcc8810ba5d8cd2/tests/test_cli_compile.py#L109-L113)
- [2](https://github.com/jazzband/pip-tools/blob/37946f212248e932d5f7022f7fcc8810ba5d8cd2/tests/test_cli_compile.py#L810-L823)
- [3](https://github.com/jazzband/pip-tools/blob/37946f212248e932d5f7022f7fcc8810ba5d8cd2/tests/test_cli_compile.py#L931-L935)
- [4](https://github.com/jazzband/pip-tools/blob/37946f212248e932d5f7022f7fcc8810ba5d8cd2/tests/test_cli_compile.py#L1062)

atugushev: Oh, I didn't notice that `dedent` is already there:

```
stream.write(dedent(content))
```

atugushev: Then, it's okay. My bad.

tests/test_cli_compile.py
```diff
@@ -1656,3 +1656,118 @@ def test_triple_equal_pinned_dependency_is_used(
     assert out.exit_code == 0, out
     for line in out_expected_content:
         assert line in out.stderr
+
+
+@pytest.mark.network
+@pytest.mark.parametrize(
+    ("fname", "content"),
+    (
+        # setuptools
```
atugushev, line 1665: Could we use `pytest.param(..., id='..')` instead of comments? That improves `pytest -v` output.

Befure:

```
tests/test_cli_compile.py::test_input_formats[setup.cfg-\n                [metadata]\n                name = sample_lib\n                author = Vincent Driessen\n                author_email = me@nvie.com\n\n                [options]\n                packages = find:\n                install_requires =\n                    small-fake-a==0.1\n                    small-fake-b==0.2\n\n                [options.extras_require]\n                dev =\n                    small-fake-c==0.3\n                    small-fake-d==0.4\n                test =\n                    small-fake-e==0.5\n                    small-fake-f==0.6\n            ] PASSED [ 25%]
tests/test_cli_compile.py::test_input_formats[setup.py-\n                from setuptools import setup\n\n                setup(\n                    name="sample_lib",\n                    version=0.1,\n                    install_requires=["small-fake-a==0.1", "small-fake-b==0.2"],\n                    extras_require={\n                        "dev": ["small-fake-c==0.3", "small-fake-d==0.4"],\n                        "test": ["small-fake-e==0.5", "small-fake-f==0.6"],\n                    },\n                )\n            ] PASSED [ 50%]
tests/test_cli_compile.py::test_input_formats[pyproject.toml-\n                [build-system]\n                requires = ["flit_core >=2,<4"]\n                build-backend = "flit_core.buildapi"\n\n                [tool.flit.metadata]\n                module = "sample_lib"\n                author = "Vincent Driessen"\n                author-email = "me@nvie.com"\n\n                requires = ["small-fake-a==0.1", "small-fake-b==0.2"]\n\n                [tool.flit.metadata.requires-extra]\n                dev  = ["small-fake-c==0.3", "small-fake-d==0.4"]\n                test = ["small-fake-e==0.5", "small-fake-f==0.6"]\n            ] PASSED [ 75%]
tests/test_cli_compile.py::test_input_formats[pyproject.toml-\n                [build-system]\n                requires = ["poetry_core>=1.0.0"]\n                build-backend = "poetry.core.masonry.api"\n\n                [tool.poetry]\n                name = "sample_lib"\n                version = "0.1.0"\n                description = ""\n                authors = ["Vincent Driessen <me@nvie.com>"]\n\n                [tool.poetry.dependencies]\n                python = "*"\n                small-fake-a = "0.1"\n                small-fake-b = "0.2"\n\n                small-fake-c = "0.3"\n                small-fake-d = "0.4"\n                small-fake-e = "0.5"\n                small-fake-f = "0.6"\n\n                [tool.poetry.extras]\n                dev  = ["small-fake-c", "small-fake-d"]\n                test = ["small-fake-e", "small-fake-f"]\n            ] PASSED [100%]
```

After:

```
tests/test_cli_compile.py::test_input_formats[setup.cfg] PASSED                                                                                                                                      [ 25%]
tests/test_cli_compile.py::test_input_formats[setup.py] PASSED
tests/test_cli_compile.py::test_input_formats[flit] PASSED
tests/test_cli_compile.py::test_input_formats[poetry] PASSED
```

orsinium: Yep, great idea. Fixed 👍🏿

webknjaz: FYI another nice way to do this would be to add `ids=('setup.cfg', 'setup.py')`. Sometimes it's easier than embedding the ids into each param. But of course, it depends on the situation.

tests/test_cli_compile.py
```diff
@@ -1656,3 +1656,119 @@ def test_triple_equal_pinned_dependency_is_used(
...
+
+                [tool.poetry.extras]
+                dev  = ["small-fake-c", "small-fake-d"]
+                test = ["small-fake-e", "small-fake-f"]
+            """,
+            id="poetry",
+        ),
+    ),
+)
+def test_input_formats(runner, tmpdir, fname, content):
```
webknjaz, line 1757: `tmpdir` is sorta deprecated (well, at least there's a mention of it being replaced). It's always better to use `tmp_path` which does the same except it injects `pathlib.Path` objects instead of `py.path` which are nicer to work with.

orsinium: Yep, I know. But all other tests in this file use `tmpdir`. I think it's better to keep it consistent and migrate all at once.

tests/test_cli_compile.py
orsinium: Good to know, thank you 👍🏿 I think in this situation it looks nice to have it per case.

orsinium:
Interesting. Why did it fail? Seems to be flaky 🤔

orsinium:
Can't reproduce locally:

```
while tox -e py37-piplatest-coverage -- tests/test_cli_compile.py::test_input_formats -v --numprocesses=4  ; do :; done
```

Probably, some kind of side-effect. Pip cache poisoning?

atugushev (approved):
LGTM 👍🏻

webknjaz (approved):

atugushev:
@orsinium thanks for the contribution! Awesome job! 🎉

PR #1238, opened 2020-11-22
Remove unnecessary envvar configuration in click options

jdufresne:
These are environment variables consumed by pip. As environment
variables are passed through to subprocesses they're already handled. No
special processing in pip-tools is required so simplify the CLI
definition

##### Contributor checklist

- [x] Provided the tests for the changes.
- [x] Gave a clear one-line description in the PR (that the maintainers can add to CHANGELOG.md on release).
- [x] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #1357, opened 2021-03-18
Parallelize pytest runs

webknjaz:
$sbj. This change makes a single pytest run almost 4x faster on my machine:
|        command         |       time       |
|------------------------|------------------|
| $ tox -e py -- -n auto | 17.90s           |
| $ tox -e py -- -n 0    | 68.54s (0:01:08) |

I'm pretty sure it'll have a positive impact on the CI completion time too. UPD: in GHA it's about 2x faster in this mode.

**Changelog-friendly one-liner**: (misc/non user facing?) Enabled pytest to run tests in parallel.

##### Contributor checklist

- [x] (not exactly but...)Provided the tests for the changes.
- [x] (likely irrelevant)Gave a clear one-line description in the PR (that the maintainers can add to CHANGELOG.md on release).
- [ ] (irrelevant)Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,6 +40,7 @@ exclude = tests
 testing =
     pytest
     pytest-rerunfailures
+    pytest-xdist
 coverage = pytest-cov
 
 [options.entry_points]
@@ -48,6 +49,9 @@ console_scripts =
     pip-sync = piptools.scripts.sync:cli
 
 [tool:pytest]
+addopts =
+    # `pytest-xdist`:
+    --numprocesses=auto
 norecursedirs = .* build dist venv test_data piptools/_compat/*
 testpaths = tests piptools
 filterwarnings =


atugushev (approved):
Thanks for bringing this! I remember we made a while ago some preparation to make this possible. See https://github.com/jazzband/pip-tools/pull/1022#pullrequestreview-336352189.

atugushev:
Thanks @webknjaz!

AndydeCleyre:
Hey, I wasn't invited to the review party, but don't think this should have gone through without setting unique cache dirs for parallel `pip-compile` invocations, via `--cache-dir`.

AndydeCleyre:
All the tests are unreliable in this state. I also notice this, which may be a problem:

```python
class CliRunner(object):
    """The CLI runner provides functionality to invoke a Click command line
    script for unittesting purposes in a isolated environment.  This only
    works in single-threaded systems without any concurrency as it changes the
    global interpreter state.
    """
```

AndydeCleyre:
And I see here we have a no-longer-helpful [fixture for this](https://github.com/richafrank/pip-tools/blob/master/tests/test_cli_compile.py#L17):

```python
@pytest.fixture(autouse=True)
def _temp_dep_cache(tmpdir, monkeypatch):
    monkeypatch.setenv("PIP_TOOLS_CACHE_DIR", str(tmpdir / "cache"))
```

But that env var was removed in d786232cf0ba4d4e9cf11e9d50b6665d78e330c3, merged in #1238, which broke silently broke this fixture.

webknjaz:
I think it's good that this got merged because it reveals problems with tests that may depend on the shared state/order. This gives us a chance to identify the flaky tests and actually fix them over time. Plus, going forward it'll make the test authors more cautious and will encourage everyone to consider the test stability...

webknjaz:
> ```python
>     works in single-threaded systems without any concurrency as it changes the
>     global interpreter state.
> ```

I'm pretty sure pytest-xdist distributes tests across subprocesses, not threads. So this shouldn't be a problem.

AndydeCleyre:
I'm thinking the real bug here is in #1238, where the intent seemed to be only to remove env vars handled by pip itself, and `PIP_TOOLS_CACHE_DIR`'s removal was an oopsie. I propose we just re-enable the env var, and then I think there's nothing wrong with this parallel business here.

PR #1363, opened 2021-03-23
Extra requirement support (extras_require)

### LINKED ISSUE #625: Pip-compile ignores extras_require when parsing setup.py

orsinium:
**Changelog-friendly one-liner**: Add `--extra` to specify `extras_require` dependencies.

Supports setup.py and anything PEP-517 compatible. Tested for setup.py, setup.cfg, poetry, and flit.

Close #625.

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -24,7 +24,7 @@ from ..utils import (
     dedup,
     is_pinned_requirement,
     key_from_ireq,
-    req_is_in_extras,
+    req_check_markers,
 )
 from ..writer import OutputWriter
 
@@ -398,8 +398,7 @@ def cli(
         ireq for key, ireq in upgrade_install_reqs.items() if key in allowed_upgrades
     )
 
-    # Filter out pip environment markers which do not match (PEP496)
-    constraints = [req for req in constraints if req_is_in_extras(req, extras=extras)]
+    constraints = [req for req in constraints if req_check_markers(req, extras=extras)]
 
     log.debug("Using indexes:")
     with log.indentation():
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -71,7 +71,7 @@ def _get_default_option(option_name: str) -> Any:
     "--extra",
     "extras",
     multiple=True,
-    help="names of extras_require to install",
+    help="Names of extras_require to install",
 )
 @click.option(
     "-f",
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -393,7 +393,7 @@ def cli(
                 )
             )
 
-    if not setup_file_found:
+    if extras and not setup_file_found:
         msg = "--extra has effect only with setup.py and PEP-517 input formats"
         raise click.BadParameter(msg)
 
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -2,12 +2,13 @@ import os
 import shlex
 import sys
 import tempfile
-from typing import Any, BinaryIO, Optional, Tuple, cast
+from typing import Any, BinaryIO, List, Optional, Tuple, cast
 
 import click
 from click.utils import LazyFile, safecall
 from pep517 import meta
 from pip._internal.commands import create_command
+from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
 from pip._internal.utils.misc import redact_auth_from_url
 
@@ -19,13 +20,7 @@ from ..logging import log
 from ..repositories import LocalRequirementsRepository, PyPIRepository
 from ..repositories.base import BaseRepository
 from ..resolver import Resolver
-from ..utils import (
-    UNSAFE_PACKAGES,
-    dedup,
-    is_pinned_requirement,
-    key_from_ireq,
-    req_check_markers,
-)
+from ..utils import UNSAFE_PACKAGES, dedup, is_pinned_requirement, key_from_ireq
 from ..writer import OutputWriter
 
 DEFAULT_REQUIREMENTS_FILE = "requirements.in"
@@ -349,7 +344,7 @@ def cli(
     # Parsing/collecting initial requirements
     ###
 
-    constraints = []
+    constraints: List[InstallRequirement] = []
     setup_file_found = False
     for src_file in src_files:
         is_setup_file = os.path.basename(src_file) in METADATA_FILENAMES
@@ -406,7 +401,7 @@ def cli(
         ireq for key, ireq in upgrade_install_reqs.items() if key in allowed_upgrades
     )
 
-    constraints = [req for req in constraints if req_check_markers(req, extras=extras)]
+    constraints = [req for req in constraints if req.match_markers(extras)]
 
     log.debug("Using indexes:")
     with log.indentation():
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -19,7 +19,13 @@ from ..logging import log
 from ..repositories import LocalRequirementsRepository, PyPIRepository
 from ..repositories.base import BaseRepository
 from ..resolver import Resolver
-from ..utils import UNSAFE_PACKAGES, dedup, is_pinned_requirement, key_from_ireq
+from ..utils import (
+    UNSAFE_PACKAGES,
+    dedup,
+    is_pinned_requirement,
+    key_from_ireq,
+    req_is_in_extras,
+)
 from ..writer import OutputWriter
 
 DEFAULT_REQUIREMENTS_FILE = "requirements.in"
@@ -60,6 +66,12 @@ def _get_default_option(option_name: str) -> Any:
     "--rebuild",
     is_flag=True,
     help="Clear any caches upfront, rebuild from scratch",
+)
+@click.option(
+    "-e",
+    "--extras",
+    multiple=True,
+    help="names of extras_require to install",
 )
 @click.option(
     "-f",
@@ -204,22 +216,23 @@ def cli(
     dry_run: bool,
     pre: bool,
     rebuild: bool,
-    find_links: Tuple[str],
+    extras: Tuple[str, ...],
+    find_links: Tuple[str, ...],
     index_url: str,
-    extra_index_url: Tuple[str],
+    extra_index_url: Tuple[str, ...],
     cert: Optional[str],
     client_cert: Optional[str],
-    trusted_host: Tuple[str],
+    trusted_host: Tuple[str, ...],
     header: bool,
     emit_trusted_host: bool,
     annotate: bool,
     upgrade: bool,
-    upgrade_packages: Tuple[str],
+    upgrade_packages: Tuple[str, ...],
     output_file: Optional[LazyFile],
     allow_unsafe: bool,
     generate_hashes: bool,
     reuse_hashes: bool,
-    src_files: Tuple[str],
+    src_files: Tuple[str, ...],
     max_rounds: int,
     build_isolation: bool,
     emit_find_links: bool,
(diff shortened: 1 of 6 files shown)


orsinium:
I hope, we'll be able to take it into #1362, it's closely related to #1356, they play nice together.

atugushev:
> Ready for review. I hope, we'll be able to take it into #1362, it's closely related to #1356, they play nice together.

Ok, I've added it to `6.1.0`.

piptools/scripts/compile.py
```diff
@@ -61,6 +67,12 @@ def _get_default_option(option_name: str) -> Any:
     is_flag=True,
     help="Clear any caches upfront, rebuild from scratch",
 )
+@click.option(
+    "-e",
```
atugushev, line 71:
```suggestion
    "-E",
```

I'would like to suggest `-E` instead of `-e`, which clashes with `pip install -e`.

orsinium: Good catch. Let's just drop the short key and leave only `--extras`

orsinium: Fixed :+1:

orsinium:
Is there anything that remains to do?

atugushev:
Could you add a test for `pip-compile --extra=whatever requirements.in`?

atugushev:
@orsinium IMO there should be an error.

orsinium:
Done 👍🏿

tests/conftest.py
```diff
@@ -322,3 +322,39 @@ def _make_sdist(package_dir, dist_dir, *args):
...
+        path = os.path.join(tmpdir, fname)
+        with open(path, "w") as stream:
+            stream.write(dedent(content))
+        return path
+
+    return _make_module
+
+
+@pytest.fixture
+def fake_dists(tmpdir, make_package, make_wheel):
```
webknjaz, line 348: Please add a docstring to improve the UX of the `pytest --fixtures` invocation.

tests/test_cli_compile.py
```diff
@@ -1658,123 +1658,161 @@ def test_triple_equal_pinned_dependency_is_used(
...
+            description = ""
+            authors = ["Vincent Driessen <me@nvie.com>"]
+
+            [tool.poetry.dependencies]
+            python = "*"
+            small-fake-a = "0.1"
+            small-fake-b = "0.2"
+
+            small-fake-c = "0.3"
+            small-fake-d = "0.4"
```
webknjaz, line 1742: Is it intentional that these are duplicated in here and in the extras?

tests/test_cli_compile.py
```diff
@@ -1658,123 +1658,161 @@ def test_triple_equal_pinned_dependency_is_used(
...
+    out = runner.invoke(
+        cli, ["-n", "--extra", "dev", "--find-links", fake_dists, meta_path]
+    )
+    assert out.exit_code == 0, out.stderr
+    assert "small-fake-a==0.1" in out.stderr
+    assert "small-fake-b==0.2" in out.stderr
+    assert "small-fake-c==0.3" in out.stderr
+    assert "small-fake-d==0.4" in out.stderr
+    assert "small-fake-e" not in out.stderr
+    assert "small-fake-f" not in out.stderr
```
webknjaz, line 1782: I don't fully understand. Is "extra" supposed to only select the dists that are listed in extras but not the main runtime deps?

piptools/scripts/compile.py
```diff
@@ -61,6 +67,12 @@ def _get_default_option(option_name: str) -> Any:
     is_flag=True,
     help="Clear any caches upfront, rebuild from scratch",
 )
+@click.option(
+    "--extra",
+    "extras",
+    multiple=True,
+    help="names of extras_require to install",
```
webknjaz, line 74: 1) `s/n/N`
2) If this is supposed to install only extras but not the default runtime deps, can this be renamed into `--extra-only`?
3) If it's supposed to merge those extras with the main runtime deps list, can we add an additional flag `--extras-only`/`--exclude-default-deps` or something?

tests/test_cli_compile.py
```diff
@@ -1658,123 +1658,161 @@ def test_triple_equal_pinned_dependency_is_used(
...
-        stream.write("'example module'\n__version__ = '1.2.3'")
-    path = os.path.join(tmpdir, fname)
-    with open(path, "w") as stream:
-        stream.write(dedent(content))
 
-    out = runner.invoke(cli, ["-n", "--find-links", dists_path, path])
+
+@pytest.mark.network
+@pytest.mark.parametrize(("fname", "content"), METADATA_TEST_CASES)
+def test_input_formats(fake_dists, runner, make_module, fname, content):
```
webknjaz, line 1757: It'd be great if you could add docstrings to the tests so that it'd be clear what expectations they assume.

tests/test_cli_compile.py
orsinium: small-fake-e and small-fake-f are from another extra, the `test` one, we pass only `dev`.

tests/test_cli_compile.py
orsinium: Yes, this is how poetry works. Here is the specification, in `extras` only names.

piptools/scripts/compile.py
webknjaz: It's actually a long-standing feature request in pip and I think there's been no opposition, just nobody had time to come up with an implementation. The main motivation was that people often supply test/dev/docs extras for different processes they have and some of them don't really require the dist itself to be installed. Of course, there's another camp saying that this sort of deps should be put elsewhere because they aren't necessary for the runtime and there should be env pins anyway.
But if pip-tools were to implement this flag, it'd facilitate a hybrid of these two approaches with the direct open-ended deps in the dist metadata configs and the pins existing in separate constraints files.
But yes, this could probably be brought up in a separate discussion.

webknjaz (approved):
I think it's merge-ready, any other comments can be addressed separately.

atugushev:
I've run `pip-compile setup.py --extra=testing` on [setup.py](https://github.com/jazzband/pip-tools/blob/master/setup.cfg) and noticed `extra` markers.


<details>

 <summary>Output</summary>

```ini
#
# To update, run:
#
#    pip-compile --extra=testing setup.py
#
apipkg==1.5
    # via execnet
attrs==20.3.0
    # via pytest
click==7.1.2
    # via pip-tools (setup.py)
execnet==1.8.0
    # via pytest-xdist
iniconfig==1.1.1
    # via pytest
packaging==20.9
    # via pytest
pep517==0.10.0
    # via pip-tools (setup.py)
pluggy==0.13.1
    # via pytest
py==1.10.0
    # via
    #   pytest
    #   pytest-forked
pyparsing==2.4.7
    # via packaging
pytest-forked==1.3.0
    # via pytest-xdist
pytest-rerunfailures==9.1.1 ; extra == "testing"
    # via pip-tools (setup.py)
pytest-xdist==2.2.1 ; extra == "testing"
    # via pip-tools (setup.py)
pytest==6.2.2 ; extra == "testing"
    # via
    #   pip-tools (setup.py)
    #   pytest-forked
    #   pytest-rerunfailures
    #   pytest-xdist
toml==0.10.2
    # via
    #   pep517
    #   pytest

# pip
# setuptools
```

</details>


Note `pytest==6.2.2 ; extra == "testing"`.

I've tried to install the result `requirements.txt` and got the following "warnings":

```console
$ pip install -r requirements.txt
Ignoring pytest-rerunfailures: markers 'extra == "testing"' don't match your environment
Ignoring pytest-xdist: markers 'extra == "testing"' don't match your environment
Ignoring pytest: markers 'extra == "testing"' don't match your environment
...
```

This is FYI, I haven't researched it yet deeply.

webknjaz:
> Note `pytest==6.2.2 ; extra == "testing"`.

Ouch... You can't use extras as markers. Only a restricted subset of vars/functions is available there.

The table of available markers is here: https://www.python.org/dev/peps/pep-0508/#environment-markers.

For `extra` it specifically says:
> An error except when defined by the context interpreting the specification.

And so I haven't heard of any tool implementing this. At least, I'm fairly sure that pip doesn't have any interface to specify extras in requirements files.

atugushev:
@webknjaz thanks for the info! Perhaps, "extra" should be filtered out in [format_requirement](https://github.com/jazzband/pip-tools/blob/019516ae8a16de739370b8e5329580fcb8fa283c/piptools/utils.py#L117) somehow. Ideas?

piptools/scripts/compile.py
```diff
@@ -387,16 +406,7 @@ def cli(
...
-    constraints = [
-        req
-        for req in constraints
-        if req.markers is None
-        # We explicitly set extra=None to filter out optional requirements
-        # since evaluating an extra marker with no environment raises UndefinedEnvironmentName
-        # (see https://packaging.pypa.io/en/latest/markers.html#usage)
-        or req.markers.evaluate({"extra": None})
-    ]
+    constraints = [req for req in constraints if req_check_markers(req, extras=extras)]
```
atugushev, line 409: [InstallRequirement.match_markers(extras)](https://github.com/pypa/pip/blob/7ec0fa5142466f8402bb866a1dcfb8ca6ffdc66b/src/pip/_internal/req/req_install.py#L268) could be used instead of `req_check_markers()`. Looks like it works the same way under the hoods.
```suggestion
    constraints = [req for req in constraints if req.match_markers(extras)]
```

piptools/scripts/compile.py
orsinium: I'm from that camp too. I have a separate environment for every tool. However, most of the tools still need dependencies: `pytest`  for sure, `mypy` for type-checking of third-party libs usage, `pylint` for inference-based tasks. In my projects, only `flake8` works nicely without knowing the project dependencies. And I keep a separate `requirements.txt` for it. Yep, makes the root of the project a bit messier but works. This is why I made [dephell](https://github.com/dephell/dephell/) a long time ago, to handle all this zoo :)

orsinium:
Possible solutions I see for removing `extra` marker:

1. Hacking with regexp
2. Hacking through inspection of `packaging.Marker._marker`. It has an interesting structure of tuples in lists inside. If we assume that `extra` always comes at the top level, we can easily find and remove it. That means getting into a "private" method but `packaging` should be very stable and unlikely to change.
3. https://github.com/dephell/dephell_markers has `.remove` method. I made it a long time ago for dephell, its well well tested and works. Still, a third-party dependency for a small task.

atugushev:
Approach №2 is fine by me 👍🏻

orsinium:
I've implemented and integrated `drop_extras` function which removes `extra` from markers. Unit tests are included and all seems to be working fine. However, I can't cover one line somehow 🤔

orsinium:
Huh. I've put a breakpoint there and it IS covered, for sure. I have a deja vu. Coverage.py seems to be quite flaky about branch coverage for if conditions. I definitely faced something like this in [deal](https://github.com/life4/deal) but I don't remember the exact solution. Probably, I've just rewritten the condition back then.

orsinium:
I decided to hack it around here too 🙃

ssbarnea (approved):

ssbarnea:
@atugushev @orsinium @webknjaz Is there anything blocking progress on this PR? I do find it very useful and a blocker for the new release.

orsinium:
Nothing from my side. There is a new logic (https://github.com/jazzband/pip-tools/pull/1363/commits/3b63aa94370185094b6a827defcd2bc9b2551b22) that needs a review from maintainers.

atugushev:
Hello @ssbarnea! Thanks for pinging. I'm going to look into this shortly.

ssbarnea:
I should be the owe thanking you. That is indeed a very important feature, one that prevented further adoption from more people.

atugushev (approved):
@orsinium thanks for fixing the bug with `extras` 🙏🏻 The PR's all good!

atugushev:
`pip-tools v6.1.0` is released 🚀

Issue #1358, opened 2021-03-18
pip-compile ignores existing output file if --output-file is specified


mikepqr:
#### What's the problem this feature will solve?

This may be a supported workflow, in which case I'd be happy to submit a PR to add it to the README if someone can help me understand what the intended steps are!

Suppose I have:
```
# requirements.in
foo==1.0
bar==2.0
```
`foo` depends on `baz>=3.0` and `bar` depends on `qux>=4.0` so that when I run `pip-compile requirements.in` I get:
```
# requirements.txt
foo==1.0  # via requirements.in
bar==2.0  # via requirements.in
baz==3.0  # via foo
qux==4.0  # via bar
```
But then there's a CVE against `bar==3.0` and I want to upgrade it to 3.1. So I run `pip-compile --upgrade-package baz==3.1`, which yields:
```
# requirements.txt
foo==1.0  # via requirements.in
bar==2.0  # via requirements.in
baz==3.1  # via foo
qux==4.2  # via baz
```
baz has been upgraded as requested, but foo and bar remain pinned as in the requirements file. That's all great. But `qux` has been upgraded too, because bar follows best practices by not overpinning its qux dependency, and qux happens to have released a new version. This is not what I want in this case. I want to upgrade _only_ baz.

#### Describe the solution you'd like

I guess I'm asking for an `--upgrade-only-package` option that does what I want, but this seems like such a common use case that I assume I'm missing something!

#### Alternative Solutions

Right now I simply edit the `baz` line in requirements.txt by hand in cases where it doesn't contain hashes.

When requirements.txt does contain hashes I do an insane set of manual steps that involves:

1. Copying requirements.txt over requirements.in
2. Deleting all the hashes from the file.
3. Pinning `baz==3.1` in the new requirements.in
4. Rerunning pip-compile to generate the hashed requirements.txt file for this fully pinned set of requirements.
5. Restore the old requirements.in file

AndydeCleyre:
Thanks for the report!

I think you mixed up `bar` and `baz` a number of times here, but I'm following.

But the behavior you want is what I usually see. Can you provide a testable example?

`reqs.in`:
```ini
httpx
```
```console
$ pip-compile --no-header reqs.in
```
`reqs.txt`:
```ini
certifi==2020.12.5
    # via httpx
h11==0.12.0
    # via httpcore
httpcore==0.12.3
    # via httpx
httpx==0.17.1
    # via -r reqs.in
idna==3.1
    # via rfc3986
[...]

Issue #744, opened 2019-02-26
Readme: "requirements without setup.py" section doesn't warn that it considers existing requirements.txt file


m-aciek:
I just've lost couple of hours trying to figure out why `pip-compile` resolves my `requirements.in` to some old versions of packages. I'd like to propose adding a note to "requirements without setup.py" readme section, that would mention how to migrate from "plain `requirements.txt` workflow" to "`requirements.in` with pip-tools" without confusion.

##### Steps to replicate

1. Have requirements.txt with some package pinned.
2. Follow Readme in Example usage of `pip-compile` -> Without `setup.py`.

##### Expected result

Readme in this section warns the reader that his existing `requirements.txt` might interfere.

##### Actual result

Readme doesn't warn and reader is confused with the result of pip-compile.

atugushev:
Hello @m-aciek,

Thank you for the report! It's a good idea. I think the "[without setup.py](https://github.com/jazzband/pip-tools#without-setuppy)" section could be rephrased something like:
> If you don't use setup.py (it's easy to write one), you can create a requirements.in **(make sure you don't have a requirement.txt file, otherwise... etc)** file to declare the Flask dependency:

@vphilippon WDYT? English native speaker needed here :)

hramezani:
Can I work on this? seems easy :smile:

PR #1369, opened 2021-03-28
Better explain role of existing requirements.txt

mikepqr:
This commit adds an explanation of the role of existing requirements.txt
files to the README.

I found this very unclear and opened #1358 because of my confusion. The
existing short note that requirements.txt "might interfere" didn't
really help me. I'm not sure how common my confusion is, but I hope my
changes makes things clearer to new users (assuming what I've written is
in fact correct!)

(In addition to my additions, I moved the "Updating requirements"
section up in the doc, since it seems like a core workflow.)

##### Contributor checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,7 +30,6 @@ even when you've pinned them.  You do pin them, right? (In building your Python 
    :target: https://pypi.org/project/pip-tools/
 .. _You do pin them, right?: http://nvie.com/posts/pin-your-packages/
 
-
 Installation
 ============
 
@@ -62,9 +61,11 @@ project so conditional dependencies that require a specific Python version,
 or other environment markers, resolve relative to your project's
 environment.
 
-**Note**: ensure you don't have ``requirements.txt`` if you compile
-``setup.py`` or ``requirements.in`` from scratch, otherwise, it might
-interfere.
+**Note**: If ``pip-compile`` finds an existing ``requirements.txt`` file that
+fulfils the dependencies then no changes will be made, even if updates are
+available. To compile from scratch, first delete the existing
+``requirements.txt`` file, or see `Updating requirements`_ for alternative
+approaches.
 
 Requirements from ``setup.py``
 ------------------------------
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -62,10 +61,6 @@ project so conditional dependencies that require a specific Python version,
 or other environment markers, resolve relative to your project's
 environment.
 
-**Note**: ensure you don't have ``requirements.txt`` if you compile
-``setup.py`` or ``requirements.in`` from scratch, otherwise, it might
-interfere.
-
```
atugushev, line 68: Could you elaborate on these deletions above? There is a reason for this note, see https://github.com/jazzband/pip-tools/issues/744.

mikepqr: I found "from scratch" and "might interfere" very vague ("from scratch" in what sense? interfere how? why only "might"?).

I think the new text covers how an existing requirements.txt affects pip-compile better in the sense that explains what will happen more precisely and completely. But I'm happy to restore it if you prefer?

atugushev: > I found "from scratch" and "might interfere" very vague ("from scratch" in what sense? interfere how? why only "might"?).

I'd prefer to keep the note. I like the explanation in the "Updating requirements" section.  Would you like to improve the note somehow?

mikepqr: How about "Note: To compile setup.py or requirements.in from scratch, you must first delete any existing requirements.txt file"?

atugushev: That's a good direction. Perhaps we should clarify shortly why `requirements.txt` must be deleted.

mikepqr: Something like this (basically copy/pasted from the new text below)?

"Note: If ``pip-compile`` finds an existing ``requirements.txt`` file, and that file fulfils the dependencies, then no changes will be made, even if updates are available. To compile setup.py or requirements.in from scratch, first delete any existing requirements.txt file, or see "Updating requirements" for alternative approaches."

mikepqr: Pushed another revision. Please let me know if this is more along the right lines.

atugushev: Thanks! Looks good 👍🏻

atugushev (approved):
LGTM 👍

atugushev:
Thanks @mikepqr 🙏🏻

mikepqr:
Thank you for your help!

PR #1380, opened 2021-04-08
Add support for target Python pip-args to PyPIRepository

### LINKED ISSUE #1382: Pip-Compile should support the target python arguments when passed through pip-args

Pip provides support for targeting an environment through the use of the 4 parameters: python version, python implementation, abi tag, and wheel platform. Currently, pip-compile will ignore these parameters if they're passed through the pip-args command. These args should be enabled to take advantage of pip's logic for targeting a specific environment.

Passing these arguments will let pip compile target the described environment. Ex:

pip compile --pip-args "--platform manylinux2014_x86_64"

This would allow users to recompile requirements for a specific linux platform, regardless of their current environment. 

Pip download allows users to download a set of packages based on a requirements file, which can be used to control the packages used in a build scenario. Adding this support enables developers to quickly update the requirements file without running the desired environment; for example, developers on a Windows environment could use the args to update a requirements file without the overhead of starting a virtual machine or otherwise emulating the desired linux environment.

skylera93:
Resolves #1382.

This change adds support for using the target python arguments in pip (platform, version, implementation, and abi), by utilizing the TargetPython class from Pip. This should enable targeting environments different than the current python environment by pip tools commands. 

ex: pip compile --pip-args "--platform manylinux2014_x86_64"

Thank you!

Adds support for target python pip-args (version, platform, implementation, and abi) to PyPIRepository

##### Contributor checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -10,6 +10,7 @@ from typing import Any, ContextManager, Dict, Iterator, List, Optional, Set, cas
 
 from click import progressbar
 from pip._internal.cache import WheelCache
+from pip._internal.cli.cmdoptions import make_target_python
 from pip._internal.cli.progress_bars import BAR_TYPES
 from pip._internal.commands import create_command
 from pip._internal.commands.install import InstallCommand
@@ -17,6 +18,7 @@ from pip._internal.index.package_finder import PackageFinder
 from pip._internal.models.candidate import InstallationCandidate
 from pip._internal.models.index import PackageIndex
 from pip._internal.models.link import Link
+from pip._internal.models.target_python import TargetPython
 from pip._internal.models.wheel import Wheel
 from pip._internal.network.session import PipSession
 from pip._internal.req import InstallRequirement, RequirementSet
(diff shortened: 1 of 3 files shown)


skylera93:
This would address issue #1382

piptools/repositories/pypi.py
```diff
@@ -71,8 +73,11 @@ def __init__(self, pip_args: List[str], cache_dir: str):
 
         self._options: optparse.Values = options
         self._session = self.command._build_session(options)
+        self._target_python = make_target_python(options)
         self._finder = self.command._build_package_finder(
-            options=options, session=self.session
+            options=options,
+            session=self.session,
+            target_python=self._target_python,
```
atugushev, line 80:
```suggestion
            target_python=make_target_python(options),
```

Could you elaborate on `self._target_python` usage? I don't see any usage except on tests.

skylera93: The usage was within the package finder, but after further testing the package finder is not able to use the target python in the way I thought.

skylera93:
The target python and the associated args do not function expected, and do not support the pip compile logic I thought they would

PR #1389, opened 2021-04-22
Refactor producing cache keys in GHA jobs

webknjaz:
This change makes the GHA cache dirs and keys calculated dynamically per recommendation @ https://github.com/jazzband/pip-tools/pull/1378#issuecomment-824278917.

**Changelog-friendly one-liner**: $sbj (not user-facing)

##### Contributor checklist

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -33,30 +33,55 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Pip cache
+
+      - name: Calculate cache metadata
+        id: cache-meta
+        shell: python
+        run: |
+          import hashlib
+          import subprocess
+          import sys
+
+          cache_dir_cmd = sys.executable, '-m', 'pip', 'cache', 'dir'
+          cache_dir = subprocess.check_output(
+              cache_dir_cmd,
+              universal_newlines=True,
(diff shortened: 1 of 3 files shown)


.github/workflows/cron.yml
```diff
@@ -33,30 +33,55 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Pip cache
+
+      - name: Calculate cache metadata
+        id: cache-meta
+        shell: python
```
atugushev, line 39: I wonder can we put this script into a python file and run it instead?

webknjaz: I don't think it's a good idea. Moving the context out of the workflow makes it harder to follow and forces people to look up external sources. This would be different if it was needed outside of GHA, of course.

webknjaz: It would be possible to make it a composite action but it looks like an overkill.

ssbarnea:
@webknjaz Please make it draft until is ready for review and merge.

hugovk:
pip can now be cached via `actions/setup-python` to save a _lot_ of config:

https://github.com/actions/setup-python#caching-packages-dependencies

Please see https://github.com/jazzband/pip-tools/pull/1571 for an alternative.

Issue #1391, opened 2021-04-23
pip-compile fails AttributeError: 'NoneType' object has no attribute 'specifier'


fj-dt:
Hello everyone,

My project is depending on a lot of git repositories. These packages are also dependent on other git repositories.
When I generate one `requirements.txt` then I do `pip-sync` it works fine. But when I generate one `requirements.txt` and one `requirements-dev.txt`, when I do `pip-sync requirements.txt requirements-dev.txt` then I get this error:

```
Traceback (most recent call last):
  File "/Users/fj/PycharmProjects/my-project/.venv/bin/pip-sync", line 8, in <module>
    sys.exit(cli())
  File "/Users/fj/PycharmProjects/my-project/.venv/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/fj/PycharmProjects/my-project/.venv/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/fj/PycharmProjects/my-project/.venv/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/fj/PycharmProjects/my-project/.venv/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/fj/PycharmProjects/my-project/.venv/lib/python3.7/site-packages/piptools/scripts/sync.py", line 123, in cli
    merged_requirements = sync.merge(requirements, ignore_conflicts=force)
  File "/Users/fj/PycharmProjects/my-project/.venv/lib/python3.7/site-packages/piptools/sync.py", line 116, in merge
    if ireq.specifier != existing_ireq.specifier:
  File "/Users/fj/PycharmProjects/my-project/.venv/lib/python3.7/site-packages/pip/_internal/req/req_install.py", line 263, in specifier
    return self.req.specifier
AttributeError: 'NoneType' object has no attribute 'specifier'
```

`requirements.txt` contains:
```
git+ssh://git@bitbucket.org/private-bucket/enums.git@4.0.0
```
and `requirements-dev.txt` contains the same package which is a sub-dependency of another git repository.

So I tried to debug it, by adding a print into `piptools/sync.py` line 113:
```python
                    # NOTE: We check equality here since we can assume that the
                    # requirements are all pinned
[...]

PR #1329, opened 2021-02-22
Fixup relative and absolute path handling

AndydeCleyre:
<details><summary>Initial Summary (Outdated)</summary>

```

Rewrite input file paths as relative to the output file, or as absolutes if using stdout:

- Change click arg output_file from click.File to click.Path,
  so we can get its absolute path
- Change to output file's parent folder before opening it,
  helping upstream pip code to use correct relative paths
- Return to the starting dir after compilation,
  for test or other calls from code which don't expect dir changes
- Rewrite src_files as absolute paths as soon as we have them,
  to resolve relative paths properly (CLI args are relative to CWD)
- When deriving the output path from a single input path, stay safer and more predictable,
  particularly if the basename has no dot, and the path does (see example in #1067)
- Rewrite src_files as relative to the output path once that's known,
  unless output is stdout
- Don't overwrite an input file ending in '.txt' when deriving the output file path
- Add tests:
    - test_annotation_relative_paths
    - test_input_file_with_txt_extension
    - test_input_file_without_extension_and_dotted_path

Fixes #1107

Contributes to #1067

Minor tag-alongs:
  - fix some comment typos
  - git-ignore sublime text project and workspace files

QUESTIONS:
1. If output is stdout, should paths (in annotations) be absolute,
  or relative to current working folder? 
  With this PR, they're absolute, and I think that is appropriate.
2. With this PR, `output_file` changes its type, early on. 
  Does that bother anyone? It starts as a `click.Path`, then 
  after paths are resolved by our logic, it's replaced by a `click.File`. 
  It's a small window of `click.Path`-ness. An example consequence
  is seen in this PR's change to `test_writer.py`.
3. What is the best result of using `name.txt` as an input file, without specifying the output file?
  With this PR, it outputs to `name.txt.txt`, which is the best I can think of.
4. What additional tests would be good to have, if any? 
  More annotation variants, like with `-c`? More complicated relative paths?

**Changelog-friendly one-liners**: 
- Rewrite input file paths as relative to the output file, or as absolutes if using stdout
- Don't overwrite an input file ending in '.txt' when deriving the output file path
- Don't confuse dots in folder names with file extensions when deriving the output file path

```
</details>

<details><summary>Summary, Take 2 (also outdated)</summary>

Fixup relative and absolute path handling:

These changes have been made with the general guideline of
storing paths as absolute as soon as we can, and rendering
them as relative or absolute as needed.

| Path | Initial Interpretation | Output Format (file) | Output Format (stdout) |
[...]

PR #1392, opened 2021-04-25
PEP 440 Direct Reference support

FlorentJeannot:
This PR adds [PEP 440 Direct References](https://www.python.org/dev/peps/pep-0440/#direct-references) support to fix https://github.com/jazzband/pip-tools/issues/1391. 

`pip` can support direct references, so the egg can be omitted inside the URL (see [the code from pip](https://github.com/pypa/pip/blob/7ec0fa5142466f8402bb866a1dcfb8ca6ffdc66b/src/pip/_vendor/packaging/requirements.py#L149)).

In my projects which have git repositories specified as dependencies, without a direct reference or an egg, pip was not able to determine the package name, which was causing an issue during `pip-sync` (inside the merge, comparing two identical requirements from two different files).

Also, inside `setup.py`, `install_requires` must be written with a direct reference (`pip` fails to install the dependencies if it is not using a direct reference). `pip-tools` was ignoring the direct reference, so it was not generating the `requirements.txt` correctly. See my example:

```
# setup.py
install_requires=["package @ git+ssh://url@version"]

# requirements.txt (from pip-tools)
git+ssh://url@version
```

**Changelog-friendly one-liner**: PEP 440 Direct References support 

##### Contributor checklist


FlorentJeannot:
FYI I installed `pip-tools` in my project using my fork and I generated and compared the `requirements.txt` and `dev-requirements.txt` files with the new ones:

- there was no regression
- the direct references were generated when it was only needed
- one dependency of a dependency generated a URL with an egg, without direct reference (which is fine: it keeps the original behavior of the tool)

And `pip-sync` worked (it was not without this fix; see https://github.com/jazzband/pip-tools/issues/1391)

FlorentJeannot:
To help you understand this fix/enhancement, from the master branch, try the following tests:

### Test 1

In `pip-tools/tests/test_data/packages/fake_with_deps/setup.py`, add in `install_requires`:
```python
"requests @ git+git://github.com/psf/requests@v2.25.1"
```

In `test_realistic_complex_sub_dependencies`, add the following at the end:
```python
print(out.stderr)
assert "requests @ git+git://github.com/psf/requests@v2.25.1" in out.stderr
```

This test will fail because it is missing the direct reference, inside the output there is:
```
git+git://github.com/psf/requests@v2.25.1
```

Also, note that there is no `egg` specified.

### Test 2

In `pip-tools/tests/test_resolver.py`, add the following:

```python
def test_combine_install_requirements_direct_ref(repository, from_line):
    first_dep = from_line("git+ssh://****@bitbucket.org/private-bucket/enums.git@4.0.0", comes_from="-r requirements.in")
    second_dep = from_line("git+ssh://****@bitbucket.org/private-bucket/enums.git@4.0.0", comes_from="-r dev-requirements.in")

    combined = combine_install_requirements(repository, [first_dep, second_dep])
    assert combined.comes_from == first_dep.comes_from  # shortest string
    assert set(combined._source_ireqs) == {first_dep, second_dep}
    assert str(combined.req.specifier) == ""
```

This is failing because both names are missing a specifier. You can make the test pass by specifying a direct reference or an egg:

- `enums @ git+ssh://****@bitbucket.org/private-bucket/enums.git@4.0.0`
- `git+ssh://****@bitbucket.org/private-bucket/enums.git@4.0.0#egg=enums`

For example this works:

```python
def test_combine_install_requirements_direct_ref(repository, from_line):
    first_dep = from_line("enums @ git+ssh://****@bitbucket.org/private-bucket/enums.git@4.0.0", comes_from="-r requirements.in")
    second_dep = from_line("git+ssh://****@bitbucket.org/private-bucket/enums.git@4.0.0#egg=enums", comes_from="-r dev-requirements.in")

    combined = combine_install_requirements(repository, [first_dep, second_dep])
    assert combined.comes_from == first_dep.comes_from  # shortest string
    assert set(combined._source_ireqs) == {first_dep, second_dep}
    assert str(combined.req.specifier) == ""
```

### Conclusion

- Test 1 does not generate the requirement with the direct reference, even though it was specified in `install_requires `
- Test 2 fails if a specifier (a direct reference or an egg) is missing

Because `pip-tools` does not handle direct references, when:

- `requirements.in` contains a direct reference
- `dev-requirements.in` contains the same direct reference

Then the resolver fails when entering `combine_install_requirements`.

I hope this helps to make it clearer and I am looking forward to reading your feedback!

tests/test_data/packages/fake_with_deps/setup.py
```diff
@@ -18,5 +18,6 @@
         "SQLAlchemy!=0.9.5,<2.0.0,>=0.7.8,>=1.0.0",
         "python-memcached>=1.57,<2.0",
         "xmltodict<=0.11,>=0.4.6",
+        "requests @ git+git://github.com/psf/requests@v2.25.1",
```
FlorentJeannot, line 21: This new dependency does not bring anything more by itself, so I could either:

1. remove it
2. assert inside `test_realistic_complex_sub_dependencies` that `out.stderr` contains a direct reference

AndydeCleyre:
Hey, you reached out last week over at #1329, and today I was able to start looking at this. It's *possible* that, aside from changes to tests, all that's needed on top of my WIP PR is:

```diff
diff --git a/piptools/utils.py b/piptools/utils.py
index 4f15729..2f81af1 100644
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -1,6 +1,7 @@
 import collections
 import itertools
 import os
+import re
 import shlex
 from contextlib import contextmanager, suppress
 from typing import (
@@ -137,6 +138,8 @@ def format_requirement(
                         ireq.local_file_path, from_dir
                     ).replace(os.path.sep, "/")
             ireq_url = f"{path_url}{fragment_string(ireq)}"
+        if not re.match(r"^#(|.*&)egg=.*", fragment_string(ireq)) and ireq.name:
+            ireq_url = f"{ireq.name} @ {ireq_url}"
         line = f"-e {ireq_url}" if ireq.editable else ireq_url
 
     if marker:
```
I'll leave this diff right here, and if #1329 is ever done this will help me merge.

FlorentJeannot:
@AndydeCleyre Thanks for your suggestion. Yeah it looks like that would work, the tests I've added will help you find out. :)

luzfcb:
@FlorentJeannot I guess that's necessary to include these two test cases to ensure this continues to work in the future. Anyway looks good to me.

The output should be the same as the input:

```
django-debug-toolbar@https://github.com/jazzband/django-debug-toolbar/archive/6ec6bfe0ba5912d7fd0a2b98176a215847157696.zip

django-taggit@https://github.com/jazzband/django-taggit/archive/74460fdc4a8b8bf2d36fb8a59691403cde7b4528.zip#sha1=594b7dd32bec37d8bf70a6ffa8866d30e93f3c42

```

FlorentJeannot:
Hey @luzfcb, this PR would definitely fix your issue! :)

Concerning your suggestion, I think the existing tests are enough. Could you explain your point of view?

@atugushev Do you know when this PR will be reviewed?

Thanks!

piptools/utils.py
```diff
@@ -115,7 +115,16 @@ def format_requirement(
     if ireq.editable:
         line = f"-e {ireq.link.url}"
     elif is_url_requirement(ireq):
-        line = ireq.link.url
+        if not ireq.name:
+            line = ireq.link.url
+        elif ireq.link.url.startswith("file:./"):
```
atugushev, line 120: How `file:.` would work?

FlorentJeannot: This is to preserve an existing "hack" that allows us to specify a relative path, as mentioned [here](https://github.com/jazzband/pip-tools/issues/204#issuecomment-550051424).
Although, I can see this is not good enough: if the relative path does not start with `"./"`, such as `file:src/foo#egg=foo` then this does not work.
I'm going to update this.

FlorentJeannot: The logic has been updated, so I am resolving this conversation if you don't mind.

piptools/utils.py
```diff
@@ -115,7 +115,18 @@ def format_requirement(
     if ireq.editable:
         line = f"-e {ireq.link.url}"
     elif is_url_requirement(ireq):
-        line = ireq.link.url
+        # if the requirement has no name then it's just a URL
+        if not ireq.name:
+            line = ireq.link.url
+        # if it starts with "file:", PEP508 does not support direct reference
+        elif ireq.link.url.startswith("file:"):
```
atugushev, line 122: > PEP508 does not support direct reference

I couldn't find any information about that in the PEP. I've checked the grammar and couldn't found any restrictions on `name @ file:...` too.

I've tried locally absolute paths and it works:

```console
$ pip install 'pip-tools @ file:///Users/albert/Projects/pip-tools'
...
Installing collected packages: pip-tools
  Attempting uninstall: pip-tools
    Found existing installation: pip-tools 6.1.1.dev72+g39856be
    Uninstalling pip-tools-6.1.1.dev72+g39856be:
      Successfully uninstalled pip-tools-6.1.1.dev72+g39856be
Successfully installed pip-tools-6.1.1.dev72+g39856be
```

Local path doesn't work though:

```
$ pip install 'pip-tools @ file:./'
ERROR: Invalid requirement: 'pip-tools @ file:./'

$ pip install 'pip-tools @ file:.'
ERROR: Invalid requirement: 'pip-tools @ file:.'

$ pip install 'pip-tools @ file:pip-tools'
ERROR: Invalid requirement: 'pip-tools @ file:pip-tools'
```

tests/test_utils.py
```diff
@@ -32,9 +32,54 @@ def test_format_requirement(from_line):
...
+            "example @ https://example.com/example.zip",
+            id="direct reference",
+        ),
+        pytest.param(
+            "Example @ https://example.com/example.zip",
+            "example @ https://example.com/example.zip",
+            id="direct reference lowered case",
+        ),
+        pytest.param(
+            "exemple @ https://example.com/example.zip#egg=example",
```
atugushev, line 54:
```suggestion
            "example @ https://example.com/example.zip#egg=example",
```

Fixes a typo

FlorentJeannot: Thanks for catching this. 👍

piptools/utils.py
FlorentJeannot: Sorry about that, you're definitely right. There's even an example in pip's documentation about it: https://pip.pypa.io/en/stable/cli/pip_install/#requirement-specifiers

I have updated the code again. It looks simpler IMO now. Let me know what you think.
Also I'd love to add a test in `test_cli_compile.py` about a direct reference with a relative path in `file:` but it seems tricky to me because as far as I know the unit tests are working in a temp dir.

FlorentJeannot: @atugushev I have updated the tests in `test_cli_compile.py`.

piptools/utils.py
```diff
@@ -115,7 +116,16 @@ def format_requirement(
...
         line = f"-e {ireq.link.url}"
     elif is_url_requirement(ireq):
-        line = ireq.link.url
+        # if the requirement has no name then it's just a URL
+        if not ireq.name:
+            line = ireq.link.url
+        # otherwise parse the URL
+        else:
+            parsed_url = urlparse(ireq.link.url)
+            if "egg=" in parsed_url.fragment:
```
atugushev, line 125:
```suggestion
            if ireq.link.egg_fragment:
```

`egg_fragment` can be used

piptools/utils.py
```diff
@@ -115,7 +116,16 @@ def format_requirement(
     if ireq.editable:
         line = f"-e {ireq.link.url}"
     elif is_url_requirement(ireq):
-        line = ireq.link.url
+        # if the requirement has no name then it's just a URL
+        if not ireq.name:
```
atugushev, line 120: For `if-else` statements I would recommend the following best practice:

> Put the case you normally expect to process first. This is in line with the general principle of putting code that results from a decision as close as possible to the decision.

```python
if ok:
    do_ok_stuff()
else:
    do_not_ok_stuff()
```

FlorentJeannot: Thanks for your feedback, I've updated the logic.

piptools/utils.py
FlorentJeannot: Nice catch!

atugushev (approved):
Thanks for addressing suggestions :+1:

---

As a side note, this transformation feels confusing:
```
example @ https://example.com/example.zip#egg=example
→
https://example.com/example.zip#egg=example
```
 
 I'd prefer the opposed instead:
```
https://example.com/example.zip#egg=example
→
example @ https://example.com/example.zip
```

AFAIK egg fragment appears to be [discouraged](https://github.com/pypa/pip/issues/1289#issuecomment-757402359) for direct references. However, pip is going to refactor this within https://github.com/pypa/pip/issues/1289, so it would make sense to address that later, as soon as the issue is resolved.

FlorentJeannot:
> Thanks for addressing suggestions 👍
> 
> As a side note, this transformation feels confusing:
> 
> ```
> example @ https://example.com/example.zip#egg=example
> →
> https://example.com/example.zip#egg=example
> ```
> 
> I'd prefer the opposed instead:
> 
> ```
> https://example.com/example.zip#egg=example
> →
> example @ https://example.com/example.zip
> ```
> 
> AFAIK egg fragment appears to be [discouraged](https://github.com/pypa/pip/issues/1289#issuecomment-757402359) for direct references. However, pip is going to refactor this within [pypa/pip#1289](https://github.com/pypa/pip/issues/1289), so it would make sense to address that later, as soon as the issue is resolved.

Hello @atugushev,

Thank you for your feedback and your approval for this PR.
I suggest to mention "PEP 508 support" in the `CHANGELOG.md` instead of PEP 440 as it seems to be more appropriate.

Regarding your side note, I agree with you and I'll happily provide a PR for this when https://github.com/pypa/pip/issues/1289 will be resolved.

atugushev:
@FlorentJeannot 

We should also support extras, see https://github.com/jazzband/pip-tools/issues/1450 for details. Would you like to address it here or in a following up PR?

FlorentJeannot:
> @FlorentJeannot
> 
> We should also support extras, see #1450 for details. Would you like to address it here or in a following up PR?

@atugushev 

I will do it in another PR since this one has been approved.

Also I have another question. It seems that `pip` already converts URLs to the PEP 508 format. See https://github.com/pypa/pip/blob/7da05f33345de4c59930398aefa011ed7ccdeca3/src/pip/_vendor/packaging/requirements.py#L148
Is there a good reason to wait for https://github.com/pypa/pip/issues/1289 or could we do it like them now?

atugushev:
@FlorentJeannot 

Thanks for the PR! Awesome job 👍🏻 

> Is there a good reason to wait for [pypa/pip#1289](https://github.com/pypa/pip/issues/1289) or could we do it like them now?

Could you elaborate on how code in [`Requirement.__str__`()](https://github.com/pypa/pip/blob/7da05f33345de4c59930398aefa011ed7ccdeca3/src/pip/_vendor/packaging/requirements.py#L148) related to egg fragment refactoring in pypa/pip#1289?

FlorentJeannot:
@atugushev Thanks 👍  Happy to contribute to this project 😄 

So I believe `Requirement.__str__()` is what is used to format the requirement in `pip`. I think that's the equivalent to the function I modified in `piptools` to support the direct reference.
In their code they seem to always convert requirements to the direct reference format when they have a URL with a name.

So for example if you do this:
`pip install 'git+https://github.com/encode/uvicorn.git@0.14.0#egg=uvicorn[standard]'`

Even though it was specified with an egg, the output of `pip freeze` will be a direct reference for `uvicorn` (which makes me wonder if we shouldn't do like them already):

```
asgiref==3.4.1
click==8.0.1
h11==0.12.0
httptools==0.2.0
pep517==0.10.0
python-dotenv==0.18.0
PyYAML==5.4.1
toml==0.10.2
uvicorn @ git+https://github.com/encode/uvicorn.git@87da6cf4082cd28306bdb78d7d42552d131cc179
uvloop==0.15.3
watchgod==0.7
websockets==9.1
```

atugushev:
@FlorentJeannot 

> if we shouldn't do like them already

Ideally, we should. Feel free to shoot a PR if think it's doable.

FlorentJeannot:
> @FlorentJeannot
> 
> > if we shouldn't do like them already
> 
> Ideally, we should. Feel free to shoot a PR if you think it's doable.

Will do!

PR #1394, opened 2021-04-27
Improve instructions for new contributors

FlorentJeannot:
When I made my first PR, I was expecting tox to be in `testing`, because the doc says:

```
- Install pip-tools in development mode and its test dependencies with pip install -e .[testing].
- Check with `tox -e checkqa` to see your changes are not breaking the style conventions.
```

**Changelog-friendly one-liner**: Improve instructions for new contributors

##### Contributor checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,10 @@ and follow the [guidelines](https://jazzband.co/about/guidelines).
 
 Here are a few additional or emphasized guidelines to follow when contributing to `pip-tools`:
 
-- Always provide tests for your changes.
+- To set up your project's environment, have a look at [Python Packaging User Guide](https://packaging.python.org)
+  and [Real Python Tutorials](https://realpython.com)
+- Install pip-tools in development mode and its test dependencies with `pip install -e '.[testing]'`.
+- Check with `tox -e checkqa` to see your changes are not breaking the style conventions.
 - Give a clear one-line description in the PR (that the maintainers can add to [CHANGELOG](CHANGELOG.md) afterwards).
 - Wait for the review of at least one other contributor before merging (even if you're a Jazzband member).
 - Before merging, assign the PR to a milestone for a version to help with the release process.
@@ -17,75 +20,6 @@ The only exception to those guidelines is for trivial changes, such as
 documentation corrections or contributions that do not change pip-tools itself.
 
 Contributions following these guidelines are always welcomed, encouraged and appreciated.
-
-## Development Guidelines
-
-### Installation
-
-This project requires Python 3.6 or higher.
-
-At the project's root, create a virtual environment, using your current Python 3 version:
-
-```console
-$ python3 -m venv .venv
-```
-
-Activate the virtual environment:
-
-```console
-$ source .venv/bin/activate
-```
-
-Install `pip-tools` in development mode and its test dependencies:
-
-```console
-$ pip install -e .[testing]
-```
-
-If you are using `zsh`, you need to escape the brackets: `\[testing\]`.
-
-
-### Testing with pytest
-
-This project uses [pytest](https://docs.pytest.org/en/stable/contents.html) to run the tests.
-
-Inside your virtual environment, at the project's root, use:
-
-```console
-$ pytest
-```
-
-
-### Testing with tox
-
-This project uses [tox](https://tox.readthedocs.io/en/latest/) to test different environments.
-
-To install several Python versions, use [pyenv](https://github.com/pyenv/pyenv).
-To help `tox` find out which Python versions are installed, at the root of the project, declare the versions using
-`pyenv local  <version> <version2> ...`. For example:
-
-```console
-$ pyenv local 3.7.10 3.8.9 3.9.4
-```
-
-Inside your virtual environment, install `tox`:
-
-```console
-$ pip install tox
-```
-
-Prepare `tox` to use the Python versions:
-
-```console
-$ tox --notest -p auto --parallel-live
-```
-
-Run `tox` in parallel:
-
-```console
-$ tox -p auto
-```
-
 
 ## Project Release Process
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ and follow the [guidelines](https://jazzband.co/about/guidelines).
 Here are a few additional or emphasized guidelines to follow when contributing to `pip-tools`:
 
 - If you need to have a virtualenv outside of `tox`, it is possible to reuse its configuration to provision it as [described in the docs](https://tox.readthedocs.io/en/latest/example/devenv.html#creating-development-environments-using-the-devenv-option).
-- Always provide tests for your changes and run `tox -p all` to make sure they are passing the checks.
+- Always provide tests for your changes and run `tox -p all` to make sure they are passing the checks locally.
 - Give a clear one-line description in the PR (that the maintainers can add to [CHANGELOG](CHANGELOG.md) afterwards).
 - Wait for the review of at least one other contributor before merging (even if you're a Jazzband member).
 - Before merging, assign the PR to a milestone for a version to help with the release process.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ using your current python version:
 
 Activate the virtual environment:
 
-    source ./.venv/bin/activate
+    source .venv/bin/activate
 
 ## Project Contribution Guidelines
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,7 +69,3 @@ To help keeping track of the releases and their changes, here's the current rele
 Please be mindful of other before and when performing a release, and use this access responsibly.
 
 Do not hesitate to ask questions if you have any before performing a release.
-
-```
-
-```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,6 +12,7 @@ Here are a few additional or emphasized guidelines to follow when contributing t
   and [Real Python Tutorials](https://realpython.com)
 - Install pip-tools in development mode and its test dependencies with `pip install -e '.[testing]'`.
 - Check with `tox -e checkqa` to see your changes are not breaking the style conventions.
+- Always provide tests for your changes.
 - Give a clear one-line description in the PR (that the maintainers can add to [CHANGELOG](CHANGELOG.md) afterwards).
 - Wait for the review of at least one other contributor before merging (even if you're a Jazzband member).
 - Before merging, assign the PR to a milestone for a version to help with the release process.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ and follow the [guidelines](https://jazzband.co/about/guidelines).
 
 Here are a few additional or emphasized guidelines to follow when contributing to `pip-tools`:
 
-- Use `tox` to create a [development environment](https://tox.readthedocs.io/en/latest/example/devenv.html#creating-development-environments-using-the-devenv-option)
+- Use `tox` to create a [development environment](https://tox.readthedocs.io/en/latest/example/devenv.html#creating-development-environments-using-the-devenv-option).
 - Always provide tests for your changes and run `tox -p all` to make sure they are passing the checks.
 - Give a clear one-line description in the PR (that the maintainers can add to [CHANGELOG](CHANGELOG.md) afterwards).
 - Wait for the review of at least one other contributor before merging (even if you're a Jazzband member).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ and follow the [guidelines](https://jazzband.co/about/guidelines).
 
 Here are a few additional or emphasized guidelines to follow when contributing to `pip-tools`:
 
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -41,6 +41,7 @@ testing =
     pytest
     pytest-rerunfailures
     pytest-xdist
+    tox
```
atugushev, line 44: This would install `tox` inside `tox`'s virtualenvs. Would it make more sense to update the contribution guide to suggest install tox before using it?

webknjaz (changes requested):
Tox is a central piece of automation but it is not a test dependency. It manages envs and their deps. Instead, the instructions should be updated to use tox. One can use `--devenv` if a regular virtualenv is needed.

FlorentJeannot:
Thanks for your feedback, I made the changes. Let me know if you think we should tweak the instructions.

CONTRIBUTING.md
```diff
@@ -4,13 +4,27 @@ This is a [Jazzband](https://jazzband.co/) project. By contributing you agree
...
+    pip install tox
+
+Create a virtual environment, with `pip-tools` in development mode and its test dependencies,
+using your current python version:
+
+    tox --devenv .venv -e py
+
+Activate the virtual environment:
+
+    source ./.venv/bin/activate
```
webknjaz, line 20: I think it's cleaner like this:
```suggestion
    source .venv/bin/activate
```

CONTRIBUTING.md
```diff
@@ -4,13 +4,27 @@ This is a [Jazzband](https://jazzband.co/) project. By contributing you agree
 to abide by the [Contributor Code of Conduct](https://jazzband.co/about/conduct)
 and follow the [guidelines](https://jazzband.co/about/guidelines).
 
+## Getting started
+
+Install tox:
+
+    pip install tox
```
webknjaz, line 11: Since it's not RST, we should probably use markdown + maybe mark it with a `console` lexer:
```suggestion
```console
$ python -m pip install tox --user
```\
```

webknjaz: Although I'm not sure if this should dive into the details of using tools like pip and tox — maybe it's better to point to the respective docs rather than attempting to maintain copies of their guides?

CONTRIBUTING.md
```diff
@@ -4,13 +4,27 @@ This is a [Jazzband](https://jazzband.co/) project. By contributing you agree
...
+    source ./.venv/bin/activate
+
 ## Project Contribution Guidelines
 
 Here are a few additional or emphasized guidelines to follow when contributing to pip-tools:
 
-- Install pip-tools in development mode and its test dependencies with `pip install -e .[testing]`.
 - Check with `tox -e checkqa` to see your changes are not breaking the style conventions.
-- Always provide tests for your changes.
+- Always provide tests for your changes, use `tox -e coverage` to run the tests with coverage reports.
```
webknjaz, line 27: Honestly, running `tox` will run all the test envs with coverage + linters. Not sure if mentioning just `coverage` is needed. Besides, codecov will report back to PRs if it's decreased, I think. Maybe this sort of detail belongs to a development guide rather than contribution instructions. I see such details as non-mandatory things that may help during the development process but should be used if the developer actually needs them. Same with use `tox -e py` to run on test env with the current Python, use `tox -e pyXY` to specify other Python, and so on — we could basically end up documenting how tox works but that would probably be too overwhelming.
Mentioning `tox -av` along the lines might be useful, OTOH.

webknjaz:
@FlorentJeannot this PR probably deserves a title change now that it doesn't do what the said title says :man_shrugging:

CONTRIBUTING.md
FlorentJeannot: Agreed

CONTRIBUTING.md
FlorentJeannot: Markdown seems to understand the format I used though (I agree it's missing the console lexer though)

image

CONTRIBUTING.md
FlorentJeannot: `tox -e checkqa` was already there when I started contributing to `pip-tools`, so it already looked like a mix of contribution and development guide to me. 
We could eventually regroup these instructions in a Development guide section of this file? I think making another file for this is too much. I personally like a README that tells me exactly what to do to get started quickly. Later I can go to the tool's website to understand it better (we could provide the links too).

Also, when I ran `tox -e coverage` and `tox -e checkqa`, it ran the checks and the tests only for my version of python. Do you suggest that we mention to run `tox` only, to run everything once?

By the way what I did for my first PR:
- Ensured the tests were passing for my python version, checked the code coverage reports about my new tests
- Run the code linters which fixed my files
- See in my PR that the CI succeeded for every python versions

FlorentJeannot:
> @FlorentJeannot this PR probably deserves a title change now that it doesn't do what the said title says 🤷‍♂️

Agreed, I updated the PR's title. However I'm not sure I can update the branch name without any trouble. If you know how to please let me know.
Thanks!

CONTRIBUTING.md
atugushev: Maybe `tox -p all`?

CONTRIBUTING.md
```diff
@@ -4,13 +4,33 @@ This is a [Jazzband](https://jazzband.co/) project. By contributing you agree
...
+```console
+$ source .venv/bin/activate
+```
+
 ## Project Contribution Guidelines
 
 Here are a few additional or emphasized guidelines to follow when contributing to pip-tools:
 
-- Install pip-tools in development mode and its test dependencies with `pip install -e .[testing]`.
 - Check with `tox -e checkqa` to see your changes are not breaking the style conventions.
```
atugushev, line 32: We could drop this line since `pre-commit.ci` checks and autofixes QA issues. What do you think?

CONTRIBUTING.md
FlorentJeannot: FYI `tox -p all` is really slow on my old macbook air:

```
✔ OK py39-piplatest-coverage in 9 minutes, 36.482 seconds
✔ OK py39-pipprevious-coverage in 10 minutes, 20.35 seconds
✔ OK py39-pip20.3-coverage in 10 minutes, 20.76 seconds
✔ OK py39-pipmain-coverage in 10 minutes, 40.551 seconds
```

So do you guys consider running `tox` with all test environments mandatory?
Also it seems to run only for `py39`, the current version I am using. Does that mean devs should install other python versions on their machine so `tox` can use them as well? If so, this is something we should mention I think? (it's my first time contributing to an open source project and I am not used to how `tox` works)

According to what you say, I suggest to change `Getting started` to `Development Guidelines`. 
Inside, we could suggest command lines to get things done correctly. But it should be kept simple, as @webknjaz said, we don't want to write a full doc about the tools and all possibilities.

Inside the contributing guidelines, we could have:

```
- Always provide tests for your changes.
- Check with `tox` that your changes are not breaking the style conventions.
```

And inside `Development Guidelines`, we could have what we wrote in `Getting started` and add a line or two about testing the branch, probably `$ tox -p all`.

What do you think?

CONTRIBUTING.md
FlorentJeannot: Yes that's a nice catch.
Also if we suggest to run `tox -p all` then I think it will still be run?

CONTRIBUTING.md
```diff
@@ -4,13 +4,33 @@ This is a [Jazzband](https://jazzband.co/) project. By contributing you agree
...
+using your current python version:
+
+```console
+$ tox --devenv .venv -e py
+```
+
+Activate the virtual environment:
+
+```console
+$ source .venv/bin/activate
```
webknjaz, line 25: One thing I'm concerned about is that it doesn't make it clear that running tests through tox will not use this env, it can be used for development detached from tox and if one wants to run pytest there, they should do so directly.

FlorentJeannot:
@webknjaz @atugushev I updated the contribution guidelines. I think it's better to tell users to install everything inside the project's virtualenv. I ran the command lines to double-check that it works. Let me know what you think.

Also, if you could take some time to review https://github.com/jazzband/pip-tools/pull/1392 that would be kind of you. I described the issue as much as I could to help you understand it. Thanks in advance!

webknjaz:
The whole workflow is tox-centric, as almost any Python project of this sort today. Throwing that away and maintaining a copy of instructions that tox has embedded keeping it in sync all the time is a thankless job.
This is why I'm huge :-1: on removing it from the guide. The whole idea of tox is that it provides the same experience in the CI and locally.

webknjaz:
By the way, instead of guessing how to escape the brackets for zsh, it's easier to just wrap that with single quotes.

webknjaz:
The typical experience is to just do `tox -e py -- <extra args for pytest>` without having a raw venv for testing. That venv can be useful for developing, although I just reuse what tox already has — `.tox/py/bin/python` gets me in just as easily.

webknjaz:
We should not maintain a usage guide on every possible way to set up the env/python/etc — there are great writeups describing a million ways to do this already, we could just point at https://packaging.python.org and https://realpython.com instead.

FlorentJeannot:
> We should not maintain a usage guide on every possible way to set up the env/python/etc — there are great writeups describing a million ways to do this already, we could just point at https://packaging.python.org and https://realpython.com instead.

Ok let's do that instead.

CONTRIBUTING.md
```diff
@@ -6,9 +6,11 @@ and follow the [guidelines](https://jazzband.co/about/guidelines).
 
 ## Project Contribution Guidelines
 
-Here are a few additional or emphasized guidelines to follow when contributing to pip-tools:
+Here are a few additional or emphasized guidelines to follow when contributing to `pip-tools`:
 
-- Install pip-tools in development mode and its test dependencies with `pip install -e .[testing]`.
+- To set up your project's environment, have a look at [Python Packaging User Guide](https://packaging.python.org)
+  and [Real Python Tutorials](https://realpython.com)
```
webknjaz, line 12: Could you link to specific articles rather than website homepages?

CONTRIBUTING.md
```diff
@@ -6,9 +6,11 @@ and follow the [guidelines](https://jazzband.co/about/guidelines).
...
+- Install pip-tools in development mode and its test dependencies with `pip install -e '.[testing]'`.
```
webknjaz, line 13: Let's avoid telling people to directly access the extras. We don't have a linter that would enforce a person changing the name of the extra in `setup.cfg` or some setting in `tox.ini` to also make sure to change the guide and keep it in sync at all times. Why did you drop the `--devenv` thing anyway? It would create a venv with deps identical to what tox has.

CONTRIBUTING.md
webknjaz: I think it's a good idea to have `-p` there. The ultimate goal is to let people know how to run the same test matrix that the CI runs locally. And running linters locally can also be useful because people often want to reproduce the offenses in their env to be able to address them. Rerunning the whole CI on every typo fix push is wasteful.

CONTRIBUTING.md
FlorentJeannot: I couldn't find any good article about `tox` so I provided a link to tox's documentation instead.

CONTRIBUTING.md
FlorentJeannot: I now suggest to install tox and use the devenv option specified in tox's documentation. I am only suggesting one command line so we won't have to keep this doc up to date.

CONTRIBUTING.md
```diff
@@ -6,11 +6,10 @@ and follow the [guidelines](https://jazzband.co/about/guidelines).
...
 ## Project Contribution Guidelines
 
-Here are a few additional or emphasized guidelines to follow when contributing to pip-tools:
+Here are a few additional or emphasized guidelines to follow when contributing to `pip-tools`:
 
-- Install pip-tools in development mode and its test dependencies with `pip install -e .[testing]`.
-- Check with `tox -e checkqa` to see your changes are not breaking the style conventions.
-- Always provide tests for your changes.
+- Use `tox` to create a [development environment](https://tox.readthedocs.io/en/latest/example/devenv.html#creating-development-environments-using-the-devenv-option).
+- Always provide tests for your changes and run `tox -p all` to make sure they are passing the checks.
```
webknjaz, line 12:
```suggestion
- Always provide tests for your changes and run `tox -p all` to make sure they are passing the checks locally.
```

CONTRIBUTING.md
```diff
@@ -6,11 +6,10 @@ and follow the [guidelines](https://jazzband.co/about/guidelines).
 
 ## Project Contribution Guidelines
 
-Here are a few additional or emphasized guidelines to follow when contributing to pip-tools:
+Here are a few additional or emphasized guidelines to follow when contributing to `pip-tools`:
 
-- Install pip-tools in development mode and its test dependencies with `pip install -e .[testing]`.
-- Check with `tox -e checkqa` to see your changes are not breaking the style conventions.
-- Always provide tests for your changes.
+- Use `tox` to create a [development environment](https://tox.readthedocs.io/en/latest/example/devenv.html#creating-development-environments-using-the-devenv-option).
```
webknjaz, line 11: Maybe something like this?
```suggestion
- If you need to have a virtualenv outside of `tox`, it is possible to reuse its configuration to provision it as [described in the docs](https://tox.readthedocs.io/en/latest/example/devenv.html#creating-development-environments-using-the-devenv-option).
```

atugushev (approved):
Looks great! Thanks @FlorentJeannot for working on this patiently and @webknjaz for reviews!

FlorentJeannot:
@atugushev No problem! Happy to contribute to this project!
Thanks guys for your reviews and your help!

webknjaz (approved):

Issue #1398, opened 2021-05-14
piptools fails with newest click release


nicoa:
Hey,

not sure if I'm correct here or over at click issues, please let me know if information is missing or this is the wrong place!

#### Environment Versions

1. docker `python:3.8-buster` and debian
1. Python version: `Python 3.8.10` and `Python 3.9.5`
1. pip version: `21.1.1`
1. pip-tools version: `pip-tools-6.1.0`

#### Steps to replicate

````console
pip install --upgrade pip click
pip install pip-tools
pip-compile --output-file requirements.txt requirements.in
````

#### Expected result
This worked under click 7.1.2, so I'd expect it to work after upgrading as well as the dependency is not pinned.

#### Actual result
After using newest click (click-8.0.0), the following error was raised:

````python-traceback
pip 21.1.1 from /usr/local/lib/python3.8/site-packages/pip (python 3.8)
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/click/decorators.py", line 384, in callback
    version = metadata.version(package_name)  # type: ignore
  File "/usr/local/lib/python3.8/importlib/metadata.py", line 530, in version
    return distribution(distribution_name).version
  File "/usr/local/lib/python3.8/importlib/metadata.py", line 503, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/local/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: piptools
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/local/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1134, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1058, in main
    with self.make_context(prog_name, args, **extra) as ctx:
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 927, in make_context
    self.parse_args(ctx, args)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1376, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 2352, in handle_parse_result
    value = self.process_value(ctx, value)
[...]

PR #1400, opened 2021-05-19
pass package_name explicitly in click.version_option decorators

nicoa:
Closes https://github.com/jazzband/pip-tools/issues/1398 .

Unfortunately mypy threw some errors during pre-commit run, where two of them seem to be valid (coming from CI as well).

Further, some of the tests failed, but this seems to be related to a pip.conf file adding extra indices to the compiled requirements, I'll re-run them again to see if this is related.

**Changelog-friendly one-liner**: pass package_name explicitly in click.version_option decorators for compatibility with click>=8.0

##### Contributor checklist

### DIFF
--- a/piptools/_compat/__init__.py
+++ b/piptools/_compat/__init__.py
@@ -1,4 +1,4 @@
-from .pip_compat import PIP_VERSION, parse_requirements
 from .click_compat import IS_CLICK_VER_8_PLUS
+from .pip_compat import PIP_VERSION, parse_requirements
 
 __all__ = ["PIP_VERSION", "IS_CLICK_VER_8_PLUS", "parse_requirements"]
--- a/piptools/_compat/__init__.py
+++ b/piptools/_compat/__init__.py
@@ -1,3 +1,4 @@
 from .pip_compat import PIP_VERSION, parse_requirements
+from .click_compat import IS_CLICK_VER_8_PLUS
 
-__all__ = ["PIP_VERSION", "parse_requirements"]
+__all__ = ["PIP_VERSION", "IS_CLICK_VER_8_PLUS", "parse_requirements"]
--- a/piptools/_compat/__init__.py
+++ b/piptools/_compat/__init__.py
@@ -1,4 +1,4 @@
-from .click_compat import IS_CLICK_VER_8_PLUS
+from .click import IS_CLICK_VER_8_PLUS
 from .pip_compat import PIP_VERSION, parse_requirements
 
 __all__ = ["PIP_VERSION", "IS_CLICK_VER_8_PLUS", "parse_requirements"]
--- a/piptools/_compat/click_compat.py
+++ b/piptools/_compat/click_compat.py
@@ -1,0 +1,5 @@
+import click
+from pip._vendor.packaging.version import Version
+
+CLICK_VERSION = Version(click.__version__).major
+IS_CLICK_VER_8_PLUS = CLICK_VERSION > 7
--- a/piptools/_compat/click_compat.py
+++ b/piptools/_compat/click_compat.py
@@ -1,5 +1,5 @@
 import click
 from pip._vendor.packaging.version import Version
 
-CLICK_VERSION = Version(click.__version__).major
-IS_CLICK_VER_8_PLUS = CLICK_VERSION > 7
+CLICK_MAJOR_VERSION = Version(click.__version__).major
+IS_CLICK_VER_8_PLUS = CLICK_MAJOR_VERSION > 7
--- a/piptools/_compat/click_compat.py
+++ b/piptools/_compat/click_compat.py
@@ -1,5 +1,0 @@
-import click
-from pip._vendor.packaging.version import Version
-
-CLICK_MAJOR_VERSION = Version(click.__version__).major
-IS_CLICK_VER_8_PLUS = CLICK_MAJOR_VERSION > 7
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -12,7 +12,7 @@ from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
 from pip._internal.utils.misc import redact_auth_from_url
 
-from .._compat import parse_requirements
+from .._compat import parse_requirements, IS_CLICK_VER_8_PLUS
 from ..cache import DependencyCache
 from ..exceptions import PipToolsError
 from ..locations import CACHE_DIR
@@ -34,11 +34,7 @@ DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
 METADATA_FILENAMES = frozenset({"setup.py", "setup.cfg", "pyproject.toml"})
 
 # TODO: drop click 7 and remove this block, pass directly to version_option
-if click.__version__[0] == "7":
-    ver_kwargs = {}
-else:
-    # this was introduced in click8 and if not passed would break this package
-    ver_kwargs = {"package_name": "pip-tools"}
+version_option_kwargs = {} if IS_CLICK_VER_8_PLUS else {"package_name": "pip-tools"}
 
 
 def _get_default_option(option_name: str) -> Any:
@@ -52,7 +48,7 @@ def _get_default_option(option_name: str) -> Any:
 
 
 @click.command(context_settings={"help_option_names": ("-h", "--help")})
-@click.version_option(**ver_kwargs)
+@click.version_option(**version_option_kwargs)
 @click.pass_context
 @click.option("-v", "--verbose", count=True, help="Show more output")
 @click.option("-q", "--quiet", count=True, help="Give less output")
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -33,6 +33,13 @@ DEFAULT_REQUIREMENTS_FILE = "requirements.in"
 DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
 METADATA_FILENAMES = frozenset({"setup.py", "setup.cfg", "pyproject.toml"})
 
+# TODO: drop click 7 and remove this block, pass directly to version_option
+if click.__version__.split(".")[0] == "7":
+    ver_kwargs = {}
+else:
+    # this was introduced in click8 and if not passed would break this package
+    ver_kwargs = {"package_name": "pip-tools"}
+
 
 def _get_default_option(option_name: str) -> Any:
     """
@@ -45,7 +52,7 @@ def _get_default_option(option_name: str) -> Any:
 
 
 @click.command(context_settings={"help_option_names": ("-h", "--help")})
-@click.version_option(package_name="pip-tools")
+@click.version_option(**ver_kwargs)
 @click.pass_context
 @click.option("-v", "--verbose", count=True, help="Show more output")
 @click.option("-q", "--quiet", count=True, help="Give less output")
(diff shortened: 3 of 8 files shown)


nicoa:
the failing mypy CI is due to click not being upgraded yet in typeshed. What do you think, should I add a `# type: ignore` to the two respective lines or wait for typeshed to incorporate click 8 changes?

nicoa:
Update: see https://github.com/python/typeshed/issues/5503, this seems to be an issue with click directly. same questions still applies: ignore that typing?

davidism:
Click 8 provides typing directly, typeshed stubs are not used anymore. Not sure how the pre-commit mypy hook works, are you sure you're getting the latest click version in that environment?

nicoa:
Hey, are any maintainers around here and reading this? It seems that the pre-commit CI is optional for merging, would be great to get some comment on whether it's preferred to first merge that or first fix the failing CI job.

Regarding the mypy hook (the failing one): it seems that `additional_dependencies: [click>=8.0.0]` in pre-commit-config fixes the wrong click version, but introduces a whole lot of new errors (probably due to switch to new click version, not sure). I can provide them here if required. Would it make sense to open a ticket to fix them?

atugushev:
> Regarding the mypy hook (the failing one): it seems that `additional_dependencies: [click>=8.0.0]`

Aw, that makes sense now, since we cache installed dependencies in CI. @webknjaz Is there a decent way to drop the cache (other than update the cache key or example)?

nicoa:
> > Regarding the mypy hook (the failing one): it seems that `additional_dependencies: [click>=8.0.0]`
> 
> Aw, that makes sense now, since we cache installed dependencies in CI. @webknjaz Is there a decent way to drop the cache (other than update the cache key or example)?

Still, there would be other errors introduced then: I'd open a separate ticket for that, would you agree with that procedure?

atugushev:
> Still, there would be other errors introduced then with the new click version: I'd open a separate ticket for that, would you agree with that procedure?

Yeah, let's track this in a separate issue. Thanks! In addition to click, I'm inclined to think that the latest pip update brought new types and we need to address that also.

webknjaz:
> Is there a decent way to drop the cache (other than update the cache key or example)?

Not that I know of. Although, you seem to be assuming the wrong check. The one that is failing is not coming from GHA. It comes from the pre-commit.ci service.

nicoa:
The newest pre-commit ci run seemed to use the proper version and raised the same errors I experienced when explicitly adding the new dependency (compare #1401 and #1402 ). I'd say this can be merged while ignoring the failing pre-commit.ci check, what do you think @webknjaz ?

webknjaz:
Yes, unless we want to have broad compatibility with older click, we should bump the dep too.

Regarding broken typing, if it's directly connected to this change, it must be fixed as a part of it.

nicoa:
will bump the dependency then.

Regarding the typing: no, it's due to new click and is on latest pre-commit-CI as well, this just didn't had it due to some caching apparently. Would be solved in #1402 .

webknjaz:
But bumping click is directly related to this then. This PR cannot be accepted with breaking typing.

nicoa:
Sorry, that was a typo... I wanted to write that it's due to new pip version - so not related in any way to click and the current CI setup is broken with the same errors if running now as well already (as you can see in #1401 ).

webknjaz:
Oh, I misunderstood that. Thinking more about dropping the older click, I think it's probably preferable to support those. This is because click is fairly popular and there's a good chance that some users will have other tools in the same env that may pin click more strictly and so it's a potentially serious conflict we'd cause.

piptools/scripts/compile.py
```diff
@@ -45,7 +45,7 @@ def _get_default_option(option_name: str) -> Any:
 
 
 @click.command(context_settings={"help_option_names": ("-h", "--help")})
-@click.version_option()
+@click.version_option(package_name="pip-tools")
```
nicoa, line 48: @webknjaz should we rather not still stick to click >= 7 and pass this option only on click version 8?

nicoa:
fair point, but not sure how to handle as the current setup is broken sometimes under click 8. Made some inline suggestion in the code to keep click 7 but handle 8 separately then.

webknjaz:
There's a few ways:
1)
```python
ver_kwargs = {} if click7 else {'package_name': 'piptools'} 
...
@click.version_option(**ver_kwargs)
def cli(...): ...
```
2)
```python
...
def cli(...): ...

if click8:
    cli = click.version_option(package_name='piptools')(cli)
```
3)
```python
from itertools import partial
ver_opt = click.version_option if click7 else partial(click.version_option, package_name='piptools')
...
@ver_opt()
def cli(...): ...
```

nicoa:
That's what I suggested, sorry if that was not clear :-)  was just waiting for the okay to do this version-based differencing, would implement option 1 then. Thanks!

piptools/scripts/compile.py
```diff
@@ -33,6 +33,13 @@
 DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
 METADATA_FILENAMES = frozenset({"setup.py", "setup.cfg", "pyproject.toml"})
 
+# TODO: drop click 7 and remove this block, pass directly to version_option
+if click.__version__[0] == "7":
```
webknjaz, line 37: This check is not very reliable: what if the version is 70?

nicoa: true, I wrongly expected this to be removed sooner, but better to be safe. What do you think about `click.__version__.split(".")[0] == "7"`? That looks uglier but would do what it should do.

webknjaz: Almost. Except you should turn that into an integer and use `>` in the comparison. You may also want to move this into a compat module and declare some variable like `IS_CLICK_VER_8_PLUS`.

The reason for sooner vs later is not the best way of thinking about it is because over time the end users may use a release of pip-tools with this code which allows for any version of click above 7. So they may end up with old pinned pip-tools but a very modern click. I seem to recall stories of how win devs used a comparison with "9" for win 95/98 and then suddenly it was 2000, and then win 7/8 which started driving nuts old software that still works on new platforms because people never considered that versioning could change, back in the day.

nicoa: Definitely makes sense, thanks for the words!

implemented this, hope it is the proper place and implementation :)

piptools/_compat/__init__.py
```diff
@@ -1,3 +1,4 @@
+from .click_compat import IS_CLICK_VER_8_PLUS
 from .pip_compat import PIP_VERSION, parse_requirements
 
-__all__ = ["PIP_VERSION", "parse_requirements"]
+__all__ = ["PIP_VERSION", "IS_CLICK_VER_8_PLUS", "parse_requirements"]
```
webknjaz, line 4: Side note: why is this a list? It's usually recommended to make `__all__` immutable, hence use a tuple.

P.S. This is out of the scope of this PR, just wanted to make a note.

piptools/_compat/click_compat.py
```diff
@@ -0,0 +1,5 @@
+import click
+from pip._vendor.packaging.version import Version
+
+CLICK_VERSION = Version(click.__version__).major
```
webknjaz, line 4: I don't think the var name corresponds to its contents. Either add that it's major or only store the version.

piptools/_compat/click_compat.py
```diff
@@ -0,0 +1,5 @@
+import click
+from pip._vendor.packaging.version import Version
```
webknjaz, line 2: I don't really like relying on how and whether pip vendors its deps. Plus it may cause problems when repackaged in distros that unbundle vendored software.

piptools/_compat/__init__.py
nicoa: to be fair, this is a list as well in the python tutorial on modules: https://docs.python.org/3/tutorial/modules.html#importing-from-a-package and the pip package has this format as well. 
Do you want to change this here or should I leave as it is? I actually don't have any preference here.

piptools/_compat/click_compat.py
```diff
@@ -0,0 +1,5 @@
+import click
```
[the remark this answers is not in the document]
nicoa: thought it makes sense to have this analog to `pip_compat.py`, but can change this if you prefer `click` over that

piptools/_compat/click_compat.py
nicoa: sure, was just in analogy to `pip_compat` module - would you agree with using `int(click.__version__.split(".")[0])`?

piptools/_compat/__init__.py
webknjaz: No, don't touch this because it's out of the scope. But in the future, this should probably be added. I've added this comment for history.

P.S. Normally, pylint would warn about this problem but I don't think it's integrated.

piptools/_compat/click_compat.py
webknjaz: Yeah, that should be enough.

webknjaz: Although, you may want to multi-line it to keep the inline complexity low

webknjaz (approved):

piptools/_compat/click_compat.py
nicoa: not sure what you mean with multi-lining, as for me black would merge this into one line if not forcing to not do on the following: 
```
int(
    click.__version__.split(".")[0]
)
```

piptools/_compat/__init__.py
nicoa: at least pycodestyle seems to do so, compare https://github.com/PyCQA/pydocstyle/issues/62

piptools/_compat/click_compat.py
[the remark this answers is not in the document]
nicoa: fair point, would adapt then 👍

piptools/_compat/click_compat.py
nicoa: nvmd, will simply add a comment then in between the brackeets

nicoa:
Hey @webknjaz thanks for your patience with me during this PR! For me it looks like it is ready to get into a next release. Whats the progress from here? would you merge it to master directly and/or add a milestone (I am not able to do any of those two)? And do I have to do anything from my side or does everything look good now?

piptools/_compat/click_compat.py
webknjaz: Yeah, that's one of the reasons I dislike black...

As for the line complexity, I like this metric that https://wemake-python-stylegui.de introduced called Jones complexity. It really makes a difference maintainability/readability-wise!

Here's some more info on this https://sobolevn.me/2019/10/complexity-waterfall#lines.

I wish this set of linter plugins was integrated here.

webknjaz:
So it looks like the CI is failing because of an unrelated problem so that would need to be fixed before merging this PR.

Regarding the release, I don't do those so you'll have to wait.

piptools/_compat/click_compat.py
nicoa: didn't knew the blog entry regarding the complexity until now, found it a good read and would recommend to colleagues, thanks a lot for that, very interesting!

nicoa:
Alright, would wait then, thanks for the information! Yes, for the CI there is issue #1403 .

nicoa:
Hey @webknjaz , as you already reviewed this I'd merge it as well, but not sure if this is only fine for more senior persons on the project. Should I proceed or do you / others want to keep the hands on what gets to master?

webknjaz:
@nicoa feel free to merge

PR #1405, opened 2021-05-28
fix failing tox tests

### LINKED ISSUE #1403: tox tests are failing due to new pip release
There is a deprecation leading to failing CI since a few days, namely the one discussed in https://github.com/pypa/pip/issues/7555:
```
  DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
   pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
```

I'm not sure about the options here - i.e. if this should be filtered out or tested as suggested by pip maintainers in that message and where / how to do this in the tests when the cli is invoked.

Further, a second problem are the pip deprecation warnings for the "previous" pip test setup.

So, the obvious solution would be to do the following on each pip-call: use feature flag in recent pip (as suggested in the warning above) and suppress version warnings in previous pip (`export PIP_DISABLE_PIP_VERSION_CHECK=1`).

Unfortunately I'm not sure how to proceed here or how to move the warnings from the pytest error section to the pytest warnings section.

If the solution is easy to implement and one could guide me to the proper place and has a bigger vision of what should happen and has the time for this, I'd be happy to get some guidance to file a PR.

### Links
* latest successful: https://github.com/jazzband/pip-tools/actions/runs/870122483
* first failing: https://github.com/jazzband/pip-tools/actions/runs/873544759

nicoa:
Closes #1403 .

**Changelog-friendly one-liner**: fix failing tox tests by pinning setuptools to old version and disabling pip version check on old version and using new feature "in-tree-build" in new version

##### Contributor checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,6 +17,8 @@ deps =
     pip20.3: pip==20.3.*
 setenv =
     coverage: PYTEST_ADDOPTS=--strict --doctest-modules --cov --cov-report=term-missing --cov-report=xml {env:PYTEST_ADDOPTS:}
+        pipprevious,pip20.3: PIP_DISABLE_PIP_VERSION_CHECK=1
+        piplatest: PIP_USE_FEATURE=in-tree-build
 commands_pre =
     piplatest: python -m pip install -U pip
     pip --version
--- a/tox.ini
+++ b/tox.ini
@@ -11,11 +11,11 @@ extras =
     testing
     coverage: coverage
 deps =
+    setuptools<57
     pipprevious: pip==20.3.*
     piplatest: pip
     pipmain: -e git+https://github.com/pypa/pip.git@main#egg=pip
     pip20.3: pip==20.3.*
-    setuptools<=57
 setenv =
     coverage: PYTEST_ADDOPTS=--strict --doctest-modules --cov --cov-report=term-missing --cov-report=xml {env:PYTEST_ADDOPTS:}
         pipprevious,pip20.3: PIP_DISABLE_PIP_VERSION_CHECK=1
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -11,12 +11,15 @@ extras =
...
     coverage: coverage
 deps =
+    setuptools<57
     pipprevious: pip==20.3.*
     piplatest: pip
     pipmain: -e git+https://github.com/pypa/pip.git@main#egg=pip
     pip20.3: pip==20.3.*
 setenv =
     coverage: PYTEST_ADDOPTS=--strict --doctest-modules --cov --cov-report=term-missing --cov-report=xml {env:PYTEST_ADDOPTS:}
+        pipprevious,pip20.3: PIP_DISABLE_PIP_VERSION_CHECK=1
```
nicoa, line 21: those are probably not necessary - should we still use them anyways (at least the feature flag)?

tox.ini
```diff
@@ -11,12 +11,15 @@ extras =
     testing
     coverage: coverage
 deps =
+    setuptools<57
```
webknjaz, line 14: This requires a comment explaining the motivation

tox.ini
webknjaz, line 21: This indent is not aligned with the rest of the lines

tox.ini
nicoa: added a comment

ssbarnea (approved):

PR #1408, opened 2021-06-02
Fixed typing

ssbarnea:
**Changelog-friendly one-liner**: Fixed typing

##### Contributor checklist

pip 21.0 has internal changes that break at least our typing and we will need significant changes to support 21.x for typing. For the moment we ceil pip version for mypy and assure running mypy does get us consistentresults inside and outside pre-commit.

Fixes type change introduced by click 8

A previous version of this patch attempted to pin down pip version in setup.cfg but I soon found that piplatest tox env
started to fail as `in-tree-build` feature was unknown to it. I suppose we will have to bump minimal pip version required
to >=21 as soon we upgrade type hints .

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,13 @@ repos:
       - id: mypy
         # Avoid error: Duplicate module named 'setup'
         # https://github.com/python/mypy/issues/4008
+        # Keep exclude in sync with mypy own excludes
         exclude: ^tests/test_data/
+        additional_dependencies:
+          - click==8.0.1
+          - pep517==0.10.0
+          - toml==0.10.2
+          - pip==20.3.4
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.0
     hooks:
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -2,7 +2,7 @@ import os
 import shlex
 import sys
 import tempfile
-from typing import Any, BinaryIO, List, Optional, Tuple, cast
+from typing import IO, Any, BinaryIO, List, Optional, Tuple, Union, cast
 
 import click
 from click.utils import LazyFile, safecall
@@ -231,7 +231,7 @@ def cli(
     annotate: bool,
     upgrade: bool,
     upgrade_packages: Tuple[str, ...],
-    output_file: Optional[LazyFile],
+    output_file: Union[LazyFile, IO[Any], None],
     allow_unsafe: bool,
     generate_hashes: bool,
     reuse_hashes: bool,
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -93,5 +93,9 @@ warn_return_any = true
 warn_unused_configs = true
 warn_unused_ignores = true
 
+# Avoid error: Duplicate module named 'setup'
+# https://github.com/python/mypy/issues/4008
+exclude = ^tests/test_data/
```
webknjaz, line 98: I think we should update the invocation instead.

ssbarnea: I will look into addressing this in a follow-up as this fix is unrelated to current change (the exclude was already there).  I am sure we will have more mypy work done soon, but now I need to unblock the CI/CD gates.

.pre-commit-config.yaml
```diff
@@ -25,7 +25,14 @@ repos:
       - id: mypy
         # Avoid error: Duplicate module named 'setup'
         # https://github.com/python/mypy/issues/4008
+        # Keep exclude in sync with mypy own excludes
         exclude: ^tests/test_data/
+        entry: mypy
```
asottile, line 30: this line is the default and doesn't do anything

ssbarnea:
I think we should ignore the coverage issue as covering for both versions of click 7 and 8 would make our CI pipelines too complex. It is just one line not covered. It would be much easier to just bump minimal version of click required than covering for both versions.

piptools/scripts/compile.py
```diff
@@ -231,7 +231,7 @@ def cli(
     annotate: bool,
     upgrade: bool,
     upgrade_packages: Tuple[str, ...],
-    output_file: Optional[LazyFile],
+    output_file: Union[LazyFile, Union[IO[Any]], None],
```
webknjaz, line 234: I don't understand why the internal union has just one arg. This seems pointless.

piptools/scripts/compile.py
```diff
@@ -282,7 +282,8 @@ def cli(
 
         # Close the file at the end of the context execution
         assert output_file is not None
-        ctx.call_on_close(safecall(output_file.close_intelligently))
+        if isinstance(output_file, LazyFile):
```
webknjaz, line 285: Why is this necessary?

ssbarnea:
It is needed to pass mypy because the type changed in click 8 vs 7.

piptools/scripts/compile.py
ssbarnea: That was undesired, leftover now removed.

piptools/scripts/compile.py
ssbarnea: I added an explanatory comment (click8)

webknjaz (approved):

PR #1409, opened 2021-06-05
Disable codecov comments

ssbarnea:
**Changelog-friendly one-liner**: Disable codecov PR comments

##### Contributor checklist


webknjaz:
I don't think this is necessary.

ssbarnea:
> I don't think this is necessary.

I do think is highly necessary, I don't want to be forced to scroll few screens in order to be able to reply to someone comment, just because codecov dumps a huge report as a comment. I don't want to spend time marking each of this comments off-topic in order to hide them. They have zero value.

Please note that I have nothing against coverage and failing checks when it goes down, is only that it should appear as part of the 10 checks we current have. There is no reason why coverage would need to be more visible than all the other CI jobs. Comment feature was created before GHA had support for checks, is a historical bit.

webknjaz:
But codecov usually doesn't post separate comments — it updates the first one it creates. Besides, the diagrams it posts are adjustable, I'd just configure that.

atugushev (approved):
Personally, I'd disable this comment. In fact, I tend to hide those comments. They take too much spaces in the feed and I usually do care only about `codecov / patch` metric, which can be found in checks.

Also it make some noise on email and post incorrect stats, which usually corrected later when all jobs have done. Perhabs, this can be fixed by configuration.

PR #1410, opened 2021-06-08
Fix click v7 version_option compatibility

FuegoFro:
It looks like the ternary was committed backwards in https://github.com/jazzband/pip-tools/pull/1400, since the `package_name` option was add in click 8, but is only being passed in click 7. That caused `pip-tools` to error on import when installed alongside click 7.

**Changelog-friendly one-liner**: Follow up https://github.com/jazzband/pip-tools/pull/1400 to fix click v7 compatibility.

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -34,7 +34,7 @@ DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
 METADATA_FILENAMES = frozenset({"setup.py", "setup.cfg", "pyproject.toml"})
 
 # TODO: drop click 7 and remove this block, pass directly to version_option
-version_option_kwargs = {} if IS_CLICK_VER_8_PLUS else {"package_name": "pip-tools"}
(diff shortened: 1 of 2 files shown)


FuegoFro:
Hey there! I wasn't sure how to set a milestone or what milestone should even be set (possibly 6.1.1, just since that's the milestone which was set on https://github.com/jazzband/pip-tools/pull/1400?). Would love some help with that!

Also, I wasn't sure whether I should add a new `tox` env to tests this, let me know if you think that'd be valuable!

nicoa (approved):

nicoa:
you're totally right, good catch! As far as I know, only members of the jazzband organisation can set milestones, I added 6.1.1 for now, agreeing that this makes sense.

Regarding the env: I honestly don't know, maybe others have a strong opinion on this.

webknjaz (approved):

webknjaz:
This will probably need to be synced with master to unblock the merge button.

FuegoFro:
Thank you for the quick review! 😀 I've gone ahead and rebased the branch on latest `master` 🙂

PR #1419, opened 2021-06-12
Fix pip-compile package ordering

### LINKED ISSUE #1414: Order of packages in requirements.txt -- pip-compile vs. pip freeze
The issue regards packages with names containing `-` character or a digit. This may cause an unexpected order of packages in the output file, different than with `pip freeze`. It seems that the output file is sorted by the whole lines, together with `==` and version name, instead of just by the package names.

#### Environment Versions

1. Windows 10
2. Python version: 3.7.6
3. pip version: 21.1.2
4. pip-tools version: 6.1.0

#### Steps to replicate

`pip-compile` with `requirements.in` file:

```
django
django-redis
django-sendfile
django-sendfile2
djangorestframework
```

#### Expected result

(without comments and additional libraries, for clarity)

```
django==3.2.4
django-redis==5.0.0
django-sendfile==0.3.11
django-sendfile2==0.6.0
djangorestframework==3.12.4
```

#### Actual result

```
django-redis==5.0.0
django-sendfile2==0.6.0
django-sendfile==0.3.11
django==3.2.4
djangorestframework==3.12.4
```

adamsol:
Closes #1414.

**Changelog-friendly one-liner**: Fix `pip-compile` package ordering to match that of `pip freeze`.

##### Contributor checklist

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -89,7 +89,7 @@ class OutputWriter:
         self.emit_find_links = emit_find_links
 
     def _sort_key(self, ireq: InstallRequirement) -> Tuple[bool, str]:
-        return (not ireq.editable, str(ireq.req).lower())
+        return (not ireq.editable, key_from_ireq(ireq))
 
     def write_header(self) -> Iterator[str]:
         if self.emit_header:
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -827,10 +827,10 @@ def test_generate_hashes_with_annotations(runner):
 @pytest.mark.network
 def test_generate_hashes_with_long_annotations(runner):
     with open("requirements.in", "w") as fp:
+        fp.write("Django==1.11.29\n")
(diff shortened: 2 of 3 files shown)


ssbarnea:
I am a little bit against this change because it would break sorting using cli sort or editor sorting, which does the sorting line by line. In fact I would support `pip freeze` update to address this and making it compliant. I am wondering what others think.

adamsol:
I think that the current `pip freeze`'s behaviour makes more sense, and generic sorting utilities aren't the right tool for this job anyway. I doubt anyone would expect or want the `=` character to take part in the sorting. And even ignoring the lexicographic order of package names, those utilities won't handle urls or comments in the file properly. Also note that the order of packages after running a line-by-line sort on `requirements.in` (without version specifiers) can be different than for the resulting `requirements.txt`.

atugushev (approved):
LGTM 👍

PR #1424, opened 2021-06-21
Fix changelog category title template

atugushev:
To be consistent with the current template of the [changelog](https://github.com/jazzband/pip-tools/blob/master/CHANGELOG.md) and [release page](https://github.com/jazzband/pip-tools/releases).

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -23,7 +23,7 @@ categories:
       - "dependency"
       - "docs"
 
-category-template: "$TITLE: "
+category-template: "$TITLE:"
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

ssbarnea (changes requested):
The extra space at the end looks like a bug to me, we will likely endup with extra spacing that is not needed.

.github/release-drafter.yml
```diff
@@ -23,6 +23,7 @@ categories:
       - "dependency"
       - "docs"
 
+category-template: "$TITLE: "
```
atugushev, line 26:
```suggestion
category-template: "$TITLE:"
```

@ssbarnea that's a bug. Thanks for noticing.

ssbarnea (approved):

PR #1425, opened 2021-06-21
Prettify CHANGELOG file

atugushev:
We tend to wrap lines in [CHANGELOG](https://github.com/jazzband/pip-tools/blob/master/CHANGELOG.md) to ~88. Let `prettier` automatically do wrap lines, so that we don't have to do it manually since we [copy](https://github.com/jazzband/pip-tools/blob/master/CONTRIBUTING.md) changelogs from release drafts now:

> Update the CHANGELOG with the version, date and add the text from drafter release.

### DIFF
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,0 +1,5 @@
+printWidth: 88
+overrides:
+  - files: "CHANGELOG.md"
+    options:
+      proseWrap: always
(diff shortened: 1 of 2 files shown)


ssbarnea (approved):

atugushev:
Thanks @ssbarnea 🙏🏻

PR #1426, opened 2021-06-21
Make GitHub recognise BSD 3-clause license

atugushev:
See https://github.com/atugushev/pip-tools for the reference


webknjaz:
FTR I like making it markdown: https://github.com/IQAndreas/markdown-licenses/blob/master/bsd-3.md.

Also, here's the tool GH uses to detect the licenses https://licensee.github.io/licensee/command-line-usage/. You may run it locally to verify what it'll "see".

webknjaz (approved):

PR #1438, opened 2021-06-28
Allow passing --no-upgrade option

ssbarnea:
This makes this option similar with the other boolean ones and allows
calling the tools with explicit --no-upgrade.

This change ease integration with other tools like CI/CD pipelines
where user may want to change the default behavior related to
upgrades.

##### Contributor checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -125,7 +125,7 @@ def _get_default_option(option_name: str) -> Any:
 )
 @click.option(
     "-U",
-    "--upgrade",
+    "--upgrade/--no-upgrade",
     is_flag=True,
     default=False,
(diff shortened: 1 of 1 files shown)


atugushev (approved):
LGTM 👍  Maybe "enhancement"?

ssbarnea:
> LGTM 👍 Maybe "enhancement"?

I see it a as a bugfix, and I do not want to wait until 6.3.0 is ready, especially as lack of it prevents use with tox posargs where I want to do something like

```
command = pip-compile {posargs:--upgrade} ....
```

This construct allows running that tox env with only compilation and still allow user to call it with `tox -e envname -- --no-upgrade` to disable the implicit upgrade (something very useful for CI pipeline, where we only want to check if compile deps are valid, not to upgrade all deps).

PR #1460, opened 2021-08-01
Change log level of hash message

plannigan:
Minor change. It seemed odd that this was the only log message in the entire code bash using `log()` directly. As a result, it was not effected by the verbosity level.

I'm open to using a different log level. I went with `info()` as it would not change the currently default behavior.

**Changelog-friendly one-liner**: Adjust log level for PyPI hash message.

##### Contributor checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -331,7 +331,7 @@ class PyPIRepository(BaseRepository):
         with log.indentation():
             hashes = self._get_hashes_from_pypi(ireq)
             if hashes is None:
-                log.log("Couldn't get hashes from PyPI, fallback to hashing files")
+                log.info("Couldn't get hashes from PyPI, fallback to hashing files")
                 return self._get_hashes_from_files(ireq)
 
         return hashes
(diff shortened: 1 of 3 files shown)


piptools/repositories/pypi.py
```diff
@@ -331,7 +331,7 @@ def get_hashes(self, ireq: InstallRequirement) -> Set[str]:
         with log.indentation():
             hashes = self._get_hashes_from_pypi(ireq)
             if hashes is None:
-                log.log("Couldn't get hashes from PyPI, fallback to hashing files")
+                log.info("Couldn't get hashes from PyPI, fallback to hashing files")
```
atugushev, line 334:
```suggestion
                log.debug("Couldn't get hashes from PyPI, fallback to hashing files")
```

atugushev (commented):
Nice catch! I'd prefer `debug` though:

atugushev (approved):
LGTM 👍

plannigan:
There were a few test cases that explicitly checked for the log message. With the move to debug, they were failing because it was no longer being written. I updated the test cases to check for no message being written as it is the smallest change. A different option would to to remove the output check, as I don't think explicitly checking for log messages provide much value.

PR #1469, opened 2021-08-12
Add changelog to the Sphinx docs

webknjaz:
This patch enables Sphinx to include the changelog document as a top-level web page. Preview:

pip-tools-sphinx-changelog

Live demo: https://pip-tools--1469.org.readthedocs.build/en/1469/changelog/

P.S. The _only_ goal of this PR is to include `CHANGELOG.md` into the Sphinx setup.

##### Contributor checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -320,7 +320,7 @@ Improved Documentation:
   environment to avoid confusion
   ([#1023](https://github.com/jazzband/pip-tools/pull/1023)). Thanks @tekumara
 - Add a note to the `README` about how to install requirements on different stages to
-  [Workflow for layered requirements](https://github.com/jazzband/pip-tools#workflow-for-layered-requirements)
+  [Workflow for layered requirements](https://pip-tools.rtfd.io/en/latest/#workflow-for-layered-requirements)
   section ([#1044](https://github.com/jazzband/pip-tools/pull/1044)). Thanks @hramezani
 
 # 4.4.0 (2020-01-21)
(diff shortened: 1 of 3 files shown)


jdufresne:
IMO, we'd be better off converting the top-level document to rst and then moving it into the docs directory. It would then become a first class document in the docs infrastructure.

ssbarnea:
@jdufresne I find markdown more contributor friendly than ReST, especially for less experienced python developers. As sphinx already got decent md support I see a migrations towards md everywhere as better in long term.

I personally can write both, but I based on what I seend as incoming pull requests from others, people struggle less writing markdown than rst.

jdufresne:
> As sphinx already got decent md support I see a migrations towards md everywhere as better in long term.

I still prefer rst, but am not against switching to md.

Either way, this sounds like an orthogonal issue to me. First step would be moving the changelog into the official docs in whatever format they are currently being written in.

webknjaz:
I also prefer RST but it's out of the scope here. Plus it's probably hard with the current integration that relies on GH PRs, hence MD.

jdufresne (changes requested):
@webknjaz I'm not clear on your position, can you clarify?

Are you saying moving the changelog to the `docs/` directory as rst is out of scope? If so, I disagree. If we're going to better integrate the changelog then I think we should do so completely by moving it fully to the docs.

> Plus it's probably hard with the current integration that relies on GH PRs, hence MD.

Can you be explicit what this means? I don't know what you're referring to.

We can always rework or reevaluate tools if they are making other work more difficult.

webknjaz:
It's out of the scope of *this PR*. Because it's about including one file into the sphinx setup and *nothing else*.

Fully remaking the changelog generation procedure is something that does not fit within the scope.

webknjaz:
(rebased after merging #1480)

ssbarnea (approved):

PR #1474, opened 2021-08-26
Convert docs/contributing.md to rst to match Sphinx default format

jdufresne:
This obviates the need to use myst-parser when building docs as now all
doc files are in a single format, rst. This slightly simplifies the docs
overall by reducing the number of formats contributors and tools need to
manage.

##### Contributor checklist

##### Maintainer checklist


docs/contributing.rst
```diff
@@ -0,0 +1,84 @@
...
+   milestone.
+
+Please be mindful of other before and when performing a release, and use
+this access responsibly.
+
+Do not hesitate to ask questions if you have any before performing a
+release.
+
+.. |Jazzband| image:: https://jazzband.co/static/img/jazzband.svg
+   :target: https://jazzband.co/
```
AndydeCleyre, line 84: Currently this substitution is defined, but unused, and near-identical to the required header but missing the `:alt:`. I would think we should complete this definition here, and use `|Jazzband|` where we need it.

AndydeCleyre (changes requested):
I agree that we should only use one markup format if it's simpler, and it is.

If we're definitely using Sphinx, that format should be reStructuredText.

There's a partial and unused substitution definition though at the moment, please see my other comment.

Issue #1377, opened 2021-04-07
Support outputting pip-compile in a structured format (e.g. json)


cjerdonek:
Support outputting pip-compile in a structured format (e.g. json)

It would be useful if, in addition to the `requirements.txt` format, `pip-compile` also had the option of outputting the requirements and what it knows about each requirement in a standard format like json. This would give users much greater ability to create custom `requirements.txt` files, without having to rely on fragile parsing of `pip-compile`'s `requirements.txt` output.

The structured data could include things like:

* for each requirement:
  * the "via" values in a structured format
  * the requirement's hashes (for `--generate-hashes`), as well as distinguishing the "exact" hash (e.g. for #1330)
  * whether the requirement is "safe" in the sense of `--allow-unsafe`
  * whether the requirement is hashable or editable
  * the annotation line as constructed by `pip-compile` (for convenience, since these would also be constructible manually from other data)
  * the requirement line as constructed by `pip-compile` (for convenience, if the user would like to include it as is)

This feature would give users a way to address a number of issues in this tracker, in the absence of a fix in `pip-tools`. Here is a partial list:

* #1330 ("Provide a way to limit to a single hash per requirement when generating hashes")
* #1306 ("Compact annotations")
* #1107 ("annotations should be relative to file they are in")
* #1084 ("Add an option to not include paths in annotations")
* #994 ("Ignore hashes for version control repositories")
* #817 ("Allow to skip dependencies")
* #391 ("Include comments from input files in output file")
* #333 ("Exclude package?")
* #215 ("allow overriding/forcing a certain version")

woodruffw:
This would _potentially_ be useful for #1484 as well, although not strictly necessary! 

For our purposes, a schema like this (at the absolute minimum) would be sufficient:

```json
[
  {
    "name": "some-package-name",
    "version": "1.0.0",
  },
  ...
]
```

No particular order or other information would benefit our particular use case, but additional constraints wouldn't hurt us either!

con-f-use:
As I said in the [duplicate](https://github.com/jazzband/pip-tools/issues/1875), I have a fork, that filthy hack part of this. So my question is, would a proper implementation be welcome? Don't wanna put time into something, that will not be reviewed or is out of scope.

con-f-use:
For reference:
- https://github.com/pypa/pip/pull/10771
- https://github.com/jazzband/pip-tools/issues/1654

webknjaz:
Would #2124 supersede this?

PR #1477, opened 2021-08-29
Enable single-line annotations with `pip-compile --annotation-style=line`

### LINKED ISSUE #1306: Request: Compact annotations
**Primary:**

I would like to see a lot more information on my screen without having scroll up and down, when viewing annotated `pip-compile` output. At least as much as was the case before annotations became split into many lines.

**Secondary:**

I would like to use simple line-oriented shell tools to casually inspect properties of the `pip-compile` output, without resorting to multi-line regular expressions.

I would like the ability, through a flag or any other means, to have `pip-compile`'s annotated output place annotations for a pinned requirement on the same line as said requirement.

My screen is wide and short, as is common for computers. It's a more efficient use of this space to use more width and less height.

Example of the different use of space, for the `magic-wormhole` package from PyPI:

Example

#### Alternative Solutions
- I could skip annotations, but then I'd lose a lot of valuable information.
- I could add a post-processor to re-write the `pip-compile` output every time I compile.
- I could revert #1237 and poorly maintain a fork of this project, but that would make me sad.

#### Additional context
- #1297
- #1237
- #1377

AndydeCleyre:
`pip-compile` gains `--annotation-style=line` which uses the recently discontinued single-line annotation format.

The current default behavior is unchanged, but can be explicitly passed as `--annotation-style=split`.

This aims to fix #1306. Please provide feedback!

##### Contributor checklist

##### Maintainer checklist


tests/test_cli_compile.py
```diff
@@ -946,6 +946,54 @@ def test_generate_hashes_with_long_annotations(runner):
...
+        #
+        # This file is autogenerated by pip-compile with python \
+{sys.version_info.major}.{sys.version_info.minor}
+        # To update, run:
+        #
+        #    pip-compile --annotation-style=line --generate-hashes
+        #
+        django==1.11.29 \\
+            --hash=sha256:014e3392058d94f40569206a24523ce254d55ad2f9f46c6550b0fe2e4f94cf3f \\
+            --hash=sha256:4200aefb6678019a0acf0005cd14cfce3a5e6b9b90d06145fcdd2e474ad4329c
```
merwok, line 971: There used to be a backslash at the end of each hash line

AndydeCleyre: Yeah, but was that a mistake, or unnecessary? As in, the next line really has no business or need to be on a line continuation?

Just let me know what makes sense.

AndydeCleyre: And it _must_ not have a continuation if there's no annotation.

AndydeCleyre: Also, thanks for taking a look! Looking forward to this feature's return.

AndydeCleyre: In the spirit of line-oriented parsing, if someone uses a simple `sed` expression to strip comment lines, we don't need the complication of breaking the resulting syntax because one req continues into another.

reinout:
Small detail: the term "split". Would "verbose" or "multiline" be friendlier?

The verbose, multiline output is the default, so it seems a bit weird to use the term "split" as it is only "split" as opposed to the non-default "line" version.

Would a `--single-line` option without parameter be clearer? Then you only need to explain that it restores the old single-line-behaviour. You don't need to figure out a name for the default behaviour.

tests/test_cli_compile.py
merwok: The backlash was technically unnecessary, as it preceded a comment line, but you could argue that it was making one logical line with all hashes and the `via` comment, and more importantly it avoided unnecessary diff noise when new hashes were added.  If an option is added to restore the older, nicer format, then please restore it fully :)

AndydeCleyre: For me personally the point here is to be friendly to line oriented processing, so I don't wish to make that harder with the pre-comment continuation. But I will make it so if there's a consensus.

AndydeCleyre:
Thanks for the feedback! 

> Small detail: the term "split". Would "verbose" or "multiline" be friendlier?
> 

Multiline seems fine, but I don't prefer verbose, as we could have a more verbose single line annotation, theoretically. 

Whatever maintainers and community want to call it. 

> Would a `--single-line` option without parameter be clearer? 

Well not that exactly, because that doesn't indicate that this only affects the annotation (hashes still get their own lines). But the question of a single flag vs a parameter style argument is a good one. 

---

I don't personally care about these details, and am going to mark this as ready for review to hopefully gather answers to these and other questions from other maintainers.

jdufresne:
Any consideration given to developing the JSON intermediate format as discussed in #1377? The different annotations styles (as well as new ones) could be a consumer of this feature.

AndydeCleyre:
> Any consideration given to developing the JSON intermediate format as discussed in #1377? The different annotations styles (as well as new ones) could be a consumer of this feature.

Yes, but:

1. We need this regression fixed faster than we will build a consensus on json structure. 
2. It's suggested that the non structured output be included as a value in the structured output, in which case we'd still need to specify annotation format independently of output format.
3. The implementation can be later changed to use an intermediate structured object.

jdufresne:
Can you elaborate on point 2? I don't follow what that is suggesting or why it should prevent the work going ahead. Where exactly was this suggested. The purpose of the JSON format would be mean we don't need the formatting up front.

AndydeCleyre:
@jdufresne 

Sure, in the issue for structured output, #1377:

> The structured data could include things like:

> ...

> - the annotation line(s) as constructed by pip-compile (for convenience, since these would also be constructible manually from other data)
> - the requirement line(s) as constructed by pip-compile (for convenience, if the user would like to include them as is)

I am not at all saying work shouldn't go ahead on structured output. But that is a separate issue. Yes, it may later be used by this former and future feature.

yhovio (approved):

AndydeCleyre:
Note: yhovio's "approval" is meaningless; they're a spam account.

rsalmaso:
Nice! I was thinking about the same approach as https://github.com/jazzband/pip-tools/pull/1297 was really a not good feature.

tests/test_cli_compile.py
merwok: I don’t fully understand your use cases.  Either line-oriented tools work with physical lines, in which case my request doesn’t change anything, or they interpret backslashes to make logical lines and we’ll still have one line per dependency.

If you want to remove hashes, it would be simpler to call pip-compiles with different options.

My request is to keep the same format that we had before to avoid huge diffs.  This includes the backslash on each hash line 🙂

AndydeCleyre: I'm on mobile and busy right now, but the example is removing comments, not hashes. A naive comment remover would combine non comment lines, resulting in invalid syntax.

AndydeCleyre: So an explicit example of this breaking naive line-oriented expectations:

`reqs.txt`:
```ini
django-debug-toolbar==1.11 \
    --hash=sha256:89d75b60c65db363fb24688d977e5fbf0e73386c67acf562d278402a10fc3736 \
    --hash=sha256:c2b0134119a624f4ac9398b44f8e28a01c7686ac350a12a74793f3dd57a9eea0 \
    # via -r requirements.in
django-storages==1.9.1 \
    --hash=sha256:3103991c2ee8cef8a2ff096709973ffe7106183d211a79f22cf855f33533d924 \
    --hash=sha256:a59e9923cbce7068792f75344ed7727021ee4ac20f227cf17297d0d03d141e91 \
    # via -r requirements.in
```

```console
$ grep -Ev '^(\s*#|$)' reqs.txt
```

```ini
django-debug-toolbar==1.11 \
    --hash=sha256:89d75b60c65db363fb24688d977e5fbf0e73386c67acf562d278402a10fc3736 \
    --hash=sha256:c2b0134119a624f4ac9398b44f8e28a01c7686ac350a12a74793f3dd57a9eea0 \
django-storages==1.9.1 \
    --hash=sha256:3103991c2ee8cef8a2ff096709973ffe7106183d211a79f22cf855f33533d924 \
    --hash=sha256:a59e9923cbce7068792f75344ed7727021ee4ac20f227cf17297d0d03d141e91 \
```

merwok: Thanks for explaining the use case.  I don’t see any need for me to use such line procesing, and as I explained see more value in lines being diff-friendly and keeping the format that I currently have in all repositories (having pinned pip-tools for months now).

AndydeCleyre: To be clear, this concern is only about a one-time commit per `txt`, right? As in, the diffs won't be any bigger on an ongoing basis, only potentially bigger one time while getting over the `pip-tools` update hump?

AndydeCleyre: I'll explicitly note that adding that trailing slash would also mean unnecessarily larger diffs each time annotations are turned on or off. And this is not limited to a one-time `pip-tools` update hump.

piptools/writer.py
```diff
@@ -258,15 +277,14 @@ def _format_requirement(
...
-                    annotation_lines.append("    #   " + source)
-                annotation = "\n".join(annotation_lines)
-            line = f"{line}\n{comment(annotation)}"
+            annotation = {
+                "split": annotation_style_split,
+                "line": annotation_style_line,
+            }[self.annotation_style](required_by)
+            sep = {"split": "\n    ", "line": "\n    " if ireq_hashes else "  "}[
+                self.annotation_style
+            ]
```
ulope, line 286: IMO this would be much more readable if written as if blocks.
With only two options there is really no need to use dispatch tables.

tests/test_cli_compile.py
ulope: I agree with @merwok.
This PR is about adding a new (well old) output format. 
Please don't mix in a change to the current format.

ulope (commented):
Added comments.
In general looks like it's doing what it says on the tin :)

tests/test_cli_compile.py
merwok: Yes, the concern is about one commit.  I use pip-tools in many projects and don’t have a need to change options often.  I always want annotations, and I typically have hashes on from the start or enable the option once then keep it.  But! you’re the one doing the work here to restore a style close to what we had, so if I’m not convincing you then I’ll say thanks and take the one-time useless commit in my repos 🙂

AndydeCleyre:
@ulope


> Please don't mix in a change to the current format.

It is necessarily a change the current format. The old one is not current, but has been blown away. 

> This PR is about adding a new (well old) output format. 

Well for me the PR is about being friendly to line oriented tools and making good use of space.

Can you confirm that the only advantage to the final trailing slash is a _one_ time potentially smaller diff for each file? And after the one time change, it would only make potentially _larger_ diffs periodically, never smaller?

It seems strange to design a long term feature around a single event in time, with demonstrable and ongoing negative consequences, which we can instead avoid.

tests/test_cli_compile.py
AndydeCleyre: Thanks. I want to emphasize that while I am still against it, I'm not rejecting this yet, especially since @ulope is in agreement with you. 

Thank you for laying out the key details with me.

I just want to make sure the potential advantages and disadvantages are written out for a proper consensus to be reached.

jdufresne:
> Can you confirm that the only advantage to the final trailing slash is a one time potentially smaller diff for each file? And after the one time change, it would only make potentially larger diffs periodically, never smaller?

The trailing slash exists so that a hash change at the end of the hash list changes 1 line instead of 2. This is the same strategy formatting tools use when placing a unnecessary comma on the last line of a list. That is, appending an item is one line change in, not two.

piptools/writer.py
```diff
@@ -258,15 +277,14 @@ def _format_requirement(
...
+            lines = f"{line:24}{sep}{comment(annotation)}".splitlines()
```
jdufresne, line 287: What does the magic number 24 mean? Maybe add a comment or give it a name to make it self-documenting.

AndydeCleyre: I took it from the previous implementation. I don't know if it's derived from anything special, so I have no knowledge to add.

AndydeCleyre:
> The trailing slash exists so that a hash change at the end of the hash list changes 1 line instead of 2.

How often is it the case that the hash changes without the locked version changing (and therefore all the hashes changing anyway)?

jdufresne:
> > The trailing slash exists so that a hash change at the end of the hash list changes 1 line instead of 2.
> 
> How often is it the case that the hash changes without the locked version changing (and therefore all the hashes changing anyway)?

This has happened to me several times for compiled packages. It seemed like some projects would roll out platform specific binary packages over a period of time. So sometimes there would be a new hashes.

This also occurred with pure Python packages. Sometimes a project would forget to upload a wheel package. After getting a bug report they would then upload it days later, causing a new hash in my requirements.txt.

I don't have hard numbers, only anecdotes.

AndydeCleyre:
Thanks, that is helpful info. I had thought hashes were generally fixed for each release.

Does anyone think it's worthwhile to have 3 styles here? One that splits across many lines (current), one that is friendly to line oriented processing (what this PR calls "line" style), and one that minimizes diff sizes in the case of new hashes being added without a new release but increases them in a bunch  of other cases ("legacy" or something)?

I have never cared about the diff size, which seems to me a cosmetic concern, only the utility of seeing more info on the screen and the utility of simple line oriented processing friendliness.

merwok:
To me it feels that having two whole formats that differ by one backslash is a bit much.
I edit locked requirement files very rarely, when there’s some bug or version inconsistency that removes an extra specifier for example, but otherwise treat them as generated artifacts that can be mostly ignored, but also a text file with a useful history.  I would probably reach for a library (requirementslib?) if I wanted to do post-processing on the file, not a format-ignorant line tool.

tests/test_cli_compile.py
AndydeCleyre: I've added the line continuations; please check if they're working as expected, thanks!

piptools/writer.py
AndydeCleyre: I've redone that as blocks.

Before, an unexpected `self.annotation_style` would naturally raise a `KeyError`, but is now unhandled. I wonder about checking against a tuple or enum in either the writer or the compile script.

piptools/writer.py
AndydeCleyre: The new and third commit adds a comment that aims to convey a reassuring shrug in more words.

piptools/writer.py
AndydeCleyre: Oh, but of course we are enforcing the correct style names via the click parameter.

piptools/writer.py
merwok: nice 👍🏽 comments are usually before the line they apply to, though, unless pip-tools uses an unusual convention for this.

merwok:
Tested the fork and it works well!

Only change is that now a `django` requirement sorts before `django-other`, which is fine.

piptools/writer.py
AndydeCleyre: Thanks, swapped the lines

atugushev (approved):
Looks good to me 👍🏻 Thanks for the work! 

---

I see the continuation backslash on the last generated `--hash` option is back again (gone in 5.5.0, see #1237). I presume there'll be some noise in users' VCS. Are we sure this is okay?

AndydeCleyre:
Thanks @atugushev!

- I don't prefer to have that final trailing slash, but the reviewers before you do. 🤷‍♂️ 
- Do you think the style names can be improved?

jdufresne:
> I see the continuation backslash on the last generated --hash option is back again (gone in 5.5.0, see #1237). I presume there'll be some noise in users' VCS. Are we sure this is okay?

Thanks for linking this issue. I forgot that we intentionally removed this when I made a comment above 🤦. Yeah, lets not include the final trailing slash. (Sorry for the back and forth on this.)

jdufresne (approved):
Rest lgtm.

AndydeCleyre:
Alright I've removed the continuation between the last hash and the annotation again, with apologies to @merwok and @ulope.

AndydeCleyre:
It looks like codecov is confused by the rebase; can I get it to take a "fresh look" without adding a dummy commit?

atugushev:
@AndydeCleyre reopening PR didn't help. I don't know the other way to re-trigger checks 🤷🏻

AndydeCleyre:
OK, having bumped the codecov check, it looks like I need to add a little more test work to exercise the line annotation style with/without (probably without) hashes. I'll try to get this done later today (but if anyone else adds it in the meantime, 👍🏼).

AndydeCleyre:
Actually shouldn't that be covered already, by `test_annotate_option` with `id="annotate line style"`?

Can someone else please have a look at the coverage complaint?

AndydeCleyre:
I think the coverage test is mistaken. Not sure what to do about that.

atugushev:
Maybe merging master helps. Let's try.

AndydeCleyre:
Ah, I got the same result locally as well but based on the CLI output I misinterpreted the meaning. So this was introduced by replacing the `dict` dispatcher with the `if` block, since we never pass an invalid style name.

Since `split` is the default, I could just do `if...="line"...` and a catch-all `else:` rather than checking for `"split"` specifically.

AndydeCleyre:
Thank you very much @atugushev! I think this one will pass coverage checks.

AndydeCleyre:
All set, unless @atugushev or @jdufresne object to the latest commit.

piptools/writer.py
```diff
@@ -258,15 +277,15 @@ def _format_requirement(
             required_by.add(_comes_from_as_string(ireq))
 
         if required_by:
-            sorted_required_by = sorted(required_by)
-            if len(sorted_required_by) == 1:
-                source = sorted_required_by[0]
-                annotation = "    # via " + source
+            if self.annotation_style == "line":
```
atugushev, line 280: I'd prefer this instead. Works fine with coverage and explicitly checks all annotation styles (even erroneous):

```python
            if self.annotation_style == "line":
                annotation = annotation_style_line(required_by)
                sep = "\n    " if ireq_hashes else "  "
            elif self.annotation_style == "split":
                annotation = annotation_style_split(required_by)
                sep = "\n    "
            else:  # pragma: no cover
                raise ValueError('Invalid value for annotation style')
```

atugushev: As an alternative option, we could add a unit test to `test_writer.py` to check `else`  branch instead of using `no cover` pragma. Up to you. Both are fine by me.

AndydeCleyre: Excellent, thanks, let's go with the `no cover` approach.

atugushev:
Thanks, @AndydeCleyre! 🎉

atugushev:
Released in `pip-tools v6.3.0` 🚀

Issue #1448, opened 2021-07-08
Add a way to choose line endings in compiled files


jack1142:
#### What's the problem this feature will solve?

I'm trying to use the pip-compile hook with my existing pre-commit configuration but one of the hooks I have is a mixed line ending auto-fixer (which I use to ensure *all* files in the repository use LF for line endings):
```yaml
repos:
  - repo: https://github.com/jazzband/pip-tools
    rev: 6.2.0
    hooks:
      - id: pip-compile
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v3.4.0
    hooks:
      - id: mixed-line-ending
        args:
          - "--fix=lf"
```
This configuration causes issues whenever I try to touch a requirements file.

#### Describe the solution you'd like

A way to set the line endings used for the generated `requirements.txt` file. The alternative would be for pip-compile to always use `\n` but then I'm guessing someone else might complain about it in the future like I'm doing now :)

#### Alternative Solutions

I can work around this issue by excluding requirement files from the check but this doesn't *really* solve it as I still end up with requirements files ending with CRLF, it just allows me to continue to use both hooks.

Another solution would be a way for `--dry-run` to return non-zero exit code when the file would have any dependencies changed (like what #882 and #1215 suggest) but this doesn't fully solve it either as then you lose the benefit of pip-compile *auto-fixing* the issue during pre-commit run.

#### Additional context

N/A

graingert:
Imho pip-compile should reuse whatever line separator it finds first in requirements.txt

AndydeCleyre:
> This configuration causes issues whenever I try to touch a requirements file.

Can you please elaborate on the problem? Do you not have a guaranteed order of hook runs (to ensure the line-ending hook runs last)?

jack1142:
Run this on Windows (due to `os.linesep` being equal to `"\r\n"` there):
```
python -m pip install -U pre-commit
git clone https://github.com/jack1142/pip-tools-issue-1448
cd pip-tools-issue-1448
pre-commit run --all-files
```
See the output:
```
pip-compile..............................................................Passed
Mixed line ending........................................................Failed
- hook id: mixed-line-ending
- exit code: 1

requirements.txt: fixed mixed line endings
```

[...]

PR #1491, opened 2021-09-27
Add --newline=[LF|CRLF|native|preserve] option to compile, to override the line separator characters used

AndydeCleyre:
`pip-compile` gains an option with ~two~ ~three~ **four** valid choices: `--newline=[LF|CRLF|native|preserve]`,
which can be used to override the guessed newline character used in the output file. The default is ~`native`~ **`preserve`**, which ~uses `os.linesep`~ tries to be consistent with an existing output file, or input file, or ~FALLBACK_VALUE (`native`, or `LF`? TBD)~ **falls back to `LF`**, in that order.

This aims to address #1448.

~Note: [poll for fallback value](https://github.com/jazzband/pip-tools/pull/1491#issuecomment-929417008)~

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -98,6 +98,7 @@ class OutputWriter:
         index_urls: Iterable[str],
         trusted_hosts: Iterable[str],
         format_control: FormatControl,
+        linesep: str,
         allow_unsafe: bool,
         find_links: List[str],
         emit_find_links: bool,
@@ -117,6 +118,7 @@ class OutputWriter:
         self.index_urls = index_urls
         self.trusted_hosts = trusted_hosts
         self.format_control = format_control
+        self.linesep = linesep
         self.allow_unsafe = allow_unsafe
         self.find_links = find_links
         self.emit_find_links = emit_find_links
@@ -215,7 +217,7 @@ class OutputWriter:
         if packages:
             for ireq in sorted(packages, key=self._sort_key):
                 if has_hashes and not hashes.get(ireq):
-                    yield MESSAGE_UNHASHED_PACKAGE
+                    yield self.linesep.join(MESSAGE_UNHASHED_PACKAGE.splitlines())
                     warn_uninstallable = True
                 line = self._format_requirement(
                     ireq, markers.get(key_from_ireq(ireq)), hashes=hashes
@@ -227,10 +229,10 @@ class OutputWriter:
             yield ""
             yielded = True
             if has_hashes and not self.allow_unsafe:
-                yield MESSAGE_UNSAFE_PACKAGES_UNPINNED
+                yield self.linesep.join(MESSAGE_UNSAFE_PACKAGES_UNPINNED.splitlines())
                 warn_uninstallable = True
             else:
-                yield MESSAGE_UNSAFE_PACKAGES
+                yield self.linesep.join(MESSAGE_UNSAFE_PACKAGES.splitlines())
 
             for ireq in sorted(unsafe_requirements, key=self._sort_key):
                 ireq_key = key_from_ireq(ireq)
@@ -264,7 +266,7 @@ class OutputWriter:
             else:
                 log.info(line)
                 self.dst_file.write(unstyle(line).encode())
-                self.dst_file.write(os.linesep.encode())
+                self.dst_file.write(self.linesep.encode())
 
     def _format_requirement(
         self,
@@ -279,7 +281,7 @@ class OutputWriter:
             line = strip_extras(line)
 
         if not self.annotate:
-            return line
+            return self.linesep.join(line.splitlines())
 
         # Annotate what packages or reqs-ins this package is required by
         required_by = set()
@@ -308,6 +310,6 @@ class OutputWriter:
                 annotation = strip_extras(annotation)
             # 24 is one reasonable column size to use here, that we've used in the past
             lines = f"{line:24}{sep}{comment(annotation)}".splitlines()
-            line = "\n".join(ln.rstrip() for ln in lines)
+            line = self.linesep.join(ln.rstrip() for ln in lines)
 
         return line
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -160,6 +160,12 @@ def _get_default_option(option_name: str) -> Any:
     ),
 )
 @click.option(
+    "--newline",
+    type=click.Choice(("LF", "CRLF", "native", "preserve"), case_sensitive=False),
+    default="preserve",
+    help="Override the newline control characters used",
+)
+@click.option(
     "--allow-unsafe/--no-allow-unsafe",
     is_flag=True,
     default=False,
@@ -266,6 +272,7 @@ def cli(
     upgrade: bool,
     upgrade_packages: Tuple[str, ...],
     output_file: Union[LazyFile, IO[Any], None],
+    newline: str,
     allow_unsafe: bool,
     strip_extras: bool,
     generate_hashes: bool,
@@ -492,6 +499,21 @@ def cli(
 
     log.debug("")
 
+    if newline == "preserve":
+        if os.path.exists(output_file.name):
+            with open(output_file.name, "rb") as existing_target:
+                existing_txt = existing_target.read().decode()
+            if "\r\n" in existing_txt:
+                newline = "CRLF"
+            elif "\n" in existing_txt:
+                newline = "LF"
+    linesep = {
+        "native": os.linesep,
+        "LF": "\n",
+        "CRLF": "\r\n",
+        "preserve": os.linesep,
+    }[newline]
+
     ##
     # Output
     ##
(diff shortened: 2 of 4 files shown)


AndydeCleyre:
If this feature goes ahead as it is, it changes the behavior for Windows to get (more consistently) CRLF output. Consequently we'll need to adjust many compile and writer tests:

The compile tests need to either expect `os.linesep`-split lines (a bunch of splitting and re-joining, mostly), or to use the new `--newline=LF` option.

Similarly the writer tests can either expect `os.linesep` or set `linesep` to `\n`.

After starting on the former (long) way, I'm leaning toward the latter.

AndydeCleyre:
Ack, the default behavior is a complication for the header's compile command determination (e.g. `--newline=LF` will not appear on a linux system when specified, as that's "default", and `--newline=CRLF` will not appear on a Windows system).

So the tests can be finessed to make it work, or the option's default behavior can change. I would not be opposed to setting LF as default . . .

AndydeCleyre:
Hey, can any Windows user chime in regarding the best default value (either `native` which becomes `\r\n` or `LF` which is `\n`)? 

@altendky @vphilippon @jack1142

AndydeCleyre:
Thank you @jack1142! I look forward to feedback on that point from maintainers, too.

OK now I am satisfied that the test covers all relevant cases (toggling hashes, annotations, annotation styles, and including the potential warning comments), but I fear it adds too many slow tests. 

So I'm going to mark this as ready for review, but hope to get feedback or commits regarding:

- what's the best default value?
- can/should we improve the test performance here?

webknjaz:
I think this may be problematic when different team members use different OSs. Have you considered this case? IIUC, the default native value will be different for different people, right?

AndydeCleyre:
> I think this may be problematic when different team members use different OSs. Have you considered this case? IIUC, the default native value will be different for different people, right?

Yes, the "native" value means whatever Python's `os.linesep` is, so CRLF on Windows and LF on other stuff, pretty much.

Yes, I am considering the use case, but need feedback from Windows users, as I have no idea what they want here. So I have suggested either keeping default native, or setting default to LF. Do you have an inclination?

AndydeCleyre:
Also pinging @davidovich for Windows use case insight.

OK here's a poll for the default value:

- `LF`: 🚀 
- `CRLF`: 🎉 
- `native` (`CRLF` on Windows): 👀 

I may be using the wrong rocket and "tada" but close enough.

altendky:
I can see arguments for both `LF` and `native`.  I personally use `LF` everywhere, but I also try to get myself to be explicit about it even in cases where the underlying code or program does what I want anyways.

I guess I'll vote `LF` above, but it doesn't super matter to me and backwards compatibility is a relevant thing.

AndydeCleyre:
I do care about backwards compatibility, but I must note that the current (and "backward" behavior) on Windows is, I think, an inconsistent mix of line endings, depending on annotations and warnings and maybe hashes. 

This is because we create some multi line "line" strings for warnings, for example, then join/print these.

altendky:
`a_file.write("\n".join("ab\ncd"))` shouldn't be an issue.  Python universal newlines  ought to make it all the same.  Unless you have some non-universal writes or explicitly `\r\n` or such.

atugushev:
> I do care about backwards compatibility, but I must note that the current (and "backward" behavior) on Windows is, I think, an inconsistent mix of line endings, depending on annotations and warnings and maybe hashes.

@AndydeCleyre this looks like a bug. We can fix that inconsistency before the `--newline` option.

AndydeCleyre:
What would the backwards compatible behavior be, for windows?

atugushev:
> What would the backwards compatible behavior be, for windows?

I'd guess, If we fix the "inconsistent mix of line endings" then it would be `linesep.os`. What would you think?

AndydeCleyre:
> I'd guess, If we fix the "inconsistent mix of line endings" then it would be `os.linesep`. What would you think?
> 

I'd like someone using Windows to confirm. If that is indeed the case, then isn't this PR, in its current state, what is desired anyway?

webknjaz:
Let's consider two major cases separately.
I think that for the case of "a file already exists and is getting updated", the default should always be "preserve whatever was present before". That would be `--newline=preserve`, right?
And for the cases where pip-compile generates a newly-created file, it could be either LF or `os.linesep`.

AndydeCleyre:
Ok, I can add a fourth option, preserve, and make it default ("first choice default"), then if invalid, use "second choice default," and we need a consensus from Windows users on that second choice default.

Sounds good @webknjaz

AndydeCleyre:
I've added the `preserve` behavior and made it default.

I'm very interested to keep hearing more about the best Windows default, and *if* it is `LF`, I wonder if we want to have `native` at all.

graingert:
I think it should be "preserve" and fallback to the newlines in the input file

AndydeCleyre:
After the poll question is settled and implemented, I'll revert the commit changing old tests, and see via CI exactly which tests really need to be changed.

AndydeCleyre:
Thanks!

I've made the change, and replaced the adjust-existing-tests commit with a new one.

ssbarnea:
I am inclined to reject this PR as I see no real benefit from it, `LF` won the battle and there is no need to add extra complexity to the tool.

I know that this may upset some but I hope they will understand it.

jack1142:
> I am inclined to reject this PR as I see no real benefit from it, `LF` won the battle and there is no need to add extra complexity to the tool.

Reject and accept a *different PR* that would only enforce `LF` without any configuration or completely reject the whole change?

ssbarnea:
> > I am inclined to reject this PR as I see no real benefit from it, `LF` won the battle and there is no need to add extra complexity to the tool.
> 
> Reject and accept a _different PR_ that would only enforce `LF` without any configuration or completely reject the whole change?

I am not sure.as I do not want to alienate users. IMHO, enforcing use of LF on all platforms would be a benefit.

AndydeCleyre:
Duplicating my own comment from the relevant issue, for visibility and feedback here:

---

Maybe the move forward is to collpase #1491 into a single new flag, `--preserve-newlines` (or similar): 

- without it, always use `LF`
- with it, and existing newlines can be detected, use those
- with it, and no existing newlines are detected, use `LF`

What do you think?

graingert:
preserve-newlines should be the default. Having no example line separator and forcing pip-compile to guess is so rare (passed from stdin) that the choice barely matters

AndydeCleyre:
So more like an `--LF` flag is called for?

jack1142:
> preserve-newlines should be the default. Having no example line separator and forcing pip-compile to guess is so rare (passed from stdin) that the choice barely matters

Perhaps it barely matters which line ending is chosen but it does matter for it to be consistent between systems which is not currently the case either.

AndydeCleyre:
I just threw up draft #1584 based on this PR and one more commit that implements this change and all it entails:

```diff
 @click.option(
-    "--newline",
-    type=click.Choice(("LF", "CRLF", "native", "preserve"), case_sensitive=False),
-    default="preserve",
-    help="Override the newline control characters used",
+    "--force-unix-newlines",
+    is_flag=True,
+    default=False,
+    help="Always use LF newlines, rather than auto-detecting from existing files.",
 )
```

Is that close to what we need?

AndydeCleyre:
I'd like to direct anyone here to the [latest poll](https://github.com/jazzband/pip-tools/issues/1448#issuecomment-1057272727) back at the source issue, about which behavior should be the default, and what other options should be available, if any. Thanks for any feedback!

AndydeCleyre:
There aren't very many votes in [the poll](https://github.com/jazzband/pip-tools/issues/1448#issuecomment-1057272727), but the current winner is still the implementation as it is here in this PR.

ssbarnea:
Whom from those with review/merge are supporting this feature? I am still -0.5 on it as it would add unrequired complexity to the code as users can control the newlines using git. If I would support something in this area it will enforcing use of `LF` newlines regardless platform, making it predictable.

@atugushev Are you? Asking because I do not want to see @AndydeCleyre spending more and more time on a task that never gets in.

As I am against, he will need two +1 reviews to get it in.

PS. Sorry, but I am only trying to protect the codebase.

piptools/writer.py
```diff
@@ -264,7 +266,7 @@ def write(
             else:
                 log.info(line)
                 self.dst_file.write(unstyle(line).encode())
-                self.dst_file.write(os.linesep.encode())
+                self.dst_file.write(self.linesep.encode())
```
graingert, line 269: I think it would be neater to use a TextIOWrapper so the existing code can continue to use `\n`, which should sharply reduce the number of lines changed here

```python
try:
    dst_file = io.TextIOWrapper(self.dst_file, encoding="utf8", newlines=self.linesep):
    for line in self._iter_lines(...):
        dst_file.write(line)
        dst_file.write("\n")
finally:
    dst_file.detatch()
```

graingert (changes requested):

graingert (changes requested):
I'm hugely in favor of this feature, and I think using a TextIOWrapper to translate the newlines will keep the changes small enough to justify it

AndydeCleyre:
@graingert 

Thanks!

Closing this in favor of #1652 

from +140 −17
to +132 −8

Issue #1505, opened 2021-10-12
Second level dependencies missing on for local file packages on >= 6.3.1


lindycoder:
Hi, thank you for the great tool!

There seems to be a regression with 6.3.1 (and 6.4.0).  Dependencies of dependencies don't make it to the final result when using local packages.

Below is script to reproduce, problem seems to be in pip-tools as any recent tested versions of pip (21.3 and 21.2.4) produce them same problem.

Note : Tested with absolute files to remove the problematic relative files handling issues.

#### Environment Versions

1. OS Type: Reproduced in OSX and a `python:3.8-buster` docker image
1. Python version: `Python 3.8.6`
1. pip version: `pip 21.3 from /private/tmp/pip-tools/venv/lib/python3.8/site-packages/pip (python 3.8)`
1. pip-tools version: `pip-compile, version 6.4.0` && `pip-compile, version 6.3.1`

#### Steps to replicate

```
#!/usr/bin/env bash -x

mkdir /tmp/pip-tools && cd /tmp/pip-tools

mkdir -p project-a
cat > project-a/setup.py <<EOF
#!/usr/bin/env python
from setuptools import setup
setup(
    name="project-a",
    version="0.1",
    install_requires=[
      "pip-tools",
    ]
)
EOF

mkdir -p project-b
cat > project-b/setup.py <<EOF
#!/usr/bin/env python
from setuptools import setup
setup(
    name="project-b",
    version="0.1",
    install_requires=[
      "project-a",
    ]
)
EOF

cat > requirements.in <<EOF
file:///tmp/pip-tools/project-a#egg=project-a
file:///tmp/pip-tools/project-b#egg=project-b
EOF

virtualenv venv
venv/bin/pip install -U "pip==21.3"

venv/bin/pip install -U "pip-tools==6.3.0"
venv/bin/pip-compile -v requirements.in  # Has all reqs

venv/bin/pip install -U "pip-tools==6.3.1"
venv/bin/pip-compile -v requirements.in  # Has no sub reqs

venv/bin/pip install -U "pip-tools==6.4.0"
venv/bin/pip-compile -v requirements.in  # Has no sub reqs
```

#### Expected result
```
# This file is autogenerated by pip-compile with python 3.8
# To update, run:
#
#    pip-compile requirements.in
#
click==8.0.3
    # via pip-tools
pep517==0.11.0
    # via pip-tools
pip-tools==6.4.0
    # via project-a
[...]

PR #1506, opened 2021-10-13
Prevent loss of deep dependencies of local requirements [Fixes #1505]

AndydeCleyre:
I've just restored the minimum use of `copy_ireq_dependencies`, which we removed as part of #1486 for some important fixes.

@richafrank Please have a look, as I don't know what any negative consequences or better alternatives might be.

TODO:
- [ ] sanity check
- [ ] add test

##### Contributor checklist

- [ ] Provided the tests for the changes.
- [ ] Assure PR title is short, clear, and good to be included in the user-oriented changelog

##### Maintainer checklist

- [ ] Assure one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #1507, opened 2021-10-13
Track dependencies for local packages that get combined

richafrank:
Fixes https://github.com/jazzband/pip-tools/issues/1505

Partially revert "Fix a regression where multiple declarations of a dependency"

(commit d720fe605b7a0c88be8a9afba5029b1408c8295b)

This brings back copy_ireq_dependencies and combines its use with that of the
proxies-for-naming. Editable and url_requirement ireqs are not cached by the
Repository, so when dependencies are queried for them once, no dependencies
will be found later for the copy made by combine_install_requirements.

copy_ireq_dependencies links those dependencies at the time of copy.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -284,6 +284,25 @@ def test_iter_dependencies_ignores_constraints(resolver, from_line):
         next(res._iter_dependencies(ireq))
 
 
+def test_iter_dependencies_after_combine_install_requirements(
+    pypi_repository, base_resolver, make_package, from_line
+):
+    res = base_resolver([], repository=pypi_repository)
+
+    sub_deps = ["click"]
+    package_a = make_package("package-a", install_requires=sub_deps)
+    package_b = make_package("package-b", install_requires=["package-a"])
+
+    local_package_a = from_line(path_to_url(package_a))
+    assert [dep.name for dep in res._iter_dependencies(local_package_a)] == sub_deps
+
+    package_a_from_b = from_line("package-a", comes_from=path_to_url(package_b))
+    combined = combine_install_requirements(
+        pypi_repository, [local_package_a, package_a_from_b]
+    )
+    assert [dep.name for dep in res._iter_dependencies(combined)] == sub_deps
+
+
 def test_combine_install_requirements(repository, from_line):
     celery30 = from_line("celery>3.0", comes_from="-r requirements.in")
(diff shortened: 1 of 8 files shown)


AndydeCleyre:
Thank you!

Do you think it would be appropriate to combine this fix with [this one](https://github.com/jazzband/pip-tools/pull/1385/files#diff-75c898902c2661a44987829ad609250f598417159bba40f90e248c7a131cc8ad) (or [an alternate form](https://github.com/jazzband/pip-tools/pull/1506/files#diff-75c898902c2661a44987829ad609250f598417159bba40f90e248c7a131cc8ad))?

tests/test_resolver.py
```diff
@@ -284,6 +284,25 @@ def test_iter_dependencies_ignores_constraints(resolver, from_line):
         next(res._iter_dependencies(ireq))
 
 
+def test_iter_dependencies_after_combine_install_requirements(
```
atugushev, line 287: Could you add an integration test to `test_cli_compile.py` instead of unit test, in case we refactor out resolver internals later (maybe)?

richafrank: Probably as expected, it'll take me a bit more time to formulate that, but I can do that. I'd say let's keep the unit test too, since (at least with the current resolver) it should be easier (clearer/faster) to debug upon failure?

richafrank: @atugushev Done!

richafrank:
> Do you think it would be appropriate to combine this fix with this one (or an alternate form)?

@AndydeCleyre I cherry-picked your version of the fix. I also tried to take the test from the [original PR](https://github.com/jazzband/pip-tools/pull/1385/files#diff-4cbff640f1fc708fb7f2d3e79ac97f24cfe22063bad7ce6f67ab97b5b5171a0b), but between updating the test to fail without the fix (very important!) and adding `make_package`, I ended up re-writing it.

If you all think this PR is too bloated now, I can split the last two commits back out.

AndydeCleyre:
I'm going to close this in favor of focusing related attention on your own #1519. Please let me know if I'm making a mistake. Thanks!

richafrank:
Works for me! (I had left it open in case we wanted it as an incremental improvement, while progressing on the bigger change.)

PR #1509, opened 2021-10-14
Support Exclude Package with custom unsafe packages

hmc-cs-mdrissi:
https://github.com/jazzband/pip-tools/issues/333 Closes this issue. Custom unsafe packages allows you to exclude packages. I named the option --unsafe-packages as all logic is customizing UNSAFE_PACKAGES (motivated by this [comment](https://github.com/jazzband/pip-tools/issues/333#issuecomment-917643576)), but if people would prefer I call it exclude that works. The one difference in the two names is whether original UNSAFE_PACKAGES remain or are gone. 

As I'd ideally like setuptools/pip/etc to be in my requirements.txt file, but want to exclude other packages I prefer some way to replace the existing unsafe packages entirely. This is not a big deal to me though so if people would prefer semantic of it only adds packages to unsafe list I can do that. That is less flexible though.

I also added a flag to control whether recursive dependencies of unsafe only packages are treated as unsafe or not. The current logic for setuptools/pip is recursive dependencies only from unsafe packages are treated unsafe, but my use case for exclude is I only want to remove the exact packages I list.

Multiple occurrences of the flag are allowed. It can be used like,

pip-compile --unsafe-package foo --unsafe-package bar ...

I added two test cases to cover recursive flag being on/off.

@nvie @ssbarnea 

##### Contributor checklist

##### Maintainer checklist


piptools/_compat/pip_compat.py
```diff
@@ -3,18 +3,18 @@
...
 from pip._internal.req.constructors import install_req_from_parsed_requirement
 from pip._vendor.packaging.version import parse as parse_version
+from pip._vendor.requests.sessions import Session
 
 PIP_VERSION = tuple(map(int, parse_version(pip.__version__).base_version.split(".")))
 
 
 def parse_requirements(
     filename: str,
-    session: PipSession,
```
hmc-cs-mdrissi, line 17: Oh PipSession/Session is from pyright marking it as a type error. I can undo, but as I was working on this did want to mention pyright detects several type errors that mypy misses. I'd estimate like 1-10 type errors per file.

I think most type errors are inconsistencies in some of the pip types used. Like InstallRequirement vs InstallCandidate in find_best_match. The return type is inconsistent with actual reveal type value and while I'm guessing the two are intended to be compatible they are not currently. Here PipSession is used but the function called works with a more generic Session and one of the callers had more generic type.

atugushev (dismissed):
Generally, it looks good. I have a concern though:

piptools/scripts/compile.py
```diff
@@ -236,6 +236,20 @@ def _get_default_option(option_name: str) -> Any:
...
+    help="List of packages to consider unsafe. Replaces "
+    f"{', '.join(sorted(UNSAFE_PACKAGES))}. Can specify multiple times.",
+)
+@click.option(
+    "--allow-unsafe-recursive/--no-allow-unsafe-recursive",
+    is_flag=True,
+    default=False,
+    help="Determines whether dependencies that solely belong to unsafe packages "
+    "should be treated as unsafe. Default is false. This only covers direct dependencies. "
+    "Dependencies of dependencies of unsafe packages will not be marked unsafe.",
```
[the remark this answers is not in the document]
ssbarnea: @atugushev Can you please remove your requested changes? I do not think that this is a good enough reason for blocking this change, especially as this behavior was already existing. As part of the review process we should try to avoid scope creeping on original PR.

IMHO, that change is fixing one problem and the other aspect is unrelated.

piptools/scripts/compile.py
```diff
@@ -462,6 +478,8 @@ def cli(
             cache=DependencyCache(cache_dir),
             clear_caches=rebuild,
             allow_unsafe=allow_unsafe,
+            unsafe_packages=set(unsafe_packages),
```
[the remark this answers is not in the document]
ssbarnea: Closing as resolved.

di:
I see this is in the [6.5.0 milestone](https://github.com/jazzband/pip-tools/milestone/53), is this blocking a 6.5.0 release?

ssbarnea:
I would really want to see this feature in ASAP as lack of it seriously limited tool use in several projects.

piptools/scripts/compile.py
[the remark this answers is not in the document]
atugushev: Removed 👍

atugushev:
It does not block the 6.5.0. Let's release it later.

hmc-cs-mdrissi:
Is thee anything that's recommended I should do for this PR? Should I deal with dependencies of dependencies logic or should I leave it as out of scope? If there's another commit I can do to help merge this that would be helpful. It'd be nice to use this without fork.

piptools/scripts/compile.py
```diff
@@ -236,6 +236,20 @@ def _get_default_option(option_name: str) -> Any:
     default=True,
     help="Add options to generated file",
 )
+@click.option(
+    "--unsafe-packages",
+    multiple=True,
+    help="List of packages to consider unsafe. Replaces "
```
AndydeCleyre, line 242: I think the help text here is a bit misleading regarding the required syntax, suggesting that maybe the user ought to provide a CSV. I would like to bring it more in line with the descriptions for our other options which specify `multiple=True`.

piptools/scripts/compile.py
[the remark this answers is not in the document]
AndydeCleyre: It's hard for me to be sure about details for this feature, as I haven't seen many use cases up close, but I think "only want to remove the exact packages I list" is *probably* the most useful mode here. I wouldn't be opposed to dropping the recursive flag....

But if it stays, and it only covers one level of deps, I would prefer an alternative name to "recursive," which implies more depth.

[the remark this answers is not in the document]
hmc-cs-mdrissi: My personal use is I have an exact list I want to remove and there dependencies I don't intend to remove. The recursive flag was mainly added as current code partly looked at recursive dependencies. I would be happy to remove recursive flag and leave that as an out of scope feature for future if anyone has a use case for it.

piptools/scripts/compile.py
hmc-cs-mdrissi: Done. I adjusted the help to be more similar to --upgrade-package.

piptools/scripts/compile.py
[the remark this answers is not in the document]
hmc-cs-mdrissi: I've added a commit that drops the recursive flag. I'm happy to revert if there's a preference to keep recursive logic.

tests/conftest.py
```diff
@@ -234,6 +234,11 @@ def pip_with_index_conf(make_pip_conf):
     )
 
 
+@pytest.fixture(autouse=True)
+def pip_with_empty_conf(make_pip_conf):
```
hmc-cs-mdrissi, line 238: This was added because many unit tests make assumptions about default pip.conf. Specifically if you run unit tests on master with a pip.conf that has an extra-index-url the url will get added to output of many tests causing them to fail. This makes all tests that don't specify pip.conf file use empty one and avoids needing to modify your global pip.conf for tests.

tests/test_cache.py
```diff
@@ -116,6 +115,3 @@ def test_reverse_dependencies(from_line, tmpdir):
         "bottom": {"middle", "top"},
         "bonus": {"top"},
     }
-
-    # Clean up our temp directory
-    rmtree(tmpdir)
```
hmc-cs-mdrissi, line 121: pytest deletes tempdirs already. The extra delete here causes one test to fail after making all tests default to use empty pip.conf. The delete here causes the other fixture to fail [here](https://github.com/jazzband/pip-tools/blob/73a8a6e4c91e2f9a90642ed46294fbe94b2cf419/tests/conftest.py#L208).

piptools/scripts/compile.py
AndydeCleyre: Thanks! Maybe the user-facing `--unsafe-packages` can be renamed as singular, `--unsafe-package`?

I'm not tied to the "unsafe" term, but either way I think it should be singular.

hmc-cs-mdrissi: Done. I'm neutral on name and other names like --exclude-package are fine. Only specific pro to keep unsafe-package is it replaces existing unsafe package list (default excludes) of pip/setuptools.

ssbarnea:
Any chance to resolve current conflicts? I would really want to see this feature in.

hmc-cs-mdrissi:
I've resolved conflicts and I think tests should pass now with 2020 resolver too.

ssbarnea:
@atugushev Please have a look at this and if ok-ish lets get it in before it needs other rebases.

piptools/scripts/compile.py
```diff
@@ -245,6 +245,12 @@ def _get_default_option(option_name: str) -> Any:
     default=True,
     help="Add options to generated file",
 )
+@click.option(
+    "--unsafe-package",
+    multiple=True,
+    help="Specify a package to consider unsafe; may be used more than once. "
+    f"Replaces default unsafe packages, {', '.join(sorted(UNSAFE_PACKAGES))}",
```
atugushev, line 252:
```suggestion
    f"Replaces default unsafe packages: {', '.join(sorted(UNSAFE_PACKAGES))}",
```

Shouldn't there be a colon?

atugushev (approved):
Otherwise LGTM 👍🏻

ssbarnea (approved):

atugushev:
@hmc-cs-mdrissi thanks for cleaning up `reverse_dependencies()`. Could you also consider this https://github.com/hmc-cs-mdrissi/pip-tools/pull/1?

ssbarnea:
@atugushev I would merge it as is and let him fix after. Because that the first contribution even CI is not running until one of us is approving it.

atugushev:
> I would merge it as is and let him fix after. Because that the first contribution even CI is not running until one of us is approving it.

@ssbarnea sure! Feel free to.

atugushev (approved):

ssbarnea:
@atugushev That pip cache step is hilarous almost two hours and still running https://github.com/jazzband/pip-tools/runs/7379969275?check_suite_focus=true

Issue #1511, opened 2021-10-16
can pip-compile handle multiple extras of the same package properly


zyxue:
#### Environment Versions

1. OS Type: Mac OS or Linux
1. Python version: 3.8.2
1. pip version: 21.3
1. pip-tools version: 6.4.1.dev1+geb9162f

#### Steps to replicate

1. put the following two top-level deps in `requirements.in`
```
cat requirements.in 
ray[default]==1.4.0
tune-sklearn==0.4.0
```
2. `pip-compile requirements.in --rebuild`
3. the output is wrong

My debugging leads to the following:

[`tune-sklearn`](https://github.com/ray-project/tune-sklearn/) depends on [`ray[tune]`](https://github.com/ray-project/ray) as specified in its [setup.py](https://github.com/ray-project/tune-sklearn/blob/master/setup.py#L29), then pip-compile [groups constraints](https://github.com/jazzband/pip-tools/blob/eb9162fdbd6c289ea0ea41699c46fddb3715b242/piptools/resolver.py#L81) into `ray[default,tune]`, which leads to zero dependencies

```
jq .dependencies.ray ~/Library/Caches/pip-tools/depcache-cp3.8.json
{
  "1.4.0[default,tune]": [],
  "1.4.0[default]": [
    "aiohttp",
    "aiohttp-cors",
    "aioredis",
    "click>=7.0",
    "colorama",
    "colorful; extra == \"default\"",
    "filelock",
    "gpustat",
    "grpcio>=1.28.1",
    "jsonschema",
    "msgpack<2.0.0,>=1.0.0",
    "numpy>=1.16",
    "opencensus",
    "prometheus-client>=0.7.1",
    "protobuf>=3.15.3",
    "py-spy>=0.2.0",
    "pydantic>=1.8",
    "pyyaml",
    "redis>=3.5.0",
    "requests"
  ]
}
```

#### Expected result

`"1.4.0[default,tune]"` should have the deps of both `default` and `tune`.

#### Actual result

`"1.4.0[default,tune]"` has zero deps.

atugushev:
Hello @zyxue ,

Thanks for the report. That's because `pip-tools` doesn't yet have a proper dependency resolver. See #693 and the other [related issues](https://github.com/jazzband/pip-tools/issues?q=is%3Aissue+is%3Aopen+label%3Aresolver) for details.

zyxue:
[...]

PR #1512, opened 2021-10-16
Re resolve combined ireq

zyxue:
An idea for fixing https://github.com/jazzband/pip-tools/issues/1511

It will affects this line https://github.com/pypa/pip/blob/fc6d7778a5ee5616cc517ba8b9d41ddf5f204e5c/src/pip/_internal/resolution/legacy/resolver.py#L373 and leads to resolution of the dependencies of `ray[default,tune]==1.4.0` properly

Currently, this line evaluated to `True` because `combined_ireq` copied the `.prepared=True` from `ray[default]==1.4.0`, which was already prepared before.

### DIFF
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -341,7 +341,7 @@ def test_combine_install_requirements_extras_no_req(
     )
 
 
-def test_combine_install_requirements_with_multiple_extras_reset_prepared(
+def test_combine_install_requirements_for_one_package_with_multiple_extras_reset_prepared(
     repository, from_line
 ):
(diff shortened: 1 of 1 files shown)


atugushev (commented):
Neat! Could you add an integration test to `test_cli_compile.py` base on the issue details? You could use `make_package` and `make_wheel`, see `test_duplicate_reqs_combined` for example.

richafrank:
I forgot to update here: https://github.com/jazzband/pip-tools/pull/1519 is intended to resolve the conflict between these two.

AndydeCleyre:
I'm going to close this in favor of focusing related attention on #1519. Thank you for this, and please let me know if something's amiss!

PR #1531, opened 2021-11-09
Disregard existing pins in the output which can't be found by pip, rather than failing to compile

### LINKED ISSUE #1530: Unnecessary hard failure of pip-compile when a required package is already present in output with an unavailable version
This was encountered in #1529 

When a needed package is already pinned in the output file,
but has an invalid or at least unavailable version there,
the compilation will fail.

#### Environment Versions

1. Linux
1. Python version: `3.9.5`
1. pip version: `21.3.1`
1. pip-tools version: `6.4.0`

#### Steps to replicate

```console
$ echo requests >reqs.in
$ pip-compile reqs.in
$ sed -i 's/==2021/==2041/' reqs.txt
$ pip-compile reqs.in
```

#### Expected result

I'd expect that the pre-existing content of the output file is considered "low priority" -- that if something there doesn't fit the needs of the input files, it will be discarded. In other words, it should ultimately behave in this case as if `certifi` isn't in the output file to begin with, and so generate a valid output file again.

#### Actual result

```
  ERROR: Could not find a version that satisfies the requirement certifi==2041.10.8 (from requests==2.26.0->-r reqs.in (line 1)) (from versions: 0.0.1, 0.0.2, 0.0.3, 0.0.4, 0.0.5, 0.0.6, 0.0.7, 0.0.8, 1.0.0, 1.0.1, 14.5.14, 2015.4.28, 2015.9.6, 2015.9.6.1, 2015.9.6.2, 2015.11.20, 2015.11.20.1, 2016.2.28, 2016.8.2, 2016.8.8, 2016.8.31, 2016.9.26, 2017.1.23, 2017.4.17, 2017.7.27, 2017.7.27.1, 2017.11.5, 2018.1.18, 2018.4.16, 2018.8.13, 2018.8.24, 2018.10.15, 2018.11.29, 2019.3.9, 2019.6.16, 2019.9.11, 2019.11.28, 2020.4.5, 2020.4.5.1, 2020.4.5.2, 2020.6.20, 2020.11.8, 2020.12.5, 2021.5.30, 2021.10.8)
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/click/core.py", line 1128, in __call__
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/click/core.py", line 1395, in invoke
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/click/core.py", line 754, in invoke
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/click/decorators.py", line 26, in new_func
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/piptools/scripts/compile.py", line 466, in cli
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/piptools/resolver.py", line 175, in resolve
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/piptools/resolver.py", line 319, in _resolve_one_round
    their_constraints.extend(self._iter_dependencies(best_match))
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/piptools/resolver.py", line 428, in _iter_dependencies
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/piptools/repositories/local.py", line 79, in get_dependencies
    return self.repository.get_dependencies(ireq)
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/piptools/repositories/pypi.py", line 236, in get_dependencies
    self._dependencies_cache[ireq] = self.resolve_reqs(
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/piptools/repositories/pypi.py", line 199, in resolve_reqs
    results = resolver._resolve_one(reqset, ireq)
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/pip/_internal/resolution/legacy/resolver.py", line 379, in _resolve_one
    dist = self._get_dist_for(req_to_install)
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/pip/_internal/resolution/legacy/resolver.py", line 331, in _get_dist_for
    self._populate_link(req)
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/pip/_internal/resolution/legacy/resolver.py", line 300, in _populate_link
    req.link = self._find_requirement_link(req)
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/pip/_internal/resolution/legacy/resolver.py", line 266, in _find_requirement_link
    best_candidate = self.finder.find_requirement(req, upgrade)
  File "/home/andy/.local/share/venvs/42fa537b433f422253f0da1997538ecd/venv/lib/python3.9/site-packages/pip/_internal/index/package_finder.py", line 909, in find_requirement
    raise DistributionNotFound(
pip._internal.exceptions.DistributionNotFound: No matching distribution found for certifi==2041.10.8 (from requests==2.26.0->-r reqs.in (line 1))
```

AndydeCleyre:
Before running `pip-compile`, the output file may contain a pinned requirement which can't be found in PyPI or whichever repo (e.g. a revoked release). This change aims to disregard these un-find-able pins, rather than cause the compile operation to fail.

Fixes #1530

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/repositories/local.py
+++ b/piptools/repositories/local.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import optparse
-from contextlib import contextmanager, suppress
+from contextlib import contextmanager
 from typing import Iterator, Mapping, cast
 
 from pip._internal.commands.install import InstallCommand
@@ -14,6 +14,7 @@ from pip._internal.utils.hashes import FAVORITE_HASH
 from piptools.utils import key_from_ireq
 
 from ..exceptions import NoCandidateFound
+from ..logging import log
 from .base import BaseRepository
 from .pypi import PyPIRepository
 
@@ -79,8 +80,10 @@ class LocalRequirementsRepository(BaseRepository):
         key = key_from_ireq(ireq)
         existing_pin = self.existing_pins.get(key)
         if existing_pin and ireq_satisfied_by_existing_pin(ireq, existing_pin):
-            with suppress(NoCandidateFound):
+            try:
                 return self.repository.find_best_match(existing_pin, prereleases)
+            except NoCandidateFound:
+                log.warning(f"Discarding {existing_pin} to proceed the resolution")
         return self.repository.find_best_match(ireq, prereleases)
 
     def get_dependencies(self, ireq: InstallRequirement) -> set[InstallRequirement]:
--- a/piptools/repositories/local.py
+++ b/piptools/repositories/local.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import optparse
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from typing import Iterator, Mapping, cast
 
 from pip._internal.commands.install import InstallCommand
@@ -11,8 +11,9 @@ from pip._internal.network.session import PipSession
 from pip._internal.req import InstallRequirement
 from pip._internal.utils.hashes import FAVORITE_HASH
 
-from piptools.utils import as_tuple, key_from_ireq, make_install_requirement
+from piptools.utils import key_from_ireq
 
+from ..exceptions import NoCandidateFound
 from .base import BaseRepository
 from .pypi import PyPIRepository
 
@@ -78,10 +79,9 @@ class LocalRequirementsRepository(BaseRepository):
         key = key_from_ireq(ireq)
         existing_pin = self.existing_pins.get(key)
         if existing_pin and ireq_satisfied_by_existing_pin(ireq, existing_pin):
-            project, version, _ = as_tuple(existing_pin)
-            return make_install_requirement(project, version, ireq)
-        else:
-            return self.repository.find_best_match(ireq, prereleases)
+            with suppress(NoCandidateFound):
+                return self.repository.find_best_match(existing_pin, prereleases)
+        return self.repository.find_best_match(ireq, prereleases)
 
     def get_dependencies(self, ireq: InstallRequirement) -> set[InstallRequirement]:
         return self.repository.get_dependencies(ireq)
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -1971,6 +1971,12 @@ def test_ignore_compiled_unavailable_version(pip_conf, runner):
     assert "small-fake-a==" in out.stderr
     assert "small-fake-a==9999" not in out.stderr.splitlines()
 
+    assert (
+        "Discarding small-fake-a==9999 "
+        "(from -r requirements.txt (line 1)) "
+        "to proceed the resolution"
+    ) in out.stderr
+
 
 def test_prefer_binary_dist(
     pip_conf, make_package, make_sdist, make_wheel, tmpdir, runner
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -1958,6 +1958,21 @@ def test_preserve_compiled_prerelease_version(pip_conf, runner):
     assert "small-fake-a==0.3b1" in out.stderr.splitlines()
 
 
+@pytest.mark.xfail
+def test_ignore_compiled_unavailable_version(pip_conf, runner):
+    with open("requirements.in", "w") as req_in:
+        req_in.write("small-fake-a")
+
+    with open("requirements.txt", "w") as req_txt:
+        req_txt.write("small-fake-a==9999")
+
+    out = runner.invoke(cli, ["--no-annotate", "--no-header"])
+
+    assert out.exit_code == 0, out
+    assert "small-fake-a==" in out.stderr
+    assert "small-fake-a==9999" not in out.stderr.splitlines()
+
+
 def test_prefer_binary_dist(
     pip_conf, make_package, make_sdist, make_wheel, tmpdir, runner
 ):
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -1958,7 +1958,6 @@ def test_preserve_compiled_prerelease_version(pip_conf, runner):
     assert "small-fake-a==0.3b1" in out.stderr.splitlines()
 
 
-@pytest.mark.xfail
 def test_ignore_compiled_unavailable_version(pip_conf, runner):
     with open("requirements.in", "w") as req_in:
         req_in.write("small-fake-a")


AndydeCleyre:
The function could be restyled smaller, but it may be less obvious to read:

```diff
diff --git a/piptools/repositories/local.py b/piptools/repositories/local.py
index 07dd070..75a594e 100644
--- a/piptools/repositories/local.py
+++ b/piptools/repositories/local.py
@@ -1,5 +1,5 @@
 import optparse
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from typing import Iterator, Mapping, Optional, Set, cast
 
 from pip._internal.index.package_finder import PackageFinder
@@ -71,12 +71,9 @@ class LocalRequirementsRepository(BaseRepository):
         key = key_from_ireq(ireq)
         existing_pin = self.existing_pins.get(key)
         if existing_pin and ireq_satisfied_by_existing_pin(ireq, existing_pin):
-            try:
+            with suppress(NoCandidateFound):
                 return self.repository.find_best_match(existing_pin, prereleases)
-            except NoCandidateFound:
-                return self.repository.find_best_match(ireq, prereleases)
-        else:
-            return self.repository.find_best_match(ireq, prereleases)
+        return self.repository.find_best_match(ireq, prereleases)
 
     def get_dependencies(self, ireq: InstallRequirement) -> Set[InstallRequirement]:
         return self.repository.get_dependencies(ireq)
```

piptools/repositories/local.py
```diff
@@ -72,12 +72,9 @@ def find_best_match(
         existing_pin = self.existing_pins.get(key)
         if existing_pin and ireq_satisfied_by_existing_pin(ireq, existing_pin):
             try:
-                self.repository.find_best_match(existing_pin, prereleases)
+                return self.repository.find_best_match(existing_pin, prereleases)
```
AndydeCleyre, line 75: Oh, I'm now doubting this part, as maybe info from the original ireq gets lost along the way.

piptools/repositories/local.py
```diff
@@ -76,10 +77,9 @@ def find_best_match(
         key = key_from_ireq(ireq)
         existing_pin = self.existing_pins.get(key)
         if existing_pin and ireq_satisfied_by_existing_pin(ireq, existing_pin):
-            project, version, _ = as_tuple(existing_pin)
-            return make_install_requirement(project, version, ireq)
-        else:
-            return self.repository.find_best_match(ireq, prereleases)
+            with suppress(NoCandidateFound):
+                return self.repository.find_best_match(existing_pin, prereleases)
```
atugushev, line 81: BTW we can also suppress yanked releases if `best_candidate.link.is_yanked is True`. What do you think?

atugushev: Also I think we should print a warning here that existing pin is not valid/exists.

atugushev, line 81: Also note this would probably slow down the re-compile process because it would always hit index_urls and find_links.

AndydeCleyre: Yes, thanks, we shouldn't use yanked releases -- I didn't realize `find_best_match` would return them. I tested with a found yanked release (input: `gssapi`; output: `gssapi==1.6.6`), and unfortunately in this case the `link` attribute of the `InstallRequirement` candidate, as returned by `find_best_match`, is `None`, so we can't immediately check `candidate.link.is_yanked`.

It's been just over a year since I submitted this PR, and I don't readily recall how we mutate `InstallRequirements` to fill the attributes with useful info, if that's what's needed.

I also can't find any existing explicit mention/handling/testing of yanked releases in the codebase.

And yes, once we figure out the yanked business, I'll add a logged warning about discarded matches, thanks!

AndydeCleyre: Hmm, well it seems to me that in order to do the "right thing" here (avoid a hard failure on bad already-compiled versions), we *need* to consult the index about the already-compiled versions.

atugushev: The question is what's more important "speed" or "right thing, but in rare case"?

AndydeCleyre: I'm now doubting that we should reject already-pinned yanked versions. Despite a package being yanked (not deleted), a downstream package may still need that version, and understand the risks, or want to take time to handle it mindfully.

Another yanked example is `attrs` `21.1.0`, and when that is already pinned during a non-upgrading compile, we get a nice informative warning:

```
WARNING: The candidate selected for download or install is a yanked version: 'attrs' candidate (version 21.1.0 at https://files.pythonhosted.org/packages/2b/00/7f74fb9608daf3626aad5de91181026a5107b997ce2788eabffeb5d41cb1/attrs-21.1.0-py2.py3-none-any.whl (from https://pypi.org/simple/attrs/) (requires-python:>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*))
  Reason for being yanked: Installable but not importable on Python 3.4.
```

AndydeCleyre: As an aside, do you know why running `tox -e checkqa` locally always has isort move `from build import BuildBackendException` in `piptools/scripts/compile.py`, but in CI it moves back?

AndydeCleyre: Exactly the problem, thanks!

AndydeCleyre: I copied the warning that you're already using in the backtracking resolver.

AndydeCleyre: About yanked releases, I'm now thinking:

- existing pins which are yanked should indeed be discarded
- yanked releases must still be valid candidates, allowing for users to constrain their way into using a yanked release if they *need* to
- despite being "valid candidates," they must not prevail over a non-yanked alternative release satisfying the constraints

atugushev: I did some simple benchmark.

Before the fix:
```bash
$ echo web3 > requirements.in

# first run
$ time pip-compile
real	0m18.898s
user	0m11.924s
sys	0m0.993s

# second run
$ time pip-compile
real	0m0.589s
user	0m0.517s
sys	0m0.072s
```


After the fix:
```bash
$ echo web3 > requirements.in

# first run
$ time pip-compile
real	0m17.062s
user	0m12.388s
sys	0m0.970s

# second run
$ time pip-compile
real	0m3.631s
user	0m2.583s
sys	0m0.120s
```

As I expected re-compile would be slower after the fix (6 times slower on my benchmark) due to the hitting to the simple API for every compiled package.

I understand the intention to do the right thing, but the issue doesn't seem critical and slows down pip-compile on re-runs, which might not be good. Also, the legacy resolver will be removed soon, and spending much energy on it might not be worth it.

What do you think if we only merge the test which I find pretty helpful?

AndydeCleyre: Thanks for running the numbers!

> What do you think if we only merge the test which I find pretty helpful?

Sure, sounds good to me.

atugushev: Thank you! Would you like to open a separate PR?

AndydeCleyre: Sure, let's continue at #1774

Issue #1277, opened 2020-12-18
Unexpected incompatible dependencies


astrojuanlu:
#### Environment Versions

1. OS Type: Linux Mint 20
1. Python version: `$ python -V`: Python 3.8.5
1. pip version:  `$ pip --version`: pip 20.3.3 from /home/juanlu/.pyenv/versions/_nbs38/lib/python3.8/site-packages/pip (python 3.8)
1. pip-tools version: `$ pip-compile --version`: pip-compile, version 5.4.0

#### Steps to replicate

1. `echo 'jupyterlab==3.0.0rc13' > requirements.in`
2. `pip-compile`
3. 💥 

#### Expected result

Set of constraints that work:

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile
#
argon2-cffi==20.1.0       # via notebook
async-generator==1.10     # via nbclient
attrs==20.3.0             # via jsonschema
babel==2.9.0              # via jupyterlab-server
backcall==0.2.0           # via ipython
bleach==3.2.1             # via nbconvert
certifi==2020.12.5        # via requests
cffi==1.14.4              # via argon2-cffi
chardet==4.0.0            # via requests
decorator==4.4.2          # via ipython
defusedxml==0.6.0         # via nbconvert
entrypoints==0.3          # via nbconvert
idna==2.10                # via requests
ipykernel==5.4.2          # via notebook
ipython-genutils==0.2.0   # via jupyter-server, nbformat, notebook, traitlets
ipython==7.19.0           # via ipykernel, jupyterlab
jedi==0.17.2              # via ipython
jinja2==2.11.2            # via jupyter-server, jupyterlab, jupyterlab-server, nbconvert, notebook
json5==0.9.5              # via jupyterlab-server
jsonschema==3.2.0         # via jupyterlab-server, nbformat
jupyter-client==6.1.7     # via ipykernel, jupyter-server, nbclient, notebook
jupyter-core==4.7.0       # via jupyter-client, jupyter-server, jupyterlab, nbconvert, nbformat, notebook
jupyter-server==1.0.11    # via jupyterlab, jupyterlab-server, nbclassic
jupyterlab-pygments==0.1.2  # via nbconvert
jupyterlab-server==2.0.0rc7  # via -r requirements.in, jupyterlab
jupyterlab==3.0.0rc13     # via -r requirements.in
markupsafe==1.1.1         # via jinja2
mistune==0.8.4            # via nbconvert
[...]

PR #1539, opened 2021-11-22
Add support for pip's 2020 dependency resolver

atugushev:
### What's new?

Added new option `--resolver [backtracking|legacy]` to `pip-compile` (default is `legacy`). 

### How to use?

To enable [2020 dependency resolver](https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020) run `pip-compile --resolver=backtracking`.

### Backtracking resolver example

```bash
$ echo "oslo.utils==1.4.0" | pip-compile - --resolver=backtracking --allow-unsafe --annotation-style=line -qo-
#
# This file is autogenerated by pip-compile with python 3.8
# To update, run:
#
#    pip-compile --allow-unsafe --annotation-style=line --output-file=- --resolver=backtracking -
#
babel==2.10.3             # via oslo-i18n, oslo-utils
iso8601==1.0.2            # via oslo-utils
netaddr==0.8.0            # via oslo-utils
netifaces==0.11.0         # via oslo-utils
oslo-i18n==2.1.0          # via oslo-utils
oslo-utils==1.4.0         # via -r -
pbr==0.11.1               # via oslo-i18n, oslo-utils
pytz==2022.1              # via babel
six==1.16.0               # via oslo-i18n, oslo-utils

pip==22.1.2               # via pbr
```

### Legacy resolver example

```bash
$ echo "oslo.utils==1.4.0" | pip-compile - --resolver=legacy --allow-unsafe -qo-
Could not find a version that matches pbr!=0.7,!=2.1.0,<1.0,>=0.6,>=2.0.0 (from oslo.utils==1.4.0->-r -)
Tried: 0.5.2.5.g5b3e942, 0.5.0, 0.5.1, 0.5.2, 0.5.4, 0.5.5, 0.5.6, 0.5.7, 0.5.8, 0.5.10, 0.5.11, 0.5.12, 0.5.13, 0.5.14, 0.5.15, 0.5.16, 0.5.17, 0.5.18, 0.5.19, 0.5.20, 0.5.21, 0.5.22, 0.5.23, 0.6, 0.7.0, 0.8.0, 0.8.1, 0.8.2, 0.9.0, 0.9.0, 0.10.0, 0.10.0, 0.10.1, 0.10.1, 0.10.2, 0.10.2, 0.10.3, 0.10.3, 0.10.4, 0.10.4, 0.10.5, 0.10.5, 0.10.6, 0.10.6, 0.10.7, 0.10.7, 0.10.8, 0.10.8, 0.11.0, 0.11.0, 0.11.1, 0.11.1, 1.0.0, 1.0.0, 1.0.1, 1.0.1, 1.1.0, 1.1.0, 1.1.1, 1.1.1, 1.2.0, 1.2.0, 1.3.0, 1.3.0, 1.4.0, 1.4.0, 1.5.0, 1.5.0, 1.6.0, 1.6.0, 1.7.0, 1.7.0, 1.8.0, 1.8.0, 1.8.1, 1.8.1, 1.9.0, 1.9.0, 1.9.1, 1.9.1, 1.10.0, 1.10.0, 2.0.0, 2.0.0, 2.1.0, 2.1.0, 3.0.0, 3.0.0, 3.0.1, 3.0.1, 3.1.0, 3.1.0, 3.1.1, 3.1.1, 4.0.0, 4.0.0, 4.0.1, 4.0.1, 4.0.2, 4.0.2, 4.0.3, 4.0.3, 4.0.4, 4.0.4, 4.1.0, 4.1.0, 4.1.1, 4.1.1, 4.2.0, 4.2.0, 4.3.0, 4.3.0, 5.0.0, 5.0.0, 5.1.0, 5.1.0, 5.1.1, 5.1.1, 5.1.2, 5.1.2, 5.1.3, 5.1.3, 5.2.0, 5.2.0, 5.2.1, 5.2.1, 5.3.0, 5.3.0, 5.3.1, 5.3.1, 5.4.0, 5.4.0, 5.4.1, 5.4.1, 5.4.2, 5.4.2, 5.4.3, 5.4.3, 5.4.4, 5.4.4, 5.4.5, 5.4.5, 5.5.0, 5.5.0, 5.5.1, 5.5.1, 5.6.0, 5.6.0, 5.7.0, 5.7.0, 5.8.0, 5.8.0
There are incompatible versions in the resolved dependencies:
  pbr!=0.7,<1.0,>=0.6 (from oslo.utils==1.4.0->-r -)
  pbr!=2.1.0,>=2.0.0 (from oslo.i18n==5.1.0->oslo.utils==1.4.0->-r -)
```
 
##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/repositories/base.py
+++ b/piptools/repositories/base.py
@@ -3,6 +3,7 @@ from abc import ABCMeta, abstractmethod
 from contextlib import contextmanager
 from typing import Iterator, Optional, Set
 
+from pip._internal.commands.install import InstallCommand
 from pip._internal.index.package_finder import PackageFinder
 from pip._internal.models.index import PyPI
 from pip._internal.network.session import PipSession
@@ -61,3 +62,8 @@ class BaseRepository(metaclass=ABCMeta):
     @abstractmethod
     def finder(self) -> PackageFinder:
         """Returns a package finder to interact with simple repository API (PEP 503)"""
+
+    @property
+    @abstractmethod
+    def command(self) -> InstallCommand:
+        """Return an install command."""
--- a/piptools/repositories/local.py
+++ b/piptools/repositories/local.py
@@ -2,6 +2,7 @@ import optparse
 from contextlib import contextmanager
 from typing import Iterator, Mapping, Optional, Set, cast
 
+from pip._internal.commands.install import InstallCommand
 from pip._internal.index.package_finder import PackageFinder
 from pip._internal.models.candidate import InstallationCandidate
 from pip._internal.req import InstallRequirement
@@ -60,6 +61,11 @@ class LocalRequirementsRepository(BaseRepository):
     @property
     def session(self) -> Session:
         return self.repository.session
+
+    @property
+    def command(self) -> InstallCommand:
+        """Return an install command instance."""
+        return self.repository.command
 
     def clear_caches(self) -> None:
         self.repository.clear_caches()
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -73,10 +73,9 @@ class PyPIRepository(BaseRepository):
         # Use pip's parser for pip.conf management and defaults.
         # General options (find_links, index_url, extra_index_url, trusted_host,
         # and pre) are deferred to pip.
-        self.command: InstallCommand = create_command("install")
-        extra_pip_args = ["--use-deprecated", "legacy-resolver"]
+        self._command: InstallCommand = create_command("install")
 
-        options, _ = self.command.parse_args(pip_args + extra_pip_args)
+        options, _ = self.command.parse_args(pip_args)
         if options.cache_dir:
             options.cache_dir = normalize_path(options.cache_dir)
         options.require_hashes = False
@@ -103,8 +102,7 @@ class PyPIRepository(BaseRepository):
         self._cache_dir = normalize_path(str(cache_dir))
         self._download_dir = os.path.join(self._cache_dir, "pkgs")
 
-        if PIP_VERSION[0] < 22:
-            self._setup_logging()
+        self._setup_logging()
 
     def clear_caches(self) -> None:
         rmtree(self._download_dir, ignore_errors=True)
@@ -120,6 +118,11 @@ class PyPIRepository(BaseRepository):
     @property
     def finder(self) -> PackageFinder:
         return self._finder
+
+    @property
+    def command(self) -> InstallCommand:
+        """Return an install command instance."""
+        return self._command
 
     def find_all_candidates(self, req_name: str) -> List[InstallationCandidate]:
         if req_name not in self._available_candidates_cache:
@@ -463,6 +466,9 @@ class PyPIRepository(BaseRepository):
             no_color=self.options.no_color,
             user_log_file=self.options.log,
         )
+
+        if PIP_VERSION[0] >= 22:
+            return
 
         # Sync pip's console handler stream with LogContext.stream
         logger = logging.getLogger()
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -139,24 +139,14 @@ def combine_install_requirements(
             key=lambda x: (len(str(x)), str(x)),
         )
 
-    combined_ireq = InstallRequirement(
+    combined_ireq = copy_install_requirement(
+        template=source_ireqs[0],
         req=req,
         comes_from=comes_from,
-        editable=source_ireqs[0].editable,
-        link=link_attrs["link"],
-        markers=source_ireqs[0].markers,
-        use_pep517=source_ireqs[0].use_pep517,
-        isolated=source_ireqs[0].isolated,
-        install_options=source_ireqs[0].install_options,
-        global_options=source_ireqs[0].global_options,
-        hash_options=source_ireqs[0].hash_options,
         constraint=constraint,
         extras=extras,
-        user_supplied=source_ireqs[0].user_supplied,
+        **link_attrs,
     )
-    # e.g. If the original_link was None, keep it so. Passing `link` as an
-    # argument to `InstallRequirement` sets it as the original_link:
-    combined_ireq.original_link = link_attrs["original_link"]
     combined_ireq._source_ireqs = source_ireqs
 
     return combined_ireq
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -1,24 +1,53 @@
+import collections
 import copy
+from abc import ABCMeta, abstractmethod
 from functools import partial
 from itertools import chain, count, groupby
-from typing import Dict, Iterable, Iterator, List, Optional, Set, Tuple
+from typing import (
+    Any,
+    DefaultDict,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+)
 
 import click
+from pip._internal.cache import WheelCache
+from pip._internal.exceptions import DistributionNotFound
 from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
+from pip._internal.resolution.resolvelib.base import Candidate
+from pip._internal.resolution.resolvelib.candidates import ExtrasCandidate
+from pip._internal.resolution.resolvelib.resolver import Resolver
+from pip._internal.utils.logging import indent_log
+from pip._internal.utils.temp_dir import TempDirectory, global_tempdir_manager
+from pip._vendor.packaging.specifiers import SpecifierSet
+from pip._vendor.packaging.utils import canonicalize_name
+from pip._vendor.resolvelib.resolvers import ResolutionImpossible, Result
 
 from piptools.cache import DependencyCache
 from piptools.repositories.base import BaseRepository
 
-from ._compat.pip_compat import update_env_context_manager
+from ._compat import PIP_VERSION
+from ._compat.pip_compat import get_build_tracker, update_env_context_manager
+from .exceptions import PipToolsError
 from .logging import log
 from .utils import (
     UNSAFE_PACKAGES,
+    as_tuple,
+    copy_install_requirement,
     format_requirement,
     format_specifier,
     is_pinned_requirement,
     is_url_requirement,
     key_from_ireq,
+    key_from_req,
+    remove_value,
+    strip_extras,
 )
 
 green = partial(click.style, fg="green")
@@ -133,10 +162,55 @@ def combine_install_requirements(
     return combined_ireq
 
 
-class Resolver:
+class BaseResolver(metaclass=ABCMeta):
+    repository: BaseRepository
+    unsafe_constraints: Set[InstallRequirement]
+
+    @abstractmethod
+    def resolve(self, max_rounds: int) -> Set[InstallRequirement]:
+        """
+        Find concrete package versions for all the given InstallRequirements
+        and their recursive dependencies and return a set of pinned
+        ``InstallRequirement``'s.
+        """
+
+    def resolve_hashes(
+        self, ireqs: Set[InstallRequirement]
+    ) -> Dict[InstallRequirement, Set[str]]:
+        """Find acceptable hashes for all of the given ``InstallRequirement``s."""
+        log.debug("")
+        log.debug("Generating hashes:")
+        with self.repository.allow_all_wheels(), log.indentation():
+            return {ireq: self.repository.get_hashes(ireq) for ireq in ireqs}
+
+    def _filter_out_unsafe_constraints(
+        self,
+        ireqs: Set[InstallRequirement],
+        reverse_dependencies: Dict[str, Set[str]],
+    ) -> None:
+        """
+        Remove from a given set of ``InstallRequirement``'s unsafe constraints.
+
+        Reverse_dependencies is used to filter out packages that are only
+        required by unsafe packages. This logic is incomplete, as it would
+        fail to filter sub-sub-dependencies of unsafe packages. None of the
+        UNSAFE_PACKAGES currently have any dependencies at all (which makes
+        sense for installation tools) so this seems sufficient.
+        """
+        for req in ireqs.copy():
+            required_by = reverse_dependencies.get(req.name.lower(), set())
+            if req.name in UNSAFE_PACKAGES or (
+                required_by and all(name in UNSAFE_PACKAGES for name in required_by)
+            ):
+                self.unsafe_constraints.add(req)
+                ireqs.remove(req)
+
+
+class LegacyResolver(BaseResolver):
     def __init__(
         self,
         constraints: Iterable[InstallRequirement],
+        existing_constraints: Dict[str, InstallRequirement],
         repository: BaseRepository,
         cache: DependencyCache,
         prereleases: Optional[bool] = False,
@@ -157,28 +231,26 @@ class Resolver:
         self.allow_unsafe = allow_unsafe
         self.unsafe_constraints: Set[InstallRequirement] = set()
 
+        options = self.repository.options
+        if "legacy-resolver" not in options.deprecated_features_enabled:
+            raise PipToolsError("Legacy resolver deprecated feature must be enabled.")
+
+        # Make sure there is no enabled 2020-resolver
+        options.features_enabled = remove_value(
+            options.features_enabled, "2020-resolver"
+        )
+
     @property
     def constraints(self) -> Set[InstallRequirement]:
         return set(
             self._group_constraints(chain(self.our_constraints, self.their_constraints))
         )
 
-    def resolve_hashes(
-        self, ireqs: Set[InstallRequirement]
-    ) -> Dict[InstallRequirement, Set[str]]:
-        """
-        Finds acceptable hashes for all of the given InstallRequirements.
-        """
-        log.debug("")
-        log.debug("Generating hashes:")
-        with self.repository.allow_all_wheels(), log.indentation():
-            return {ireq: self.repository.get_hashes(ireq) for ireq in ireqs}
-
     def resolve(self, max_rounds: int = 10) -> Set[InstallRequirement]:
         """
-        Finds concrete package versions for all the given InstallRequirements
-        and their recursive dependencies.  The end result is a flat list of
-        (name, version) tuples.  (Or an editable package.)
+        Find concrete package versions for all the given InstallRequirements
+        and their recursive dependencies and return a set of pinned
+        ``InstallRequirement``'s.
 
         Resolves constraints one round at a time, until they don't change
         anymore.  Protects against infinite loops by breaking out after a max
@@ -216,21 +288,11 @@ class Resolver:
         results = {req for req in best_matches if not req.constraint}
 
         # Filter out unsafe requirements.
-        self.unsafe_constraints = set()
         if not self.allow_unsafe:
-            # reverse_dependencies is used to filter out packages that are only
-            # required by unsafe packages. This logic is incomplete, as it would
-            # fail to filter sub-sub-dependencies of unsafe packages. None of the
-            # UNSAFE_PACKAGES currently have any dependencies at all (which makes
-            # sense for installation tools) so this seems sufficient.
-            reverse_dependencies = self.reverse_dependencies(results)
-            for req in results.copy():
-                required_by = reverse_dependencies.get(req.name.lower(), set())
-                if req.name in UNSAFE_PACKAGES or (
-                    required_by and all(name in UNSAFE_PACKAGES for name in required_by)
-                ):
-                    self.unsafe_constraints.add(req)
-                    results.remove(req)
+            self._filter_out_unsafe_constraints(
+                ireqs=results,
+                reverse_dependencies=self.reverse_dependencies(results),
+            )
 
         return results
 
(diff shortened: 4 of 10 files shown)


astrojuanlu:
I confirm that this fixes my original use case in #1277 🚀 

<details>

```
(.venv) juanlu at arion2 in /tmp/reqs
$ echo 'jupyterlab==3.0.0rc13' > requirements.in
(.venv) juanlu at arion2 in /tmp/reqs
$ pip-compile
Could not find a version that matches jupyter-server~=1.0.1,~=1.1,~=1.5 (from jupyterlab==3.0.0rc13->-r requirements.in (line 1))
Tried: 0.0.1, 0.0.1, 0.0.2, 0.0.2, 0.0.3, 0.0.3, 0.0.4, 0.0.4, 0.0.5, 0.0.5, 0.1.0, 0.1.0, 0.1.1, 0.1.1, 0.2.0, 0.2.0, 0.2.1, 0.2.1, 0.3.0, 0.3.0, 1.0.0, 1.0.0, 1.0.1, 1.0.1, 1.0.2, 1.0.2, 1.0.3, 1.0.3, 1.0.4, 1.0.4, 1.0.5, 1.0.5, 1.0.6, 1.0.6, 1.0.7, 1.0.7, 1.0.8, 1.0.8, 1.0.9, 1.0.9, 1.0.10, 1.0.10, 1.0.11, 1.0.11, 1.1.0, 1.1.0, 1.1.1, 1.1.1, 1.1.2, 1.1.2, 1.1.3, 1.1.3, 1.1.4, 1.1.4, 1.2.0, 1.2.0, 1.2.1, 1.2.1, 1.2.2, 1.2.2, 1.2.3, 1.2.3, 1.3.0, 1.3.0, 1.4.0, 1.4.0, 1.4.1, 1.4.1, 1.5.0, 1.5.0, 1.5.1, 1.5.1, 1.6.0, 1.6.0, 1.6.1, 1.6.1, 1.6.2, 1.6.2, 1.6.3, 1.6.3, 1.6.4, 1.6.4, 1.7.0, 1.7.0, 1.8.0, 1.8.0, 1.9.0, 1.9.0, 1.10.0, 1.10.0, 1.10.1, 1.10.1, 1.10.2, 1.10.2, 1.11.0, 1.11.0, 1.11.1, 1.11.1, 1.11.2, 1.11.2
Skipped pre-versions: 1.0.0rc0, 1.0.0rc0, 1.0.0rc1, 1.0.0rc1, 1.0.0rc2, 1.0.0rc2, 1.0.0rc3, 1.0.0rc3, 1.0.0rc4, 1.0.0rc4, 1.0.0rc5, 1.0.0rc5, 1.0.0rc6, 1.0.0rc6, 1.0.0rc7, 1.0.0rc7, 1.0.0rc8, 1.0.0rc8, 1.0.0rc9, 1.0.0rc9, 1.0.0rc10, 1.0.0rc10, 1.0.0rc11, 1.0.0rc11, 1.0.0rc12, 1.0.0rc12, 1.0.0rc13, 1.0.0rc13, 1.0.0rc14, 1.0.0rc14, 1.0.0rc15, 1.0.0rc15, 1.0.0rc16, 1.0.0rc16, 1.7.0a1, 1.7.0a1, 1.7.0a2, 1.7.0a2
There are incompatible versions in the resolved dependencies:
  jupyter-server~=1.0.1 (from jupyterlab==3.0.0rc13->-r requirements.in (line 1))
  jupyter-server~=1.1 (from jupyterlab-server==2.0.0->jupyterlab==3.0.0rc13->-r requirements.in (line 1))
  jupyter-server~=1.5 (from nbclassic==0.2.8->jupyterlab==3.0.0rc13->-r requirements.in (line 1))
(.venv) juanlu at arion2 in /tmp/reqs
$ pip-compile --resolver=2020
#
# This file is autogenerated by pip-compile with python 3.9
# To update, run:
#
#    pip-compile --resolver=2020
#
argon2-cffi==21.1.0
    # via notebook
attrs==21.2.0
    # via jsonschema
babel==2.9.1
    # via jupyterlab-server
backcall==0.2.0
    # via ipython
bleach==4.1.0
    # via nbconvert
certifi==2021.10.8
    # via requests
cffi==1.15.0
    # via argon2-cffi
charset-normalizer==2.0.7
    # via requests
debugpy==1.5.1
    # via ipykernel
decorator==5.1.0
    # via ipython
defusedxml==0.7.1
    # via nbconvert
entrypoints==0.3
    # via
    #   jupyter-client
    #   nbconvert
idna==3.3
    # via requests
ipykernel==6.5.1
    # via notebook
ipython==7.29.0
    # via
    #   ipykernel
    #   jupyterlab
ipython-genutils==0.2.0
    # via
    #   jupyter-server
    #   nbformat
    #   notebook
jedi==0.18.1
    # via ipython
jinja2==3.0.3
    # via
    #   jupyter-server
    #   jupyterlab
    #   jupyterlab-server
    #   nbconvert
    #   notebook
json5==0.9.6
    # via jupyterlab-server
jsonschema==4.2.1
    # via
    #   jupyterlab-server
    #   nbformat
jupyter-client==7.0.6
    # via
    #   ipykernel
    #   jupyter-server
    #   nbclient
    #   notebook
jupyter-core==4.9.1
    # via
    #   jupyter-client
    #   jupyter-server
    #   jupyterlab
    #   nbconvert
    #   nbformat
    #   notebook
jupyter-server==1.0.11
    # via
    #   jupyterlab
    #   jupyterlab-server
    #   nbclassic
jupyterlab==3.0.0rc13
    # via -r requirements.in
jupyterlab-pygments==0.1.2
    # via nbconvert
jupyterlab-server==2.0.0rc7
    # via jupyterlab
markupsafe==2.0.1
    # via jinja2
matplotlib-inline==0.1.3
    # via
    #   ipykernel
    #   ipython
mistune==0.8.4
    # via nbconvert
nbclassic==0.2.4
    # via jupyterlab
nbclient==0.5.9
    # via nbconvert
nbconvert==6.3.0
    # via
    #   jupyter-server
    #   notebook
nbformat==5.1.3
    # via
    #   jupyter-server
    #   nbclient
    #   nbconvert
    #   notebook
nest-asyncio==1.5.1
    # via
    #   jupyter-client
    #   nbclient
    #   notebook
notebook==6.4.6
    # via nbclassic
packaging==21.3
    # via
    #   bleach
    #   jupyterlab
    #   jupyterlab-server
pandocfilters==1.5.0
    # via nbconvert
parso==0.8.2
    # via jedi
pexpect==4.8.0
    # via ipython
pickleshare==0.7.5
    # via ipython
prometheus-client==0.12.0
    # via
    #   jupyter-server
    #   notebook
prompt-toolkit==3.0.22
    # via ipython
ptyprocess==0.7.0
    # via
    #   pexpect
    #   terminado
pycparser==2.21
    # via cffi
pygments==2.10.0
    # via
    #   ipython
    #   jupyterlab-pygments
    #   nbconvert
pyparsing==3.0.6
    # via packaging
pyrsistent==0.18.0
    # via jsonschema
python-dateutil==2.8.2
    # via jupyter-client
pytz==2021.3
    # via babel
pyzmq==22.3.0
    # via
    #   jupyter-client
    #   jupyter-server
    #   notebook
requests==2.26.0
    # via jupyterlab-server
send2trash==1.8.0
    # via
    #   jupyter-server
    #   notebook
six==1.16.0
    # via
    #   bleach
    #   python-dateutil
terminado==0.12.1
    # via
    #   jupyter-server
    #   notebook
testpath==0.5.0
    # via nbconvert
tornado==6.1
    # via
    #   ipykernel
    #   jupyter-client
    #   jupyter-server
    #   jupyterlab
    #   notebook
    #   terminado
traitlets==5.1.1
    # via
    #   ipykernel
    #   ipython
    #   jupyter-client
    #   jupyter-core
    #   jupyter-server
    #   matplotlib-inline
    #   nbclient
    #   nbconvert
    #   nbformat
    #   notebook
urllib3==1.26.7
    # via requests
wcwidth==0.2.5
    # via prompt-toolkit
webencodings==0.5.1
    # via bleach

# setuptools
```
</details>

FlorentJeannot:
Hello @atugushev 

Thank you for pinging me to test your branch, and thank you for taking the time to work on this feature!

I tried it and I noted some remarks:

1. ℹ️ From now on, `-c requirements.txt` does not work if the `requirements.txt` contains for example a direct reference. It requires a package name with a constraint. This is not related to your PR, but to how `pip` works. I thought it was interesting to mention it.

2. ❓ The listing of dependencies in the generated `.txt` file is a bit weird, for example:
```
certifi==2021.10.8
    # via
    #   requests
    #   requests[socks]
```
I don't know if we want to output the same package name twice?

3. ⚠️ The generated `.txt` file does not output the same dependencies depending on the order in the `.in` file.

For example:
```
crochet==2.0.0
Scrapy==2.5.1
```

Will output:
```
[...]
twisted==21.7.0
    # via
    #   crochet
    #   twisted[http2]
[...]
```

While:
```
Scrapy==2.5.1
crochet==2.0.0
```

Will output:
```
[...]
 twisted[http2]==21.7.0
    # via
    #   crochet
    #   scrapy
    #   twisted[http2]
[...]
```

atugushev:
@FlorentJeannot thanks for the feedback and good catch! Looks like we need more tests for packages with extras. I'll address the bugs shortly.

FlorentJeannot:
Thank you @atugushev. I found another one, not sure if it's related to your changes though.

When I have a direct reference with a relative path which contains an egg, the output of the .txt is the absolute path of the file on my machine.

Let me know if you need an exemple and thank you.
Also let me know when you have made some fixes, so that I can test it again and let you know if I spot other things.

webknjaz:
@FlorentJeannot if you could send a PR with a failing test, it'd be easier to come up with a fix separately, once it's committed in Git. Per https://pganssle-talks.github.io/xfail-lightning/.

piptools/repositories/base.py
```diff
@@ -61,3 +62,8 @@ def session(self) -> PipSession:
     @abstractmethod
     def finder(self) -> PackageFinder:
         """Returns a package finder to interact with simple repository API (PEP 503)"""
+
+    @property
+    @abstractmethod
+    def command(self) -> InstallCommand:
+        """Returns an install command"""
```
webknjaz, line 69: Let's use a PEP 257 compliant docstring.
```suggestion
        """Return an install command."""
```

piptools/repositories/local.py
```diff
@@ -61,6 +62,10 @@ def finder(self) -> PackageFinder:
     def session(self) -> Session:
         return self.repository.session
 
+    @property
+    def command(self) -> InstallCommand:
```
webknjaz, line 66: """Return an install command instance."""

AndydeCleyre:
> When I declare a local file which contains an egg (for example `file:./vendored/package.zip[requests]`), the output of the `.txt` is the absolute path of the file on my machine.

@FlorentJeannot I don't know if @atugushev  wants to merge this before or after #1329

piptools/resolver.py
```diff
@@ -104,10 +117,29 @@ def combine_install_requirements(
...
+
+    def resolve(self, max_rounds: int) -> Set[InstallRequirement]:
+        raise NotImplementedError("Override in subclass")
+
+    def resolve_hashes(
+        self, ireqs: Set[InstallRequirement]
+    ) -> Dict[InstallRequirement, Set[str]]:
+        """
+        Finds acceptable hashes for all of the given InstallRequirements.
+        """
```
webknjaz, line 131: Methods/functions perform actions so the docstrings should start with a verb and be in imperative mood as opposed to objects/classes/modules/packages where the docstrings merely describe what they are.
```suggestion
        """Find acceptable hashes for all of the given ``InstallRequirement``s."""
```

piptools/resolver.py
```diff
@@ -448,3 +478,164 @@ def reverse_dependencies(
             ireq for ireq in ireqs if not (ireq.editable or is_url_requirement(ireq))
         ]
         return self.dependency_cache.reverse_dependencies(non_editable)
+
+
+class Resolver(BaseResolver):
+    """
+    Uses pip's 2020 resolver.
+    """
```
webknjaz, line 486: Classes don't do things, they are not actions. They represent a state/concept/abstraction. You should describe what it is, not what it "does" because it doesn't do anything by itself, an external actor will call methods/actions on it.
```suggestion
    """A resolver wrapper that uses resolvelib."""
```

piptools/resolver.py
```diff
@@ -448,3 +478,164 @@ def reverse_dependencies(
...
+        if "2020-resolver" not in options.features_enabled:
+            raise PipToolsError("2020 resolver feature must be enabled.")
+
+        # Make sure there is no enabled legacy resolver
+        options.deprecated_features_enabled = remove_value(
+            options.deprecated_features_enabled, "legacy-resolver"
+        )
+
+    def resolve(self, max_rounds: int = 10) -> Set[InstallRequirement]:
+        with get_requirement_tracker() as req_tracker, global_tempdir_manager(), indent_log(), update_env_context_manager(  # noqa: E501
```
webknjaz, line 518: Why don't you just fix the line length instead of ignoring it? You could reorder the CMs to put the update env first (unless it's order-sensitive).
Also, https://pypi.org/project/flake8-length/ exists. It could allow occasional length violations.

piptools/resolver.py
```diff
@@ -448,3 +478,164 @@ def reverse_dependencies(
...
+            PIP_EXISTS_ACTION="i"
+        ):
+            # Mark direct/primary/user_supplied packages
+            for ireq in self.constraints:
+                ireq.user_supplied = True
+
+            # Pass compiled requirements from `requirements.txt` as constraints to resolver
+            existing_constraints = list(self.existing_constraints.values())
```
webknjaz, line 526: maybe use a set?
```suggestion
            existing_constraints = set(self.existing_constraints.values())
```
Or maybe omit it and add each in the loop?

piptools/resolver.py
```diff
@@ -448,3 +478,164 @@ def reverse_dependencies(
...
+            for ireq in existing_constraints:
+                ireq.extras = set()  # pip does not support extras in constraints
+                ireq.constraint = True
+                ireq.user_supplied = False
+                ireq._is_existing_pin = True
```
webknjaz, line 531: How about just doing
```suggestion
                ireq._is_existing_pin = True
                self.constraints.append(ireq)
```
and you wouldn't need to maintain an extra variable

piptools/resolver.py
```diff
@@ -448,3 +478,164 @@ def reverse_dependencies(
...
+        options.deprecated_features_enabled = remove_value(
+            options.deprecated_features_enabled, "legacy-resolver"
+        )
+
+    def resolve(self, max_rounds: int = 10) -> Set[InstallRequirement]:
+        with get_requirement_tracker() as req_tracker, global_tempdir_manager(), indent_log(), update_env_context_manager(  # noqa: E501
+            PIP_EXISTS_ACTION="i"
+        ):
+            # Mark direct/primary/user_supplied packages
+            for ireq in self.constraints:
```
webknjaz, line 522: I don't understand why not apply this conversion to the constraints list during the initialization. Also, I don't really understand why this needs to be mutable. Is that expectation that it'll be modified by external callers post-initialization?

piptools/scripts/compile.py
```diff
@@ -224,6 +224,14 @@ def _get_default_option(option_name: str) -> Any:
 @click.option(
     "--pip-args", "pip_args_str", help="Arguments to pass directly to the pip command."
 )
+@click.option(
+    "--resolver",
+    "resolver_name",
+    type=click.Choice(("legacy", "2020")),
```
webknjaz, line 230: I don't think it's very user-friendly to call a resolver `"2020"`. It refers to a pip version which is an implementation detail. Maybe call it `"backtracking"`? Or maybe call it `resolvelib` (but that's also too internal).

piptools/utils.py
```diff
@@ -437,3 +437,10 @@ def get_sys_path_for_python_executable(python_executable: str) -> List[str]:
     assert isinstance(paths, list)
     assert all(isinstance(i, str) for i in paths)
     return [os.path.abspath(path) for path in paths]
+
+
+def remove_value(lst: List[_T], value: _T) -> List[_T]:
+    """
+    Returns new list without a given value.
+    """
+    return [item for item in lst if item != value]
```
webknjaz, line 446: The function name implies that it will mutate the incoming list which is not true judging from the implementation. Choose a better one.
```suggestion
def omit_list_value(lst: List[_T], value: _T) -> List[_T]:
    """Produce a new list with a given value skipped."""
    return [item for item in lst if item != value]
```

piptools/writer.py
```diff
@@ -269,12 +281,17 @@ def _format_requirement(
...
             }
-        elif ireq.comes_from:
-            required_by.add(_comes_from_as_string(ireq))
+
+        if ireq.comes_from:
+            required_by.add(_comes_from_as_string(ireq.comes_from))
+
+        if hasattr(ireq, "_required_by"):
+            for name in ireq._required_by:
+                required_by.add(name)
```
webknjaz, line 294:
```suggestion
        required_by |= set(getattr(ireq, "_required_by", set()))
```

AndydeCleyre:
I wonder if the varying extras behavior noted by @FlorentJeannot is resolved by @richafrank's sitting PRs.

@atugushev Do you think we'll be finalizing this one before any other PRs, or are there some you plan to review and merge first?

tests/test_cli_compile.py
```diff
@@ -427,11 +451,21 @@ def test_editable_package_without_non_editable_duplicate(pip_conf, runner):
...
     without also adding a duplicate "non-editable" requirement variation
     """
+
+    if current_resolver != RESOLVER_LEGACY:
+        pytest.skip(
+            "This test is actual only for legacy resolver. "
+            "Constraints refactored in 2020 resolver. "
+            "See https://github.com/pypa/pip/issues/9020 for details."
+        )
+
```
webknjaz, line 468: It's way cleaner like this:
```suggestion
@pytest.mark.parametrize(
    'current_resolver',
    (
       RESOLVER_LEGACY,
    ),
    indirect=('current_resolver', ),
)
@pytest.mark.usefixtures('current_resolver')
def test_editable_package_constraint_without_non_editable_duplicate(pip_conf, runner):
    """
    piptools keeps editable constraint,
    without also adding a duplicate "non-editable" requirement variation
    """
```
and won't produce unnecessary skipped statuses in the pytest report.

webknjaz: You could also make it reusable like this:
```suggestion

legacy_resolver_only = pytest.mark.parametrize(
    'current_resolver',
    (
       RESOLVER_LEGACY,
    ),
    indirect=('current_resolver', ),
)

@legacy_resolver_only
@pytest.mark.usefixtures('current_resolver')
def test_editable_package_constraint_without_non_editable_duplicate(pip_conf, runner):
    """
    piptools keeps editable constraint,
    without also adding a duplicate "non-editable" requirement variation
    """
```

jdimatteo:
I'd love for this PR to be merged soon! I was using this but this branch is missing fixes like https://github.com/jazzband/pip-tools/pull/1567

saulshanabrook:
Would folks be open to someone else taking on this PR to address the pending comments and rebase off of master?

pradyunsg:
> Would folks be open to someone else taking on this PR to address the pending comments and rebase off of master?

Yes, that would be welcome.

piptools/repositories/base.py
atugushev: Addressed in 81419f9f1090bda0761020e7e9832ba279a5fd4d

piptools/repositories/local.py
atugushev: Addressed in 81419f9f1090bda0761020e7e9832ba279a5fd4d

piptools/repositories/pypi.py
```diff
@@ -120,6 +119,10 @@ def session(self) -> PipSession:
     def finder(self) -> PackageFinder:
         return self._finder
 
+    @property
+    def command(self) -> InstallCommand:
```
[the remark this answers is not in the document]
atugushev: Addressed in 81419f9f1090bda0761020e7e9832ba279a5fd4d

piptools/resolver.py
atugushev: Addressed in 81419f9f1090bda0761020e7e9832ba279a5fd4d

piptools/resolver.py
atugushev: Addressed in 81419f9f1090bda0761020e7e9832ba279a5fd4d

tests/test_cli_compile.py
atugushev: Addressed in c6f0efbfff7ab513c6e6c2e25e90700527f797d8

piptools/resolver.py
atugushev: Black can't format long `with` statements. See https://github.com/psf/black/issues/664

piptools/resolver.py
atugushev: Addressed in 7035524

piptools/resolver.py
atugushev: Addressed in 7035524

piptools/scripts/compile.py
atugushev: Backtracking is okay. Addressed in c5fdc7e.

piptools/writer.py
atugushev: Addressed in e361fa5

piptools/utils.py
atugushev: Addressed in e361fa5

atugushev:
Hey @FlorentJeannot! 

Could you please check the latest version where I've fixed the [bug with annotations](https://github.com/jazzband/pip-tools/pull/1539#issuecomment-975542600) (2 and 3 points)?

FlorentJeannot:
Hello @atugushev !

Thank you for your update and pinging me about this. I tried your changes and it looks good!

However, I found some other issues. I am trying to reproduce them as I cannot share the files with you, but so far I could not.

Basically the bug is that for some dependencies, local files or from a VCS, instead of showing the correct path or URL in the output, it shows the absolute path of a wheel file from pip's cache. I'll continue investigating tomorrow. If I can't reproduce I'll share more information.

piptools/resolver.py
atugushev: Addressed in ac2f5cb24b5bbd4da06fe1c8a43b362aa3c0b868

FlorentJeannot:
@atugushev
TL;DR: Looks good to me 👍 

I had an issue which would show this kind of output:

```
file:///Users/fj/Library/Caches/pip/wheels/2d/87/4b/d4a9edf4de8fdfbebd07dad937ddd46e0cf9a73740fdd64706/package-version-py3-none-any.whl
```

I fixed it by deleting the cache of pip. If we have people opening issue tickets for this, we should ask them to try `pip cache purge`. Also note that I did `rm -rf /Users/florent/Library/Caches/pip/*` for myself, but I expect pip's command line to do the same.

Also there's a discrepancy between legacy and backtracking about the extras. If the extra comes from a package in the `.in` file, then it's displayed in the `.txt` file; otherwise it's omitted.
I think it's somewhat fine, since we're heading in the right direction (e.g. only specify extras in `.in` files and do not show them in compiled files to mimic `pip freeze` behavior).

Before
```
twisted[http2]==21.7.0
    # via
    #   crochet
    #   scrapy
```

After

```
twisted==21.7.0
    # via
    #   crochet
    #   scrapy
```

piptools/resolver.py
atugushev: > I don't understand why not apply this conversion to the constraints list during the initialization. 

Do you mean inside `Resolver.__init__()`?

atugushev:
I've just checked all [reported resolver bugs](https://github.com/jazzband/pip-tools/issues?q=is%3Aissue+is%3Aopen+label%3Aresolver+label%3Abug) (16 open) - new resolver fixes them all 🔥 It looks ready to go live.

AndydeCleyre:
I'm not sure if this is specific for this branch, because the input (`https://github.com/vmware/vsphere-automation-sdk-python/archive/v7.0.3.0.tar.gz`) doesn't compile at all in master, but over at [1597 (comment)](https://github.com/jazzband/pip-tools/issues/1597#issuecomment-1062718003) we're seeing for example:

```
nsx-policy-python-sdk @ file://localhost//tmp/pip-resolve-ehumdwah/vsphere-automation-sdk_a3f4013e6ddf4bc2a7c6a2cb49639288/lib/nsx-policy-python-sdk/nsx_policy_python_sdk-3.1.2.1.1-py2.py3-none-any.whl
```

I tried working around it with `pip cache purge`, but the output is the same.

atugushev:
> I'm not sure if this is specific for this branch, because the input (`https://github.com/vmware/vsphere-automation-sdk-python/archive/v7.0.3.0.tar.gz`) doesn't compile at all in master, but over at [1597 (comment)](https://github.com/jazzband/pip-tools/issues/1597#issuecomment-1062718003) we're seeing for example:

@AndydeCleyre this looks out of the scope of this PR.

AndydeCleyre:
Agreed now that you've shown what that package is doing. Thanks for having a look!

AndydeCleyre:
This branch is failing some tests locally but I have to run and will check it out again in like 8 hours.

AndydeCleyre:
A pretty straightforward merge from master fixes my local tests, just as you can see `CI / Ubuntu / 3.7 / main` can't use `pip._internal.req.req_tracker`.

atugushev:
> I'm happy to do a rebase but I don't want to disrupt ongoing work and review.

@AndydeCleyre please, feel free to.

atugushev:
@AndydeCleyre I'd avoid rebase though and prefer merge. Feel free to push commits to the branch to address the issue with req_tracker.

atugushev:
The `pipmain` still doesn't pass CI. I've tracked down bad commit https://github.com/pypa/pip/commit/0c284520c6d068cb25ac89d9dbee0456c2eba23a using bisect.

q0w:
@atugushev could you also update `repositories/pypi.py` because of this change https://github.com/pypa/pip/pull/11014

AndydeCleyre:
Would it be worthwhile to cut a small release with the `pip` version capped (both in the readme, and in `setup.cfg`)?

atugushev:
@q0w updated. Thanks for noticing!

atugushev:
> Would it be worthwhile to cut a small release with the `pip` version capped (both in the readme, and in `setup.cfg`)?

@AndydeCleyre after `pip-22.1` is released we can upgrade minimal requirements to `pip>=22.0`.

AndydeCleyre:
@thomas-riccardi 

Thanks! Reproduced here as well.

atugushev:
@AndydeCleyre thanks for the attached tests 🙏🏻

atugushev:
Hey @thomas-riccardi,

Thanks for the good catch and simple reproducer! I've fixed the bug and tested it on docker - looks good. Could you confirm whether the last commit ([56d0ab3](https://github.com/jazzband/pip-tools/pull/1539/commits/56d0ab303b75f44e966bddf62f518135b2261e66)) fixes your upgrade workflow?

thomas-riccardi:
Hello @atugushev , thanks for the fast reply!
I confirm that e2b2c911b15bdb72d3de5bd8f7b9cfbee8609e3a fixes the simple scenario I reported here; but it doesn't cover our most complex scenario: sub-dependencies conflicts. Indeed your fix handles direct conflicts with requirements.in, but not indirect ones.

Here is an updated repro (using the latest version of this PR, including your fix):
### Reproduction 2
<details>
  <summary>Click to expand</summary>

```Dockerfile
# Usage: docker build . --progress=plain --build-arg CACHE_BREAKER=$RANDOM --build-arg RESOLVER=backtracking

FROM python:3.7

RUN pip --version
RUN pip install pip==21.*  

#RUN pip install pip-tools@git+https://github.com/atugushev/pip-tools@new-resolver
RUN pip install pip-tools@git+https://github.com/atugushev/pip-tools@56d0ab303b75f44e966bddf62f518135b2261e66  # pin commit for stability

ARG CACHE_BREAKER

ARG RESOLVER=legacy
RUN echo "flake8==3.8.4" > requirements.in
RUN pip-compile --verbose requirements.in -o requirements.txt --resolver=$RESOLVER
RUN cat requirements.txt

# simulate minimal upgrade: update requirements.in, keep old requirements.txt
RUN echo "flake8==3.9.2" > requirements.in
RUN pip-compile --verbose requirements.in -o requirements.txt --resolver=$RESOLVER
RUN cat requirements.txt

# result:
# expected: both resolvers should work
# actual: backtracking resolover fails: it seems to newly take output requirements.txt as strict constraints
```

backtracking resolver output:
```
#12 0.991   The conflict is caused by:
#12 0.991       flake8 3.9.2 depends on pycodestyle<2.8.0 and >=2.7.0
#12 0.991       The user requested (constraint) pycodestyle==2.6.0
```

legacy resolver output:
```
#13 0.349 flake8==3.9.2                                                                       
#13 0.349     # via -r requirements.in                                                        
#13 0.349 importlib-metadata==4.11.3                                                          
#13 0.349     # via flake8                                                                    
#13 0.349 mccabe==0.6.1                                                                       
#13 0.349     # via flake8                                                                                                                                                                   
#13 0.349 pycodestyle==2.7.0                                                                                                                                                                 
#13 0.349     # via flake8                                                                                                                                                                   
#13 0.349 pyflakes==2.3.1                                                                     
#13 0.349     # via flake8                                                                    
#13 0.349 typing-extensions==4.2.0                                                            
#13 0.349     # via importlib-metadata                                                        
#13 0.349 zipp==3.8.0                                                                         
#13 0.349     # via importlib-metadata                                                        
#13 DONE 0.4s
```
</details>

LouisAumaitre:
Hi!

I'm getting an issue with extras using this resolver.

When I attempt to compile a file containing `-c other_file.txt`, if that other file contains extras, I get `pip._internal.exceptions.InstallationError: Constraints cannot have extras`. (I don't get the error when I compile the other file)

Example:

`file_a.txt` contains `coverage[toml]==6.3.2`

`file_b.txt` contains `-c file_a.txt`

Compiling `file_a.txt` works, compiling `file_b.txt` throws an `InstallationError`.

AndydeCleyre:
As long as we are offering `--strip-extras` and not offering its negative, I'd guess that the default output line format would include extras (where this PR currently strips them).

Now that the constraints syntax is stricter (with the backtracking resolver), I expect it will be *much* more common for folks to need files without the extras. So I would support a separate PR to do that by default, while offering a new option to include them, e.g. `--no-strip-extras`.

#1613

AndydeCleyre:
> > Would it be worthwhile to cut a small release with the `pip` version capped (both in the readme, and in `setup.cfg`)?
> 
> @AndydeCleyre after `pip-22.1` is released we can upgrade minimal requirements to `pip>=22.0`.

Looks like we ran out of time here...

ryanhiebert:
@AndydeCleyre : It may be a myopia of limited viewpoint, but I don't know what benefit having the extras in the output gives, and I'd be perfectly fine with there being no option to have them at all. The biggest downside I currently know of is that it would change existing output files, but since they are auto-generated lock files, that doesn't seem like a very big deal to me.

AndydeCleyre:
@ryanhiebert 

I'm not saying it's good or common practice, but my understanding is that extras can be used to enforce installation order, and the set of packages already installed can be used by `setup.py`'s `install` to follow different code paths accordingly.

A hypothetical example:

- we have a package, `mypackage`
- it defines an extra, `interactive`
- the same package author also provides a kind of dummy package, `mypackage-interactive-installation`, required by `mypackages`'s `interactive` extra
- during `mypackage`'s `install`, if and only if `mypackage-interactive-installation` is already installed, the user is prompted to make some choices which will affect the installation

AndydeCleyre:
@atugushev 

I'm excited for this PR to get merged, but if the final 1% of the problem isn't an easy solve, can we hammer out some of the other PRs that have been piling up behind this one? I'd be happy to help merge/rebase this afterward.

atugushev:
> I'm excited for this PR to get merged, but if the final 1% of the problem isn't an easy solve, can we hammer out some of the other PRs that have been piling up behind this one? I'd be happy to help merge/rebase this afterward.

@AndydeCleyre 

I agree, let's not block other PRs by this one. 👍🏻 Thanks for helping!

AndydeCleyre:
I merged master back in, so now in addition to our need to get `test_ignore_incompatible_existing_pins` passing, the backtracking resolver breaks master's relatively new `test_combine_different_extras_of_the_same_package`:

```diff
- ay[default]==0.1
+ ay[default,tune]==0.1
```

AndydeCleyre:
I don't know if we can easily discard incompatible existing pins during resolution, like I *think* we do with the legacy resolver.

What does look easy enough is discarding all the existing pins and trying again, in the case where including them results in a resolution error.

In other words, this change would get the `test_ignore_incompatible_existing_pins` test passing, and not break existing tests, but would change the behavior in some cases I think:

```diff
diff --git a/piptools/resolver.py b/piptools/resolver.py
index 1022306..2b8ef36 100644
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -17,6 +17,7 @@ from typing import (
 
 import click
 from pip._internal.cache import WheelCache
+from pip._internal.exceptions import DistributionNotFound
 from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
 from pip._internal.resolution.resolvelib.base import Candidate
@@ -538,6 +539,7 @@ class Resolver(BaseResolver):
 
             # Pass compiled requirements from `requirements.txt`
             # as constraints to resolver
+            compatible_existing_constraints = []
             for ireq in self.existing_constraints.values():
                 # Skip if the compiled install requirement conflicts with
                 # the primary install requirement.
@@ -551,7 +553,7 @@ class Resolver(BaseResolver):
                 ireq.extras = set()  # pip does not support extras in constraints
                 ireq.constraint = True
                 ireq.user_supplied = False
-                self.constraints.append(ireq)
+                compatible_existing_constraints.append(ireq)
 
             wheel_cache = WheelCache(
                 self.options.cache_dir, self.options.format_control
@@ -593,10 +595,18 @@ class Resolver(BaseResolver):
 
             self.command.trace_basic_info(self.finder)
 
-            resolver.resolve(
-                root_reqs=self.constraints,
-                check_supported_wheels=not self.options.target_dir,
-            )
+            try:
+                resolver.resolve(
+                    root_reqs=self.constraints + compatible_existing_constraints,
+                    check_supported_wheels=not self.options.target_dir,
+                )
+            except DistributionNotFound as e:
+                log.info(str(e))
+                log.info("Discarding existing pins and trying again")
+                resolver.resolve(
+                    root_reqs=self.constraints,
+                    check_supported_wheels=not self.options.target_dir,
+                )
 
         return self._get_install_requirements(resolver._result)

```

atugushev:
> What does look easy enough is discarding all the existing pins and trying again, in the case where including them results in a resolution error.

@AndydeCleyre excellent idea, this let me figure out how to fix the issue properly! Thanks a lot 🙏🏻 

Hey @thomas-riccardi! Could you review and confirm the fix?

atugushev:
> I merged master back in, so now in addition to our need to get `test_ignore_incompatible_existing_pins` passing, the backtracking resolver breaks master's relatively new `test_combine_different_extras_of_the_same_package`:
> 
> ```diff
> - ay[default]==0.1
> + ay[default,tune]==0.1
> ```

@AndydeCleyre thanks for the test! Addressed in 868052d1a06b5fd6adc44209a778d1713d9954f0.

AndydeCleyre:
This is looking great!

I think the CI failures for pip's unreleased main branch are only encountered by the backtracking resolver, so that might be one more issue worth tackling before a merge.

atugushev:
> I think the CI failures for pip's unreleased main branch are only encountered by the backtracking resolver, so that might be one more issue worth tackling before a merge.

@AndydeCleyre yeah, this looks related to https://github.com/jazzband/pip-tools/pull/1632.

piptools/utils.py
```diff
@@ -441,3 +442,41 @@ def get_sys_path_for_python_executable(python_executable: str) -> List[str]:
...
+_strip_extras_re = re.compile(r"\[.+?\]")
+
+
+def strip_extras(name: str) -> str:
+    """Strip extras from package name, e.g. pytest[testing] -> pytest."""
+    return re.sub(_strip_extras_re, "", name)
+
+
+def copy_install_requirement(template: InstallRequirement) -> InstallRequirement:
+    """Make a copy of an ``InstallRequirement``."""
```
richafrank, line 461: Thoughts on `combine_install_requirements` delegating some of its logic to this new function?

atugushev: Agreed. Would you like to open a PR against my `new-resolver` branch I'd be happy to merge?

atugushev:
All checks (except `codecov/project`, which is fine) are green. This is ready for final review.

piptools/utils.py
richafrank: Sure thing. Not sure if I'll make it in time for your PR's merge, but it can be a separate cleanup, if so.

atugushev:
@jazzband/pip-tools 

Pip's legacy resolver is going to be removed in the nearest future according to https://github.com/pypa/pip/issues/10946, so I need some volunteers to push this through so that users would have enough time to test the backtracking resolver.

ssbarnea:
@atugushev You have my support but there is a problem. That code is so old that pip behavior changed few times. For example the new resolver is default and passing the 2020-resolver flag is a very bad idea, read the waring.
```
WARNING: --use-feature=2020-resolver no longer has any effect, since it is now the default dependency resolver in pip. This will become an error in pip 21.0.
```

It is bit hilarious because that error came from pip `22.1.2` even if they declared that they will remove the flag in `21.0`.

I would say the we need to drop support for pip versions that did not have the new resolved active by default and avoid passing arguments for it.

In fact I would go so far to say that I don't even bother about the legacy one.

piptools/utils.py
atugushev: Addressed in c8aca929bbd7d6c49907a445ed15f7e7edee18ee

atugushev:
@ssbarnea I've removed  `--use-feature=2020-resolver` usage in 4fb4610a3ca23bc94fbf3076bae9e1ab367eb0e1

ssbarnea (approved):

atugushev:
@thomas-riccardi thanks for the catch! That looks easy to fix. I'll do it shortly.

atugushev:
@thomas-riccardi fixed in 697ec256a331bbd1514a4aa63a1a9817e15779d0. Could you try it?

thomas-riccardi:
@atugushev 
> @thomas-riccardi fixed in [697ec25](https://github.com/jazzband/pip-tools/commit/697ec256a331bbd1514a4aa63a1a9817e15779d0). Could you try it?

I confirm this fixes both my repro v3 *and* our real use-case (after 3 rounds), thank you!

atugushev:
@thomas-riccardi Excellent! Thanks for your help - that was super helpful 🙏🏻 

--- 
I'm going to squash some commits and prepare PR for merging.

atugushev:
Huh, squashing all but preserving a few commits is a nightmare. I think I'd go with "Create merge commit" if nobody minds.

ssbarnea:
Squash it! I do not want to see 65 commits polluting our history.

atugushev:
I'm so happy this finally merged. Many thanks to my heroes @AndydeCleyre, @FlorentJeannot,  @richafrank, @ssbarnea, @thomas-riccardi, @webknjaz, and all others involved for your support.

atugushev:
@gschaffner thanks for the report! Could you open a separate issue, please?

deifactor:
One thing I noticed is that this is *significantly* slower than the legacy resolver, presumably because it's not using the pip-compile cache.

ssbarnea:
@deifactor It is not a bug, just an opportunity for you to make a pull-request and address this performance regression! ;)

astrojuanlu:
Worth opening an issue for wider visibility?

taeefnajib:
Hi @atugushev I have a question! I need both `flytekit` and `bentoml` installed in the same project. `protobuf = ">=4.21.1,<5.0.0"` is required by `flytekit` and `protobuf<4.0` is required by `bentoml`. When I try `pip-compile requirements.in` I get an error. But after using your solution, i.e. `pip-compile requirements.in --resolver=backtracking` `pip-compile` generates `requirements.txt` file. But the version of `protobuf` is `3.20.3` How can I have both versions of `protobuf` in my `requirements.txt` file?

ryanhiebert:
@taeefnajib Python doesn't allow you to install two different versions of the same dependency in the same environment, so we can't support two versions of a package in a requirements.txt file.

taeefnajib:
@ryanhiebert thanks. let me see why `bentoml` locked the version of `protobuf` to `<4.0`

taeefnajib:
@ryanhiebert Just a quick question! When I put `bentoml` and `flytekit` in my `requirements.txt` file manually (without mentioning versions), and use `pip install -r requirements.txt` it installs multiple versions of `protobuf` It's `pip-compile` that cannot generate a `requirements.txt` file with multiple versions. Do you know why is that happening and if I can find a workaround?

AndydeCleyre:
@taeefnajib it's probably installing one version, then uninstalling that to install a different version. Run `pip list` to see the final state.

taeefnajib:
@AndydeCleyre you are right. I can only see ` 3.20.3`

luhn:
It looks like pip will install flytekit 1.2.11, which requires `protobuf>=3.6.1,<4`, in order to satisfy all dependencies.

```
>> echo "bentoml" >> req.txt
>> echo "flytekit" >> req.txt
>> pip install -qr req.txt
>> pip list | grep protobuf
protobuf                                     3.20.3
>> pip list | grep flytekit
flytekit                                     1.2.11
```

taeefnajib:
@luhn thanks. Yes, it's installing `1.2.11` even though I was hoping to install `1.5.0`

PR #1547, opened 2021-12-26
Improve punctuation/grammar with header.

### LINKED ISSUE #1636: The banner in the generated requirements.txt is confusing
When I use pip-compile to generate a requirements.txt file, it includes this banner at the top:

```
# This file is autogenerated by pip-compile with python 3.8
# To update, run:
#
#    pip-compile --no-emit-index-url
```

However, without the `--upgrade` argument, that command won't generate a new requirements.txt file, unless the setup.py changes. It sounds like it's the distinction between _updating_ the file vs _upgrading_ the packages pinned in the file. As a new user, I found that confusing when subsequent runs of pip-compile weren't picking up upgraded packages.

Ref: https://github.com/jazzband/pip-tools/blob/master/README.rst?plain=1#L64-L68

blueyed:
Resolves #1636.

##### Contributor checklist

There are outdated references in the README etc.

##### Maintainer checklist


piptools/writer.py
```diff
@@ -116,10 +116,10 @@ def write_header(self) -> Iterator[str]:
         if self.emit_header:
             yield comment("#")
             yield comment(
-                "# This file is autogenerated by pip-compile with python "
-                f"{sys.version_info.major}.{sys.version_info.minor}"
+                "# This file is autogenerated by pip-compile with Python "
+                f"{sys.version_info.major}.{sys.version_info.minor}."
             )
-            yield comment("# To update, run:")
+            yield comment("# To update it, run:")
```
hugovk, line 122: "Python" with capital P is good.

This change is also fine but I think the original is fine too and reads a bit better:

```suggestion
            yield comment("# To update, run:")
```

atugushev:
We might want to rephrase the whole "To update run:" line, see the details https://github.com/jazzband/pip-tools/issues/1636.

ssbarnea (approved):

atugushev:
Perhaps we should address this issue together with #1636 to update the header in one shot?

piptools/writer.py
```diff
@@ -132,8 +132,8 @@ def write_header(self) -> Iterator[str]:
         if self.emit_header:
             yield comment("#")
             yield comment(
-                "# This file is autogenerated by pip-compile with python "
-                f"{sys.version_info.major}.{sys.version_info.minor}"
+                "# This file is autogenerated by pip-compile with Python "
+                f"{sys.version_info.major}.{sys.version_info.minor}."
             )
             yield comment("# To update, run:")
```
atugushev, line 138:
```suggestion
                "# This file is autogenerated by pip-compile with Python "
                f"{sys.version_info.major}.{sys.version_info.minor}"
            )
            yield comment("# by the following command:")
```

This addresses https://github.com/jazzband/pip-tools/issues/1636. What do you think?

ssbarnea (approved):

PR #1557, opened 2022-01-29
fix: Add missing comma

mrshu:
* Add missing comma


webknjaz (changes requested):
That comma seems unnecessary. Why did you think it was needed?

mrshu:
You are indeed right @webknjaz -- apologies for the noise.

Issue #1558, opened 2022-01-30
pip-compile fails with pip==22.0


jordan-dimov:
I upgraded Pip to latest version 22.0 today and pip-tools stopped working. 

When I run pip-compile to upgrade my requirements, I now get the following stack trace:

```
$ pip-compile -U requirements/requirements.in
Traceback (most recent call last):
  File "/Users/jordan/.virtualenvs/my-project/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
  File "/Users/jordan/.virtualenvs/my-project/lib/python3.7/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jordan/.virtualenvs/my-project/lib/python3.7/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/Users/jordan/.virtualenvs/my-project/lib/python3.7/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/jordan/.virtualenvs/my-project/lib/python3.7/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/Users/jordan/.virtualenvs/my-project/lib/python3.7/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/jordan/.virtualenvs/my-project/lib/python3.7/site-packages/piptools/scripts/compile.py", line 342, in cli
    repository = PyPIRepository(pip_args, cache_dir=cache_dir)
  File "/Users/jordan/.virtualenvs/my-project/lib/python3.7/site-packages/piptools/repositories/pypi.py", line 106, in __init__
    self._setup_logging()
  File "/Users/jordan/.virtualenvs/my-project/lib/python3.7/site-packages/piptools/repositories/pypi.py", line 455, in _setup_logging
    assert isinstance(handler, logging.StreamHandler)
AssertionError

```

As soon as I downgrade pip to v. 21.3.1 for example, the above command succeeds as expected.

#### Environment Versions

1. OS Type:  MacOS
2. Python version: 3.7.10
3. pip version: 22.0
4. pip-tools version: 6.4.0

hugovk:
https://github.com/pypa/pip/issues/10726#issuecomment-1025206695:

[...]

PR #1565, opened 2022-02-02
add RichPipStreamHandler as allowed handler in repositories/pypi.py

nicoa:
adressing https://github.com/jazzband/pip-tools/issues/1558 .

Details to be filled.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -31,7 +31,7 @@ from pip._internal.network.session import PipSession
 from pip._internal.req import InstallRequirement, RequirementSet
 from pip._internal.req.req_tracker import get_requirement_tracker
 from pip._internal.utils.hashes import FAVORITE_HASH
-from pip._internal.utils.logging import indent_log, setup_logging
+from pip._internal.utils.logging import RichPipStreamHandler, indent_log, setup_logging
 from pip._internal.utils.misc import normalize_path
 from pip._internal.utils.temp_dir import TempDirectory, global_tempdir_manager
 from pip._internal.utils.urls import path_to_url, url_to_path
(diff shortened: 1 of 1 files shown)


piptools/repositories/pypi.py
```diff
@@ -31,7 +31,7 @@
 from pip._internal.req import InstallRequirement, RequirementSet
 from pip._internal.req.req_tracker import get_requirement_tracker
 from pip._internal.utils.hashes import FAVORITE_HASH
-from pip._internal.utils.logging import indent_log, setup_logging
+from pip._internal.utils.logging import RichPipStreamHandler, indent_log, setup_logging
```
nicoa, line 34: oversaw that this won't work with old pip versions - needs to be imported depending on version

piptools/repositories/pypi.py
```diff
@@ -452,7 +461,7 @@ def _setup_logging(self) -> None:
         logger = logging.getLogger()
         for handler in logger.handlers:
             if handler.name == "console":  # pragma: no branch
-                assert isinstance(handler, logging.StreamHandler)
+                assert isinstance(handler, compatible_handler)
```
davisagli, line 464: I think it's going to take more than this. The RichPipStreamHandler sends its output to a rich `console`, not to a `stream`.

It looks like this code was added in https://github.com/jazzband/pip-tools/pull/949 to fix a problem with displaying the progress bars from pip's verbose output.

But it appears that that fix is no longer needed when the progress bars are rendered using `rich`. I tested and if I simply remove lines 451-466 and run pip-compile in verbose mode, the progress bars display fine and the command completes successfully with pip 22.

So I would suggest making these lines conditional on whether RichPipStreamHandler was imported above.

nicoa: Expected something like this would come up. Thanks for checking! Sounds like a reasonable solution, while I'm not sure if the whole code introduced in #949 can be "reverted / ignored" depending on pip version or we should only skip the assignment of the "proper" stream to the respective handler during execution.

@atugushev would be happy to have your expertise here as you worked on adding this back then in case you have time.

hugovk:
This has been included in https://github.com/jazzband/pip-tools/pull/1567, and tests pass.

wbolster:
to avoid parallel work and confusion, closing this in favour of #1567

Issue #1434, opened 2021-06-23
consider transfer of pip-tools under pypa organization


ssbarnea:
While I am very happy with how @jazzband org helped with pip-tools project maintenance, I do also think that it would in pip-tools long term interest to be transitioned under @pypa organization. I will try to explain some of the benefits that I see from that:

* increase visibility of pip-tools among other packaging tools under pypa org
* more likely to get active help from other maintainers with heavy packaging experience
* the [transfer process](https://www.pypa.io/en/latest/members/#project-membership) indicate that the project owners keep control over the project and that they have ability to add/remove other maintainers, or at least this is how interpreted it
* pip-tools maintainers would be able to be able to influence other decisions made by pypa, i suppose  that would likely translate to being less likely to a tool change that break another one (pip extras inside constraints issue could have likely being avoided)
* with all packaging tools under the same work, i think it would be much easier to make introduce changes into shared libraries, like setuptools or packaging for example

This proposal comes after I see the success of [transfer of pipx under pypa](https://github.com/pypa/pipx/issues/596) too.

Does this makes sense? Feel free to upvote or downvote.

ssbarnea:
@jazzband/pip-tools Can we please get some explicit comments from pip-tools cores?

pipx move under pypa was a big success, and happened long time ago.

One extra reason for reviving this proposal is the lack of action on https://github.com/orgs/jazzband/teams/pip-tools/discussions/13 thread, which makes be believe that is hinders project maintenance, especially as this kind of tool needs lots of discussions and a way to cleanup the issue tracker (converting fake bugs into discussions).

Somehow I think future of pip-tools would be far better pypa due to extra visibility.

ssbarnea:
Before we attempt to do anything we need to clarify a thing, **who is/are the pip-tool project lead?**

That is an essential question because only leads can [approve a release](https://github.com/jazzband/pip-tools/blob/master/CONTRIBUTING.md#project-release-process) and normal contributors cannot even see the list of those. Other projects have the main maintainers/leads listed on the README page. I think we should also have the same thing.

I know I can prepare and make the release on github releases, but I have no idea whom can approve it.

ssbarnea:
I see that @nvie is the only person with access on pypa in addition to the jazzband account and he is still the top contributor, followed by @atugushev.

I guess that he should be on the magic-secret list, so I will try to contact him.

vphilippon:
[...]

PR #1609, opened 2022-04-06
Release 6.6.0

atugushev:


hramezani (approved):

atugushev:
Thanks @hramezani 🙏🏻

ssbarnea:
So time to use https://github.com/jazzband/pip-tools/releases page to convert the draft into a release?

Before doing this, can you please enable the discussions on the project? I do not have enough right to do it myself. If you do, you will see a checkbox to create a discussion thread on publish.

atugushev:
I've missed your message and already published the release, sorry.

atugushev:
> Before doing this, can you please enable the discussions on the project?

I've missed your message and already published the release, sorry. There is an open discussion about the "discussing page", see: https://github.com/orgs/jazzband/teams/pip-tools/discussions/13

atugushev:
However, I don't have such permission to enable discussions. We should confirm this with @jezdez first on the thread above.

ssbarnea:
Maybe that is the cherry that would finally push https://github.com/jazzband/pip-tools/issues/1434 to make it happen. At least projects un pypa org are not prevented from enabling discussions... not to mention that this org would be better suited for our project.

Issue #1627, opened 2022-05-23
Use of (Deprecated) pep517 Package Hides setup.py Syntax Errors


Aaron-Beal-CardinalPeak:
Running `python -m piptools compile` when the `setup.py` file has invalid Python in it leads to this unhelpful error message from the `pep517` package:

```
Traceback (most recent call last):
  File "C:\Users\ab\.pyenv\pyenv-win\versions\3.8.10\Lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\ab\.pyenv\pyenv-win\versions\3.8.10\Lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\ab\git\test\venv\lib\site-packages\piptools\__main__.py", line 17, in <module>
    cli()
  File "C:\Users\ab\git\test\venv\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\ab\git\test\venv\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "C:\Users\ab\git\test\venv\lib\site-packages\click\core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Users\ab\git\test\venv\lib\site-packages\click\core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\ab\git\test\venv\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "C:\Users\ab\git\test\venv\lib\site-packages\click\decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "C:\Users\ab\git\test\venv\lib\site-packages\piptools\scripts\compile.py", line 408, in cli
    dist = meta.load(os.path.dirname(os.path.abspath(src_file)))
  File "C:\Users\ab\git\test\venv\lib\site-packages\pep517\meta.py", line 71, in load
    path = Path(build_as_zip(builder))
  File "C:\Users\ab\git\test\venv\lib\site-packages\pep517\meta.py", line 58, in build_as_zip
    builder(dest=out_dir)
  File "C:\Users\ab\git\test\venv\lib\site-packages\pep517\meta.py", line 53, in build
    _prep_meta(hooks, env, dest)
[...]

Issue #1375, opened 2021-04-04
`test_input_formats` failing on pypy3


atugushev:
Introduced in #1356.

```console
2021-04-04T00:25:09.6159906Z         out = runner.invoke(cli, ["-n", "--find-links", dists_path, path])
2021-04-04T00:25:09.6160556Z >       assert out.exit_code == 0, out.stderr
2021-04-04T00:25:09.6161090Z E       AssertionError: 
2021-04-04T00:25:09.6161502Z E       assert 1 == 0
2021-04-04T00:25:09.6163174Z E        +  where 1 = <Result BackendUnavailable('Traceback (most recent call last):\n  File "/home/runner/work/pip-tools/pip-tools/.tox/pip... importlib._bootstrap>", line 962, in _find_and_load_unlocked\nModuleNotFoundError: No module named \'flit_core\'\n',)>.exit_code

...

2021-04-04T00:25:10.3688000Z >       assert out.exit_code == 0, out.stderr
2021-04-04T00:25:10.3688554Z E       AssertionError: 
2021-04-04T00:25:10.3689037Z E       assert 1 == 0
2021-04-04T00:25:10.3690850Z E        +  where 1 = <Result BackendUnavailable('Traceback (most recent call last):\n  File "/home/runner/work/pip-tools/pip-tools/.tox/pip...zen importlib._bootstrap>", line 962, in _find_and_load_unlocked\nModuleNotFoundError: No module named \'poetry\'\n',)>.exit_code
```

See [logs](https://github.com/jazzband/pip-tools/runs/2262017812?check_suite_focus=true) for details.

/cc: @orsinium

webknjaz:
> No module named \'poetry\'\n',)

Looks like the trailing quote and line feed get into the name during parsing?

@atugushev why don't we run pypy jobs on PRs? They are not that heavy and being executed in parallel with the other jobs, the overall workflow completion would be barely noticeable.
Besides, I noticed that there's no caching set up in GHA, fixing this should make things even faster.

webknjaz:
Oh, and if we ever get that "auto-merge" feature on, there will be even fewer reasons to care about the CI completion time. cc @jezdez

atugushev:
@webknjaz 

We moved `pypy` to the cron workflow because they were too slow. However, there were `pypy2` jobs also at that time, but now there are only two jobs. It would be great to improve speed and move these jobs to PRs test workflow.

[...]

Issue #1148, opened 2020-05-17
Xfail test_realistic_complex_sub_dependencies on pypy windows


atugushev:
>           # TODO: fix test_realistic_complex_sub_dependencies test on Windows

this could be an xfailing test

_Originally posted by @webknjaz in https://github.com/jazzband/pip-tools/pull/1147#issuecomment-629831688_

DiddiLeija:
Hi @atugushev. Is this issue resolved? If not, I can work on it. Just tell me where to go.

atugushev:
Hello @DiddiLeija,

It's not yet resolved. This test should be marked as `xfail` for the Windows platform so that we can uncomment the Windows platform on CI.

DiddiLeija:
Ok. I'll be opening a PR the next week.

DiddiLeija:
I created #1489 to fix this issue.

PR #1629, opened 2022-05-25
Replace direct usage of the `pep517` module with the `build` module, for loading project metadata

AndydeCleyre:
The `build` module is better supported at this time, and going forward.

The change looks straightforward, except for dropped support of `setup.cfg`-only projects.

From the `build` [changelog](https://pypa-build.readthedocs.io/en/latest/changelog.html?highlight=setup.cfg#id25):

> As a side-effect of [PR #260](https://github.com/pypa/build/pull/260), projects not containing either a `pyproject.toml` or `setup.py` will be reported as invalid. This affects projects specifying only a `setup.cfg`, such projects are recommended to add a `pyproject.toml`. The new behavior is on par with what pip currently does, so if you are affected by this, your project should not be pip installable.

So should pip-tools also drop support for `setup.cfg`-only projects, and adjust our test case projects accordingly?

That's what I'm going to do in this PR until/unless someone objects.

Fixes:
- #1375
- (the need for) #1148
- #1627

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -6,6 +6,7 @@ import tempfile
 from typing import IO, Any, BinaryIO, List, Optional, Tuple, Union, cast
 
 import click
+from build import BuildBackendException
 from build.util import project_wheel_metadata
 from click.utils import LazyFile, safecall
 from pip._internal.commands import create_command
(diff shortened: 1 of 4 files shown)


atugushev (approved):
Awesome! Most API in `pep517` is deprecated in favor of `build`. Thanks for the improvement 🎉

PR #1630, opened 2022-05-26
Allow wildcard argument for `--extra`

apljungquist:
The goal is to make it easy to compile version locks for all runtime
dependencies in a way that is robust to human error. By using the
wildcard extra the risk of forgetting to list any extra is removed.

Is this a use case pip-tools wants to support?
If yes, is the current approach a good one?

##### Contributor checklist

##### Maintainer checklist


tests/test_cli_compile.py
```diff
@@ -2100,6 +2100,35 @@ def test_multiple_extras(fake_dists, runner, make_module, fname, content, extra_
     assert "extra ==" not in out.stderr
 
 
+@pytest.mark.network
+@pytest.mark.parametrize(("fname", "content"), METADATA_TEST_CASES)
+@pytest.mark.xfail(is_pypy, reason="https://github.com/jazzband/pip-tools/issues/1375")
```
atugushev, line 2105:
```suggestion
```

#1375 is closed so the xfail marker could be removed.

tests/test_cli_compile.py
```diff
@@ -2100,6 +2100,35 @@ def test_multiple_extras(fake_dists, runner, make_module, fname, content, extra_
...
+        ],
+    )
+    assert out.exit_code == 0, out.stderr
+    assert "small-fake-a==0.1" in out.stderr
+    assert "small-fake-b==0.2" in out.stderr
+    assert "small-fake-c==0.3" in out.stderr
+    assert "small-fake-d==0.4" in out.stderr
+    assert "small-fake-e==0.5" in out.stderr
+    assert "small-fake-f==0.6" in out.stderr
+    assert "extra ==" not in out.stderr
```
atugushev, line 2129:
```suggestion
    assert out.stderr == expected
```

Could you please check the whole output?

atugushev (commented):
I think this is a great idea! A few comments.

tests/test_cli_compile.py
apljungquist: Certainly. Do you think it is better to save the output to a file and read it or as a string literal in the test?
A third idea is to implement a parser, if there is not one already, and do the comparison in the parsed domain. This way the formatting, and any future changes to it, does not affect the test.

Should we also update similar tests such as `test_multiple_extras`?

apljungquist: Sorry for leaving this unfinished. There's another test that uses a string literal so that is what I did as well. Anything else that can be improved?

atugushev: It's all good now! Thanks 👍🏻

piptools/scripts/compile.py
```diff
@@ -442,6 +444,14 @@ def cli(
                     for req in metadata.get_all("Requires-Dist") or []
                 ]
             )
+            # Since "*" is not a valid extra there is no risk confusing an actual extra
+            # for the wildcard.
+            # https://peps.python.org/pep-0508/#grammar
+            if "*" in extras:
```
atugushev, line 450: What do you think, shouldn't we use an explicit option (e.g., `--all-extras`) to avoid making users think that a wildcard can be used as a regexp like `--extras=foo*`? Also, as commented, "'*' is not a valid extra".

apljungquist: Good point about the pattern patching, will fix.

atugushev: Awesome, thanks!

atugushev (approved):
LGTM 👍🏻 

@apljungquist could you please update the PR title?

apljungquist:
Certainly.

Thanks for being patient and guiding me through the process.

tests/test_cli_compile.py
```diff
@@ -2190,6 +2190,70 @@ def test_multiple_extras(fake_dists, runner, make_module, fname, content, extra_
...
+        dedent(
+            """\
+        small-fake-a==0.1
+        small-fake-b==0.2
+        small-fake-c==0.3
+        small-fake-d==0.4
+        small-fake-e==0.5
+        small-fake-f==0.6
+        Dry-run, so nothing updated.
+        """
```
webknjaz, line 2224: This should probably be indented.
```suggestion
            small-fake-a==0.1
            small-fake-b==0.2
            small-fake-c==0.3
            small-fake-d==0.4
            small-fake-e==0.5
            small-fake-f==0.6
            Dry-run, so nothing updated.
            """
```

ssbarnea (approved):

Issue #1633, opened 2022-05-30
I would like the output of help to be written in READEME.


29rou:
#### What's the problem this feature will solve?

This is so that those who are wondering whether to use this tool can understand how to use it at a glance. I also believe that this will encourage more people to use this tool.

#### Describe the solution you'd like

Describe the output of help in READEME.

#### Alternative Solutions

Add a man page to the documentation.

#1634

atugushev:
@29rou FYI, there is a PR (#1934) where CLI reference has been included in the pip-tools' documentation.

PR #1634, opened 2022-05-30
Describe the output of help in reademe.

29rou:
Fixes 
- Add pip-compile help in README. 
- Add pip-sync help in README.

#1633 

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -560,6 +560,32 @@ Usage
                                       Add options to generated file
      -h, --help                       Show this message and exit.
 
+::
+
+   Usage: 
+       pip-sync [OPTIONS] [SRC_FILES]...
+
+        Synchronize virtual environment with
+        requirements.txt.
+
+   Options:
+      --version                 Show the version and exit.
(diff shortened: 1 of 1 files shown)


atugushev (commented):
I'm not sure this is a good idea. At least this should be generated automatically somehow. Otherwise, the text would be outdated soon.

29rou:
I agree, it would be better to be able to automatically put the output of help in the README.

29rou:
I agree with using pre-commit-hooks. Therefore, I will try to set up pre-commit-hooks.

merwok:
What about including only the usage line in README?  Full help output would make a long README even longer + description on PyPI.

(If you do decide to go for full help output, then [cog](https://nedbatchelder.com/code/cog) may help)

ssbarnea:
I am going to close this for the stated reasons. I am not against including output of --help into docs but not on readme, unless the implementation (lint?) will update the readme automatically. I do not want anyone to have to remember that this output needs to updated on each merge.

PR #1635, opened 2022-05-31
Remove duplicate PyPIRepository().finder.find_all_candidates cache

gschaffner:
Hi! I believe that `PyPIRepository()._available_candidates_cache` is redundant, as `PackageFinder().find_all_candidates` has been `lru_cache`d since Pip 20.3 (https://github.com/pypa/pip/commit/1dd6d562789b06dc1508212505e32f36bd71d310).

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -5,7 +5,9 @@ import hashlib
 import itertools
 import optparse
 import os
+import sys
 from contextlib import contextmanager
+from functools import lru_cache
 from shutil import rmtree
 from typing import Any, BinaryIO, ContextManager, Iterator, NamedTuple
 
@@ -78,11 +80,6 @@ class PyPIRepository(BaseRepository):
         )
 
         # Caches
-        # stores project_name => InstallationCandidate mappings for all
-        # versions reported by PyPI, so we only have to ask once for each
-        # project
-        self._available_candidates_cache: dict[str, list[InstallationCandidate]] = {}
-
         # stores InstallRequirement => list(InstallRequirement) mappings
         # of all secondary dependencies for the given requirement, so we
         # only have to go to disk once for each requirement
@@ -119,12 +116,6 @@ class PyPIRepository(BaseRepository):
         """Return an install command instance."""
         return self._command
 
-    def find_all_candidates(self, req_name: str) -> list[InstallationCandidate]:
-        if req_name not in self._available_candidates_cache:
-            candidates = self.finder.find_all_candidates(req_name)
-            self._available_candidates_cache[req_name] = candidates
-        return self._available_candidates_cache[req_name]
-
     def find_best_match(
         self, ireq: InstallRequirement, prereleases: bool | None = None
     ) -> InstallRequirement:
@@ -135,7 +126,7 @@ class PyPIRepository(BaseRepository):
         if ireq.editable or is_url_requirement(ireq):
             return ireq  # return itself as the best match
 
-        all_candidates = self.find_all_candidates(ireq.name)
+        all_candidates = self.finder.find_all_candidates(ireq.name)
         candidates_by_version = lookup_table(all_candidates, key=candidate_version)
         matching_versions = ireq.specifier.filter(
             (candidate.version for candidate in all_candidates), prereleases=prereleases
@@ -381,7 +372,7 @@ class PyPIRepository(BaseRepository):
         # We need to get all of the candidates that match our current version
         # pin, these will represent all of the files that could possibly
         # satisfy this constraint.
-        all_candidates = self.find_all_candidates(ireq.name)
+        all_candidates = self.finder.find_all_candidates(ireq.name)
         candidates_by_version = lookup_table(all_candidates, key=candidate_version)
         matching_versions = list(
             ireq.specifier.filter(candidate.version for candidate in all_candidates)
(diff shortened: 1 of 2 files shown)


piptools/repositories/pypi.py
```diff
@@ -434,22 +424,23 @@ def _wheel_support_index_min(self: Wheel, tags: List[Tag]) -> int:
...
         Wheel.support_index_min = _wheel_support_index_min
-        self._available_candidates_cache = {}
-
         # If we don't clear this cache then it can contain results from an
         # earlier call when allow_all_wheels wasn't active. See GH-1532
-        self.finder.find_all_candidates.cache_clear()
+        # TODO: Python >= 3.9: Use `lru_cache().cache_parameters()`.
+        self.finder.find_all_candidates = lru_cache(
+            maxsize=self.finder.find_all_candidates.cache_info().maxsize
+        )(self.finder.find_all_candidates.__wrapped__).__get__(self.finder)
```
atugushev, line 436: Could you elaborate on this block? This part looks tricky.

gschaffner: Sure!

`self.finder.find_all_candidates` is an `lru_cached` instance method with signature `(req_name: str) -> ...`. The function that the the cache wraps, i.e. `self.finder.find_all_candidates.__wrapped__` (https://github.com/pypa/pip/blob/ca2d9f41931a449b8c1b27d02031199d91af93e7/src/pip/_internal/index/package_finder.py#L810-L860; note that L809 is not included in this highlight) has signature `(self: PackageFinder, req_name: str) -> ...`. Note the additional `self` there—the L810–L860 function that `@lru_cache(maxsize=None)` wraps has `self` in its signature still.<sup>1</sup>

`x = lru_cache(maxsize=self.finder.find_all_candidates.cache_info().maxsize)(self.finder.find_all_candidates.__wrapped__)` creates a new `lru_cache`d version of the same underlying wrapped function (L810–L860) with the same caching parameters, so that we have a fresh cache to use while the context manager `allow_all_wheels` is open. Note that `x` has the same signature as the wrapped function, i.e. it still needs `self` as its first argument. If we just did `finder.x = x`, we'd have to call it as `finder.x(finder, "foo")`, since it still needs `self` as an argument. To actually bind `x` as a method of `finder`, we need to invoke it as a descriptor `finder.x = x.__get__(finder)`.

If you prefer, `finder.x = types.MethodType(finder, x)` would be equivalent to `finder.x = x.__get__(finder)` here.

[Descriptors](https://docs.python.org/3/glossary.html#term-descriptor) can definitely be a bit tricky, but they're pretty core to the language. https://docs.python.org/3/howto/descriptor.html#simple-example-a-descriptor-that-returns-a-constant and https://docs.python.org/3/howto/descriptor.html#functions-and-methods definitely explains `__get__` descriptors better than I can if you haven't seen them (explicitly) before. (You have _definitely_ seen them implicitly though–every instance/class method is the result of a `__get__` binding!!)

<sup>1</sup>: <details>
<summary>Click to expand footnote</summary>
  
Aside: `lru_cache`ing instance/class methods can cause garbage collector deadlocks if one is not careful! E.g.

```
class Foo:
   @lru_cache(maxsize=None)
   def bar(self, x: int): ...
```

This creates a single cache, not a separate cache for each instance of `Foo`! The single cache is keyed on `(self, x)`, though, not just on `(x,)`.

Doing

```
foo = Foo()
print(foo)  # repr(foo) == <Foo object at 0x1234>
foo.bar(0)
```

will add new entry to the cache, keyed on `(foo, 0)`. Note that the instance `foo` is part of the new cache key. Now suppose we either do `foo = something_else`, or we leave the context of the variable `foo`. Both of these would *usually* make `<Foo object at 0x1234>` ready to be garbage collected. Here, however, our cache is still keeping around a reference to `Foo` and prevent garbage collection. This is a memory leak if `maxsize=None`!
</details>

piptools/repositories/pypi.py
```diff
@@ -434,22 +424,23 @@ def _wheel_support_index_min(self: Wheel, tags: List[Tag]) -> int:
...
 
         try:
             yield
         finally:
             Wheel.supported = original_wheel_supported
             Wheel.support_index_min = original_support_index_min
-            self._available_candidates_cache = original_cache
+            self.finder.find_all_candidates = original_candidates_cache
```
gschaffner, line 443: A minor note:

`self.finder.find_all_candidates = original_candidates_cache` is not really excellent code IMO, as it doesn't always leave `self.finder` in the same state that `self.finder` was in before `allow_all_wheels(...)` was called. `self.finder.find_all_candidates = original_candidates_cache` adds `{"find_all_candidates": original_candidates_cache}` to `self.finder.__dict__`. For the original `self.finder`, though, `self.finder.__dict__` was empty, and `self.finder.find_all_candidates` would have been resolved from the MRO instead.

`self.finder.find_all_candidates = original_candidates_cache` could be replaced with `del self.finder.find_all_candidates` to avoid polluting `self.finder.__dict__`. Functionally these are the same, but I do think `del` is a whole lot less readable.

ssbarnea:
@gschaffner Can you please revive or close this one? We are trying to cleanup the PR queue.

atugushev (commented):
This change is related to a method that is currently only utilized by the legacy resolver, which has been deprecated and could possibly be removed in the future. Considering this, it seems impractical to improve this deprecated code. Based on these reasons, I'm going to close the PR. However, I greatly appreciate your work and your contribution. Thank you!

PR #1649, opened 2022-06-30
Fix compile cached vcs packages

### LINKED ISSUE #1647: `@ git+` requirements with backtracking resolver can result in pins to local wheel cache
If an `@ git+` requirement is present in pip's wheel cache, `pip-compile --resolver backtracking` will pin to `@ file://....whl`.

(Moved from https://github.com/jazzband/pip-tools/pull/1539#issuecomment-1170845933. `@ git+` pinned package switched from `pip-tools` to `typeguard` for clarity.)

#### Environment Versions

1. OS Type: Manjaro 21.3.1
1. Python version: CPython 3.10.5
1. pip version: 22.1.2
1. pip-tools version: 1e05d0028ce16d4b2de53ae38a33df7feb84e800

#### Steps to replicate

```bash
#!/usr/bin/env bash

set -Eeuo pipefail
shopt -s inherit_errexit

cd "$(mktemp -d)"

echo 'typeguard @ git+https://github.com/agronholm/typeguard.git@a298d6cde249fe7ad46b90b94aa3bdd246e60417' > requirements.in
# cache ~/.cache/pip/wheels/a3/ac/4a/1aa2eba1bf7a5f0259fbc2f4160621fe8ea72613653e32c6fe/typeguard-2.13.3.post48-py3-none-any.whl
python -m venv --upgrade-deps .venv && . .venv/bin/activate
mv ~/.cache/pip/wheels{,.bak}
pip install typeguard==2.13.3
pip install -r requirements.in
# pip-compile
rm -r .venv
python -m venv --upgrade-deps .venv && . .venv/bin/activate
pip install 'pip-tools @ git+https://github.com/jazzband/pip-tools.git@1e05d0028ce16d4b2de53ae38a33df7feb84e800'
pip-compile --resolver backtracking
# restore cache
rm -r ~/.cache/pip/wheels
mv ~/.cache/pip/wheels{.bak,}
```

#### Expected result
```
typeguard @ git+https://github.com/agronholm/typeguard.git@a298d6cde249fe7ad46b90b94aa3bdd246e60417
```

#### Actual result

```
< ... truncated for brevity ... >
typeguard @ file:///home/ganden/.cache/pip/wheels/a3/ac/4a/1aa2eba1bf7a5f0259fbc2f4160621fe8ea72613653e32c6fe/typeguard-2.13.3.post48-py3-none-any.whl
    # via -r requirements.in
```

atugushev:
Fixes #1647 

##### Contributor checklist

##### Maintainer checklist


piptools/resolver.py
```diff
@@ -763,7 +763,14 @@ def _get_install_requirement_from_candidate(
...
         # Prepare pinned install requirement. Copy it from candidate's install
         # requirement so that it could be mutated later.
-        pinned_ireq = copy_install_requirement(ireq)
+        pinned_ireq = copy_install_requirement(
+            template=ireq,
+            # The link this candidate "originates" from. This is different
+            # from ``ireq.link`` when the link is found in the wheel cache.
+            # ``ireq.link`` would point to the wheel cache, while this points
+            # to the found remote link (e.g. from pypi.org).
+            link=candidate.source_link,
```
richafrank, line 772: Glad your `copy_install_requirement` extensibility made this easy!

atugushev: That's very handy indeed!

huajie2020 (approved):

tests/test_cli_compile.py
```diff
@@ -511,6 +511,39 @@ def test_editable_package_vcs(runner):
...
+    # TODO: install package in isolated environment
+    subprocess.run(
+        [
+            sys.executable,
+            "-m" "pip",
+            "install",
+            vcs_package,
+        ],
+        check=True,
+    )
```
gschaffner, line 536: On my machine, this doesn't cause a wheel for six to be cached in `~/.cache/pip/wheels`. I did a quick sanity with `[sys.executable, "-m", "pip", "cache", "dir"]` here to confirm that `~/.cache/pip/wheels` is the relevant cache when running under `tox -e py310-piplatest-coverage`.

I had a similar issue when writing the MWE script in #1647. Pip would not cache the wheel unless I installed another version of the VCS package first.

Doing that doesn't seem to fix the issue here though. On my machine, replacing `typeguard @ ...` with `six @ git+https://github.com/benjaminp/six@65486e4383f9f411da95937451205d3c7b61b9e1` in my MWE script doesn't actually cause the issue to occur—for some reason I can't get Pip to cache a wheel for this version of six.

gschaffner: Suggested changes https://github.com/jazzband/pip-tools/compare/atugushev:fix-compile-cached-vcs-1647...bb2e91edbd3d0f1bdd86c41e84c88621f8d35690 to check that the wheel was actually cached.

bb2e91edbd3d0f1bdd86c41e84c88621f8d35690 is failing though—I can't find a way to get Pip to cache `six-1.16.0-py2.py3*.whl`. I am not sure why<sup>1</sup>, but the easiest fix is probably to change to another package that supports 3.7–3.10, has no dependencies for this range of Pythons, and that Pip will cache a wheel for, e.g. Plumbum 1.7.2 (suggested changes: 2f311c89dbcf9cfa9ebc3f28758081859f43d1c2). As far as I can tell, there isn't another candidate for this that's already used in pip-tools's suite elsewhere (the list of `git+`'s used elsewhere seems to be pip-tools, django, pip, and billiard).

<sup>1</sup>: My guess is that this is because six does not use `setuptools_scm`, but I haven't looked into this.

atugushev: The test works on CI. As you can see it [fails](https://github.com/jazzband/pip-tools/runs/7222931098?check_suite_focus=true#step:10:75) without the fix:
```
E       AssertionError: assert 'six @ git+ht...05d3c7b61b9e1' == 'six @ file:/...-none-any.whl'
[75](https://github.com/jazzband/pip-tools/runs/7222931098?check_suite_focus=true#step:10:76)
E         - six @ file:///home/runner/.cache/pip/wheels/59/4b/31/545573687bd39ded5dd79f7707ebb72b5029fd67236734b138/six-1.16.0-py2.py3-none-any.whl
[76](https://github.com/jazzband/pip-tools/runs/7222931098?check_suite_focus=true#step:10:77)
E         + six @ git+https://github.com/benjaminp/six@65486e4383f9f411da95937451205d3c7b61b9e1
```

I think that should be enough for now.

atugushev: @gschaffner what would you think?

gschaffner: Sorry for the delay!

Without the fix, the test mistakenly passes on my laptop when run as

```shellsession
$ git checkout 53f36d88bbaaad3ece3e964b015015690c6792e9  # or 835b7e58c4fb929e114d88fce7e1f911aef2c4cd
$ rm -r ~/.cache/pip/wheels
$ tox -e py310-piplatest-coverage -- --no-cov tests/test_cli_compile.py::test_compile_cached_vcs_package  # or tox -e py310-piplatest-coverage
```

Can you replicate? It seems like the test is not putting six into the cache.

(My initial guess was that the test was failing (as it should) on GHA because six was already in the wheels cache loaded by

https://github.com/jazzband/pip-tools/blob/53f36d88bbaaad3ece3e964b015015690c6792e9/.github/workflows/ci.yml#L61-L68

This doesn't seem to be the case however—see that there is no wheel for six loaded in https://github.com/jazzband/pip-tools/runs/7439890847?check_suite_focus=true#step:7:223.)

AndydeCleyre: Copying the test into master locally also passes for me, but I found that it properly fails due to a local cache path being compiled if I change the package to:

```python
    vcs_package = (
        "plumbum @ git+https://github.com/tomerfiliba/plumbum@"
        "70380ddc5ad346ab6758506f5000864260a9c418"
    )
```

gschaffner: In case you do decide to use https://github.com/atugushev/pip-tools/compare/fix-compile-cached-vcs-1647...gschaffner:pip-tools:fix-compile-cached-vcs-1647 (since it has a check to ensure the wheel is cached), I just switched it from plumbum to typing-extensions to avoid third party.

I think pip does not create and cache a wheel for six because six does not specify a wheel-supporting PEP 517 `build-system`, so `pip install six @ git+...` essentially just runs `git clone ... && setup.py install`, which doesn't produce a wheel.

AndydeCleyre: I'm very eager for this to go through and get released, and am happy to approve once the test is made more reliable.

atugushev: > Without the fix, the test mistakenly passes
> Can you replicate? It seems like the test is not putting six into the cache.

Can't reproduce. The test fails as expected:

```
$ git checkout 53f36d88bbaaad3ece3e964b015015690c6792e9
$ rm -fr $(pip cache dir)
$ tox -e py38-piplatest-coverage -- --no-cov tests/test_cli_compile.py::test_compile_cached_vcs_package
>       assert vcs_package == out.stderr.strip()
E       AssertionError: assert 'six @ git+ht...05d3c7b61b9e1' == 'six @ file:/...-none-any.whl'
E         - six @ file:///Users/albert/Library/Caches/pip/wheels/b4/42/6d/bab90ef8caaba6e66b7c1b7e35d83fbe1b67ae9a1101839ff2/six-1.16.0-py2.py3-none-any.whl
E         + six @ git+https://github.com/benjaminp/six@65486e4383f9f411da95937451205d3c7b61b9e1

/Users/albert/Projects/pip-tools/tests/test_cli_compile.py:544: AssertionError
```

atugushev: @AndydeCleyre we can release a fix without a test (the fix is pretty trivial) if you feel that the test is not reliable.

gschaffner: This container seems to be a reproducible way to make the test pass when it should not:

```shellsession
$ [podman|docker] run --rm -it nixery.dev/shell/git/python310 bash <<EOF
set -Eeuo pipefail
shopt -s inherit_errexit

git init -b main pip-tools
cd pip-tools
git remote add origin https://github.com/jazzband/pip-tools.git
git pull --depth 1 origin 53f36d88bbaaad3ece3e964b015015690c6792e9
git log -1

python -m venv --upgrade-deps .venv
. .venv/bin/activate
pip install tox

tox -e py310-piplatest-coverage -- -n 0 --no-cov -v tests/test_cli_compile.py::test_compile_cached_vcs_package
exit
EOF
```

atugushev: Thanks for the reproducer! I'll look into it.

atugushev: I'd like to suggest merging this fix without tests and addressing it in a separate PR. What would you think? Don't want to block this pretty easy one-line fix due to complicated flaky tests.

AndydeCleyre: Ok, but isn't it all fine if you just replace six with any other package?

atugushev: That was my first thought. However, while the test works on CI, it can [mistakenly pass](https://github.com/jazzband/pip-tools/pull/1649#discussion_r926146005) in a [different](https://github.com/jazzband/pip-tools/pull/1649#discussion_r939591090) setup. This makes me think that the test is flaky and must be reworked in a way as suggested [here](https://github.com/jazzband/pip-tools/pull/1649#discussion_r912303168).

AndydeCleyre: Ok, thanks, yeah I'm on board to get the fix out ahead of the test.

atugushev: Cross-link https://github.com/jazzband/pip-tools/issues/1686#issue-1393116474

atugushev: @AndydeCleyre okay, all set.

gschaffner (approved):
LGTM. If it's helpful, I'd be happy open a PR for #1686 based on https://github.com/jazzband/pip-tools/commit/6f0d7dc5a8dbcd2dddfee47e21772b3d99e6ea41.

atugushev:
@gschaffner thanks! Feel free to.

PR #1652, opened 2022-07-03
Add --newline=[LF|CRLF|native|preserve] to compile

AndydeCleyre:
`pip-compile` gains an option with four valid choices: `--newline=[LF|CRLF|native|preserve]`,
which can be used to override the guessed newline character used in the output file. The default is `preserve`, which tries to be consistent with an existing output file, or input file, or falls back to `LF`, in that order.

This aims to address #1448.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -49,6 +49,35 @@ def _get_default_option(option_name: str) -> Any:
     return getattr(default_values, option_name)
 
 
+def _determine_linesep(
+    strategy: str = "preserve", filenames: Tuple[str, ...] = ()
+) -> str:
+    """
+    Determine and return linesep string for OutputWriter to use.
+    Valid strategies: "LF", "CRLF", "native", "preserve"
+    When preserving, files are checked in order for existing newlines.
+    """
+    if strategy == "preserve":
+        for fname in filenames:
+            try:
+                with open(fname, "rb") as existing_file:
+                    existing_text = existing_file.read()
+            except FileNotFoundError:
+                continue
+            if b"\r\n" in existing_text:
+                strategy = "CRLF"
+                break
+            elif b"\n" in existing_text:
+                strategy = "LF"
+                break
+    return {
+        "native": os.linesep,
+        "LF": "\n",
+        "CRLF": "\r\n",
+        "preserve": "\n",
+    }[strategy]
+
+
 @click.command(context_settings={"help_option_names": ("-h", "--help")})
 @click.version_option(**version_option_kwargs)
 @click.pass_context
@@ -510,24 +539,9 @@ def cli(
 
     log.debug("")
 
-    # Determine linesep for OutputWriter to use
-    if newline == "preserve":
-        for fname in (output_file.name, *src_files):
-            if os.path.exists(fname):
-                with open(fname, "rb") as existing_file:
-                    existing_text = existing_file.read().decode()
-                if "\r\n" in existing_text:
-                    newline = "CRLF"
-                    break
-                elif "\n" in existing_text:
-                    newline = "LF"
-                    break
-    linesep = {
-        "native": os.linesep,
-        "LF": "\n",
-        "CRLF": "\r\n",
-        "preserve": "\n",
-    }[newline]
+    linesep = _determine_linesep(
+        strategy=newline, filenames=(output_file.name, *src_files)
+    )
 
     ##
     # Output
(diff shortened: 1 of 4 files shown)


AndydeCleyre:
See also: #1584, which adds a commit on top of this, reducing the options to a single `--force-unix-newlines` flag.

graingert:
My review was requested on this, and I really want the feature and the code looks great to me - but I don't feel like I'm in a position to approve this PR

ssbarnea:
@graingert If all those conditions are met, feel free to make your review official. As I said, I am bit of 0.5 on this, so we will need two approvals to merge it. I do not want to alienate the author, but at the same time I cannot refrain from seeing this feature as adding useless complexity, as use of git attributes looks like a valid workaround.

AndydeCleyre:
@ssbarnea Please don't worry about alienating me; this is not even a feature I'm interested in personally. And don't forget there's the slightly simpler alternative at #1584.

piptools/scripts/compile.py
```diff
@@ -500,6 +507,25 @@ def cli(
 
     log.debug("")
 
+    # Determine linesep for OutputWriter to use
+    if newline == "preserve":
+        for fname in (output_file.name, *src_files):
+            if os.path.exists(fname):
+                with open(fname, "rb") as existing_file:
+                    existing_text = existing_file.read().decode()
```
graingert, line 515: this keeps a reference to the existing_text for the duration of the program - which might be quite large

graingert:
```python
def determine_linesep(newline, output_file, src_files):
    # Determine linesep for OutputWriter to use
    if newline == "preserve":
        for fname in (output_file.name, *src_files):
            try:
                with open(fname, "rb") as existing_file:
                    existing_text = existing_file.read().decode()
            except FileNotFoundError:
                break
            else:
                if "\r\n" in existing_text:
                    return "\r\n"
                if "\n" in existing_text:
                    return "\n"
    return {
        "native": os.linesep,
        "LF": "\n",
        "CRLF": "\r\n",
        "preserve": "\n",
    }[newline]
```

graingert (changes requested):

graingert (approved):

piptools/writer.py
```diff
@@ -257,14 +260,25 @@ def write(
...
 
-        for line in self._iter_lines(results, unsafe_requirements, markers, hashes):
-            if self.dry_run:
-                # Bypass the log level to always print this during a dry run
-                log.log(line)
-            else:
-                log.info(line)
-                self.dst_file.write(unstyle(line).encode())
-                self.dst_file.write(os.linesep.encode())
+        if not self.dry_run:
```
webknjaz, line 263: FWIW this could be cleaner if implemented with a context manager conditionally returning a dummy I/O object.

tests/test_cli_compile.py
```diff
@@ -936,6 +936,88 @@ def test_generate_hashes_with_annotations(runner):
...
+    ),
+)
+def test_override_newline(
+    runner, gen_hashes, annotate_options, nl_options, must_include, must_exclude
+):
+    opts = annotate_options + nl_options
+    if gen_hashes:
+        opts += ("--generate-hashes",)
+
+    with open("requirements.in", "w") as req_in:
```
webknjaz, line 969: I feel like with modern Pythons it'd be nicer if `Pathlib.write_text()` was used. But more importantly, why don't you use the built-in `tmp_path` fixture to write temporary files?

webknjaz (approved):
Aside for a few refactoring suggestions, this LGTM.

tests/test_cli_compile.py
AndydeCleyre: Feel free to make any changes you like, and thanks! This change isn't for me personally, and I have COVID right now and can't bring my full focus to the task.

graingert: I'll try and make a PR with this fix in!

piptools/writer.py
graingert: this refactor requires `from __future__ import annotations` to get flake8 to pass: https://github.com/jazzband/pip-tools/pull/1696

tests/test_cli_compile.py
graingert: done here: https://github.com/jazzband/pip-tools/pull/1694

PR #1693, opened 2022-10-05
Update changelog for 6.9.0 release

ssbarnea:


atugushev (approved):

webknjaz (commented):
I think the changelog should be well-formatted and use the past tense that appeared/changed since the previous release. It shouldn't use imperative mood. Its consumers are the end-users, after all.

PR #1724, opened 2022-11-11
Deprecate `pip-compile --resolver=legacy`

### LINKED ISSUE #1659: Deprecate legacy resolver

Pip's legacy resolver is going to be removed any time soon, let's prepare for it.

Deprecate the legacy resolver and remove it as soon as pip removes its own legacy resolver.

#### Alternative Solutions

Vendoring?. We've gone that path before but that didn't help.

#### Additional context

- https://github.com/pypa/pip/issues/10946
- https://github.com/pypa/pip/milestone/47

atugushev:
Resolves #1659 

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -327,7 +327,7 @@ def cli(
     emit_find_links: bool,
     cache_dir: str,
     pip_args_str: str | None,
-    resolver_name: str | None,
+    resolver_name: str,
     emit_index_url: bool,
     emit_options: bool,
     unsafe_package: tuple[str, ...],
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -8,13 +8,14 @@ import tempfile
 from typing import IO, Any, BinaryIO, cast
 
 import click
-from build import BuildBackendException
 from build.util import project_wheel_metadata
 from click.utils import LazyFile, safecall
 from pip._internal.commands import create_command
 from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
 from pip._internal.utils.misc import redact_auth_from_url
+
+from build import BuildBackendException
 
 from .._compat import IS_CLICK_VER_8_PLUS, parse_requirements
 from ..cache import DependencyCache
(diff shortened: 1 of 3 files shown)


atugushev:
Now, tests that check the exact output in stderr are failing due to the deprecation warning. I'll address that in a separate PR.

webknjaz:
Maybe try using `warnings.warn(FutureDeprecationWarning)`? And test it using `with pytest.deprecated()`?

atugushev:
> Maybe try using `warnings.warn(FutureDeprecationWarning)`? And test it using `with pytest.deprecated()`?

@webknjaz if so, what's the motivation?

ssbarnea (approved):

piptools/scripts/compile.py
```diff
@@ -378,6 +378,13 @@ def cli(
         if isinstance(output_file, LazyFile):  # pragma: no cover
             ctx.call_on_close(safecall(output_file.close_intelligently))
 
+    if resolver_name == "legacy":
+        log.warning(
```
ssbarnea, line 382: Why not using Deprecation warnings from warning module? Somehow I find it better.

atugushev: https://docs.python.org/3/library/exceptions.html#DeprecationWarning

> DeprecationWarning
> Base class for warnings about deprecated features when those warnings are intended for _other Python developers_.

`DeprecaetionWarning` is intended for developers, but pip-tools prints warning for end-users. Also, it is "Ignored by the default warning filters", so it should be `Warning` or something else.

image

vs

image


Personally, I prefer the latter.

atugushev: But I don't mind using the warning module if you feel that's better.

ssbarnea: I +1 because it was only an observation. I am ok with both approaches. The real benefit of warnings is that they are very easy to silence from outside the application.

webknjaz:
@atugushev the motivation is that it's easier to test. Plus if you actually put the warning in the resolution code, it'd be easier to trace what to remove later.
It could be caught and logged differently in the outer CLI layer, I suppose.
Also, there are third-party libs that make it easier to manage depreciations, they are worth exploring.

PR #1728, opened 2022-11-13
Test on Python 3.12

atugushev:
##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -17,20 +17,32 @@ jobs:
           - Windows
           - MacOS
         python-version:
+          - "3.11"
           - "3.10"
           - "3.9"
           - "3.8"
           - "3.7"
         pip-version:
           - main
+        include:
+          - os: Ubuntu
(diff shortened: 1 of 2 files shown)


ssbarnea (approved):

webknjaz:
@atugushev instead of deadsnakes you could just use `~3.12.0-0` from GH

atugushev:
@webknjaz I followed the pattern in `ci.yaml`. Isn't "deadsnakes"  build's main python nightly which is why we use it?

webknjaz:
Yes and no. Deadsnakes builds CPython's repo branch every day. It is truly unstable.
GitHub builds tagged releases, they are updated less frequently, providing less surprises every day.
Originally, deadsnakes were used because GH was slow to provide alphas and to emulate what Travis CI used to have.
So depending on the motivation, you might want to use one or the other.

atugushev:
@webknjaz thanks for the info. I guess we can switch to GHA completely then. I'll open a PR shortly.

Issue #1671, opened 2022-08-20
CI fails on Windows against pip main branch


atugushev:
It fails on install pip from git `pip install -e git+https://github.com/pypa/pip.git@main#egg=pip`:

```
2022-08-20T00:23:46.5591707Z ERROR: Exception:
2022-08-20T00:23:46.5592085Z Traceback (most recent call last):
2022-08-20T00:23:46.5602182Z   File "D:\a\pip-tools\pip-tools\.tox\pipmain\lib\site-packages\pip\_internal\cli\base_command.py", line 167, in exc_logging_wrapper
2022-08-20T00:23:46.5603563Z     status = run_func(*args)
2022-08-20T00:23:46.5604779Z   File "D:\a\pip-tools\pip-tools\.tox\pipmain\lib\site-packages\pip\_internal\cli\req_command.py", line 247, in wrapper
2022-08-20T00:23:46.5606413Z     return func(self, options, args)
2022-08-20T00:23:46.5607264Z   File "D:\a\pip-tools\pip-tools\.tox\pipmain\lib\site-packages\pip\_internal\commands\install.py", line 461, in run
2022-08-20T00:23:46.5607827Z     installed = install_given_reqs(
2022-08-20T00:23:46.5608584Z   File "D:\a\pip-tools\pip-tools\.tox\pipmain\lib\site-packages\pip\_internal\req\__init__.py", line 73, in install_given_reqs
2022-08-20T00:23:46.5609135Z     requirement.install(
2022-08-20T00:23:46.5609866Z   File "D:\a\pip-tools\pip-tools\.tox\pipmain\lib\site-packages\pip\_internal\req\req_install.py", line 790, in install
2022-08-20T00:23:46.5610661Z     install_wheel(
2022-08-20T00:23:46.5611440Z   File "D:\a\pip-tools\pip-tools\.tox\pipmain\lib\site-packages\pip\_internal\operations\install\wheel.py", line 727, in install_wheel
2022-08-20T00:23:46.5611993Z     _install_wheel(
2022-08-20T00:23:46.5612745Z   File "D:\a\pip-tools\pip-tools\.tox\pipmain\lib\site-packages\pip\_internal\operations\install\wheel.py", line 644, in _install_wheel
2022-08-20T00:23:46.5613393Z     generated_console_scripts = maker.make_multiple(scripts_to_generate)
[...]

PR #1736, opened 2022-11-20
Fix CI tests on Windows against pip main branch

atugushev:
Fixes https://github.com/jazzband/pip-tools/issues/1671.

[Tested](https://github.com/jazzband/pip-tools/actions/runs/3505927878/jobs/5872595359) on 0081f1fded86828f90d80561f7f824281973d2a6.

##### Contributor checklist

##### Maintainer checklist


atugushev:
One more thing to address - don't use cache when install pip main zip archive.

tox.ini
```diff
@@ -13,7 +13,7 @@ extras =
 deps =
     pipprevious: pip==22.2.*
     piplatest: pip
-    pipmain: -e git+https://github.com/pypa/pip.git@main#egg=pip
+    pipmain: https://github.com/pypa/pip/archive/main.zip
```
ssbarnea, line 16: That looks interesting! I never knew this was possible. What are the benefits? Will it install newer pip when something new is updated in pip? Will it avoid reinstalling pip when the archive is not changed? Any side-effects on packaging?

-- Mainly asking because I want to know if I should use the same approach in my projects.

atugushev: > I never knew this was possible.

I've [seen](https://sourcegraph.com/search?q=context:global+file:%5Etox%5C.ini%24+%28master%7Cmain%29.zip&patternType=regexp&sm=1) this mostly in Django projects.

> What are the benefits?

Back in the day it was [faster](https://stackoverflow.com/a/24811490/1529164) than installing from git, especially with submodules. I've tested locally for pip-tools and it's slightly faster.

> Will it install newer pip when something new is updated in pip?

I've tested on my own project. It re-installs only if version is updated. I think that's okay for `tox` where you can always recreate environment using `tox --recreate` and get latest updates.

> Will it avoid reinstalling pip when the archive is not changed?

It won't be reinstalled If version is not updated.

> Any side-effects on packaging, versioning in particular?

Not that I know of.

---

Unfortunately I failed to research why pip fails to re-install itself from git on Windows. It's certainly a bug, but I spent too much time and gave up. I've tried to install the zip and it worked like a charm, hence this PR.

webknjaz: @atugushev in your log it's visible that `-e` and the pip repo URL are parsed as a single argument while they should be separate.

webknjaz: So it could also be a tox bug too.

ssbarnea (approved):

atugushev:
Thanks @ssbarnea!

PR #1768, opened 2022-12-11
Enable TOML reformatting in prettier

ssbarnea:
- enable TOML reformatting in prettier

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,3 +42,6 @@ repos:
     rev: v3.0.0-alpha.4
     hooks:
       - id: prettier
+        additional_dependencies:
+          - "prettier"
+          - "prettier-plugin-toml"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,13 +36,14 @@ classifiers = [
 ]
 keywords = ["pip", "requirements", "packaging"]
 dependencies = [
-    # direct dependencies
-    "build",
-    "click >= 8",
-    "pip >= 22.2",
-    # indirect dependencies
-    "setuptools",  # typically needed when pip-tools invokes setup.py
-    "wheel",  # pip plugin needed by pip-tools
+  # direct dependencies
+  "build",
+  "click >= 8",
+  "pip >= 22.2",
+  # indirect dependencies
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -36,9 +36,14 @@ repos:
     rev: 1.7.4
     hooks:
       - id: bandit
-        args: [--ini, .bandit]
+        args: [-c, pyproject.toml]
```
atugushev, line 39: I wonder what's your opinion on 3rd party configs on: having all in `pyproject.toml` or separate files (e.g. `.flake8`, `.coveragerc`, etc)?

webknjaz: It's best to have separate configs because it is hard to maintain a combined mess in a single one.

ssbarnea: If tool authors would have considered that "combined mess" is a real problem, they would have prevented enablement of that feature, but [almost all adopted it](https://github.com/orgs/stdize/discussions/1).

I (personally) find the benefit of decluttering the project root folder as adding more value than the downside of keeping everything in one file, especially as these settings do not need lots of updates and we already have a formatter for it. In fact I suspect that passing the name of the config file in pre-commit config is not even needed.

In general TOML is far more safe file format than old INIs, so the change of bugs with them is considerably lower. I am sure that bugs about inline comments and multiline entries are not unknown to you.

atugushev: I agree on `ini`-mess, however long `pyproject.toml` with ombined everything can become huge if we adopt more and more tools that require some configuration. This is just my personal opinion based on a habit to organise code namespaces/modules as much as possible. 

What if we create a `.bandit.toml` file and  pass it `args: [-c, .bandit.toml]` to get rid of ini config at least fo bandit?

ssbarnea: Bad idea, unless that filename in particular is documented as recognized by the tool.

ssbarnea: I removed bandit change for PR due to current controversy.

atugushev (approved):

PR #1774, opened 2022-12-13
Add test: ignore unavailable versions in preexisting output

AndydeCleyre:
Before running `pip-compile`, the output file may contain a pinned requirement which can't be found in PyPI or whichever repo (e.g. a revoked release). This test checks that we gracefully discard these un-find-able pins, rather than cause the compile operation to fail.

The test only succeeds with the backtracking resolver, and is marked as xfail for legacy.

Related: 
- #1530
- #1531, which this PR replaces

I think it would be appropriate to label this with `skip-changelog`.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -23,6 +23,12 @@ legacy_resolver_only = pytest.mark.parametrize(
     indirect=("current_resolver",),
 )
 
+backtracking_resolver_only = pytest.mark.parametrize(
+    "current_resolver",
+    ("backtracking",),
+    indirect=("current_resolver",),
+)
+
 
 @pytest.fixture(
     autouse=True,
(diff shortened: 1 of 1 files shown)


tests/test_cli_compile.py
```diff
@@ -2030,6 +2030,31 @@ def test_preserve_compiled_prerelease_version(pip_conf, runner):
     assert "small-fake-a==0.3b1" in out.stderr.splitlines()
 
 
+def test_ignore_compiled_unavailable_version(pip_conf, runner, current_resolver):
+    if current_resolver == "legacy":
+        pytest.xfail(
+            "We know this is broken in the legacy resolver, but no fix is planned."
+        )
```
atugushev, line 2037:
```suggestion
backtracking_resolver_only = pytest.mark.parametrize(
    "current_resolver",
    ("backtracking",),
    indirect=("current_resolver",),
)

@backtracking_resolver_only
def test_ignore_compiled_unavailable_version(pip_conf, runner, current_resolver):
```

Can be done same way as `legacy_resolver_only`.

atugushev (approved):
Thanks!

Issue #1779, opened 2022-12-15
Add pyproject.toml as a default file name option


berislavlopac:
#### What's the problem this feature will solve?

Currently, `pyproject.toml` needs to be explicitly declared when running `pip-compile`:

```shell
> pip-compile
Usage: pip-compile [OPTIONS] [SRC_FILES]...
Try 'pip-compile -h' for help.

Error: Invalid value: If you do not specify an input file, the default is requirements.in or setup.py
>
> pip-compile pyproject.toml
#
# This file is autogenerated by pip-compile with Python 3.10
# by the following command:
#
#    pip-compile pyproject.toml
#
anyio==3.6.2
    # via httpcore
(etc)
```

#### Describe the solution you'd like

Since `pip-compile` already supports two standard file formats (`requirements.in` and `setup.py`), it would make sense to add `pyproject.toml` to the list of automatically recognised formats.

webknjaz:
Looks like `setup.cfg` would also need to be added to the list.

berislavlopac:
It turns out that it's not necessary. I'm just preparing a PR for this, and `setup.cfg` is supposed to be accompanied either by `pyproject.toml (for PEP 517 compatible builds; see https://setuptools.pypa.io/en/latest/build_meta.html), or by `setup.py` (for legacy builds; see https://setuptools.pypa.io/en/latest/userguide/declarative_config.html). 😊

atugushev:
Closed this as completed in #1780.

PR #1780, opened 2022-12-15
Add `pyproject.toml` as default input file format

berislavlopac:
Added `pyproject.toml` as a default input file name that is searched for when `pip-compile` is invoked without arguments; see https://github.com/jazzband/pip-tools/issues/1779 for details.

The default file names are searched in the following order, with the first encountered being used to compile:

1. `requirements.in`
2. `setup.py`
3. `setup.cfg`
4. `pyproject.toml`

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -33,9 +33,7 @@ from ..utils import (
 )
 from ..writer import OutputWriter
 
-DEFAULT_REQUIREMENTS_FILES = (
-    "requirements.in", "setup.py", "pyproject.toml"
-)
+DEFAULT_REQUIREMENTS_FILES = ("requirements.in", "setup.py", "pyproject.toml")
 DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
 METADATA_FILENAMES = frozenset({"setup.py", "setup.cfg", "pyproject.toml"})
 
@@ -351,8 +349,7 @@ def cli(
         else:
             raise click.BadParameter(
                 (
-                    "If you do not specify an input file, "
-                    "the default is one of: {}"
+                    "If you do not specify an input file, " "the default is one of: {}"
                 ).format(", ".join(DEFAULT_REQUIREMENTS_FILES))
             )
 
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -33,7 +33,12 @@ from ..utils import (
 )
 from ..writer import OutputWriter
 
-DEFAULT_REQUIREMENTS_FILES = ("requirements.in", "setup.py", "setup.cfg", "pyproject.toml")
+DEFAULT_REQUIREMENTS_FILES = (
+    "requirements.in",
+    "setup.py",
+    "setup.cfg",
+    "pyproject.toml",
+)
 DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
 METADATA_FILENAMES = frozenset({"setup.py", "setup.cfg", "pyproject.toml"})
 
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -36,8 +36,8 @@ from ..writer import OutputWriter
 DEFAULT_REQUIREMENTS_FILES = (
     "requirements.in",
     "setup.py",
-    "setup.cfg",
     "pyproject.toml",
+    "setup.cfg",
 )
 DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
 METADATA_FILENAMES = frozenset({"setup.py", "setup.cfg", "pyproject.toml"})
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -354,7 +354,7 @@ def cli(
         else:
             raise click.BadParameter(
                 (
-                    "If you do not specify an input file, " "the default is one of: {}"
+                    "If you do not specify an input file, the default is one of: {}"
                 ).format(", ".join(DEFAULT_REQUIREMENTS_FILES))
             )
 
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -33,7 +33,7 @@ from ..utils import (
 )
 from ..writer import OutputWriter
 
-DEFAULT_REQUIREMENTS_FILES = ("requirements.in", "setup.py", "pyproject.toml")
+DEFAULT_REQUIREMENTS_FILES = ("requirements.in", "setup.py", "setup.cfg", "pyproject.toml")
 DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
 METADATA_FILENAMES = frozenset({"setup.py", "setup.cfg", "pyproject.toml"})
 
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -33,7 +33,9 @@ from ..utils import (
 )
 from ..writer import OutputWriter
 
-DEFAULT_REQUIREMENTS_FILE = "requirements.in"
+DEFAULT_REQUIREMENTS_FILES = (
+    "requirements.in", "setup.py", "pyproject.toml"
+)
 DEFAULT_REQUIREMENTS_OUTPUT_FILE = "requirements.txt"
 METADATA_FILENAMES = frozenset({"setup.py", "setup.cfg", "pyproject.toml"})
 
@@ -342,16 +344,16 @@ def cli(
     log.verbosity = verbose - quiet
 
     if len(src_files) == 0:
-        if os.path.exists(DEFAULT_REQUIREMENTS_FILE):
-            src_files = (DEFAULT_REQUIREMENTS_FILE,)
-        elif os.path.exists("setup.py"):
-            src_files = ("setup.py",)
+        for file_path in DEFAULT_REQUIREMENTS_FILES:
+            if os.path.exists(file_path):
+                src_files = (file_path,)
+                break
         else:
             raise click.BadParameter(
                 (
                     "If you do not specify an input file, "
-                    "the default is {} or setup.py"
-                ).format(DEFAULT_REQUIREMENTS_FILE)
+                    "the default is one of: {}"
+                ).format(", ".join(DEFAULT_REQUIREMENTS_FILES))
             )
 
     if not output_file:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -381,18 +381,19 @@ def make_sdist(run_setup_file):
 def make_module(tmpdir):
     """
     Make a metadata file with the given name and content and a fake module.
+
+    By default, the file is created in the temporary directory; passing the
+    `base_dir` argument allows for fine-tuning the file's location.
     """
 
-    def _make_module(fname, content, base_dir=None):
-        if base_dir is None:
-            base_dir = tmpdir
-        path = os.path.join(base_dir, "sample_lib")
+    def _make_module(fname, content):
+        path = os.path.join(tmpdir, "sample_lib")
         os.mkdir(path)
-        path = os.path.join(base_dir, "sample_lib", "__init__.py")
+        path = os.path.join(tmpdir, "sample_lib", "__init__.py")
         with open(path, "w") as stream:
             stream.write("'example module'\n__version__ = '1.2.3'")
         if fname == "setup.cfg":
-            path = os.path.join(base_dir, "pyproject.toml")
+            path = os.path.join(tmpdir, "pyproject.toml")
             with open(path, "w") as stream:
                 stream.write(
                     "\n".join(
@@ -403,7 +404,7 @@ def make_module(tmpdir):
                         )
                     )
                 )
-        path = os.path.join(base_dir, fname)
+        path = os.path.join(tmpdir, fname)
         with open(path, "w") as stream:
             stream.write(dedent(content))
         return path
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -381,9 +381,6 @@ def make_sdist(run_setup_file):
 def make_module(tmpdir):
     """
     Make a metadata file with the given name and content and a fake module.
-
-    By default, the file is created in the temporary directory; passing the
-    `base_dir` argument allows for fine-tuning the file's location.
     """
 
     def _make_module(fname, content):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -383,14 +383,16 @@ def make_module(tmpdir):
     Make a metadata file with the given name and content and a fake module.
     """
 
-    def _make_module(fname, content):
-        path = os.path.join(tmpdir, "sample_lib")
+    def _make_module(fname, content, base_dir=None):
+        if base_dir is None:
+            base_dir = tmpdir
+        path = os.path.join(base_dir, "sample_lib")
         os.mkdir(path)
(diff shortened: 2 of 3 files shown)


piptools/scripts/compile.py
```diff
@@ -33,7 +33,7 @@
 )
 from ..writer import OutputWriter
 
-DEFAULT_REQUIREMENTS_FILE = "requirements.in"
+DEFAULT_REQUIREMENTS_FILES = ("requirements.in", "setup.py", "pyproject.toml")
```
atugushev, line 36:
```suggestion
DEFAULT_REQUIREMENTS_FILES = ("requirements.in", "setup.py", "setup.cfg", "pyproject.toml")
```

setup.cfg also can be used as a requirement file.

ssbarnea:
@berislavlopac I a very happy to see your proposed that. I was planning to do it myself for quite a few weeks but failed to find the time. Now you can be sure I will do my best to get this merged.

ssbarnea:
@berislavlopac Apparently there is a reduction in codecoverage reported by codecov, one that seems to be genuine.

berislavlopac:
@ssbarnea I'm confused - it looks like codecov is complaining that the lines in the _test_ file were not covered by tests? 🤔

webknjaz:
@berislavlopac yes, from https://app.codecov.io/gh/jazzband/pip-tools/pull/1780 it's evident that the else-block code branch doesn't get executed, meaning that `fname in default_file_names` is always true there. At least in the CI.

webknjaz:
Looking at the diff, it's probably related to the fact that you test against CWD which is a bad idea. Look at how other tests create temporary directory structure for all the cases they cover.

tests/test_cli_compile.py
```diff
@@ -1393,16 +1393,6 @@ def test_stdin_without_output_file(runner):
     assert "--output-file is required if input is from stdin" in out.stderr
 
 
-def test_not_specified_input_file(runner):
```
webknjaz, line 1396: I'd keep this as a separate test but rename it to specify that it tests the case when both the input isn't passed and the none of the fallback files are present on disk.
It is separate from testing that the input isn't passed but one (or more) of the fallback files exists.

tests/test_cli_compile.py
```diff
@@ -2457,6 +2447,27 @@ def test_triple_equal_pinned_dependency_is_used(
...
+@pytest.mark.parametrize(("fname", "content"), METADATA_TEST_CASES)
+def test_not_specified_input_file(runner, make_module, fname, content):
+    """
+    It should raise an error if there are no input files or default input files
+    such as "setup.py", "requirements.in" or "pyproject.toml".
+    """
+    default_file_names = ("requirements.in", "setup.py", "pyproject.toml", "setup.cfg")
+    make_module(fname=fname, content=content, base_dir=os.getcwd())
+    out = runner.invoke(cli)
+    if fname in default_file_names:
```
webknjaz, line 2459: Usually, the tests should be simple and test one thing only. Once logic is added, they tend to become hard to read/understand as well as easy to be misleading/buggy.

berislavlopac:
@webknjaz I agree with your points, and will look into the alternative to `cwd`. However, the issue appeared because `setup.cfg` was added to the list of allowed files.

I'm splitting the test into two separate cases, one for when there is an allowed file and another without it.

That said, I would still argue that we don't need to include `setup.cfg` in the list of automatically accepted files. This is because -- if I'm not mistaken -- on its own it doesn't do anything; it either requires `setup.py` or `pyproject.toml` to be present for `setuptools` to work properly.

webknjaz:
> on its own it doesn't do anything; it either requires `setup.py` or `pyproject.toml` to be present for `setuptools` to work properly.

It does require those, but it can still contain the deps list on its own. I don't remember exactly, but I think that pip-tools might be loading it directly.

berislavlopac:
Okay, after further thought I am okay with keeping in the explicitly defined default list, especially if it's listed after the other two. 👍

berislavlopac:
> you test against CWD which is a bad idea

Since this test is for the functionality that depends directly on some files being present (or not) in the CWD, I'm not sure what would be the right way to test it, apart from creating those files in the CWD (which the current tests do)? 🤔

webknjaz:
> I'm not sure what would be the right way to test it,

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.chdir into an empty https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html?

berislavlopac:
@webknjaz Done, at the cost of a couple more fixtured added to the test. 😄

tests/test_cli_compile.py
```diff
@@ -2457,6 +2447,38 @@ def test_triple_equal_pinned_dependency_is_used(
...
+):
+    """
+    It should raise an error if there are no input files or default input files
+    such as "setup.py", "requirements.in" or "pyproject.toml".
+    """
+    make_module(fname=fname, content=content)
+
+    monkeypatch.chdir(tmpdir)
+    out = runner.invoke(cli)
+    monkeypatch.undo()
```
webknjaz, line 2462: This is supposed to happen automatically.
```suggestion
```

tests/conftest.py
```diff
@@ -381,6 +381,9 @@ def _make_sdist(package_dir, dist_dir, *args):
 def make_module(tmpdir):
     """
     Make a metadata file with the given name and content and a fake module.
+
+    By default, the file is created in the temporary directory; passing the
+    `base_dir` argument allows for fine-tuning the file's location.
```
webknjaz, line 386: Did you forget to clean this up?

berislavlopac: I did indeed, fixed now.

tests/test_cli_compile.py
berislavlopac: And it does, but in this case it complains that the path does not exist. I'm guessing that is because the `tmpdir` is automatically removed before that happens.

piptools/scripts/compile.py
```diff
@@ -342,16 +347,15 @@ def cli(
...
+            if os.path.exists(file_path):
+                src_files = (file_path,)
+                break
         else:
             raise click.BadParameter(
                 (
-                    "If you do not specify an input file, "
-                    "the default is {} or setup.py"
-                ).format(DEFAULT_REQUIREMENTS_FILE)
+                    "If you do not specify an input file, " "the default is one of: {}"
```
atugushev, line 357:
```suggestion
                    "If you do not specify an input file, the default is one of: {}"
```

post-black fix

tests/test_cli_compile.py
```diff
@@ -2457,6 +2447,38 @@ def test_triple_equal_pinned_dependency_is_used(
...
+def test_not_specified_input_file_without_allowed_files(runner):
+    """
+    It should raise an error if there are no input files or default input files
+    such as "setup.py" or "requirements.in".
+    """
+    out = runner.invoke(cli)
+    assert "If you do not specify an input file" in out.stderr
+    assert out.exit_code == 2
+    for file_name in ("requirements.in", "setup.py", "pyproject.toml"):
+        assert file_name in out.stderr
```
atugushev, line 2479:
```suggestion
    assert out.exit_code == 2
    expected_error = (
        "Error: Invalid value: If you do not specify an input file, the default "
        "is one of: requirements.in, setup.py, pyproject.toml, setup.cfg"
    )
    assert expected_error in out.stderr.splitlines()
```

I'd prefer to check the whole line here

tests/test_cli_compile.py
```diff
@@ -2457,6 +2447,38 @@ def test_triple_equal_pinned_dependency_is_used(
...
+    runner, make_module, fname, content, tmpdir, monkeypatch
+):
+    """
+    It should raise an error if there are no input files or default input files
+    such as "setup.py", "requirements.in" or "pyproject.toml".
+    """
+    make_module(fname=fname, content=content)
+
+    monkeypatch.chdir(tmpdir)
+    out = runner.invoke(cli)
```
atugushev, line 2461:
```
assert out.exit_code == 1
```

exit code should be checked as well

tests/test_cli_compile.py
```diff
@@ -2457,6 +2447,38 @@ def test_triple_equal_pinned_dependency_is_used(
...
+    monkeypatch.undo()
+
+    assert "small-fake-a==0.1" in out.stderr
```
atugushev, line 2464:
```
(Pdb++) print(out.stderr)`
  ERROR: Could not find a version that satisfies the requirement small-fake-a==0.1 (from sample-lib (pyproject.toml)) (from versions: none)
```

>     It should raise an error if there are no input files or default input files
>     such as "setup.py", "requirements.in" or "pyproject.toml".

Could you elaborate on which exact error should be raised?

atugushev (commented):
Thanks for the PR. A few comments.

tests/test_cli_compile.py
berislavlopac: > Could you elaborate on which exact error should be raised?

Not really -- this docstring was present in the test before my changes, I just expanded the list of expected files. Perhaps it can be rewritten with more details.

atugushev: This test does not really differ from the other test. Maybe drop this test completely?

berislavlopac: Ah I see you are talking about the test when the files are present; yeah, that docstring should not be there at all, it was duplicated my mistake.

tests/test_cli_compile.py
berislavlopac: That makes sense. 👍

tests/test_cli_compile.py
berislavlopac: Sure, although I'm not a big fan of checking for messages verbatim (for example doesn't work well with i18n). But I can live with that here.

tests/test_cli_compile.py
berislavlopac: We need to test two separate functionalities when calling `pip-compile` without arguments: one that tests that a file is parsed if present, and the other one that that tests that an error is raised if none of the default files are present.

atugushev: > Test that a default-named file is parsed if present.

I don't understand why there is an error checking in the test given the description above.

berislavlopac: You're right, it shouldn't; but for some reason it's returning status code 1. 

This test, `test_not_specified_input_file`, is supposed to work exactly the same as `test_input_formats`, with the only difference being that in this one we don't pass the file path; I have literally copied it and changed the CLI invocation not to pass the filename (apparently that use case was not even tested before).

I'm trying to figure out why is it returning 1 now. 🤔

atugushev: I see. That's essentially `test_input_formats` except `monkeypatch.chdir(tmpdir)`. Why do we change the current directory by the way?

tests/test_cli_compile.py
```diff
@@ -2457,6 +2447,41 @@ def test_triple_equal_pinned_dependency_is_used(
...
+    assert out.exit_code == 1
+
+
+def test_not_specified_input_file_without_allowed_files(runner):
+    """
+    It should raise an error if there are no input files or default input files
+    such as "setup.py" or "requirements.in".
+    """
+    out = runner.invoke(cli)
+    assert "If you do not specify an input file" in out.stderr
```
atugushev, line 2476:
```suggestion
```

this can be removed now

tests/test_cli_compile.py
berislavlopac: Because the `make_module` fixture creates the files in the `tmpdir`, and we are testing the functionality which looks for the files in the CWD (see earlier discussion). So our options are:

1. Patch `make_module` to be able to create files in CWD; I did that before but reverted as recommended.
2. Create the file in `tmpdir` and change the CWD to `tmpdir` (current approach).
3. Create the file in `tmpdir`, copy it to CWD and delete after the test; I might try that next. 🤷‍♂️

atugushev: > Because the make_module fixture creates the files in the tmpdir

I see. Let's make it explicit then by:

```python
module_path = make_module(fname=fname, content=content)
monkeypatch.chdir(module_path / '..')
```

berislavlopac: OK, I think I figured it out; I was a bit too optimistic with removing all the invocation params, rather than just the filename. I also changed the test to use `chdir` to the file's actual `dirname` instead of using `tmpdir`.

tests/test_cli_compile.py
berislavlopac: 👍

tests/test_cli_compile.py
atugushev: Thanks! That looks nice now!

berislavlopac: > Let's make it explicit then

Yeah - I used `os.path` rather than `pathlib.Path` as it has been used in other tests, for consistency; I personally prefer the latter.

tests/test_cli_compile.py
```diff
@@ -2457,6 +2447,40 @@ def test_triple_equal_pinned_dependency_is_used(
...
+    meta_path = make_module(fname=fname, content=content)
+    monkeypatch.chdir(os.path.dirname(meta_path))
+    out = runner.invoke(cli, ["-n", "--no-build-isolation", "--find-links", fake_dists])
+    monkeypatch.undo()
+
+    assert out.exit_code == 0, out.stderr
+    assert "small-fake-a==0.1" in out.stderr
+    assert "small-fake-b" not in out.stderr
+    assert "small-fake-c" not in out.stderr
+    assert "extra ==" not in out.stderr
```
atugushev, line 2467:
```suggestion
    out = runner.invoke(
        cli,
        [
            "--output-file",
            "-",
            "--no-header",
            "--no-emit-options",
            "--no-annotate",
            "--no-build-isolation",
            "--find-links",
            fake_dists,
        ],
    )
    monkeypatch.undo()

    assert out.exit_code == 0, out.stderr
    assert "small-fake-a==0.1\n" == out.stdout
```

Could you please update the test so that the whole output file is checked?

tests/test_cli_compile.py
atugushev: Me too. We can refactor this in a following up PR.

tests/test_cli_compile.py
berislavlopac: Sure, but should we do the same on `test_input_formats` too?

atugushev: Actually, there are multiple tests to update to use the same pattern. Feel free to fix it in a following up PR if you like.

atugushev: However, we can fix the current test here in this PR.

atugushev: I accidentally applied the suggestion twice and github broke the code :(

atugushev (approved):
LGTM 👍 Thank you very much for your patience!

tests/test_cli_compile.py
webknjaz: Right..
This is because `monkeypatch` and `tmp_path` don't depend on each other and pytest doesn't guarantee the cleanup sequence in this case.

webknjaz: @berislavlopac so a cleaner solution could be to use a context manager (`with monkeypatch.context() as mp_ctx:`). With that, you won't need the explicit `undo()` and there would be a clear block where chdir is applied.

berislavlopac: Good point, thanks!

PR #1787, opened 2022-12-24
Raise error if input and output filenames are matched

### LINKED ISSUE #1762: `pip-compile requirements.txt` compiles dependencies to the same input file
It looks like pip-tools has a bug, when `reqs.txt` is passed as an input file then `pip-compile reqs.txt ` would resolve dependencies to the same input file. I remember we had a guard against this scenario, but apparently, we lost it somewhere in between new features.

_Originally posted by @atugushev in https://github.com/jazzband/pip-tools/issues/1760#issuecomment-1345420673_

#### Environment Versions

1. OS Type macOS
1. Python version: `3.11`
1. pip version: `22.3`
1. pip-tools version: `6.11.0`

#### Steps to replicate

1. `echo six > requirements.txt`
2. `pip-compile requirements.txt`
3. `ls requirements.*` 

#### Expected result

I would expect an error that `requirements.txt` is using as input and output file.

#### Actual result

```
$ ls requirements.*
requirements.txt
```

atugushev:
Fixes #1762.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -382,7 +382,7 @@ def cli(
             ctx.call_on_close(safecall(output_file.close_intelligently))
 
     for src_file in src_files:
-        if src_file == output_file.name:
+        if src_file != "-" and src_file == output_file.name:
             raise click.BadArgumentUsage(
                 f"input and output filenames must not be matched: {src_file}"
             )
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -381,6 +381,12 @@ def cli(
         if isinstance(output_file, LazyFile):  # pragma: no cover
             ctx.call_on_close(safecall(output_file.close_intelligently))
 
+    for src_file in src_files:
+        if src_file == output_file.name:
+            raise click.BadArgumentUsage(
+                f"input and output filenames must not be matched: {src_file}"
+            )
+
     if resolver_name == "legacy":
         log.warning(
             "WARNING: the legacy dependency resolver is deprecated and will be removed"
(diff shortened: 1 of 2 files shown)


piptools/scripts/compile.py
```diff
@@ -381,6 +381,12 @@ def cli(
         if isinstance(output_file, LazyFile):  # pragma: no cover
             ctx.call_on_close(safecall(output_file.close_intelligently))
 
+    for src_file in src_files:
+        if src_file != "-" and src_file == output_file.name:
```
webknjaz, line 385: Did you mean this?
```suggestion
    if output_file.name != "-" and output_file.name in src_files:
```

atugushev: Neat! Addressed in d7450fc26e7cca462a6f8807d30a9ee52284f08d

tests/test_cli_compile.py
```diff
@@ -2741,3 +2741,31 @@ def test_print_deprecation_warning_if_using_legacy_resolver(runner, current_reso
...
+
+
+@pytest.mark.parametrize(
+    "input_filenames",
+    (
+        pytest.param(("requirements.txt",), id="one file"),
+        pytest.param(("requirements.txt", "dev-requirements.in"), id="multiple files"),
+    ),
+)
+def test_raise_error_when_input_and_output_filenames_are_matched(
```
webknjaz, line 2753: Plz add a test for `-` too.

atugushev: We have this test: https://github.com/jazzband/pip-tools/blob/23d874e5e5d82fc27d233c58627ca5bf3bef9a1f/tests/test_cli_compile.py#L1406-L1421

webknjaz: Fair enough, resolving the thread.

webknjaz (approved):

atugushev:
Thanks @webknjaz!

PR #1788, opened 2022-12-24
Fix a regression with unsafe packages for `--allow-unsafe`

### LINKED ISSUE #1786: Unsafe requirements are no longer included in the requirements.txt generated by pip-compile
After v6.12.0 the unsafe packages pip and setuptools are no longer included in the requirements.txt when --allow-unsafe is used in pip-compile.
 
There was a change made in writer.py (https://github.com/jazzband/pip-tools/pull/1766)
I think this line should be using unsafe_packages if the user has specified allow_unsafe rather than if they have not:

unsafe_packages = unsafe_packages if **not** self.allow_unsafe else set()
I think this should be:
unsafe_packages = unsafe_packages if self.allow_unsafe else set()

#### Environment Versions

1. Ubuntu 18
1. Python version: 3.7.11
1. pip version: 22.3.1
1. pip-tools version: 6.12.1

#### Steps to replicate

Ensure requirements.in file includes pip and then run:
pip-compile --verbose --allow-unsafe --output-file requirements.txt requirements.in

#### Expected result

requirements.txt should end with this:

pip==22.3.1
    # via -r requirements.in

#### Actual result

The unsafe packages are not listed in the requirements.txt at all

q0w:
Closes #1786
##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -181,7 +181,7 @@ class OutputWriter:
         hashes: dict[InstallRequirement, set[str]] | None = None,
     ) -> Iterator[str]:
         # default values
-        unsafe_packages = unsafe_packages if not self.allow_unsafe else set()
+        unsafe_packages = unsafe_packages if self.allow_unsafe else set()
         hashes = hashes or {}
 
         # Check for unhashed or unpinned packages if at least one package does have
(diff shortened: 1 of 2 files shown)


atugushev (approved):
Awesome! 🎉

q0w:
@atugushev is it correct? https://github.com/jazzband/pip-tools/issues/1786#issuecomment-1364582961

PR #1798, opened 2023-01-14
Include package local versions in --generate-hashes

neykov:
Enables using --generate-hashes with packages that have local versions (https://peps.python.org/pep-0440/#local-version-identifiers). Some public projects, like PyTorch (https://download.pytorch.org/whl/cpu) use local versions for distributing different builds of their packages.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -383,10 +383,11 @@ class PyPIRepository(BaseRepository):
         # satisfy this constraint.
         all_candidates = self.find_all_candidates(ireq.name)
         candidates_by_version = lookup_table(all_candidates, key=candidate_version)
-        matching_versions = list(
+        matching_versions = set(
             ireq.specifier.filter(candidate.version for candidate in all_candidates)
         )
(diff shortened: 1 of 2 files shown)


webknjaz:
I'd expect this to be behind a CLI option..

neykov:
@webknjaz I'll update the wording to make it clearer, but this is a bugfix, not a behaviour change.

pip-tools won't generate correct installable hashes when local versions are involved (e.x. pytorch). When no local versions are available the behaviour is the same as the current one.

webknjaz:
How so? PEP 440 specifically say that the local versions are not to be uploaded to package indexes. This implies that these are private package versions, as far as said PEP is concerned, at least.
pip-tools retrieves the package hashes for the public packages.
I'd say it's reasonable to make this optional to prevent supply-chain attacks.
Pinning the hashes is a security feature, making it possible to hijack them with private packages implicitly makes this a desired attack vector IMHO. Hence the opt-in.

neykov:
Closing since I didn't have the time to address comments in the past year :(

PR #1815, opened 2023-02-24
Release 6.12.3

atugushev:

PR #1817, opened 2023-03-01
Fix year in changelog entry

atugushev:
Addresses https://github.com/jazzband/pip-tools/pull/1815#discussion_r1121581875


richafrank (approved):
Thanks @atugushev

PR #1822, opened 2023-03-14
Log a warning if the user specifies -P and the output file is present but empty

davidmreed:
When the user runs a command like

```
pip-compile -P some_package requirements.in
```

the expected behavior is that only `some_package` should be updated to a new version.

If the user errs by redirecting `pip-compile`'s stdout to the target file:

```
pip-compile -P some_package requirements.in > requirements.txt
```

instead of either using `-o` or allowing `pip-compile` to infer the target file name, the target file is truncated by the shell before `pip-compile` executes. This results in `pip-compile` actually updating _all_ dependencies, because while the target file is present, it's not a valid source of version constraints.

This PR adds a check in the code to detect this situation, which is almost certainly undesired, and redirect the user to a more appropriate command.

##### Contributor checklist

##### Maintainer checklist


tests/test_cli_compile.py
```diff
@@ -858,6 +858,27 @@ def test_upgrade_packages_option_no_existing_file(pip_conf, runner):
     assert out.exit_code == 0
     assert "small-fake-a==0.2" in out.stderr.splitlines()
     assert "small-fake-b==0.3" in out.stderr.splitlines()
+    assert (
+        "WARNING: the output file requirements.txt exists but is empty"
+        not in out.stderr
+    )
+
+
+def test_upgrade_packages_option_empty_existing_file_warning(pip_conf, runner):
```
chrysle, line 867:
```suggestion
def test_upgrade_packages_empty_target_file_warning(pip_conf, runner):
```
Maybe this sounds better.

chrysle (approved):
LGTM, but I'd like another person to proof-read this.

piptools/scripts/compile.py
```diff
@@ -444,6 +444,15 @@ def cli(
     # Proxy with a LocalRequirementsRepository if --upgrade is not specified
     # (= default invocation)
     if not upgrade and os.path.exists(output_file.name):
+        if upgrade_install_reqs and os.path.getsize(output_file.name) == 0:
```
webknjaz, line 447: I noticed that this conditional clause is getting complicated, which affects readability because of the perceived cognitive complexity.
How about labeling the conditions with nice variable names?
```suggestion
    output_file_exists = os.path.exists(output_file.name)
    output_file_is_empty = True if not output_file_exists else os.path.getsize(output_file.name) == 0
    if not upgrade and output_file_exists:
        if upgrade_install_reqs and output_file_is_empty:
```

tests/test_cli_compile.py
```diff
@@ -858,6 +858,27 @@ def test_upgrade_packages_option_no_existing_file(pip_conf, runner):
     assert out.exit_code == 0
     assert "small-fake-a==0.2" in out.stderr.splitlines()
     assert "small-fake-b==0.3" in out.stderr.splitlines()
+    assert (
+        "WARNING: the output file requirements.txt exists but is empty"
+        not in out.stderr
+    )
+
+
+def test_upgrade_packages_empty_target_file_warning(pip_conf, runner):
```
webknjaz, line 867: It seems we're missing tests for corner cases like `/dev/null` and `-`/stdout here...

Could you verify these for me, please? It may be useful to stick `@pytest.mark.parametrize` here.

webknjaz:
@chrysle no wording suggestions from me, but I've got a few nitpicks on the engineering side of things… Though, they are not mandatory, and it's okay to merge if @davidmreed chooses to leave things as is.

piptools/scripts/compile.py
chrysle:
```python
output_file_is_empty = not output_file_exists or os.path.getsize(output_file.name) == 0
```
I personally feel it's better like this:

```python
output_file_is_empty = os.path.getsize(output_file.name) == 0
```

Since it is already checked in the prior if condition if the file exists. But that is just a small improvement to code readability.

davidmreed: I opted for @chysle's formulation (although open to suggestion!) - I felt that moving both Booleans out of the conditional was nice, but required an `or` that was itself tricky to read.

tests/test_cli_compile.py
davidmreed: If the user does a shell redirect (`pip-compile ... > /dev/null`), there's no impact, because `pip-compile`'s normal logic for finding its output file applies.

If the user tries to do `pip-compile -o /dev/null`, they receive an error before the code-path that's changed in this PR, as that path can't be opened for writing.

I'm not sure I understand your suggestion regarding stdout. Could you say more?

piptools/scripts/compile.py
```diff
@@ -444,6 +444,16 @@ def cli(
     # Proxy with a LocalRequirementsRepository if --upgrade is not specified
     # (= default invocation)
     if not upgrade and os.path.exists(output_file.name):
+        output_file_is_empty = os.path.getsize(output_file.name) == 0
+        if upgrade_install_reqs and output_file_is_empty:
```
chrysle, line 448:
```suggestion
    output_file_exists = os.path.exists(output_file.name)
    output_file_is_empty = os.path.getsize(output_file.name) == 0
    if not upgrade and output_file_exists:
        if upgrade_install_reqs and output_file_is_empty:
```

I'm still fine with the `output_file_exists` variable though.

davidmreed: `os.path.getsize()` will throw a `FileNotFoundError` if `output_file` does not exist. That's why @webknjaz proposed 

    output_file_is_empty = not output_file_exists or os.path.getsize(output_file.name) == 0

I think that's fine too, but short-circuiting that way is not the most intuitive path to read.

chrysle: > `os.path.getsize()` will throw a `FileNotFoundError` if `output_file` does not exist.

I'm not quite sure what that has to do with the separate `output_file_exists` variable, can you explain?

tests/test_cli_compile.py
webknjaz: I wasn't sure either, that's why I asked. I'll mark this thread as resolved.

piptools/scripts/compile.py
webknjaz: Fair enough. Marking the thread as resolved.

PR #1835, opened 2023-04-01
Update examples in README

lucaswerkmeister:
The previous output had a trailing backslash on the last hash line; while this is apparently valid for both pip-sync and pip install, IMHO it looks like a bug, and it also doesn’t reflect what pip-compile currently emits (as of version 6.12.3), so update the example.

This puts the hashes example out of sync with some of the other examples in the README, but they weren’t all in sync before either (various examples were already referencing Django 4, 3 or 2), so I hope that’s not an issue.

##### Contributor checklist

- [x] Provided the tests for the changes. – N/A
- [x] Assure PR title is short, clear, and good to be included in the user-oriented changelog – I’d say so? though this could probably also be left out of the changelog altogether

##### Maintainer checklist


webknjaz (approved):

lucaswerkmeister:
I’m not sure what to make of the [CI failure](https://github.com/jazzband/pip-tools/actions/runs/4583717669/jobs/8095555593), is it something that can randomly happen?

atugushev (commented):
Thanks for the update! Would you like to fix the other examples in the README as well?

lucaswerkmeister:
Alright, I added a second commit updating the other `pip-compile` examples too.

atugushev:
Thanks!

PR #1863, opened 2023-05-06
Support config defaults using .pip-tools.toml or pyproject.toml

### LINKED ISSUE #604: Add config file
I end up writing wrapper scripts to pass options (like no-header) and support multiple .in files (#532).

If pip-compile knew to find its options in `.pip-tools.cfg` or `tox.ini` (many Python development tools such as pytest, flake8, coverage can find their config in tox.ini, even when they’re not run inside tox), I would put the options there and tell my co-workers they can simply run `pip-compile file.in`.

Idea taken from pip-tools’ fork prequ.

j00bar:
Supports use of a `.pip-tools.toml`, `pyproject.toml`, or other explicitly specified TOML file to set configuration defaults for `pip-compile` and `pip-sync`.

Fixes #604.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,6 @@ repos:
           - toml==0.10.2
           - pip==20.3.4
           - build==0.9.0
-          - types-toml==0.10.2
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.5
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,7 @@ repos:
           - toml==0.10.2
           - pip==20.3.4
           - build==0.9.0
+          - types-toml==0.10.2
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.5
     hooks:
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -27,8 +27,8 @@ from ..repositories.base import BaseRepository
 from ..resolver import BacktrackingResolver, LegacyResolver
 from ..utils import (
     UNSAFE_PACKAGES,
-    determine_config_file,
     dedup,
+    determine_config_file,
     drop_extras,
     is_pinned_requirement,
     key_from_ireq,
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -27,7 +27,7 @@ from ..repositories.base import BaseRepository
 from ..resolver import BacktrackingResolver, LegacyResolver
 from ..utils import (
     UNSAFE_PACKAGES,
-    callback_config_file_defaults,
+    determine_config_file,
     dedup,
     drop_extras,
     is_pinned_requirement,
@@ -317,7 +317,7 @@ def _determine_linesep(
     help=f"Read configuration from TOML file. By default, looks for a {CONFIG_FILE_NAME} or "
     "pyproject.toml.",
     is_eager=True,
-    callback=callback_config_file_defaults,
+    callback=determine_config_file,
 )
 def cli(
     ctx: click.Context,
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -31,6 +31,7 @@ from ..utils import (
     is_pinned_requirement,
     key_from_ireq,
     parse_requirements_from_wheel_metadata,
+    pyproject_toml_defaults_cb,
 )
 from ..writer import OutputWriter
 
@@ -302,6 +303,20 @@ def _determine_linesep(
     help="Specify a package to consider unsafe; may be used more than once. "
     f"Replaces default unsafe packages: {', '.join(sorted(UNSAFE_PACKAGES))}",
 )
+@click.option(
+    "--config",
+    type=click.Path(
+        exists=True,
+        file_okay=True,
+        dir_okay=False,
+        readable=True,
+        allow_dash=False,
+        path_type=str,
+    ),
+    help="Path to a pyproject.toml file with specialized defaults for pip-tools",
+    is_eager=True,
+    callback=pyproject_toml_defaults_cb,
+)
 def cli(
     ctx: click.Context,
     verbose: int,
@@ -340,6 +355,7 @@ def cli(
     emit_index_url: bool,
     emit_options: bool,
     unsafe_package: tuple[str, ...],
+    config: str | None,
 ) -> None:
     """
     Compiles requirements.txt from requirements.in, pyproject.toml, setup.cfg,
@@ -390,6 +406,9 @@ def cli(
         raise click.BadArgumentUsage(
             f"input and output filenames must not be matched: {output_file.name}"
         )
+
+    if config:
+        log.info(f"Using pip-tools configuration defaults found in '{config}'.")
 
     if resolver_name == "legacy":
         log.warning(
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -16,6 +16,7 @@ from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
 from pip._internal.utils.misc import redact_auth_from_url
 
+from pathlib import Path
 from .._compat import parse_requirements
 from ..cache import DependencyCache
 from ..exceptions import NoCandidateFound, PipToolsError
@@ -356,7 +357,7 @@ def cli(
     emit_index_url: bool,
     emit_options: bool,
     unsafe_package: tuple[str, ...],
-    config: str | None,
+    config: Path | None,
 ) -> None:
     """
     Compiles requirements.txt from requirements.in, pyproject.toml, setup.cfg,
@@ -409,7 +410,7 @@ def cli(
         )
 
     if config:
-        log.info(f"Using pip-tools configuration defaults found in '{config}'.")
+        log.info(f"Using pip-tools configuration defaults found in '{config !s}'.")
 
     if resolver_name == "legacy":
         log.warning(
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -28,7 +28,7 @@ from ..resolver import BacktrackingResolver, LegacyResolver
 from ..utils import (
     UNSAFE_PACKAGES,
     dedup,
-    determine_config_file,
+    override_defaults_from_config_file,
     drop_extras,
     is_pinned_requirement,
     key_from_ireq,
@@ -317,7 +317,7 @@ def _determine_linesep(
     help=f"Read configuration from TOML file. By default, looks for a {CONFIG_FILE_NAME} or "
     "pyproject.toml.",
     is_eager=True,
-    callback=determine_config_file,
+    callback=override_defaults_from_config_file,
 )
 def cli(
     ctx: click.Context,
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -28,10 +28,10 @@ from ..resolver import BacktrackingResolver, LegacyResolver
 from ..utils import (
     UNSAFE_PACKAGES,
     dedup,
-    override_defaults_from_config_file,
     drop_extras,
     is_pinned_requirement,
     key_from_ireq,
+    override_defaults_from_config_file,
     parse_requirements_from_wheel_metadata,
 )
 from ..writer import OutputWriter
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -19,19 +19,19 @@ from pip._internal.utils.misc import redact_auth_from_url
 from .._compat import parse_requirements
 from ..cache import DependencyCache
 from ..exceptions import NoCandidateFound, PipToolsError
-from ..locations import CACHE_DIR
+from ..locations import CACHE_DIR, CONFIG_FILE_NAME
 from ..logging import log
 from ..repositories import LocalRequirementsRepository, PyPIRepository
 from ..repositories.base import BaseRepository
 from ..resolver import BacktrackingResolver, LegacyResolver
 from ..utils import (
     UNSAFE_PACKAGES,
+    callback_config_file_defaults,
     dedup,
     drop_extras,
     is_pinned_requirement,
     key_from_ireq,
     parse_requirements_from_wheel_metadata,
-    pyproject_toml_defaults_cb,
 )
 from ..writer import OutputWriter
 
@@ -313,9 +313,10 @@ def _determine_linesep(
         allow_dash=False,
         path_type=str,
     ),
-    help="Path to a pyproject.toml file with specialized defaults for pip-tools",
+    help=f"Read configuration from TOML file. By default, looks for a {CONFIG_FILE_NAME} or "
+    "pyproject.toml.",
     is_eager=True,
-    callback=pyproject_toml_defaults_cb,
+    callback=callback_config_file_defaults,
 )
 def cli(
     ctx: click.Context,
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -5,6 +5,7 @@ import os
 import shlex
 import sys
 import tempfile
+from pathlib import Path
 from typing import IO, Any, BinaryIO, cast
 
 import click
@@ -16,7 +17,6 @@ from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
 from pip._internal.utils.misc import redact_auth_from_url
 
-from pathlib import Path
 from .._compat import parse_requirements
 from ..cache import DependencyCache
 from ..exceptions import NoCandidateFound, PipToolsError
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -22,11 +22,11 @@ from ..locations import CONFIG_FILE_NAME
 from ..logging import log
 from ..repositories import PyPIRepository
 from ..utils import (
-    override_defaults_from_config_file,
     flat_map,
     get_pip_version_for_python_executable,
     get_required_pip_specification,
     get_sys_path_for_python_executable,
+    override_defaults_from_config_file,
 )
 
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -17,14 +17,15 @@ from .. import sync
 from .._compat import parse_requirements
 from .._compat.pip_compat import Distribution
 from ..exceptions import PipToolsError
+from ..locations import CONFIG_FILE_NAME
 from ..logging import log
 from ..repositories import PyPIRepository
 from ..utils import (
+    callback_config_file_defaults,
     flat_map,
     get_pip_version_for_python_executable,
     get_required_pip_specification,
     get_sys_path_for_python_executable,
-    pyproject_toml_defaults_cb,
 )
 
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
@@ -97,9 +98,10 @@ DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
         allow_dash=False,
         path_type=str,
     ),
-    help="Path to a pyproject.toml file with specialized defaults for pip-tools",
+    help=f"Read configuration from TOML file. By default, looks for a {CONFIG_FILE_NAME} or "
+    "pyproject.toml.",
     is_eager=True,
-    callback=pyproject_toml_defaults_cb,
+    callback=callback_config_file_defaults,
 )
 def cli(
     ask: bool,
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -22,7 +22,7 @@ from ..locations import CONFIG_FILE_NAME
 from ..logging import log
 from ..repositories import PyPIRepository
 from ..utils import (
-    callback_config_file_defaults,
+    determine_config_file,
     flat_map,
     get_pip_version_for_python_executable,
     get_required_pip_specification,
@@ -102,7 +102,7 @@ DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
     help=f"Read configuration from TOML file. By default, looks for a {CONFIG_FILE_NAME} or "
     "pyproject.toml.",
     is_eager=True,
-    callback=callback_config_file_defaults,
+    callback=determine_config_file,
 )
 def cli(
     ask: bool,
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -22,7 +22,7 @@ from ..locations import CONFIG_FILE_NAME
 from ..logging import log
 from ..repositories import PyPIRepository
 from ..utils import (
-    determine_config_file,
+    override_defaults_from_config_file,
     flat_map,
     get_pip_version_for_python_executable,
     get_required_pip_specification,
@@ -102,7 +102,7 @@ DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
     help=f"Read configuration from TOML file. By default, looks for a {CONFIG_FILE_NAME} or "
     "pyproject.toml.",
     is_eager=True,
-    callback=determine_config_file,
+    callback=override_defaults_from_config_file,
 )
 def cli(
     ask: bool,
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -24,6 +24,7 @@ from ..utils import (
     get_pip_version_for_python_executable,
     get_required_pip_specification,
     get_sys_path_for_python_executable,
+    pyproject_toml_defaults_cb,
 )
 
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
@@ -86,6 +87,20 @@ DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
 )
 @click.argument("src_files", required=False, type=click.Path(exists=True), nargs=-1)
 @click.option("--pip-args", help="Arguments to pass directly to pip install.")
+@click.option(
+    "--config",
+    type=click.Path(
+        exists=True,
+        file_okay=True,
+        dir_okay=False,
+        readable=True,
+        allow_dash=False,
+        path_type=str,
+    ),
+    help="Path to a pyproject.toml file with specialized defaults for pip-tools",
+    is_eager=True,
+    callback=pyproject_toml_defaults_cb,
+)
 def cli(
     ask: bool,
     dry_run: bool,
@@ -103,6 +118,7 @@ def cli(
     client_cert: str | None,
     src_files: tuple[str, ...],
     pip_args: str | None,
+    config: str | None,
 ) -> None:
     """Synchronize virtual environment with requirements.txt."""
     log.verbosity = verbose - quiet
@@ -126,6 +142,9 @@ def cli(
         else:
             log.error("ERROR: " + msg)
             sys.exit(2)
+
+    if config:
+        log.info(f"Using pip-tools configuration defaults found in '{config}'.")
 
     if python_executable:
         _validate_python_executable(python_executable)
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -14,6 +14,7 @@ from pip._internal.index.package_finder import PackageFinder
 from pip._internal.metadata import get_environment
 
 from .. import sync
+from pathlib import Path
 from .._compat import parse_requirements
 from .._compat.pip_compat import Distribution
 from ..exceptions import PipToolsError
@@ -120,7 +121,7 @@ def cli(
     client_cert: str | None,
     src_files: tuple[str, ...],
     pip_args: str | None,
-    config: str | None,
+    config: Path | None,
 ) -> None:
     """Synchronize virtual environment with requirements.txt."""
     log.verbosity = verbose - quiet
@@ -146,7 +147,7 @@ def cli(
             sys.exit(2)
 
     if config:
-        log.info(f"Using pip-tools configuration defaults found in '{config}'.")
+        log.info(f"Using pip-tools configuration defaults found in '{config !s}'.")
 
     if python_executable:
         _validate_python_executable(python_executable)
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -5,6 +5,7 @@ import os
 import shlex
 import shutil
 import sys
+from pathlib import Path
 from typing import cast
 
 import click
@@ -14,7 +15,6 @@ from pip._internal.index.package_finder import PackageFinder
 from pip._internal.metadata import get_environment
 
 from .. import sync
-from pathlib import Path
 from .._compat import parse_requirements
 from .._compat.pip_compat import Distribution
 from ..exceptions import PipToolsError
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -566,10 +566,9 @@ def callback_config_file_defaults(
     if not config:
         return None
 
-    defaults: dict[str, Any] = ctx.default_map.copy() if ctx.default_map else {}
-    defaults.update(config)
-
-    ctx.default_map = defaults
+    if ctx.default_map is None:
+        ctx.default_map = {}
+    ctx.default_map.update(config)
     return config_file
 
 
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -575,17 +575,17 @@ def select_config_file(src_files: tuple[str, ...]) -> Path | None:
     """
     Returns the config file to use for defaults given `src_files` provided.
     """
-    if not src_files:
-        # If no src_files were specified, we consider the current directory the only candidate
-        candidate_dirs = [Path.cwd()]
-    else:
-        # Collect the candidate directories based on the src_file arguments provided
-        src_files_as_paths = [
-            Path(Path.cwd(), src_file).resolve() for src_file in src_files
-        ]
-        candidate_dirs = [
-            src if src.is_dir() else src.parent for src in src_files_as_paths
-        ]
+    # NOTE: If no src_files were specified, consider the current directory the
+    # NOTE: only config file lookup candidate. This usually happens when a
+    # NOTE: pip-tools invocation gets its incoming requirements from standard
+    # NOTE: input.
+    src_files_as_paths = [
+        Path(Path.cwd(), src_file).resolve()
+        for src_file in src_files or ('.', )
+    ]
+    candidate_dirs = [
+        src if src.is_dir() else src.parent for src in src_files_as_paths
+    ]
     config_file_path = next(
         (
             candidate_dir / config_file
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -579,13 +579,14 @@ def select_config_file(src_files: tuple[str, ...]) -> Path | None:
     # NOTE: only config file lookup candidate. This usually happens when a
     # NOTE: pip-tools invocation gets its incoming requirements from standard
     # NOTE: input.
-    src_files_as_paths = [
+    src_files_as_paths = (
         Path(Path.cwd(), src_file).resolve()
         for src_file in src_files or ('.', )
-    ]
-    candidate_dirs = [
-        src if src.is_dir() else src.parent for src in src_files_as_paths
-    ]
+    )
+    candidate_dirs = (
+        src if src.is_dir() else src.parent
+        for src in src_files_as_paths
+    )
     config_file_path = next(
         (
             candidate_dir / config_file
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -533,7 +533,7 @@ def parse_requirements_from_wheel_metadata(
         )
 
 
-def callback_config_file_defaults(
+def determine_config_file(
     ctx: click.Context, param: click.Parameter, value: str | None
 ) -> Path | None:
     """
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -536,9 +536,9 @@ def parse_requirements_from_wheel_metadata(
 def determine_config_file(
     ctx: click.Context, param: click.Parameter, value: str | None
 ) -> Path | None:
-    """
-    Returns the path to the config file with defaults being used, or `None` if no such file is
-    found.
+    """Return the config file path.
+
+    ``None`` is returned if no such file is found.
 
     Defaults for `click.Command` parameters should be override-able in a config file. `pip-tools`
     will use the first file found, searching in this order: an explicitly given config file, a
@@ -555,8 +555,6 @@ def determine_config_file(
     config = parse_config_file(config_file)
     if config:
         _assign_config_to_cli_context(ctx, config)
-    else:
-        return None
 
     return config_file
 
@@ -625,7 +623,7 @@ MULTIPLE_VALUE_OPTIONS = [
 ]
 
 
-@functools.lru_cache()
+@functools.lru_cache
 def parse_config_file(config_file: Path) -> dict[str, Any]:
     try:
         config = tomllib.loads(config_file.read_text(encoding="utf-8"))
@@ -640,11 +638,8 @@ def parse_config_file(config_file: Path) -> dict[str, Any]:
             hint=f"Could not parse '{config_file !s}': {value_err !s}",
         )
 
-    # In a pyproject.toml file, we expect the config to be under `[tool.pip-tools]`, but in our
-    # native configuration, it would be just `[pip-tools]`.
-    if config_file.name == "pyproject.toml":
-        config = config.get("tool", {})
-    piptools_config: dict[str, Any] = config.get("pip-tools", {})
+    # In a TOML file, we expect the config to be under `[tool.pip-tools]`
+    piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})
     piptools_config = {
         get_click_dest_for_option(k): v for k, v in piptools_config.items()
     }
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -533,18 +533,20 @@ def parse_requirements_from_wheel_metadata(
         )
 
 
-def determine_config_file(
+def override_defaults_from_config_file(
     ctx: click.Context, param: click.Parameter, value: str | None
 ) -> Path | None:
-    """Return the config file path.
+    """
+    Overrides ``click.Command`` defaults based on specified or discovered config
+    file, returning the ``pathlib.Path`` of that config file if specified or
+    discovered.
 
     ``None`` is returned if no such file is found.
 
-    Defaults for ``click.Command`` parameters should be override-able in a config
-    file. ``pip-tools`` will use the first file found, searching in this order:
-    an explicitly given config file, a ``.pip-tools.toml``, a ``pyproject.toml``
+    ``pip-tools`` will use the first config file found, searching in this order:
+    an explicitly given config file, a d``.pip-tools.toml``, a ``pyproject.toml``
     file. Those files are searched for in the same directory as the requirements
-    input file.
+    input file, or the current working directory if requirements come via stdin.
     """
     if value is None:
         config_file = select_config_file(ctx.params.get("src_files", ()))
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -2,14 +2,17 @@ from __future__ import annotations
 
 import collections
 import copy
+import functools
 import itertools
 import json
 import os
 import re
 import shlex
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Iterable, Iterator, TypeVar, cast
 
 import click
+import toml
 from click.utils import LazyFile
 from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line, parse_req_from_line
@@ -522,3 +525,104 @@ def parse_requirements_from_wheel_metadata(
             markers=parts.markers,
             extras=parts.extras,
         )
+
+
+def pyproject_toml_defaults_cb(
+    ctx: click.Context, param: click.Parameter, value: str | None
+) -> str | None:
+    """
+    Defaults for `click.Command` parameters should be override-able in pyproject.toml
+
+    Returns the path to the configuration file found, or None if no such file is found.
+    """
+    if value is None:
+        config_file = find_pyproject_toml(ctx.params.get("src_files", ()))
+        if config_file is None:
+            return None
+    else:
+        config_file = value
+
+    try:
+        config = parse_pyproject_toml(config_file)
+    except OSError as e:
+        raise click.FileError(
+            filename=config_file, hint=f"Could not read '{config_file}': {e}"
+        )
+    except ValueError as e:
+        raise click.FileError(
+            filename=config_file, hint=f"Could not parse '{config_file}': {e}"
+        )
+
+    if not config:
+        return None
+
+    defaults: dict[str, Any] = ctx.default_map.copy() if ctx.default_map else {}
+    defaults.update(config)
+
+    ctx.default_map = defaults
+    return config_file
+
+
+def find_pyproject_toml(src_files: tuple[str, ...]) -> str | None:
+    if not src_files:
+        # If no src_files were specified, we consider the current directory the only candidate
+        candidates = [Path.cwd()]
+    else:
+        # Collect the candidate directories based on the src_file arguments provided
+        src_files_as_paths = [
+            Path(Path.cwd(), src_file).resolve() for src_file in src_files
+        ]
+        candidates = [src if src.is_dir() else src.parent for src in src_files_as_paths]
+    pyproject_toml_path = next(
+        (
+            str(candidate / "pyproject.toml")
+            for candidate in candidates
+            if (candidate / "pyproject.toml").is_file()
+        ),
+        None,
+    )
+    return pyproject_toml_path
+
+
+# Some of the defined click options have different `dest` values than the defaults
+NON_STANDARD_OPTION_DEST_MAP: dict[str, str] = {
+    "extra": "extras",
+    "upgrade_package": "upgrade_packages",
+    "resolver": "resolver_name",
+    "user": "user_only",
+}
+
+
+def mutate_option_to_click_dest(option_name: str) -> str:
+    "Mutates an option from how click/pyproject.toml expect them to the click `dest` value"
+    # Format the keys properly
+    option_name = option_name.lstrip("-").replace("-", "_").lower()
+    # Some options have dest values that are overrides from the click generated default
+    option_name = NON_STANDARD_OPTION_DEST_MAP.get(option_name, option_name)
+    return option_name
+
+
+# Ensure that any default overrides for these click options are lists, supporting multiple values
+MULTIPLE_VALUE_OPTIONS = [
+    "extras",
+    "upgrade_packages",
+    "unsafe_package",
+    "find_links",
+    "extra_index_url",
+    "trusted_host",
+]
+
+
+@functools.lru_cache()
+def parse_pyproject_toml(config_file: str) -> dict[str, Any]:
+    pyproject_toml = toml.load(config_file)
+    config: dict[str, Any] = pyproject_toml.get("tool", {}).get("pip-tools", {})
+    config = {mutate_option_to_click_dest(k): v for k, v in config.items()}
+    # Any option with multiple values needs to be a list in the pyproject.toml
+    for mv_option in MULTIPLE_VALUE_OPTIONS:
+        if not isinstance(config.get(mv_option), (list, type(None))):
+            original_option = mv_option.replace("_", "-")
+            raise click.BadOptionUsage(
+                original_option, f"Config key '{original_option}' must be a list"
+            )
+    return config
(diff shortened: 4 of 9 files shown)


webknjaz:
@j00bar you can't trigger CI until you have at least one commit in the repo..

piptools/scripts/compile.py
```diff
@@ -302,6 +303,20 @@ def _determine_linesep(
...
+    "--config",
+    type=click.Path(
+        exists=True,
+        file_okay=True,
+        dir_okay=False,
+        readable=True,
+        allow_dash=False,
+        path_type=str,
+    ),
+    help="Path to a pyproject.toml file with specialized defaults for pip-tools",
```
webknjaz, line 316: Per my comments in the issue, it is important to allow a pip-tools specific config taking preference over `pyproject.toml`. Also, the way it's implemented, it'll allow naming the file arbitrarily, so this isn't accurate with regards to the generic case, only the default. Let's try to rephrase this without implying that this filename is mandatory.

piptools/utils.py
```diff
@@ -522,3 +525,104 @@ def parse_requirements_from_wheel_metadata(
             markers=parts.markers,
             extras=parts.extras,
         )
+
+
+def pyproject_toml_defaults_cb(
```
webknjaz, line 530: This needs to be renamed for the same reason. It is inaccurate that this only reads `pyproject.toml`.
Also, since it's a function, its name should start with a verb describing the action it performs, not a contextless static state.

piptools/utils.py
```diff
@@ -522,3 +525,104 @@ def parse_requirements_from_wheel_metadata(
...
+    ctx: click.Context, param: click.Parameter, value: str | None
+) -> str | None:
+    """
+    Defaults for `click.Command` parameters should be override-able in pyproject.toml
```
webknjaz, line 534: Could you make the docstring title describe what the function does following PEP 257?

piptools/utils.py
```diff
@@ -522,3 +525,104 @@ def parse_requirements_from_wheel_metadata(
...
+NON_STANDARD_OPTION_DEST_MAP: dict[str, str] = {
+    "extra": "extras",
+    "upgrade_package": "upgrade_packages",
+    "resolver": "resolver_name",
+    "user": "user_only",
+}
+
+
+def mutate_option_to_click_dest(option_name: str) -> str:
+    "Mutates an option from how click/pyproject.toml expect them to the click `dest` value"
```
webknjaz, line 597: Use triple double backticks as customary for docstrings, imperative mood, short title, long paragraph separately.

piptools/utils.py
```diff
@@ -522,3 +525,104 @@ def parse_requirements_from_wheel_metadata(
...
+# Some of the defined click options have different `dest` values than the defaults
+NON_STANDARD_OPTION_DEST_MAP: dict[str, str] = {
+    "extra": "extras",
+    "upgrade_package": "upgrade_packages",
+    "resolver": "resolver_name",
+    "user": "user_only",
+}
+
+
+def mutate_option_to_click_dest(option_name: str) -> str:
```
webknjaz, line 596: This name is misleading, I don't see the implementation mutating the incoming object, it seems to return a new one.

piptools/utils.py
```diff
@@ -522,3 +525,104 @@ def parse_requirements_from_wheel_metadata(
...
+    "upgrade_packages",
+    "unsafe_package",
+    "find_links",
+    "extra_index_url",
+    "trusted_host",
+]
+
+
+@functools.lru_cache()
+def parse_pyproject_toml(config_file: str) -> dict[str, Any]:
```
webknjaz, line 617: Again, let's not stick references to `pyproject.toml` all over the place. Looks like this just reads a tool config, regardless of the filename.

pyproject.toml
```diff
@@ -42,6 +42,7 @@ dependencies = [
   "pip >= 22.2",
   # indirect dependencies
   "setuptools", # typically needed when pip-tools invokes setup.py
+  "toml >= 0.10.1",
```
webknjaz, line 45: This line is out of place, the section says “indirect deps” but it's actually used directly now.

Though, this is not the only problem here: it should also be conditional — stdlib features a toml parser since Python 3.11 so it doesn't need an extra dep to work.

tests/test_utils.py
```diff
@@ -540,3 +543,102 @@ def test_get_sys_path_for_python_executable():
...
+    config_file = make_pyproject_toml_conf(mv_option, "not-a-list")
+    ctx = Context(compile_cli)
+    ctx.params["src_files"] = (str(config_file),)
+    pytest.raises(BadOptionUsage, pyproject_toml_defaults_cb, ctx, "config", None)
+
+
+def test_pyproject_toml_defaults_cb_bad_toml(make_pyproject_toml_conf):
+    config_file = make_pyproject_toml_conf("verbose", True)
+    config_text = open(config_file).read()
+    open(config_file, "w").write(config_text[::-1])
```
webknjaz, line 630: This opens 2 file descriptors and never closes them. It will cause `ResourceWarning`s that are usually turned into errors in pytest when in strict mode.
It's best to use pathlib for the inline invocations like this. Otherwise, use context managers.

tests/test_utils.py
```diff
@@ -540,3 +543,102 @@ def test_get_sys_path_for_python_executable():
...
+    ctx = Context(compile_cli)
+    ctx.params["src_files"] = (str(config_file),)
+    pytest.raises(FileError, pyproject_toml_defaults_cb, ctx, "config", None)
```
webknjaz, line 633: Please, use a context manager. Also, it is a best practice to always use the `match=` argument with such test helpers.

tests/test_utils.py
```diff
@@ -540,3 +543,102 @@ def test_get_sys_path_for_python_executable():
...
+
+
+def test_pyproject_toml_defaults_cb_unreadable_toml(make_pyproject_toml_conf):
+    ctx = Context(compile_cli)
+    pytest.raises(
```
webknjaz, line 638: Same as above, plz use a CM.

webknjaz (changes requested):
Thanks for working on this! I was able to skim through the patch from mobile and noticed a few areas that might need some love...
Though, I'm sure someone will check the logic too. Reviews are easier from an actual computer.

I'll also approve the CI in just a bit to see how it's doing...

webknjaz:
> I'll also approve the CI in just a bit to see how it's doing...

The PyPy failures seem unrelated... Though, they'll probably need sorting out separately before this PR could be mergeable. I wonder if bumping the interpreter version would have any effect.

j00bar:
Thank you for the detailed feedback!

1. I've added support and preference for a `.pip-tools.toml` file.
2. I've implemented the best practices and conventions you referenced.
3. I've made the dependency on `toml` apply only pre-3.11 - the tests still require the external toml library, as `tomllib` reads only but does not yet write.
4. I've updated the README to document the new configurability.

README.rst
```diff
@@ -306,6 +306,29 @@ Any valid ``pip`` flags or arguments may be passed on with ``pip-compile``'s
 Configuration
 -------------
 
+You can define project-level defaults for ``pip-compile`` and ``pip-sync`` by
+writing them to a configuration file in the same directory as your requirements
```
webknjaz, line 310: Has this changed since the previous review? I think I saw CWD being used in a few places, not an input-related location...

README.rst
```diff
@@ -306,6 +306,29 @@ Any valid ``pip`` flags or arguments may be passed on with ``pip-compile``'s
...
+for a ``.pip-tools.toml`` file and then in your ``pyproject.toml``. You can
+also specify an alternate TOML configuration file with the ``--config`` option.
+
+For example, to by default generate ``pip`` hashes in the resulting
+requirements file output, you can specify in a configuration file
+
+.. code-block:: toml
+
+    # In a .pip-tools.toml file
+    [pip-tools]
```
webknjaz, line 321: I'm unsure about this. The section name could probably be the same as for `pyproject.toml` which would make it easier to migrate between the files... OTOH, `tool.` might be excessive here.
```suggestion
    [tool.pip-tools]
```
This is not a request to make the change right away but rather an idea to be discussed.

README.rst
j00bar: It had not, no. If you specify `src_files`, it looks in the directories of the `src_files` - if you do not specify `src_files`, it uses `cwd()`.

So `pip-compile /path/to/requirements.in` would look for `/path/to/.pip-tools.toml` or `/absolute/path/to/pyproject.toml` -- whereas `pip-compile` without arguments would look in the current working directory.

At first blush, that sounds like the behavior that would be expected, versus looking in the current directory for those files, even when compiling a source in another directory. Happy to revise if you disagree.

README.rst
j00bar: Certainly - happy to have that discussion. I was comparing with other projects and trying for consistency. You can see, for example, that in this project's `pyproject.toml` that mypy, setuptools, setuptools_scm, and pytest all have the `tool` prefix - whereas flake8 and bandit's bespoke files just use the name of the tool as the section of the file.

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
...
+            filename=str(config_file), hint=f"Could not parse '{config_file}': {e}"
+        )
+
+    if not config:
+        return None
+
+    defaults: dict[str, Any] = ctx.default_map.copy() if ctx.default_map else {}
+    defaults.update(config)
+
+    ctx.default_map = defaults
```
webknjaz, line 572: Why is this needed? Would the following be equivalent?
```suggestion
    if ctx.default_map is None:
        ctx.default_map = {}
    ctx.default_map.update(config)
```

FWIW, pieces like this could benefit from being in a dedicated function with a descriptive name.

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
...
+    return str(config_file)
```
webknjaz, line 573: Operating on strings for paths is a bad API. Use pathlib wherever possible, except for places interfacing with external things that only understand strings (like printing out to logs/console).

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
   "build",
   "click >= 8",
   "pip >= 22.2",
+  "toml >= 0.10.1; python_version<'3.11'",
```
webknjaz, line 43: Let's make it a little bit more sparse, like the first part.
```suggestion
  "toml >= 0.10.1; python_version < '3.11'",
```

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
...
+
+@functools.lru_cache()
+def parse_config_file(config_file: Path) -> dict[str, Any]:
+    if sys.version_info >= (3, 11):
+        # Python 3.11 stdlib tomllib load() requires a binary file object
+        with config_file.open("rb") as ifs:
+            config = tomllib.load(ifs)
+    else:
+        # Before 3.11, using the external toml library, load requires the filename
+        config = toml.load(str(config_file))
```
webknjaz, line 640: Having seen this, I remembered that CPython specifically chose to mimic the popular library APIs. I also recall that all the major ecosystem projects moving away from some toml libraries. Namely, `toml` was something that stopped being used, along with `tomlkit`. I think, I still have one project using `toml` that I have to migrate.
I've verified that the library you should be using is called `tomli`. I checked that it's now used by `pip`, `pypa/build`, `pytest` and `tox`, among others.

Once you fix that, it'll be possible to make this somewhat more elegant:
```suggestion
    config = tomllib.loads(config_file.read_text(encoding="utf-8"))
```

piptools/utils.py
```diff
@@ -2,14 +2,22 @@
...
+import sys
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Iterable, Iterator, TypeVar, cast
 
 import click
+
+if sys.version_info >= (3, 11):
+    import tomllib
+else:
+    import toml
```
webknjaz, line 20:
```suggestion

if sys.version_info >= (3, 11):
    import tomllib
else:
    import tomli as tomllib

import click
```

pyproject.toml
webknjaz: with the recent comments in the Python modules, this should be
```suggestion
  "tomli >= 0.10.1; python_version < '3.11'",
```

pyproject.toml
```diff
@@ -57,6 +58,7 @@ testing = [
   "pytest >= 7.2.0",
   "pytest-rerunfailures",
   "pytest-xdist",
+  "toml >= 0.10.1",
```
webknjaz, line 61:
```suggestion
  "tomli-w",
```

piptools/utils.py
```diff
@@ -405,9 +414,9 @@ def get_required_pip_specification() -> SpecifierSet:
     Returns pip version specifier requested by current pip-tools installation.
     """
     project_dist = get_distribution("pip-tools")
-    requirement = next(  # pragma: no branch
+    requirement = next(
         (r for r in project_dist.requires() if r.name == "pip"), None
-    )
+    )  # pragma: no branch
```
webknjaz, line 419: This doesn't seem related to this PR.

j00bar: This appears to be an artifact of pre-commit reorganizing code.

piptools/scripts/compile.py
```diff
@@ -340,6 +356,7 @@ def cli(
     emit_index_url: bool,
     emit_options: bool,
     unsafe_package: tuple[str, ...],
+    config: str | None,
```
webknjaz, line 359: Let's make this accept an already normalized `pathlib.Path` object:
```suggestion
    config: Path | None,
```

piptools/utils.py
j00bar: I didn't have an explicit reason for it - I was following a suggestion in the PR to look at the way black implemented this as a click callback and used their code as an inspiration, in which they reconstruct a whole new dict for `default_map` instead of updating the current one. The `git blame` seems to indicate the suggested change you've offered would run afoul of the type-checker.

https://github.com/psf/black/blame/b2ee211b5ad84b62738ac0997b73bf6ee9a74d06/src/black/__init__.py#L129-L134C1

webknjaz, line 572: Upon looking into this more, I'm now convinced that this does not belong here at all. It should happen externally, rather than being tightly coupled to the path argument validation and normalization. Being a side effect only makes the control flow more entangled.

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
...
+            Path(Path.cwd(), src_file).resolve() for src_file in src_files
+        ]
+        candidate_dirs = [
+            src if src.is_dir() else src.parent for src in src_files_as_paths
+        ]
+    config_file_path = next(
+        (
+            candidate_dir / config_file
+            for candidate_dir in candidate_dirs
+            for config_file in (CONFIG_FILE_NAME, "pyproject.toml")
```
webknjaz, line 595: It might be nice to make use of `itertools.product()` here.

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
...
+        # If no src_files were specified, we consider the current directory the only candidate
+        candidate_dirs = [Path.cwd()]
+    else:
+        # Collect the candidate directories based on the src_file arguments provided
+        src_files_as_paths = [
+            Path(Path.cwd(), src_file).resolve() for src_file in src_files
+        ]
+        candidate_dirs = [
+            src if src.is_dir() else src.parent for src in src_files_as_paths
+        ]
```
webknjaz, line 590: Let's make these lazy
```suggestion
        src_files_as_paths = (
            Path(Path.cwd(), src_file).resolve() for src_file in src_files
        )
        candidate_dirs = (
            src if src.is_dir() else src.parent for src in src_files_as_paths
        )
```

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
             markers=parts.markers,
             extras=parts.extras,
         )
+
+
+def callback_config_file_defaults(
```
webknjaz, line 536: After looking into what this function doers, it became clear why it's so hard to call it properly — it does too much and has side effects.

This should probably be a simple input-to-normalized path pure function.

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
...
+    ctx: click.Context, param: click.Parameter, value: str | None
+) -> str | None:
+    """
+    Returns the path to the config file with defaults being used, or `None` if no such file is
+    found.
```
webknjaz, line 541: Let's not forget that the docstring title should be in imperative mood, short and simple, referencing to the action being performed, not describing the entire runtime logic. PEP 8 requires such indented blocks not to exceed the limit of 72 chars:
```suggestion
    """Return the config file path.

    ``None`` is returned if no such file is found.
```
(also, we should only use RST in docstrings)

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
...
+    """
+    Returns the config file to use for defaults given `src_files` provided.
+    """
+    if not src_files:
+        # If no src_files were specified, we consider the current directory the only candidate
+        candidate_dirs = [Path.cwd()]
+    else:
+        # Collect the candidate directories based on the src_file arguments provided
+        src_files_as_paths = [
+            Path(Path.cwd(), src_file).resolve() for src_file in src_files
```
webknjaz, line 586: It took me a minute to get why you're feeding Path into Path... How about
```suggestion
            (Path.cwd() / src_file).resolve() for src_file in src_files
```

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
...
+    "extras",
+    "upgrade_packages",
+    "unsafe_package",
+    "find_links",
+    "extra_index_url",
+    "trusted_host",
+]
+
+
+@functools.lru_cache()
```
webknjaz, line 632: nit: does this need to be called if you don't pass any args?
```suggestion
@functools.lru_cache
```

webknjaz: OTOH, maybe it makes sense to limit the cache size...
```suggestion
@functools.lru_cache(max_size=1)
```

tests/conftest.py
```diff
@@ -450,3 +453,21 @@ def _reset_log():
     with other tests that depend on it.
     """
     log.reset()
+
+
+@pytest.fixture
+def make_config_file(tmpdir_cwd):
+    def _maker(pyproject_param, new_default, config_file_name=CONFIG_FILE_NAME):
```
webknjaz, line 460: could you try adding typing  here?

tests/conftest.py
```diff
@@ -450,3 +453,21 @@ def _reset_log():
     with other tests that depend on it.
     """
     log.reset()
+
+
+@pytest.fixture
+def make_config_file(tmpdir_cwd):
```
webknjaz, line 459: Definitely add a docsting — it's shown in the `pytest --fixtures` output and is a part of good DX/contributore experience.

tests/conftest.py
```diff
@@ -450,3 +453,21 @@ def _reset_log():
...
+        config_path = Path(tmpdir_cwd) / pyproject_param
+        config_file = config_path / config_file_name
+        config_path.mkdir(exist_ok=True)
+
+        config_to_dump = {"pip-tools": {pyproject_param: new_default}}
+        if config_file_name == "pyproject.toml":
+            config_to_dump = {"tool": config_to_dump}
+        with open(config_file, "w") as ofs:
+            toml.dump(config_to_dump, ofs)
+        return config_file
```
webknjaz, line 471: with tomli, it's
```suggestion
        config_file.write_text(tomli_w.dumps(config_to_dump))
        return config_file
```

tests/conftest.py
```diff
@@ -8,9 +8,11 @@
 from contextlib import contextmanager
 from dataclasses import dataclass, field
 from functools import partial
+from pathlib import Path
 from textwrap import dedent
 
 import pytest
+import toml
```
webknjaz, line 15:
```suggestion
import tomli_w
```

README.rst
webknjaz: I don't disagree, just wanted to clarify that I understand this correctly.

README.rst
webknjaz: Yeah, I mostly just wanted to bring attention to this aspect. You wanted to ensure the implementation remains simple, and I saw a possibility to make that happen.

On a more general note, I understand that the tools listed come with a lot of legacy and their tool-specific configs don't use TOML at all.

I have a precedent for you, though. Towncrier, for example, expects `[tool.towncrier]` in both of its supported config file locations — `towncrier.toml` and `pyproject.toml`. And this makes sense — what if an end-user passes a config file path with an arbitrary name of `my-cool-custom-config.toml` — would we expect the section to be `[pip-tools]` or `[tool.pip-tools]`?

piptools/utils.py
webknjaz: Honestly, I wouldn't recommend doing whatever black does, as it has countless shortcomings. To me, it's rather an indication that an idea might be bad.
I understand that in-browser code suggestions often need to be corrected later on, locally. This is still a good way of visualizing the expectations/ideas/suggestions...

README.rst
AndydeCleyre: I notice that for [ruff](https://beta.ruff.rs/docs/configuration/#using-rufftoml), when using a `ruff.toml`/`.ruff.toml`, the settings can either be under `[tool.ruff]`, or without a top level heading at all.

piptools/utils.py
webknjaz: OTOH, I was making the suggestions based on the engineering thinking rather than whether some other project does something in one way or another. Keeping a project easy-to-read/follow and maintainable sometimes requires more attention to detail. As I was progressing with the review, my perception, and understanding changed/became more accurate, which is why I added new suggestions on top of the previous ones. Making pauses helped me gain better perspective..

piptools/utils.py
webknjaz: That's interesting. I wonder why this didn't happen in the PRs bumping the hooks.. Anyway, I'll mark this thread as resolved, it's not a blocker.

webknjaz:
@j00bar I apologize that this all came off quite overwhelming — my reviews are often like this when I get excited, which causes people to think I'm too nitpicky... This is a collaboratively-maintained project, meaning that hundreds of people have maintenance privileges. Though, only a few maintain specific Jazzband projects consistently. I'm reviewing this because I care about certain aspects, plus you pinged me. It could've remained unreviewed for some while otherwise. Of course, there's always a chance somebody would step in to give extra feedback.

Some bits I commented are optional, but nice to have. Others stand firm ground. In particular, that decoupling request.
You mentioned that it'changing it would require rewriting tests substantially but I don't see why — could you clarify what you foresee? I must've missed some implications, eh?

Finally, there's no time pressure to iterate on feedback. But since you mentioned that you'd postpone further commits, would it be okay if I committed something to your branch myself to ease the burden?

.pre-commit-config.yaml
```diff
@@ -33,6 +33,7 @@ repos:
           - toml==0.10.2
           - pip==20.3.4
           - build==0.9.0
+          - types-toml==0.10.2
```
webknjaz, line 36: This will no longer be relevant with `tomli`
```suggestion
```

piptools/scripts/compile.py
```diff
@@ -19,13 +19,14 @@
 from .._compat import parse_requirements
```
webknjaz, line 19:
```suggestion
from pathlib import Path
from .._compat import parse_requirements
```

piptools/scripts/compile.py
```diff
@@ -391,6 +408,9 @@ def cli(
             f"input and output filenames must not be matched: {output_file.name}"
         )
 
+    if config:
+        log.info(f"Using pip-tools configuration defaults found in '{config}'.")
```
webknjaz, line 412:
```suggestion
        log.info(f"Using pip-tools configuration defaults found in '{config !s}'.")
```

piptools/scripts/sync.py
```diff
@@ -103,6 +120,7 @@ def cli(
     client_cert: str | None,
     src_files: tuple[str, ...],
     pip_args: str | None,
+    config: str | None,
```
webknjaz, line 123:
```suggestion
    config: Path | None,
```

piptools/scripts/sync.py
```diff
@@ -127,6 +145,9 @@ def cli(
             log.error("ERROR: " + msg)
             sys.exit(2)
 
+    if config:
+        log.info(f"Using pip-tools configuration defaults found in '{config}'.")
```
webknjaz, line 149:
```suggestion
        log.info(f"Using pip-tools configuration defaults found in '{config !s}'.")
```

piptools/scripts/sync.py
```diff
@@ -17,9 +17,11 @@
 from .._compat import parse_requirements
```
webknjaz, line 17:
```suggestion
from pathlib import Path
from .._compat import parse_requirements
```

piptools/utils.py
```diff
@@ -522,3 +531,120 @@ def parse_requirements_from_wheel_metadata(
             markers=parts.markers,
             extras=parts.extras,
         )
+
+
+def callback_config_file_defaults(
+    ctx: click.Context, param: click.Parameter, value: str | None
+) -> str | None:
```
webknjaz, line 538:
```suggestion
) -> Path | None:
```

piptools/utils.py
webknjaz:
```suggestion
    return config_file
```

tests/test_utils.py
```diff
@@ -540,3 +544,113 @@ def test_get_sys_path_for_python_executable():
...
+        ("unsafe-package", ["changed"]),
+    ),
+)
+def test_callback_config_file_defaults(pyproject_param, new_default, make_config_file):
+    config_file = make_config_file(pyproject_param, new_default)
+    # Create a "compile" run example pointing to the config file
+    ctx = Context(compile_cli)
+    ctx.params["src_files"] = (str(config_file),)
+    found_config_file = callback_config_file_defaults(ctx, "config", None)
+    assert found_config_file == str(config_file)
```
webknjaz, line 600:
```suggestion
    assert found_config_file == config_file
```

piptools/utils.py
webknjaz: I'll try to see if that 3-year old problem got solved by now..

webknjaz: > The `git blame` seems to indicate the suggested change you've offered would run afoul of the type-checker.

Good news — mypy didn't explode on this.

piptools/utils.py
```diff
@@ -522,3 +531,129 @@ def parse_requirements_from_wheel_metadata(
...
+        config_file = select_config_file(ctx.params.get("src_files", ()))
+        if config_file is None:
+            return None
+    else:
+        config_file = Path(value)
+
+    if config := parse_config_file(config_file):
+        _assign_config_to_cli_context(ctx, config)
+    else:
+        return None
```
webknjaz, line 558: @j00bar through a series of code conversions I discovered that the check here is rather weird — `if (not) config:` checks for a falsy value (which a present but empty section would be).
This seems to be semantically incorrect. If there's an empty config, we'd still want to recognize its existence.
Is this the reason you thought it'd be hard to decouple this?
Why do we even want to pretend a config file wasn't passed at all to the CLI when we don't see a config there? I don't think we should do that..

tests/test_utils.py
```diff
@@ -540,3 +544,113 @@ def test_get_sys_path_for_python_executable():
...
+
+
+def test_callback_config_file_defaults_precedence(make_config_file):
+    piptools_config_file = make_config_file("newline", "LF")
+    project_config_file = make_config_file("newline", "CRLF", "pyproject.toml")
+    ctx = Context(compile_cli)
+    ctx.params["src_files"] = (str(project_config_file),)
+    found_config_file = callback_config_file_defaults(ctx, "config", None)
+    # The pip-tools specific config file should take precedence over pyproject.toml
+    assert found_config_file == str(piptools_config_file)
```
webknjaz, line 644:
```suggestion
    assert found_config_file == piptools_config_file
```

piptools/scripts/compile.py
```diff
@@ -19,13 +20,14 @@
...
 from ..exceptions import NoCandidateFound, PipToolsError
-from ..locations import CACHE_DIR
+from ..locations import CACHE_DIR, CONFIG_FILE_NAME
 from ..logging import log
 from ..repositories import LocalRequirementsRepository, PyPIRepository
 from ..repositories.base import BaseRepository
 from ..resolver import BacktrackingResolver, LegacyResolver
 from ..utils import (
     UNSAFE_PACKAGES,
+    callback_config_file_defaults,
```
webknjaz, line 30:
```suggestion
    determine_config_file,
```

piptools/scripts/compile.py
```diff
@@ -302,6 +304,21 @@ def _determine_linesep(
...
+        file_okay=True,
+        dir_okay=False,
+        readable=True,
+        allow_dash=False,
+        path_type=str,
+    ),
+    help=f"Read configuration from TOML file. By default, looks for a {CONFIG_FILE_NAME} or "
+    "pyproject.toml.",
+    is_eager=True,
+    callback=callback_config_file_defaults,
```
webknjaz, line 320:
```suggestion
    callback=determine_config_file,
```

piptools/scripts/sync.py
```diff
@@ -17,9 +18,11 @@
...
 from .._compat.pip_compat import Distribution
 from ..exceptions import PipToolsError
+from ..locations import CONFIG_FILE_NAME
 from ..logging import log
 from ..repositories import PyPIRepository
 from ..utils import (
+    callback_config_file_defaults,
```
webknjaz, line 25:
```suggestion
    determine_config_file,
```

piptools/scripts/sync.py
```diff
@@ -86,6 +89,21 @@
...
+        file_okay=True,
+        dir_okay=False,
+        readable=True,
+        allow_dash=False,
+        path_type=str,
+    ),
+    help=f"Read configuration from TOML file. By default, looks for a {CONFIG_FILE_NAME} or "
+    "pyproject.toml.",
+    is_eager=True,
+    callback=callback_config_file_defaults,
```
webknjaz, line 105:
```suggestion
    callback=determine_config_file,
```

tests/test_utils.py
```diff
@@ -5,19 +5,23 @@
...
 
 import pip
 import pytest
+from click import BadOptionUsage, Context, FileError
 from pip._vendor.packaging.version import Version
 
 from piptools.scripts.compile import cli as compile_cli
 from piptools.utils import (
     as_tuple,
+    callback_config_file_defaults,
```
webknjaz, line 18:
```suggestion
    determine_config_file,
```

tests/test_utils.py
```diff
@@ -540,3 +544,113 @@ def test_get_sys_path_for_python_executable():
...
+        ("emit-options", False),
+        ("unsafe-package", ["changed"]),
+    ),
+)
+def test_callback_config_file_defaults(pyproject_param, new_default, make_config_file):
+    config_file = make_config_file(pyproject_param, new_default)
+    # Create a "compile" run example pointing to the config file
+    ctx = Context(compile_cli)
+    ctx.params["src_files"] = (str(config_file),)
+    found_config_file = callback_config_file_defaults(ctx, "config", None)
```
webknjaz, line 599:
```suggestion
    found_config_file = determine_config_file(ctx, "config", None)
```

tests/test_utils.py
```diff
@@ -540,3 +544,113 @@ def test_get_sys_path_for_python_executable():
...
+        "extra-index-url",
+        "trusted-host",
+    ),
+)
+def test_callback_config_file_defaults_multi_value_options(mv_option, make_config_file):
+    config_file = make_config_file(mv_option, "not-a-list")
+    ctx = Context(compile_cli)
+    ctx.params["src_files"] = (str(config_file),)
+    with pytest.raises(BadOptionUsage, match="must be a list"):
+        callback_config_file_defaults(ctx, "config", None)
```
webknjaz, line 622:
```suggestion
        determine_config_file(ctx, "config", None)
```

tests/test_utils.py
```diff
@@ -540,3 +544,113 @@ def test_get_sys_path_for_python_executable():
...
+def test_callback_config_file_defaults_bad_toml(make_config_file):
+    config_file = make_config_file("verbose", True)
+    # Simple means of making invalid TOML: have duplicate keys
+    with Path(config_file).open("r+") as fs:
+        config_text_lines = fs.readlines()
+        fs.write(config_text_lines[-1])
+    ctx = Context(compile_cli)
+    ctx.params["src_files"] = (str(config_file),)
+    with pytest.raises(FileError, match="Could not parse "):
+        callback_config_file_defaults(ctx, "config", None)
```
webknjaz, line 634:
```suggestion
        determine_config_file(ctx, "config", None)
```

tests/test_utils.py
```diff
@@ -540,3 +544,113 @@ def test_get_sys_path_for_python_executable():
...
+
+
+def test_callback_config_file_defaults_precedence(make_config_file):
+    piptools_config_file = make_config_file("newline", "LF")
+    project_config_file = make_config_file("newline", "CRLF", "pyproject.toml")
+    ctx = Context(compile_cli)
+    ctx.params["src_files"] = (str(project_config_file),)
+    found_config_file = callback_config_file_defaults(ctx, "config", None)
```
webknjaz, line 642:
```suggestion
    found_config_file = determine_config_file(ctx, "config", None)
```

tests/test_utils.py
```diff
@@ -540,3 +544,113 @@ def test_get_sys_path_for_python_executable():
...
+    # The pip-tools specific config file should take precedence over pyproject.toml
+    assert found_config_file == piptools_config_file
+    lookup_param = get_click_dest_for_option("newline")
+    assert ctx.default_map[lookup_param] == "LF"
+
+
+def test_callback_config_file_defaults_unreadable_toml(make_config_file):
+    ctx = Context(compile_cli)
+    with pytest.raises(FileError, match="Could not read "):
+        callback_config_file_defaults(
```
webknjaz, line 652:
```suggestion
        determine_config_file(
```

README.rst
webknjaz: I think this paragraph should mention the corner case of what's happening when the input is coming from stdin. Also, should we fall back to the output file directory when we read from stdin but the output is specified via a CLI option?

piptools/utils.py
```diff
@@ -522,3 +531,129 @@ def parse_requirements_from_wheel_metadata(
...
+    as the requirements input file.
+    """
+    if value is None:
+        config_file = select_config_file(ctx.params.get("src_files", ()))
+        if config_file is None:
+            return None
+    else:
+        config_file = Path(value)
+
+    if config := parse_config_file(config_file):
```
q0w, line 555: py37 is already dropped?

webknjaz, line 555: For some reason I thought we already have access to walrus... Python 3.7 is going EOL in 1.5 months so we'll have to wait.
```suggestion
    config = parse_config_file(config_file)
    if config:
```

piptools/utils.py
```diff
@@ -522,3 +531,130 @@ def parse_requirements_from_wheel_metadata(
...
+        click_context.default_map = {}
+
+    click_context.default_map.update(cli_config_mapping)
+
+
+def select_config_file(src_files: tuple[str, ...]) -> Path | None:
+    """
+    Returns the config file to use for defaults given `src_files` provided.
+    """
+    if not src_files:
```
webknjaz, line 578: I've found a way to reduce branching here. Will add a commit for this.

piptools/utils.py
webknjaz: This would probably be an overkill, marking this thread as resolved.

README.rst
webknjaz: My original motivation was to help simplify the patch and the code base in general. Adding a conditional would keep some of the complexity. Do we really want that in order to save people one line? I'm conflicted here. Though, it's not as important to me so I won't be insisting of a specific resolution of this.

webknjaz:
@j00bar I've addressed some of the comments and I'm making a pause now. Maybe I'll get back to this later, or maybe you will. There's some extra discussions that happened FYI, please check them out and respond as you have time.

piptools/scripts/compile.py
```diff
@@ -391,6 +409,9 @@ def cli(
             f"input and output filenames must not be matched: {output_file.name}"
         )
 
+    if config:
+        log.info(f"Using pip-tools configuration defaults found in '{config !s}'.")
```
chrysle, line 413: This line should get a test.

piptools/scripts/sync.py
```diff
@@ -127,6 +146,9 @@ def cli(
             log.error("ERROR: " + msg)
             sys.exit(2)
 
+    if config:
+        log.info(f"Using pip-tools configuration defaults found in '{config !s}'.")
```
chrysle, line 150: Same here.

webknjaz:
@chrysle creating merge commits in PR branches brings contributors pain. It's best not to add them for the sake or adding, especially since there's clearly more things to do here and this causes unnecessary disruptions on shared branches like this.

chrysle:
Apologies @webknjaz 

I wanted to fix the RTD build. I will remember this for the future.

webknjaz:
> I wanted to fix the RTD build.

It'd be fixed on the next commit by itself, I think. Most CI-like systems use the ephemeral merge commit of PR branches into main which combines states of both.

If anything, I'd go for using the rebase setting of the branch update, but only if the PR was about to be merged so that the contributing experience isn't messed up by force pushes.

piptools/utils.py
webknjaz: Marking this thread as resolved.

piptools/utils.py
webknjaz: Marking this thread as resolved.

piptools/utils.py
webknjaz: Marking this thread as resolved.

tests/test_utils.py
webknjaz: Marking this thread as resolved.

tests/test_utils.py
webknjaz: Marking this thread as resolved.

pyproject.toml
webknjaz: Marking this thread as resolved.

piptools/utils.py
webknjaz: Marking this thread as resolved.

piptools/utils.py
webknjaz: Marking this thread as resolved.

piptools/utils.py
webknjaz: Marking this thread as resolved.

piptools/utils.py
webknjaz: Marking this thread as resolved.

piptools/scripts/compile.py
webknjaz: Marking this thread as resolved.

README.rst
j00bar: I think my recommendation would be to maintain consistency with the rest of the ecosystem insofar as we can ascertain it. I would accept an incremental degree of code complexity to achieve a better developer experience with minimal cognitive load.

Surveying popular tools in the Python ecosystem, it seems that `[tool.pip-tools]` is the convention for TOML files and `[pip-tools]` would be the convention for INI-style files. I'll make it consistent, since we're using TOML everywhere.

piptools/utils.py
j00bar: You're correct - thank you.

piptools/utils.py
j00bar: I think this bears more discussion.

The order of precedence in determining the value for a particular option is: 1. An explicit declaration at runtime via a flag; 2. A default specified in a config file; 3. A default hard-coded in pip-tools. The configuration file's values need to override the built-in defaults but not any explicitly declared values. By the time the `cli()` function is called, explicit declarations and hard-coded defaults have already been resolved with one another, and it becomes convoluted to try to determine for which options the config file's values should be operative. The same holds true for decorators.

By using an `is_eager=True` option with a callback, we have the opportunity to selectively override hard-coded defaults with new config-file-specified defaults before any commandline explicit options are handled. I don't see a simple place to inject this order of precedence otherwise, and from Click documentation, that follows the intent of the `is_eager` flag:

> Sometimes, you want a parameter to completely change the execution flow. [ ... ] In such cases, you need two concepts: eager parameters and a callback.

If you can imagine an approach using a decorator or in the `cli()` function that can nimbly achieve this order of precedence, I'm open to hearing it.

README.rst
webknjaz: Fair enough. Marking as resolved.

tests/conftest.py
webknjaz: Marking this as resolved.

tests/test_utils.py
```diff
@@ -540,3 +544,113 @@ def test_get_sys_path_for_python_executable():
...
+        ("emit-find-links", False),
+        ("cache-dir", "changed"),
+        ("resolver", "backtracking"),
+        ("emit-index-url", False),
+        ("emit-options", False),
+        ("unsafe-package", ["changed"]),
+    ),
+)
+def test_callback_config_file_defaults(pyproject_param, new_default, make_config_file):
+    config_file = make_config_file(pyproject_param, new_default)
```
webknjaz, line 595: The need to inject this invocation into every test could be eliminated by making the params indirect (this needs the respective fixtures to depend on `request` and read `request.param`).

tests/test_utils.py
```diff
@@ -540,3 +544,113 @@ def test_get_sys_path_for_python_executable():
...
+    assert ctx.default_map[lookup_param] == "LF"
+
+
+def test_callback_config_file_defaults_unreadable_toml(make_config_file):
+    ctx = Context(compile_cli)
+    with pytest.raises(FileError, match="Could not read "):
+        determine_config_file(
+            ctx,
+            "config",
+            "/path/does/not/exist/my-config.toml",
```
webknjaz, line 655: This would probably be a better non-existence guarantee:
```suggestion
            "/dev/null/path/does/not/exist/my-config.toml",
```

webknjaz:
@j00bar I'll need to dedicate more time to reviewing this, but there's one pattern I noticed that has to be fixed — your docstrings use single backticks for things that should be marked as inline code. In RST, unlike MD, these must be double backticks. The single ones have different semantics. There's currently no sane way to have MD in docstrings and RST is the only thing virtually any Sphinx-based project is able to consume.

webknjaz:
@j00bar looks like you accidentally re-reverted [`4980a8f` (#1863)](https://github.com/jazzband/pip-tools/pull/1863/commits/4980a8f23a94e9fd67afefbfcd7f768acf048cdb). Could you re-apply/cherry-pick this commit yet again for Python 3.7 compat?

tests/test_utils.py
j00bar: It's a clever idea, and if I we were making the config file based on the key and value being passed in parametrized values in multiple tests, I'd agree it may be worth the effort. But this is the only invocation that gets all of its arguments from test parameters - the others either have some or all that are constants. Given that, indirect parameterization feels like over-engineering that makes the test scenario more opaque. Thoughts?

webknjaz: Okay, this one not as important. Marking as resolved for now.

webknjaz:
Thanks for the updates! It may take me some time to review the PR — I got ill 🤒 and my brain is rather defunct atm..
Though, if I'm not answering for too long, feel free to ping me. TIA!

j00bar:
> I got ill 🤒

Damn. Feel better, friend. The PR will wait.

webknjaz:
Thanks, y'all! I'm mostly fine now, so I'll try to review the updates.

piptools/utils.py
webknjaz: @j00bar looks like this behavior hasn't been corrected.

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
...
+    """
+    if value is None:
+        config_file = select_config_file(ctx.params.get("src_files", ()))
+        if config_file is None:
+            return None
+    else:
+        config_file = Path(value)
+
+    config = parse_config_file(config_file)
+    if config:
```
webknjaz, line 557: Per https://github.com/jazzband/pip-tools/pull/1863#discussion_r1217283171, this should probably be
```suggestion
    if config is not None:
```

piptools/utils.py
webknjaz: Okay, now I understand the challenges better. Marking this thread as resolved.

README.rst
webknjaz: Agreed. Marking this thread as resolved.

piptools/utils.py
webknjaz: @j00bar I think this is the last bit that needs clearing up... And maybe a test?

j00bar: I don't think this suggestion makes sense anymore - in a [prior commit](https://github.com/jazzband/pip-tools/pull/1863/commits/93507f9c021d4b23d0379c988c80ed5fc999a8e4#diff-4d72add98acf8c426866899357a936987e567ad30d6b9064c97d94ea96cd77baL558-L559) I removed an `else: return None`, which was the problem, as you said: an empty configuration file shouldn't be treated as if there were no configuration file found/specified.

`config`'s value comes from `parse_config_file`, which returns a `dict[str, Any]`, so `config` would never be None. And since `_assign_config_to_cli_context` basically does a `default_map.update()` using the contents of `config`, with an empty dict that would be a no-op, so this conditional saves us that call.

webknjaz:
@j00bar with the rst to md conversion, there's a conflict with the readme now that needs solving.

piptools/utils.py
```diff
@@ -522,3 +531,126 @@ def parse_requirements_from_wheel_metadata(
             markers=parts.markers,
             extras=parts.extras,
         )
+
+
+def determine_config_file(
```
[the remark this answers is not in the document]
webknjaz: Yes, love it! Marking this as resolved.

j00bar:
Rebased and translated the README changes to Markdown. :+1:

piptools/utils.py
webknjaz: Alright, I see. Marking this thread as resolved.

webknjaz: It would be nice to have a regression test with this corner case but it's not a blocker.

piptools/scripts/sync.py
webknjaz, line 150: @j00bar it'd be nice to know why isn't this covered..

webknjaz (approved):
I hereby declare this worthy! Excellent work!

Any additional concerns can be addressed in follow-ups.

AndydeCleyre:
With this, how does one ensure no arguments from configuration files interfere with what's manually run (and visible) on the command line? Do we need to specify `--config /dev/null`, if that works, to prevent surprises, and ensure consistent behavior with the current release?

AndydeCleyre:
Also, are the config-file defaults merged with what's explicitly passed on the command line, or entirely replaced if any options (or args?) are passed?

j00bar:
@AndydeCleyre The order of precedence, from highest to lowest:

* Explicit command-line parameter
* Config file
* Hard coded defaults

So if you specify something on the command-line, it overrides whatever is in the config file. If you specify something in the config file, it overrides whatever is hard coded into pip-tools.

For settings with multiple values, there's no merging. It follows the order of precedence.

Does that make sense?

AndydeCleyre:
That makes sense but it doesn't clear up the questions for me.

1. If I have the option `cert` specified in the config file, and I pass `--annotation-style=line` on the command line, will the `cert` option be used?
2. If I want to run `pip-compile` with the program's defaults, even if there exists a hidden `.pip-tools.toml` file in the current directory (i.e. I want it to act as `pip-compile` does in the current release), is it now necessary that I change all current calls of arg-less `pip-compile` to `pip-compile --config /dev/null`? Will that work? Is that the only way to achieve this?

webknjaz:
These are great points. It's probably worth adding said test cases to make the expectations explicit. It might be nice to document this too and implement a `--no-config`..

j00bar:
> 1. If I have the option `cert` specified in the config file, and I pass `--annotation-style=line` on the command line, will the `cert` option be used?

Yes. Each parameter follows the order of precedence independently. The full configuration, like most command-line utilities, uses a mosaic of command-line set arguments, config file arguments, and hardcoded defaults.

> > 2. If I want to run `pip-compile` with the program's defaults, even if there exists a hidden `.pip-tools.toml` file in the current directory (i.e. I want it to act as `pip-compile` does in the current release), is it now necessary that I change all current calls of arg-less `pip-compile` to `pip-compile --config /dev/null`? Will that work? Is that the only way to achieve this?

>  It might be nice to... implement a --no-config

I don't agree. Projects use these config files to set defaults that should apply to the whole project to make the preferred way of compiling/syncing requirements be the easiest way to compile/sync requirements. That encourages consistency transparently. As the steward of many projects and having to have created automation to fail a PR when it fails consistency in the compiled requirements, I want it as laborious as possible to deviate from the project's default settings. And if there are deviations I often make, I'm going to make an additional config file with a non-default config file name.

But you're a maintainer of this one - it's up to you if you want that - but I'm not interested in helping with that one.

AndydeCleyre:
> @AndydeCleyre I wonder why is this backwards incompatible?

I still don't know the answer to this question, but this is what I was thinking of:

> If I want to run `pip-compile` with the program's defaults, even if there exists a hidden `.pip-tools.toml` file in the current directory (i.e. I want it to act as `pip-compile` does in the current release), is it now necessary that I change all current calls of arg-less `pip-compile` to `pip-compile --config /dev/null`? Will that work? Is that the only way to achieve this?

atugushev:
> I still don't know the answer to this question

@AndydeCleyre 

I'm inclined to think that this update does not represent a breaking change. The current update does not interfere with pip-compile's default behavior, nor does it require any modifications to existing calls. Running pip-compile without any arguments will still function as it does in the current release and shouldn't break any existing pipelines, as there were no configurations before.

However, I believe we should consider adding support for `--no-config` (similar to how it's done in wget or ripgrep) in the upcoming version and resolve other concerns.

Are there any objections if I remove the corresponding label and proceed with the version 6.14.0 (#1892) release process?

AndydeCleyre:
Ok, remove the breaking label and status. 

I would really like to know how to consistently protect myself from unwanted configs once this is released. 

And yes, that absolute path is going to be a paper cut for any teams that include the header and run pip-compile on different systems. IMO all paths in the txt output should be relative to the txt file.

atugushev:
@AndydeCleyre thanks!

I've prepared two PRs to address [my](https://github.com/jazzband/pip-tools/pull/1863#issuecomment-1611728125):
-  https://github.com/jazzband/pip-tools/pull/1893
- https://github.com/jazzband/pip-tools/pull/1894


Let's get this merged before release.

atugushev:
Released as part of pip-tools v6.14.0.

webknjaz:
Hey @j00bar, FYI there's some post-release adjustment and bugfixing efforts going on, in case you wanted to keep an eye on the changes, here's some:
- https://github.com/jazzband/pip-tools/pull/1893
- https://github.com/jazzband/pip-tools/pull/1896
- https://github.com/jazzband/pip-tools/issues/1902
- https://github.com/jazzband/pip-tools/issues/1903
- https://github.com/jazzband/pip-tools/pull/1904

AndydeCleyre:
When using `.pip-tools.toml`, does there need to be a `[tool.pip-tools]` section, or should the values be top-level with no section header? Whichever the case, can it be noted somehow in the README?

chrysle:
@AndydeCleyre 

> When using `.pip-tools.toml`, does there need to be a `[tool.pip-tools]` section, or should the values be top-level with no section header? 

See these lines:

https://github.com/jazzband/pip-tools/blob/776c3fa3d8b2da155eed7eb9a5b164f78937987a/piptools/utils.py#L656-L657

Maybe this could be simplified for `.pip-tools.toml`.

> Whichever the case, can it be noted somehow in the README?

I think it's explained properly under [the configuration section](https://github.com/jazzband/pip-tools#configuration).

j00bar:
> Maybe this could be simplified for .pip-tools.toml.

That was the original implementation which, in review, was rejected in favor of consistency.

AndydeCleyre:
Thanks. I thought there was a discussion about making it like ruff, but I can't find it so maybe dreamed it. From ruff's docs:

> As an alternative to `pyproject.toml`, Ruff will also respect a `ruff.toml` (or `.ruff.toml`) file, which implements an equivalent schema (though in the `ruff.toml` and `.ruff.toml` versions, the `[tool.ruff]` header is omitted).

webknjaz:
> Thanks. I thought there was a discussion about making it like ruff, but I can't find it so maybe dreamed it. From ruff's docs:
> 
> > As an alternative to `pyproject.toml`, Ruff will also respect a `ruff.toml` (or `.ruff.toml`) file, which implements an equivalent schema (though in the `ruff.toml` and `.ruff.toml` versions, the `[tool.ruff]` header is omitted).

FTR, that discussion that you're recalling was about the need to have the section header in all of the files or not. It was you who mentioned Ruff in that context: https://github.com/jazzband/pip-tools/pull/1863#discussion_r1187752607.

And I made an argument that it's more consistent to have this header in all files: https://github.com/jazzband/pip-tools/pull/1863#discussion_r1188761535. It makes it easier to copy the section over between different files. And now I think that when the end-users want to have several sections in their non-pyproject config, they can only do so when there are sections in the file.

PR #1867, opened 2023-05-10
Improve warning for `pip-compile` if no `--allow-unsafe` was passed

### LINKED ISSUE #1337: Improve --allow-unsafe warning comment
Currently, if `pip-compile` suppresses a requirement because the `--allow-unsafe` flag wasn't passed, it includes a warning that reads something like this:

```
# WARNING: The following packages were not pinned, but pip requires them to be
# pinned when the requirements file includes hashes. Consider using the --allow-unsafe flag.
# pip
```

This wording is a little incorrect because it's not always the case that pip requires them to be pinned. Namely, if the requirement is already satisfied by a package that's already installed, then pip won't check any hashes (because it won't need to install anything for that requirement). Because of this, a user seeing the current message can get the mistaken idea that they don't have any options other than passing `--allow-unsafe`.

Thus, I'd like to suggest that the wording be modified slightly to read something like:

```
# WARNING: The following packages were not pinned, but pip requires them to be
# pinned when the requirements file includes hashes and the requirement isn't
# already satisfied by a package already installed. Consider using the
# --allow-unsafe flag.
# pip
```

(As a side note, there is a third option in addition to (1) commenting out the "unsafe" requirement, and (2) including the pinned requirement with hashes. Namely, the pinned requirement can be included without hashes. Doing that third option would enforce that the user has the correct version already installed by erroring out if they don't, while not risking doing an "unsafe" operation by uninstalling / installing the package in question. The reason is that, by not including the hashes, it prevents a new install from taking place.)

chrysle:
Based on a suggestion from and fixes #1337

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -32,7 +32,7 @@ MESSAGE_UNHASHED_PACKAGE = comment(
 
 MESSAGE_UNSAFE_PACKAGES_UNPINNED = comment(
     "# WARNING: The following packages were not pinned, but pip requires them to be"
-    "\n# pinned when the requirements file includes hashes and the requirement is not "
+    "\n# pinned when the requirements file includes hashes and the requirement is not"
     "\n# satisfied by a package already installed. "
     "Consider using the --allow-unsafe flag."
 )


webknjaz:
Look like GitHub is drunk today — there's a lot of internal GHA crushes. It might be time to take a break and come back to restarting the jobs later today or tomorrow..

piptools/writer.py
```diff
@@ -32,7 +32,8 @@
 
 MESSAGE_UNSAFE_PACKAGES_UNPINNED = comment(
     "# WARNING: The following packages were not pinned, but pip requires them to be"
-    "\n# pinned when the requirements file includes hashes. "
+    "\n# pinned when the requirements file includes hashes and the requirement is not "
```
webknjaz, line 35: Let's not print out a trailing space — the line continuation adds a newline right away.
```suggestion
    "\n# pinned when the requirements file includes hashes and the requirement is not"
```

chrysle:
Looks like GitHub is cured again ;-)

webknjaz (approved):

PR #1862, opened 2023-05-05
Convert the README from rst to md

shatakshiiii:
This PR converts the documentation from README.rst to README.md
Related: https://github.com/jazzband/pip-tools/issues/1856
##### Contributor checklist

- [ ] Provided the tests for the changes.
- [x] Assure PR title is short, clear, and good to be included in the user-oriented changelog

##### Maintainer checklist

- [ ] Assure one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

Issue #1845, opened 2023-04-06
Refactor README


atugushev:
#### What's the problem this feature will solve?

Since the README file has become quite large, perhaps it's time we consider refactoring it. 

#### Describe the solution you'd like

- Make README small and concise focusing on important things
- Keep on README description, svg image, installation and basic usage of the pip-tools
- Move the other sections to https://pip-tools.rtfd.io/

References:
- https://github.com/pypa/pip
- https://github.com/pypa/build
- https://github.com/pypa/virtualenv
- https://github.com/pallets/click
- https://github.com/pytest-dev/pytest
- https://github.com/pytest-dev/pytest-asyncio

#### Alternative Solutions

Any suggestions?

#### Additional context

- https://fosstodon.org/@jamescooke/109954634904286238
- https://fosstodon.org/@jamescooke/110125938514902869

/cc @jamescooke

chrysle:
>     * Keep on README description, svg image, installation and basic usage of the pip-tools
>     * Move the other sections to https://pip-tools.rtfd.io/

I'm +1 on this! I'd suggest a style like in the [python packaging guide](https://packaging.python.org/en/latest/), with different Markdown files for the sections.

I also think intersphinx mapping should be configured for the many project references in the docs. 

There is already an attempt to convert the README to Markdown (#1862), I would wait until this is finished. I'd refactor the index.rst file and convert it to Markdown then.

I would like to help with that, should I work on a PR?

webknjaz:
> I would like to help with that, should I work on a PR?

You could try picking up #1862 as it's not clear if @shatakshiiii wants to iterate on it anytime soon. Though, if you want to push commits into their branch, make sure to clearly communicate your intent in that PR comments and only add new commits on top without force-pushing the branch (unless they consent). Alternatively, you could try starting off of their branch (to get it, add their remote via `git remote add shatakshiiii git@github.com:shatakshiiii/pip-tools.git` to your local repo and you'll be able to fetch it (though using `git fetch git@github.com:shatakshiiii/pip-tools.git` would probably work too) — then, you'd push your own branch to your fork and make a separate PR.

> I would wait until this is finished. I'd refactor the index.rst file and convert it to Markdown then.

[...]

PR #1871, opened 2023-05-12
Convert README to Markdown

chrysle:
Part of #1845, picks up work from and closes #1862 

##### Contributor checklist

##### Maintainer checklist


README.md
```diff
@@ -0,0 +1,580 @@
...
+`pip-compile` should be run from the same virtual environment as your
+project so conditional dependencies that require a specific Python version,
+or other environment markers, resolve relative to your project's
+environment.
+
+**Note**: If `pip-compile` finds an existing `requirements.txt` file that
+fulfils the dependencies then no changes will be made, even if updates are
+available. To compile from scratch, first delete the existing
+`requirements.txt` file, or see
+[Updating requirements](https://github.com/jazzband/pip-tools#updating-requirements)
```
chrysle, line 47: I think I'm going to make a relative link out of that.

pyproject.toml
```diff
@@ -8,7 +8,7 @@ requires-python = ">=3.7"
 dynamic = ["version"]
 name = "pip-tools"
 description = "pip-tools keeps your pinned dependencies fresh."
-readme = "README.rst"
+readme = "README.md"
```
webknjaz, line 11: I wonder if this will need a long description type set explicitly...

webknjaz: Try building the dist via `python -m dist` (there might be a tox env for this) and inspect it with `twine --strict`.

docs/index.md
webknjaz, line 1: I'd probably put the Sphinx stuff into a separate PR if possible.

docs/conf.py
```diff
@@ -52,3 +52,4 @@
 # -------------------------------------------------------------------------
 default_role = "any"
 nitpicky = True
+suppress_warnings = ["myst.xref_missing"]
```
webknjaz, line 55: Is it possible not to ignore this? (also a separate PR candidate)

README.md
```diff
@@ -0,0 +1,580 @@
...
+You might use `pip-compile` as a hook for the [pre-commit](https://github.com/pre-commit/pre-commit).
+See [pre-commit docs](https://pre-commit.com/) for instructions.
+Sample `.pre-commit-config.yaml`:
+
+```yaml
+repos:
+    - repo: https://github.com/jazzband/pip-tools
+    rev: 6.13.0
+    hooks:
+        - id: pip-compile
```
webknjaz, line 395: YAML indentation is broken here, let's try this:
```suggestion
- repo: https://github.com/jazzband/pip-tools
  rev: 6.13.0
  hooks:
  - id: pip-compile
```

README.md
```diff
@@ -0,0 +1,580 @@
...
+```
+
+You might want to customize `pip-compile` args by configuring `args` and/or `files`, for example:
+
+```yaml
+repos:
+    - repo: https://github.com/jazzband/pip-tools
```
webknjaz, line 402: Same here

README.md
```diff
@@ -0,0 +1,580 @@
...
+        - id: pip-compile
+        files: ^requirements/production\.(in|txt)$
+        args: [--index-url=https://example.com, requirements/production.in]
+```
+
+If you have multiple requirement files make sure you create a hook for each file.
+
+```yaml
+repos:
+    - repo: https://github.com/jazzband/pip-tools
```
webknjaz, line 414: and here

README.md
```diff
@@ -0,0 +1,580 @@
...
+We suggest to use the `{env}-requirements.txt` format
+(ex: `win32-py3.7-requirements.txt`, `macos-py3.10-requirements.txt`, etc.).
+
+### Cross-environment usage of `requirements.in`/`requirements.txt` and `pip-compile`
+
+The dependencies of a package can change depending on the Python environment in which it
+is installed. Here, we define a Python environment as the combination of Operating
+System, Python version (3.7, 3.8, etc.), and Python implementation (CPython, PyPy,
+etc.). For an exact definition, refer to the possible combinations of [PEP 508
+environment markers](https://peps.python.org/pep-0508/#environment-markers).
```
webknjaz, line 501: Plz move this to a detached reference.

README.md
```diff
@@ -0,0 +1,580 @@
...
+        args: [requirements-dev.in]
+        files: ^requirements-dev\.(in|txt)$
+        - id: pip-compile
+        name: pip-compile requirements-lint.in
+        args: [requirements-lint.in]
+        files: ^requirements-lint\.(in|txt)$
+        - id: pip-compile
+        name: pip-compile requirements.txt
+        args: [requirements.txt]
+        files: ^requirements\.(in|txt)$
```
webknjaz, line 431: Make sure to incorporate https://github.com/jazzband/pip-tools/pull/1847 and other commits that changed `README.rst` post #1862: https://github.com/jazzband/pip-tools/commits/main/README.rst.

pyproject.toml
chrysle: > I wonder if this will need a long description type set explicitly...

I changed this line because I received a traceback about missing long description.

docs/conf.py
chrysle: It'd be needed to create [cross-reference targets](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html) for all the detached links and images.

docs/index.md
chrysle: This would break the tests?

README.md
chrysle: `pre-commit` caught the indentation for us ;-)

pyproject.toml
webknjaz: I know. I'm just not sure if that should be paired with another setting.

docs/index.md
webknjaz: Not necessarily. You could include the new file instead of rewriting an extra document entirely…

docs/conf.py
webknjaz: How so?

pyproject.toml
webknjaz: Look at the table @ https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata:
It says:
> If `content-type` is not given, `"text/x-rst"` is used by default.

So I'm almost sure this would render poorly on PyPI. Could you build the package off of this branch, then unpack the tarball (`python -m tarfile`) and the wheel (`wheel unpack`) and look into the metadata files to see if there's a mention of the markdown type? If it doesn't say markdown, maybe we have to set that explicitly.

docs/index.md
webknjaz, line 1: This could also be submitted separately and just include the RST file.

pyproject.toml
chrysle: The METADATA file of the wheel and the PKG-INFO of the built package contain the following line:

```
Description-Content-Type: text/markdown
```

This seems to point out that explicitly specifying the content type in the pyproject.toml is not necessary. I would also be surprised about everything else, since the [python packaging guide](https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata) recommends this configuration.

docs/conf.py
chrysle: It's fixed here, but there will be some extra work once I convert the index in the other PR. Let's merge this first?

docs/index.md
chrysle: I've moved the conversion onto another branch. But I'd wait until this is merged and submit the PR including the newly converted Markdown README.

pyproject.toml
webknjaz: Marking the thread as resolved.

docs/index.rst
```diff
@@ -7,7 +7,8 @@
 Welcome to pip-tools' documentation!
 ====================================
 
-.. include:: ../README.rst
+.. include:: ../README.md
+   :parser: markdown
```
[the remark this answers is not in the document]
chrysle: Apparently not... I tried that out, but it didn't work even when I adjusted the [configuration](https://www.sphinx-doc.org/en/master/usage/markdown.html#configuration).

> Also, did you compare the rendering of the current RTD front page vs. this PR?

They differ a little, the size of the captions is not exactly the same, and the new version has a table of contents for each page. But I think it's all right (you can also check the build result yourself one last time).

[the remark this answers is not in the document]
webknjaz: Ah, maybe that setting doesn't apply to includes..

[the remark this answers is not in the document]
webknjaz: As for the title sizes, I think MyST has an option to offset that.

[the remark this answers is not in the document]
chrysle: There are some more problems, as I realized while scanning through the RTD build.

Regarding the badges and picture at the beginning of the README, the reference-style links don't render correctly.
Furthermore, syntax and console highlighting seem broken in some places. Also, the version table a the end is broken.

I do not know why all this renders correctly locally, but not here.

README.md
```diff
@@ -0,0 +1,583 @@
+[![jazzband][jazzband-image]][jazzband]
```
webknjaz, line 1: I'm almost sure this should be something like
```suggestion
[![jazzband-image]][jazzband]
```

docs/index.rst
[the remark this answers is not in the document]
webknjaz: Added a comment on the relevant line.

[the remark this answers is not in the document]
webknjaz: > Also, the version table at the end is broken.

That would have to be fixed before merging. Tables can be tricky (both in MD and RST). I'd start by copying a table from elsewhere and adopting it. You could also try removing some lines to see if that makes it work — this would mean that only some lines might be broken and not the entire thing.

README.md
```diff
@@ -0,0 +1,583 @@
...
+
+### Versions and compatibility
+
+The table below summarizes the latest `pip-tools` versions with the required
+`pip` and Python versions. Generally, `pip-tools` supports the same Python
+versions as the required `pip` versions.
+
+| pip-tools      | pip            | Python         |
+| -------------- | -------------- | -------------- |
+| 4.5.\*         | 8.1.3 - 20.0.2 | 2.7, 3.5 - 3.8 |
```
webknjaz, line 561: Could the `*` be breaking this table?

README.md
```diff
@@ -0,0 +1,583 @@
...
+even when you've pinned them. You do pin them, right? (In building your Python application and its dependencies for production, you want to make sure that your builds are predictable and deterministic.)
+
+[![pip-tools overview for phase II][pip-tools-overview]][pip-tools-overview]
+
+## Installation
+
+Similar to `pip`, `pip-tools` must be installed in each of your project's
+[virtual environments](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments):
+
+```console
```
webknjaz, line 20: Oh, I see that this isn't highlighted at all.. I wonder if it's a MyST bug and should be reported to @chrisjsewell. Also, could it be because of the inclusion of MD-into-RST and not MD-into-MD? I wonder if we converted `docs/index.rst` to `docs/index.md` first, would it break the same way (with `README.rst` remaining the same)?

README.md
chrysle: But what's about the syntax for the [alt text](https://www.w3schools.io/file/markdown-images/#how-to-add-an-image-with-a-link)?

README.md
chrysle: But it's escaped?

README.md
chrysle: Yes, #1874 seems broken....

README.md
webknjaz: Yes, though I'm not sure Markdown has a concept of escaping like RST. So it might be parser-dependent. GH uses GFM and Sphinx has MyST integrated. This is why more experimentation may be needed to make sure it works universally.

README.md
webknjaz: The alt text is what's in the first pair of brackets, the second one is the link to the thing. If you want different alt text, you could change the link definition, I think. Also, Markdown has various extensions that may or may not be supported by different MD parsers. I'm pretty sure MyST is mostly bare-bones, except for its own addition for Sphinx/RST compat...

ssbarnea (changes requested):
It looks almost ready but that tables is borked. I would find acceptable to get rid of the table and replace it with something like a simple list. Another alternative would be to make the table plain text and escape it as a code block.

In general I am for avoiding tables as they do not render well on mobile devices. When possible to avoid tables, I usually do.

chrysle:
Everything looks good now!

webknjaz (approved):

chrysle:
@ssbarnea Could you take another look?

ssbarnea (approved):

PR #1872, opened 2023-05-12
Add docs preview env for `tox`

### LINKED ISSUE #1869: Preview docs?
Is it possible to build and preview the docs using `tox` or another tool?

chrysle:
Closes #1869 

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -105,11 +105,12 @@ commands =
   -git fetch --tags
 
   # Build the html docs with sphinx-autobuild:
-  sphinx-autobuild \
+  {envpython} -m sphinx_autobuild \
     -j auto \
     -b html \
     -n \
     -W \
+    -d "{temp_dir}/.doctrees" \
     . \
     --watch ../README.rst \
     --watch ../CHANGELOG.md \
--- a/tox.ini
+++ b/tox.ini
@@ -92,6 +92,41 @@ skip_install = true
 allowlist_externals =
   git
 
+
+[testenv:preview]
+description = preview the docs
+deps =
+  sphinx-autobuild
+  -r{toxinidir}/docs/requirements.txt
+  # FIXME: re-enable the "-r" + "-c" paradigm once the pip bug is fixed.
+  # Ref: https://github.com/pypa/pip/issues/9243
+  # -r{toxinidir}/docs/requirements.in
+  # -c{toxinidir}/docs/requirements.txt
+commands_pre =
+commands =
+  # Retrieve possibly missing commits:
+  -git fetch --unshallow
+  -git fetch --tags
+
+  # Build the html docs with sphinx-autobuild:
+  sphinx-autobuild \
+    -j auto \
+    -b html \
+    -n \
+    -W \
+    . \
+    --watch .. \
+    "{envdir}/docs_out"
+
+changedir = {toxinidir}/docs
+isolated_build = true
+passenv =
+  SSH_AUTH_SOCK
+skip_install = true
+allowlist_externals =
+  git
+
+
 [testenv:linkcheck-docs]
 description = check links in the documentation
 deps =
--- a/tox.ini
+++ b/tox.ini
@@ -93,7 +93,7 @@ allowlist_externals =
   git
 
 
-[testenv:preview]
+[testenv:preview-docs]
 description = preview the docs
 deps =
   sphinx-autobuild
--- a/tox.ini
+++ b/tox.ini
@@ -115,13 +115,11 @@ commands =
     --watch ../CHANGELOG.md \
     "{envdir}/docs_out"
 
-changedir = {toxinidir}/docs
-isolated_build = true
-passenv =
-  SSH_AUTH_SOCK
-skip_install = true
-allowlist_externals =
-  git
+changedir = {[testenv:build-docs]changedir}
+isolated_build = {[testenv:build-docs]isolated_build}
+passenv = {[testenv:build-docs]passenv}
+skip_install = {[testenv:build-docs]skip_install}
+allowlist_externals = {[testenv:build-docs]allowlist_externals}
 
 
 [testenv:linkcheck-docs]
--- a/tox.ini
+++ b/tox.ini
@@ -115,7 +115,8 @@ commands =
     -n \
     -W \
     . \
-    --watch .. \
+    --watch ../README.rst \
+    --watch ../CHANGELOG.md \
     "{envdir}/docs_out"
 
 changedir = {toxinidir}/docs


webknjaz:
@chrysle if you're using some code editor with plugins for Git commits, I highly recommend integrating `gitlint` which help you watch out for the best practices.

chrysle:
Thanks for the tip!

I suppose you refer to me writing my commit messages in past tense. I wonder why I'm doing that all the time myself!

I'm going to rename that commit if that's common practice (also, I can fix that `venv -> env` typo) .

webknjaz:
Well, venv and env can be used interchangeably in some contexts. People also use toxenv (because it resembles the env var tox relies on, even though the config sections are actually testenvs 🤷‍♂️). So I wouldn't say it's a typo. But it may be a misconception.

Tox does create virtualenvs internally, but what you declare is a toxenv — a more high-level concert.

As for the commit title, yes, the non-use of imperative mood was what drew my attention. If you're interested in learning more around Git and practices surrounding its usage, I've got a few collections of links on the topic:
- https://gist.github.com/webknjaz/a5a4fb374b7579de827e6bedb93a5220
- https://gist.github.com/webknjaz/cb7d7bf62c3dda4b1342d639d0e78d79
- https://gist.github.com/webknjaz/a7362787a80067af8621a85a71746ca1
- https://gist.github.com/webknjaz/2aa9c7a43b51c1385260ff87e0cbb9e3

I'll also have some comments on the patch contents later, once I get to the laptop — it's not as convenient from the mobile app..

tox.ini
```diff
@@ -92,6 +92,41 @@ skip_install = true
...
+  -git fetch --tags
+
+  # Build the html docs with sphinx-autobuild:
+  sphinx-autobuild \
+    -j auto \
+    -b html \
+    -n \
+    -W \
+    . \
+    --watch .. \
```
chrysle, line 118: I had to watch the whole root directory of the project because it contains the CHANGELOG and README. 

Can I create symlinks of those files into `docs/`? Then we can remove this line, which would probably be better.

webknjaz: No, you don't need that. @pradyunsg shown me recently that you can add an extra CLI option to watch more files. Just need to google that and make use of it...

tox.ini
```diff
@@ -92,6 +92,41 @@ skip_install = true
...
+
+[testenv:preview]
+description = preview the docs
+deps =
+  sphinx-autobuild
+  -r{toxinidir}/docs/requirements.txt
+  # FIXME: re-enable the "-r" + "-c" paradigm once the pip bug is fixed.
+  # Ref: https://github.com/pypa/pip/issues/9243
+  # -r{toxinidir}/docs/requirements.in
+  # -c{toxinidir}/docs/requirements.txt
```
webknjaz, line 104: Try reusing the values from the previous section like this:
```suggestion
  {[testenv:build-docs]deps}
```

tox.ini
chrysle: Can you share your wisdom?

webknjaz: */me just googles that*

webknjaz: There's nothing clear in the README, but the CLI parser spec says `--watch` is incremental, so we could probably just use `--watch docs/ --watch CHANGELOG.md --watch README.rst`. Though, it's unclear if it would stomach a file path in place of a directory path..

webknjaz: I think that should work. I checked that https://github.com/executablebooks/sphinx-autobuild/blob/ea4d04f/src/sphinx_autobuild/cli.py#L187-L191 calls `Server.watch()` on all the corresponding CLI options passed. And https://github.com/lepture/python-livereload/blob/d042fd3/livereload/server.py#L225C26-L226 implies that the values could be file/dir paths or even path wildcards.

`sphinx-autobuild` other possible values which could be a chance to contribute it.

chrysle: > There's nothing clear in the README, but the CLI parser spec says `--watch` is incremental, so we could probably just use `--watch docs/ --watch CHANGELOG.md --watch README.rst`. Though, it's unclear if it would stomach a file path in place of a directory path..

That works! Hm, I could have sworn I tested that and received some traceback about the given path not being a directory.

webknjaz: Maybe you tried that under a different chdir?

tox.ini
```diff
@@ -92,6 +92,38 @@ skip_install = true
...
+  -git fetch --unshallow
+  -git fetch --tags
+
+  # Build the html docs with sphinx-autobuild:
+  sphinx-autobuild \
+    -j auto \
+    -b html \
+    -n \
+    -W \
+    . \
```
webknjaz, line 113: Can we set a `-d` too?

tox.ini
```diff
@@ -92,6 +92,38 @@ skip_install = true
...
+    --watch ../CHANGELOG.md \
+    "{envdir}/docs_out"
+
+changedir = {toxinidir}/docs
+isolated_build = true
+passenv =
+  SSH_AUTH_SOCK
+skip_install = true
+allowlist_externals =
+  git
```
webknjaz, line 124: Let's reuse all these values from the other env, thinking ahead about the maintainability..

tox.ini
```diff
@@ -92,6 +92,38 @@ skip_install = true
 allowlist_externals =
   git
 
+
+[testenv:preview]
```
webknjaz, line 96: Other envs also have this suffix:
```suggestion
[testenv:preview-docs]
```
Let's have it here too, for consistency.

webknjaz (commented):
Do we have any references to these envs in the contributing doc? If yes, it could be worth adding this one too.

tox.ini
chrysle: As far as I've understood, [-d](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-d) only points to the directory where the parsed source files are stored.

Is this of any relevance, and if so, where should it point to?

chrysle:
@webknjaz Do you have any further suggestions?

tox.ini
webknjaz: Yes, it points to the doctree cache. Use `-d "{temp_dir}/.doctrees"` just like others. Ideally, the sphinx commands should make use of mostly the same sets of arguments where possible.

tox.ini
```diff
@@ -92,6 +92,36 @@ skip_install = true
...
+  sphinx-autobuild
+  {[testenv:build-docs]deps}
+commands_pre =
+commands =
+  # Retrieve possibly missing commits:
+  -git fetch --unshallow
+  -git fetch --tags
+
+  # Build the html docs with sphinx-autobuild:
+  sphinx-autobuild \
```
webknjaz, line 108: Can this be invoked via the runpy interface? It'd be nice to align it with `{envpython} -m` that the rest of the places use.

chrysle: > Can this be invoked via the runpy interface?

I tried that first, but `runpy` can't find `sphinx-autobuild`. I think that's because it's not installed via _requirements.txt_. Can I add it there?

webknjaz: No, that's unrelated. `python -m` only accepts importables. Importables can't have dashes — they must be valid Python identifiers. What you wrote is a script name, the filename on disk that is created as an entry point.

webknjaz: Looking into https://github.com/executablebooks/sphinx-autobuild/tree/main/src, I see that `sphinx_autobuild` is the importable we're looking for — this is what's installed into `site-packages/`. We can also confirm that it contains a `__main__.py` which makes it an entry point for runpy: https://github.com/executablebooks/sphinx-autobuild/tree/main/src/sphinx_autobuild.
If it didn't contain, `__main__.py`, the invocation would be `python -m sphinx_autobuild.<module_name>` but with that magic module, the invocation is simpler — it's `python -m sphinx_autobuild`.

webknjaz:
```suggestion
  {envpython} -m sphinx_autobuild \
```

chrysle: Thank you for explaining, researching and taking time to review in general!

webknjaz (approved):
@chrysle why didn't you copy over the `--color` thing? Add it, if it's supported. Otherwise, merge the PR as is.

chrysle:
> @chrysle why didn't you copy over the `--color` thing?

They've added that to `sphinx-build`, but it isn't yet implemented in `sphinx-autobuild`, which therefore defaults to coloring output.

Before merging, I'd like to ask if I need to add changelog and assign a milestone? I read about that in the contributing docs.

PR #1885, opened 2023-06-14
Sync direct references with hashes

### LINKED ISSUE #1884: Direct references should use direct_url hashes as key for sync
`pip-sync` currently assumes that the URL has an egg fragment with a version specifier [1], but that behavior was never in the specification, and so can not be relied upon. Moreover, `pip-compile` actually actively removes egg fragments to create a direct reference [2]. Direct references rely on hashes rather than versions as the specifier and there is no way to specify a version for them anyway. We should be using `direct_url.hash` attribute of the installed dist [3] and comparing that with the hash fragment of the direct reference in `requirements.txt`. The current distribution type that is being used in `sync.diff` is a narrower type [4] that does not have this information, but a small refactor can make the information available.

[1] https://github.com/jazzband/pip-tools/blob/6d06be8dbd88298a9e62df1415a8da28a14eadc0/piptools/sync.py#L122
[2] https://github.com/jazzband/pip-tools/blob/6d06be8dbd88298a9e62df1415a8da28a14eadc0/piptools/utils.py#L144
[3] https://github.com/pypa/pip/blob/8a1eea4aaedb1fb1c6b4c652cd0c43502f05ff37/src/pip/_internal/models/direct_url.py#L105
[4] https://github.com/jazzband/pip-tools/blob/6d06be8dbd88298a9e62df1415a8da28a14eadc0/piptools/scripts/sync.py#L306

#### Environment Versions

1. OS Type: Debian 12
1. Python version: `$ python -V`: `Python 3.11.4`
1. pip version: `$ pip --version`: `pip 23.1.2 ...`
1. pip-tools version: `$ pip-compile --version`: `pip-compile, version 6.13.0`

#### Steps to replicate

1. Create a requirements.txt file:
```
urllib3 @ https://github.com/urllib3/urllib3/archive/refs/tags/2.0.3.zip#sha1=bc7f779fbdba2befc592eb35a6353cff0109d622
```
2. In a clean virtual environment, `pip-sync`.
3. Check if the installed dist and the specified requirement match: `pip-sync -n`.

#### Expected result

Expected that the rerun of `pip-sync` will not try to re-install the req.

#### Actual result

`pip-sync` re-installs.

siddharthab:
When a direct reference is provided with hashes, use the direct reference string as the unique key for a requirement. The key can also be used for uninstalling the dist.

See the attached issue for more details on why we want to do this.

Fixes #1884.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -17,7 +17,7 @@ from pip._internal.utils.direct_url_helpers import (
     direct_url_from_link,
 )
 
-from ._compat.pip_compat import Distribution, dist_requires
+from ._compat import Distribution
 from .exceptions import IncompatibleRequirements
 from .logging import log
 from .utils import (
@@ -66,7 +66,7 @@ def dependency_tree(
 
         dependencies.add(key)
 
-        for dep_specifier in dist_requires(v):
+        for dep_specifier in v.requires:
             dep_name = key_from_req(dep_specifier)
             if dep_name in installed_keys:
                 dep = installed_keys[dep_name]
@@ -131,12 +131,6 @@ def diff_key_from_ireq(ireq: InstallRequirement) -> str:
     if the contents at the URL have changed but the version has not.
     """
     if is_url_requirement(ireq):
-        if (
-            ireq.req
-            and (getattr(ireq.req, "key", None) or getattr(ireq.req, "name", None))
-            and ireq.specifier
-        ):
-            return key_from_ireq(ireq)
         if getattr(ireq.req, "name", None) and ireq.link.has_hash:
             return str(
                 direct_url_as_pep440_direct_reference(
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -60,7 +60,7 @@ def dependency_tree(
 
     while queue:
         v = queue.popleft()
-        key = key_from_req(v)
+        key = v.key
         if key in dependencies:
             continue
 
@@ -86,7 +86,7 @@ def get_dists_to_ignore(installed: Iterable[Distribution]) -> list[str]:
     locally, click should also be installed/uninstalled depending on the given
     requirements.
     """
-    installed_keys = {key_from_req(r): r for r in installed}
+    installed_keys = {r.key: r for r in installed}
     return list(
         flat_map(lambda req: dependency_tree(installed_keys, req), PACKAGES_TO_IGNORE)
     )
@@ -144,10 +144,9 @@ def diff_key_from_ireq(ireq: InstallRequirement) -> str:
 
 def diff_key_from_req(req: Distribution) -> str:
     """Get a unique key for the requirement."""
-    key = key_from_req(req)
+    key = req.key
     if (
-        hasattr(req, "direct_url")
-        and req.direct_url
+        req.direct_url
         and type(req.direct_url.info) == ArchiveInfo
         and req.direct_url.info.hash
     ):
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -147,7 +147,7 @@ def diff_key_from_req(req: Distribution) -> str:
     key = req.key
     if (
         req.direct_url
-        and type(req.direct_url.info) == ArchiveInfo
+        and isinstance(req.direct_url.info, ArchiveInfo)
         and req.direct_url.info.hash
     ):
         key = direct_url_as_pep440_direct_reference(req.direct_url, key)
--- a/piptools/sync.py
+++ b/piptools/sync.py
@@ -9,8 +9,13 @@ from typing import Deque, Iterable, Mapping, ValuesView
 
 import click
 from pip._internal.commands.freeze import DEV_PKGS
+from pip._internal.models.direct_url import ArchiveInfo
 from pip._internal.req import InstallRequirement
 from pip._internal.utils.compat import stdlib_pkgs
+from pip._internal.utils.direct_url_helpers import (
+    direct_url_as_pep440_direct_reference,
+    direct_url_from_link,
+)
 
 from ._compat.pip_compat import Distribution, dist_requires
 from .exceptions import IncompatibleRequirements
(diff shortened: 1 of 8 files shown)


piptools/utils.py
```diff
@@ -72,7 +72,10 @@ def key_from_req(req: InstallRequirement | Distribution | Requirement) -> str:
         key = req.key
     else:
         # from packaging, such as install requirements from requirements.txt
-        key = req.name
+        if hasattr(req, "canonical_name"):
```
chrysle, line 75: Where should that attribute come from?

siddharthab: This comes from pip's internal classes derived from `pip._internal.metadata.BaseDistribution`. Previously, `importlib.metadata.Distribution` or its pkg_resources equivalent was being used. Now added a comment.

With this change, we do start relying on another internal implementation detail of pip, but this package seems to be coupled with the internals enough already, so I thought it was OK. The alternative would be to roll out something like `pip._internal.metadata.BaseDistribution.direct_url` ourselves.

siddharthab: I can also have the exact behavior as before by using `._dist.name` here. Let me know.

tests/conftest.py
```diff
@@ -153,6 +152,22 @@ def requires(self):
...
+    def __init__(self, line, deps=None):
+        req = Requirement.parse(line)
+        if "==" in line:
+            version = line.split("==")[1]
+        else:
+            version = "0+unknown"
+
+        self._dist = InnerFakeInstalledDistribution(req, version, deps)
+        self.key = self._dist.key
+        self.canonical_name = self._dist.req.project_name
```
chrysle, line 165: Is this still needed, then?

piptools/utils.py
chrysle: > I don't think this change should be backwards incompatible in any way now.

Well, the behaviour of sync watching the `#egg=` fragments has been dropped...

chrysle: I think that'll be alright.

tests/conftest.py
```diff
@@ -153,6 +154,20 @@ def requires(self):
             return self.dep_strs
 
 
+class FakeInstalledDistribution:
```
chrysle, line 157: Could you explain for what reasons exactly you derived this from the original `FakeInstalledDistribution`?

siddharthab: I added some more comments, but I am not sure I understood exactly what you meant.

If you want, I can clean up this pattern by using an adapter class in the compatibility layer, which will have the attributes name, version, key, requires and direct_url. Then the typing will be much cleaner in both the code and the tests.

chrysle: That would be great, thanks.

I meant to ask why you created inner and outer `FakeInstalledDistribution`.

siddharthab: Alright, I made an adapter class. I think the new pattern is much cleaner, but let me know if you would like me to revert to the old one.

siddharthab: Tests are failing for Python 3.7 because `importlib-metadata` is missing from the test environment. Should I add it to setup.cfg. Not sure how you would like me to proceed here.

webknjaz: It can be added with an env marker so it's installed conditionally.

siddharthab: Alright. I did something. We will know for sure when the checks run.

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
   # direct dependencies
   "build",
   "click >= 8",
+  "importlib-metadata == 1.4; python_version < '3.8'",
```
webknjaz, line 42: Why this specific version? Normally, you shouldn't restrict library runtime deps.

siddharthab: This was the version in Python 3.8, the first time it was included as stdlib. Because it is being used with Python 3.7, I thought we should keep it as close as possible. Let me know if you want me to remove the limit.

siddharthab: I removed the dependency on importlib now. We just don't get type hints in a very tiny part of the code, but that's OK.

siddharthab:
@webknjaz happy to rework this in whichever form you would like. Please let me know.

webknjaz:
I'm on vacation this week. Ping me the next one.

piptools/sync.py
```diff
@@ -120,22 +125,36 @@ def diff_key_from_ireq(ireq: InstallRequirement) -> str:
...
         return str(ireq.link)
     return key_from_ireq(ireq)
 
 
+def diff_key_from_req(req: Distribution) -> str:
+    """Get a unique key for the requirement."""
+    key = req.key
+    if (
+        req.direct_url
+        and type(req.direct_url.info) == ArchiveInfo
```
atugushev, line 150:
```suggestion
        and isinstance(req.direct_url.info, ArchiveInfo)
```

Generally `isinstance()` provides a better alternative to `type()` in many scenarios as it takes class inheritance into account.

tests/test_sync.py
```diff
@@ -263,27 +263,41 @@ def test_diff_with_editable(fake_dist, from_editable):
...
     assert to_install == set(reqs)
     assert to_uninstall == {"example"}
 
 
+def test_diff_with_unequal_url_hash(fake_dist, from_line):
+    # if URL hashes mismatch, assume the contents have
+    # changed and reinstall
+    line = "example@file:///example.zip#"
+    installed = [fake_dist(line + "#sha1=abc")]
+    reqs = [from_line(line + "#sha1=def")]
```
atugushev, line 294:
```suggestion
    installed = [fake_dist(line + "sha1=abc")]
    reqs = [from_line(line + "sha1=def")]
```

Fixes double hash.

tests/conftest.py
```diff
@@ -162,7 +149,7 @@ def pytest_collection_modifyitems(config, items):
 
 @pytest.fixture
 def fake_dist():
-    return FakeInstalledDistribution
+    return parse_fake_distribution
```
atugushev, line 152:
```suggestion
    def _fake_dist(line, deps=[]):
        req = Requirement.parse(line)
        key = req.key
        if "==" in line:
            version = line.split("==")[1]
        else:
            version = "0+unknown"
        requires = [Requirement.parse(d) for d in deps]
        direct_url = None
        if req.url:
            direct_url = direct_url_from_link(Link(req.url))
        return Distribution(key, version, requires, direct_url)
    return _fake_dist
```

Thanks for refactor this. Wouldn't it makes sense to create a [fixture factory](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#factories-as-fixtures) since the `parse_fake_distribution` function only used by `fake_dist` fixture?

atugushev (approved):
LGTM 👍

PR #1888, opened 2023-06-16
Specify build formats in .readthedocs.yaml

### LINKED ISSUE #1887: [RTD] Drop epub and latex from docs builds
$sbj fails likely because of `formats: all` in the RTD config we have: https://readthedocs.org/projects/pip-tools/builds/21037894/. This is not visible in PR builds, since they only build HTML. We should probably change it to stop building epub and latex. @chrysle are you up for a PR?

chrysle:
Fixes #1887 

@webknjaz Another solution would be the approach in IQSS/dataverse#3542, if you want to keep building EPUB.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,7 +8,9 @@ sphinx:
   configuration: docs/conf.py
   fail_on_warning: true
 
-formats: all
+formats:
+  - pdf
+  - htmlzip
 
 build:
   os: ubuntu-22.04


webknjaz:
> if you want to keep building EPUB.

There's also latex. But I don't think we ever cared about targeting those. So I tend to just build html everywhere..

webknjaz (approved):

chrysle:
> There's also latex.

Right, that's strangely not listed in the [relevant section](https://docs.readthedocs.io/en/stable/config-file/v2.html#formats).

webknjaz:
> Right, that's strangely not listed in the relevant section.

That's a step in making a pdf.

Issue #1853, opened 2023-04-26
pip-compile drops annotations when using `--resolver=backtracking`


jhines-ocient:
#### Summary

pip-compile drops annotations when using `--resolver=backtracking`. Source `requirements.in` files are missing under the relevant `via` section. Which source file is missing is dependent on the order of the arguments.

Running the below tests with more files, it appears that only the first and last files are included in the annotations.

#### Environment Versions

1. OS Type: Linux 4.15.0-182-generic Ubuntu x86_64 GNU/Linux
1. Python version: `$ python -V`: Python 3.8.16
1. pip version: `$ pip --version`: pip 22.0.3 from python 3.8
1. pip-tools version: `$ pip-compile --version`: pip-compile, version 6.10.0

#### Steps to replicate

1. Create 3 requirements.in files with the same single requirement (`a.in`, `b.in`, and `c.in`). I used `requests`.
```
$ cat a.in
requests
$ cat b.in
requests
$ cat c.in
requests
```
2. Run `pip-compile --resolver=legacy --output-file=legacy-requirements.txt a.in b.in c.in` to generate the expected result
3. Run `pip-compile --resolver=backtracking --output-file=backtracking-requirements.txt a.in b.in c.in` to generate the first actual result
4. Run `pip-compile --resolver=backtracking --output-file=backtracking-reorder-requirements.txt a.in c.in b.in` to generate the second actual result

#### Expected result

Generated by Step 2, specifically the via section of requests. I expect that the annotations section would always contain every source for the relevant requirement. 

```
#
# This file is autogenerated by pip-compile with python 3.8
# To update, run:
#
#    pip-compile --output-file=legacy-requirements.txt a.in b.in c.in
#
certifi==2022.12.7
    # via requests
charset-normalizer==3.1.0
    # via requests
idna==3.4
    # via requests
requests==2.28.2
    # via
    #   -r a.in
    #   -r b.in
    #   -r c.in
urllib3==1.26.15
    # via requests
```

#### First Actual result

Generated by Step 3, note `b.in` is missing from the via section of requests

```
#
# This file is autogenerated by pip-compile with python 3.8
# To update, run:
#
#    pip-compile --output-file=backtracking-requirements.txt --resolver=backtracking a.in b.in c.in
#
certifi==2022.12.7
    # via requests
charset-normalizer==3.1.0
    # via requests
idna==3.4
    # via requests
requests==2.28.2
    # via
    #   -r a.in
    #   -r c.in
[...]

PR #1890, opened 2023-06-26
Fix missing `via`s when more than two input files are used

lpulley:
Fixes https://github.com/jazzband/pip-tools/issues/1853.

The gist of the problem is that

```py
self._constraints_map = {key_from_ireq(ireq): ireq for ireq in constraints}
```

only kept the last `ireq` for each key, rather than combining them (and thus their sources).

There are two changes here:

- `self._constraints_map` is no longer constructed directly from `constraints`; rather, the items in `constraints` are first grouped by key into sets, then those sets are "collapsed" via `combine_install_requirements()` into single `InstallRequirement`s.
- `pinned_ireq._source_ireqs` is no longer always a singleton list of the value from `self._constraints_map`; rather, if the constraint has `_source_ireqs`, then that list is used as `pinned_ireq._source_ireqs`. This effectively "unwraps" `_source_ireqs` (once) if possible.
  - A more sophisticated solution might be to just have `writer.py` recurse on `_source_ireqs` [when constructing `required_by`](https://github.com/lpulley/pip-tools/blob/1479ad95ca71b131e657e64b49862e58ceb14d59/piptools/writer.py#L294C50-L294C50), but that touches more code paths than are necessary to fix this, so I opted for this solution instead. Such recursion would enable combined `InstallRequirement`s to be combined further without issue, but that doesn't seem to be a requirement right now.

The new `test_many_inputs_includes_all_annotations` is parametrized for 2, 3, and 10 inputs; the 2 case was not broken but should still be tested.

---

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -1916,11 +1916,9 @@ def test_many_inputs_includes_all_annotations(pip_conf, runner, num_inputs):
     annotation.
     See: https://github.com/jazzband/pip-tools/issues/1853
     """
-    in_files = [f"requirements{n:02d}.in" for n in range(1, num_inputs + 1)]
-
+    in_files = [tmp_path / f"requirements{n:02d}.in" for n in range(num_inputs)]
     for in_file in in_files:
-        with open(in_file, "w") as req_in:
-            req_in.write("small-fake-a==0.1\n")
+        req_in.write_text("small-fake-a==0.1\n")
 
     out = runner.invoke(
         cli,
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -1908,6 +1908,7 @@ def test_upgrade_package_doesnt_remove_annotation(pip_conf, runner):
             """
         )
 
+
 @pytest.mark.parametrize(("num_inputs"), (2, 3, 10))
 def test_many_inputs_includes_all_annotations(pip_conf, runner, num_inputs):
     """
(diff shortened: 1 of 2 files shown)


daveboutcher (approved):

tests/test_cli_compile.py
```diff
@@ -1909,6 +1909,44 @@ def test_upgrade_package_doesnt_remove_annotation(pip_conf, runner):
...
+    """
+    Tests that an entry required by multiple input files is attributed to all of them in the
+    annotation.
+    See: https://github.com/jazzband/pip-tools/issues/1853
+    """
+    in_files = [f"requirements{n:02d}.in" for n in range(1, num_inputs + 1)]
+
+    for in_file in in_files:
+        with open(in_file, "w") as req_in:
+            req_in.write("small-fake-a==0.1\n")
```
atugushev, line 1923: This would create requirement files in project dir. I'd suggest using the `tmp_path` fixture:

```suggestion
    in_files = [tmp_path  / f"requirements{n:02d}.in" for n in range(num_inputs)]
    for in_file in in_files:
        req_in.write_text("small-fake-a==0.1\n")
```

atugushev (approved):
Otherwise, great work! Thanks for the fix 🎉

tests/test_cli_compile.py
lpulley: Aha! So there is a `Path` floating around for this. Excellent.

lpulley:
@atugushev I see this just missed inclusion in 6.14.0 -- any chance of a 6.14.1 to release this sooner rather than later?

And thanks for your review 🙂

atugushev:
@lpulley unfortunately no. The next release would be 7.0.0.

PR #1894, opened 2023-06-29
Downgrade log level for configuration defaults message

atugushev:
Resolves https://github.com/jazzband/pip-tools/pull/1863#issuecomment-1611728125

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -410,7 +410,7 @@ def cli(
         )
 
     if config:
-        log.info(f"Using pip-tools configuration defaults found in '{config !s}'.")
+        log.debug(f"Using pip-tools configuration defaults found in '{config !s}'.")
 
     if resolver_name == "legacy":
         log.warning(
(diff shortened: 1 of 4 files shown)


webknjaz:
I think there's a similar log message in `sync.py`.

webknjaz:
As for the failing PyPy, I saw it being temporarily disabled in the attrs' CI today, we should probably do the same.

atugushev:
> As for the failing PyPy, I saw it being temporarily disabled in the attrs' CI today, we should probably do the same.

@webknjaz disabled pypy and updated `check` job.

webknjaz:
Instead of dropping the dependency, just add it to allowed skips or even allowed failures. It'd be a one-liner instead of coordinated changes in several places..

webknjaz:
Also, it's best to make it a separate PR so it's easy to revert with a single click.

atugushev:
> Instead of dropping the dependency, just add it to allowed skips or even allowed failures. It'd be a one-liner instead of coordinated changes in several places..

@webknjaz could you please provide an example or even better push the requested change?

webknjaz:
Here you go https://github.com/jazzband/pip-tools/pull/1895

chrysle (approved):

PR #1879, opened 2023-06-12
Drop support for Python 3.7

chrysle:
Drop support for Python 3.7 which is EOL in the end of June.

##### Contributor checklist

- [ ] Provided the tests for the changes.
- [x] Assure PR title is short, clear, and good to be included in the user-oriented changelog

##### Maintainer checklist

- [x] Assure one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #1895, opened 2023-06-30
🔧 Ignore PyPy 3.7 test outcome in CI

webknjaz:
This is supposed to be temporary until said interpreter is fixed or upgraded to PyPy 3.8 (via https://github.com/jazzband/pip-tools/pull/1879) as `pip-tools` is going to drop support for Python 3.7 in the upcoming release.


chrysle (approved):

chrysle:
Great tool, your `alls-green`!

PR #1896, opened 2023-07-01
Add `--no-config` option

atugushev:
Addresses https://github.com/jazzband/pip-tools/pull/1863#issuecomment-1612960012.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -2972,4 +2972,12 @@ def test_config_option(pip_conf, runner, tmp_path, make_config_file):
     out = runner.invoke(cli, [req_in.as_posix(), "--config", config_file.as_posix()])
 
     assert out.exit_code == 0
-    assert "Dry-run, so nothing updated" in out.stderr
+    dry_run_message = "Dry-run, so nothing updated"
+    assert dry_run_message in out.stderr
+
+    out = runner.invoke(
+        cli, [req_in.as_posix(), "--no-config", "--config", config_file.as_posix()]
+    )
+
+    assert out.exit_code == 0
+    assert dry_run_message not in out.stderr
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -2972,12 +2972,18 @@ def test_config_option(pip_conf, runner, tmp_path, make_config_file):
     out = runner.invoke(cli, [req_in.as_posix(), "--config", config_file.as_posix()])
 
     assert out.exit_code == 0
-    dry_run_message = "Dry-run, so nothing updated"
-    assert dry_run_message in out.stderr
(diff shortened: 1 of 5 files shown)


tests/test_cli_sync.py
```diff
@@ -384,4 +384,10 @@ def test_config_option(run, runner, make_config_file):
     out = runner.invoke(cli, ["--config", config_file.as_posix()])
 
     assert out.exit_code == 1
-    assert "Would install:" in out.stdout
+    dry_run_message = "Would install:"
+    assert dry_run_message in out.stdout
+
+    out = runner.invoke(cli, ["--no-config", "--config", config_file.as_posix()])
```
webknjaz, line 390: This should be a separate test with its own name or at least a parametrized case..

atugushev: I prefer triangulation technique here instead.

webknjaz: I don't see how it's applicable in this case. You're clearly testing two separate behaviors that must appear as separate entries in the test report.
The only way of having two things reported from the same test function is to use `pytest-subtest`. But it still feels wrong since you're not checking two properties of the same behavior outcome.

webknjaz: Also, for property-based testing, we should really integrate Hypothesis..

atugushev: Here you go 9a474fbb07913dbbf3c14b276294bdf91d8b8f45

webknjaz: Thanks!

tests/test_cli_compile.py
```diff
@@ -2973,3 +2973,17 @@ def test_config_option(pip_conf, runner, tmp_path, make_config_file):
 
     assert out.exit_code == 0
     assert "Dry-run, so nothing updated" in out.stderr
+
+
+def test_no_config_option(pip_conf, runner, tmp_path, make_config_file):
```
webknjaz, line 2978:
```suggestion
def test_no_config_option_overrides_config_with_defaults(pip_conf, runner, tmp_path, make_config_file):
```
Would this be more accurate?

tests/test_cli_sync.py
```diff
@@ -385,3 +385,16 @@ def test_config_option(run, runner, make_config_file):
 
     assert out.exit_code == 1
     assert "Would install:" in out.stdout
+
+
+@mock.patch("piptools.sync.run")
+def test_no_config_option(run, runner, make_config_file):
```
webknjaz, line 391: How about
```suggestion
def test_no_config_option_overrides_config_with_defaults(run, runner, make_config_file):
```

webknjaz (commented):
Let's try making the test names reflect what it is we're testing...

webknjaz (approved):

PR #1897, opened 2023-07-01
Default to `--resolver=backtracking`

atugushev:
Since 7.0.0 is the next release, it's time to switch the default resolver as promised.
Refs: https://github.com/jazzband/pip-tools/issues/1659

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -282,7 +282,7 @@ def _determine_linesep(
     "--resolver",
     "resolver_name",
     type=click.Choice(("legacy", "backtracking")),
-    default="legacy",
+    default="backtracking",
     envvar="PIP_TOOLS_RESOLVER",
     help="Choose the dependency resolver.",
 )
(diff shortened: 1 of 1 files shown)


webknjaz:
Do we want the next version to be 7? I don't think dropping a Python version needs a major bump since `python_requires` already serves the role of making the installers omit incompatible versions and  it's not an API change at all, SemVer-wise.

Though if we were to flip the default, I'd expect it to be aligned with having the pip version that did this minimum supported. And such a change would indeed require a major version bump.

atugushev:
I believe we should change the default resolver. It's been a year since pip-tools released a new resolver, and since then we haven't received much feedback, so I guess it's about time.

Even though I don't have the stats, I presume that most users still use the legacy resolver, given the minimal feedback on the backtracking resolver. This will give users sufficient time to adapt to the changes before pip drops the legacy resolver, as well as give us time to fix bugs.

webknjaz (approved):
Approving but I'm concerned that this isn't covered by tests..

Issue #1654, opened 2022-07-17
Explore decoupling from `pip`'s internals, using the new `--dry-run` and `--report` flags


pradyunsg:
#### What's the problem this feature will solve?

Currently, pip-tools depends on the internals of pip which is not a supported mode of use of pip.

#### Describe the solution you'd like

See pypa/pip#53 -- pip 22.2 (upcoming release) is going to add two new flags, that'll make it possible to get the resolution result from pip without performing an installation.

#### Alternative Solutions

Continue the current dependence on the fickle internals of pip, that I keep evolving and breaking pip-tools. :)

#### Additional context

Uhm... the issues on this repository, every time there's a new pip release? ;)

sbidoul:
One thing that is not addressed with the current implementation of `--report` is the "multi platform" `--generate-hashes`, since `pip install --dry-run --report` only exposes the hashes of the artifacts it would)actually install, and not the hashes of other artifacts found on the index for the same version.

Yet, that should not stop us exploring further -- the current implementation should be sufficient for generating a single platform lock file, and I'm personally keen to better understand the needs and workflows around multi platform lock files, and if it is something that pip could help with, at least in principle.

graingert:
I think this is a duplicate of https://github.com/jazzband/pip-tools/issues/1526

Rather than add interfaces for pip-tools, pip should vendor this project and support both the constraints.txt format and the PEP format

AndydeCleyre:
Here's a primitive proof of concept using `wheezy.template`, which doesn't account for a *lot*, but demonstrates some potential of a template-oriented approach:

`requirements.in`:
```ini
requests
httpx
```
`requirements.txt.wz`:
```python
@require(install)
@for pkg in sorted(install, key=lambda p: p['metadata']['name']):
@(
  required_by = sorted(
      needy_pkg['metadata']['name']
      for needy_pkg in install
      if pkg['metadata']['name'] in (
          dist.split()[0].split('[')[0]
          for dist in needy_pkg['metadata'].get('requires_dist', [])
      )
  )
  if pkg['requested']:
      required_by.insert(0, "-r requirements.in")

  annotation = f"# via {', '.join(required_by)}"

  pin = f"{pkg['metadata']['name']}=={pkg['metadata']['version']}"

  spaces = ' ' * max(26 - len(pin), 2)
)\
@{pin}@{spaces}@{annotation}
@end
```

```console
$ pip install -U pip-tools pip wheezy.template
[...]

PR #1898, opened 2023-07-01
Consider python specific options via `--pip-args`

atugushev:
##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -11,6 +11,7 @@ from typing import Any, BinaryIO, ContextManager, Iterator, NamedTuple
 
 from click import progressbar
 from pip._internal.cache import WheelCache
+from pip._internal.cli.cmdoptions import make_target_python
 from pip._internal.commands import create_command
 from pip._internal.commands.install import InstallCommand
 from pip._internal.index.package_finder import PackageFinder
@@ -74,8 +75,9 @@ class PyPIRepository(BaseRepository):
 
         self._options: optparse.Values = options
         self._session = self.command._build_session(options)
+        target_python = make_target_python(options)
         self._finder = self.command._build_package_finder(
-            options=options, session=self.session
+            options=options, session=self.session, target_python=target_python
         )
 
         # Caches


atugushev:
This allows target specific python version:

```
$ echo six | pip-compile - -qo- --no-header --no-annotate
six==1.16.0

$ echo six | pip-compile - -qo- --no-header --no-annotate --pip-args='--python-version=3.1.0'
six==1.11.0
```

cc: @mitsuhiko

chrysle:
Looks good, but this will need a test.

chrysle:
@atugushev This would need merging for the feature worked on in #1998. Could you add a test?

atugushev:
@chrysle I'm inclined to close this, as is may not work as expected.

georgek:
@atugushev would love to know more about what you don't think would work as expected. I was looking to build upon this to enable the correct default behaviour of pip-tools when installed by pipx or as a system package etc.

atugushev:
My initial intent was to target a specific Python interpreter. I thought `--python-version` was the solution. However, it is only used for wheel and "Requires-Python" compatibility checks, as stated in the documentation:

```
$ pip install --help
...
  --python-version <python_version>
                              The Python interpreter version to use for wheel and "Requires-Python"
                              compatibility checks. Defaults to a version derived from the running
                              interpreter. The version can be specified using up to three dot-
                              separated integers (e.g. "3" for 3.0.0, "3.7" for 3.7.0, or "3.7.3").
                              A major-minor version can also be given as a string without dots (e.g.
                              "37" for 3.7.0).
```

I know there is a `--python <python>` option, however it runs pip in a subprocess which doesn't work well with the pip-tools resolver.

georgek:
It sounds like this approach might be ruled out for the purposes of #1998 then. But just for my own curiosity and perhaps to help others look for this:

What difference is there in practice between targeting a specific interpreter vs specifying `--python-version`?

What is the problem with pip-tools and `--python`? Could pip-tools be changed to work better with this option?

chrysle:
> What is the problem with pip-tools and `--python`? Could pip-tools be changed to work better with this option?

I think @atugushev refers to the problem that we too heavily rely on `pip`'s internals; it doesn't help us to pass the Python option to `pip`, as we create a command representation manually via the corresponding `InstallCommand` only to utilize it to parse and return the options to us, and then invoke the resolution process manually – which means our setting doesn't have any impact on the actual resolution result. In order to benefit from the option, we would instead need to run `pip` and parse its results, as done for `pip-sync`. There is an open issue for that (#1654) but actually decoupling would require huge changes; I can look into it, but won't promise anything, especially since activity here has been largely reduced by most maintainers and `uv` is likely to kick us totally out of business (no hard feelings).

PR #1899, opened 2023-07-04
👷 Deduplicate cron GHA through workflow reuse

webknjaz:
This patch melts most of the separate cron workflow into the main CI workflow which it then includes as reusable.
Such an approach allows for less duplication of the same CI steps while keeping the cron workflow as a separate entity.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,9 +6,14 @@ on:
     branches:
       - main
     tags:
-  schedule:
-    # Run everyday at 03:53 UTC
-    - cron: 53 3 * * *
+  workflow_call:
+    inputs:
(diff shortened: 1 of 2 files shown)


chrysle (approved):
LGTM, but quite complex.

webknjaz:
It is a bit complex due to the slight differences. Hopefully, we'll rethink those later. I didn't know that cron tested against the pip main branch but there was also another cron in the main workflow that added the usual latest+previous. Plus, it's not really clear why we skip coverage in one of the cron runs. Maybe we could get rid of that conditional as well. That's if someone wants a follow up improvement..

webknjaz:
I think I managed to simple it a bit.

Issue #1902, opened 2023-07-06
annotation --config is added even there is no configuration


q0w:
pip-tools 6.14.0 adds --config annotation even there is no configuration.
#### Environment Versions

1. OS Type: Linux
1. Python version: `$ python -V`: 3.11.3
1. pip version: `$ pip --version`:  22.3.1
1. pip-tools version: `$ pip-compile --version`: 6.14.0

#### Steps to replicate
```
mkdir requirements
touch pyproject.toml
echo astpretty > requirements/test.in
pip-compile --allow-unsafe --generate-hashes --output-file=requirements/test.txt --resolver=backtracking --strip-extras requirements/test.in
```

#### Expected result
```
#
# This file is autogenerated by pip-compile with Python 3.11
# by the following command:
#
#    pip-compile --allow-unsafe --generate-hashes --output-file=requirements/test.txt --resolver=backtracking --strip-extras requirements/test.in
#
astpretty==3.0.0 \
    --hash=sha256:15bfd47593667169485a1fa7938b8de9445b11057d6f2b6e214b2f70667f94b6 \
    --hash=sha256:b08c95f32e5994454ea99882ff3c4a0afc8254c38998a0ed4b479dba448dc581
    # via -r requirements/test.in
```

#### Actual result
```
#
# This file is autogenerated by pip-compile with Python 3.11
# by the following command:
#
#    pip-compile --allow-unsafe --config=pyproject.toml --generate-hashes --output-file=requirements/test.txt --resolver=backtracking --strip-extras requirements/test.in
#
astpretty==3.0.0 \
    --hash=sha256:15bfd47593667169485a1fa7938b8de9445b11057d6f2b6e214b2f70667f94b6 \
    --hash=sha256:b08c95f32e5994454ea99882ff3c4a0afc8254c38998a0ed4b479dba448dc581
    # via -r requirements/test.in
```

q0w:
> When running pip-compile click context has no src_files, so select_config_file returns pyproject.toml, but after compiling click context has src_files as requirements/test.in, so select_config_file returns None and so pyproject.toml != None
Making --config is_eager=False works

atugushev:
Perhaps, `src_files` should be eager instead of `config`.

q0w:
oh maybe in `override_defaults_from_config_file` it should return None if there is no config in the file
```
def override_defaults_from_config_file(
    ctx: click.Context, param: click.Parameter, value: str | None
) -> Path | None:
[...]

PR #1904, opened 2023-07-08
Fix revealed default config in header if requirements in subfolder

### LINKED ISSUE #1903: pip-compile reveals default `--config=pyproject.toml` option in header
There is a bug with pip 6.14.0. I just use `pip-compile --allow-unsafe --generate-hashes --output-file=requirements/test.txt --resolver=backtracking --strip-extras requirements/test.in`
but now it adds `--config=pyproject.toml` even there is no configuration in pyproject.toml
When running pip-compile click context has no src_files, so select_config_file returns pyproject.toml, but after compiling click context has src_files as requirements/test.in, so select_config_file returns None and so pyproject.toml != None

Making `--config` is_eager=False works

_Originally posted by @q0w in https://github.com/jazzband/pip-tools/issues/1893#issuecomment-1624405135_

atugushev:
Ensure `src_files` are handled prior to `--config`, as `override_defaults_from_config_file` is dependent on `src_files`.

Fixes #1903.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -461,7 +461,7 @@ def make_config_file(tmpdir_cwd):
         config_dir.mkdir(exist_ok=True, parents=True)
 
         # Make a config file with this one config default override
-        config_file = Path(tmpdir_cwd / config_file_name)
+        config_file = tmpdir_cwd / config_file_name
         config_to_dump = {"tool": {"pip-tools": {pyproject_param: new_default}}}
         config_file.write_text(tomli_w.dumps(config_to_dump))
         return cast(Path, config_file.relative_to(tmpdir_cwd))
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -454,11 +454,12 @@ def make_config_file(tmpdir_cwd):
     def _maker(
         pyproject_param: str, new_default: Any, config_file_name: str = CONFIG_FILE_NAME
     ) -> Path:
-        # Make a config file with this one config default override
-        config_path = Path(tmpdir_cwd) / pyproject_param
-        config_file = config_path / config_file_name
-        config_path.mkdir(exist_ok=True)
+        # Create a nested directory structure if config_file_name includes directories
+        config_dir = Path(tmpdir_cwd / config_file_name).parent
+        config_dir.mkdir(exist_ok=True, parents=True)
 
+        # Make a config file with this one config default override
+        config_file = Path(tmpdir_cwd / config_file_name)
         config_to_dump = {"tool": {"pip-tools": {pyproject_param: new_default}}}
         config_file.write_text(tomli_w.dumps(config_to_dump))
         return config_file.relative_to(tmpdir_cwd)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -457,7 +457,7 @@ def make_config_file(tmpdir_cwd):
         config_file_name: str = DEFAULT_CONFIG_FILE_NAMES[0],
     ) -> Path:
         # Create a nested directory structure if config_file_name includes directories
-        config_dir = Path(tmpdir_cwd / config_file_name).parent
+        config_dir = (tmpdir_cwd / config_file_name).parent
         config_dir.mkdir(exist_ok=True, parents=True)
 
         # Make a config file with this one config default override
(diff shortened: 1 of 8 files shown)


atugushev:
@q0w this should fix your issue too #1902.

q0w:
what about this. I think it should pickup config from pyproject.toml
https://github.com/jazzband/pip-tools/issues/1902#issuecomment-1624429055

tests/test_utils.py
```diff
@@ -410,6 +410,29 @@ def test_get_compile_command_with_config(tmpdir_cwd, config_file, expected_comma
...
+    """
+    default_config_file = Path("pyproject.toml")
+    default_config_file.touch()
+
+    (tmpdir_cwd / "subdir").mkdir()
+    req_file = Path("subdir/requirements.in")
+    req_file.touch()
+
+    with open(req_file, "w"):
+        pass
```
webknjaz, line 429: Wouldn't this be simpler?
```suggestion
    req_file.write_bytes(b"")
```

tests/test_utils.py
```diff
@@ -410,6 +410,29 @@ def test_get_compile_command_with_config(tmpdir_cwd, config_file, expected_comma
...
+def test_get_compile_command_does_not_include_default_config_if_reqs_file_in_subdir(
+    tmpdir_cwd,
+):
+    """
+    Test that ``get_compile_command`` does not include default config file
+    if requirements file is in a subdirectory.
+    Regression test for issue GH-1903.
+    """
+    default_config_file = Path("pyproject.toml")
+    default_config_file.touch()
```
webknjaz, line 422: Can we also test the case when the `[tool.pip-tools]` section is present and empty?

Also, how about a `.pip-tools.toml` with an empty or missing `[tool.pip-tools]` and `pyproject.toml` with a non-empty one? I think that the `pyproject.toml` shouldn't be taken into account in this case.

atugushev: Okay, I'll add those combinations to parametrized tests 👍🏻

tests/test_utils.py
atugushev: Yes, indeed. Applied.

tests/test_utils.py
atugushev: Addressed in 64445d00524f2b76af6825c2f79424b179d914bc

webknjaz: I was hoping to see what happens if both files exist too..

atugushev: This looks like out of scope of `get_compile_command`. We can add more unit tests to `select_config_file()`.

webknjaz: If it's out of the scope, it'd be good to add the tests separately.

atugushev: Addressed In 783a6e0cb37ed52c2cdf07a88c5a5e5bdfca8d06

chrysle:
I guess #1906 is to be merged first?

atugushev:
Looks like the bug is fixed now. Ready for another round of review.

atugushev:
FTR, resolved some conflicts after merging with the main branch.

tests/conftest.py
```diff
@@ -452,13 +452,16 @@ def make_config_file(tmpdir_cwd):
...
+        pyproject_param: str,
+        new_default: Any,
+        config_file_name: str = DEFAULT_CONFIG_FILE_NAMES[0],
     ) -> Path:
-        # Make a config file with this one config default override
-        config_path = tmpdir_cwd / pyproject_param
-        config_file = config_path / config_file_name
-        config_path.mkdir(exist_ok=True)
+        # Create a nested directory structure if config_file_name includes directories
+        config_dir = Path(tmpdir_cwd / config_file_name).parent
```
theryanwalker, line 460:
```suggestion
        config_dir = (tmpdir_cwd / config_file_name).parent
```

The `tmpdir_cwd` already yields a Path-like object, so it's unnecessary.

theryanwalker (approved):
Left a minor comment. Otherwise, LGTM!

tests/conftest.py
atugushev: Ah, no. Can', because `as_posix()` shows full path and breaks tests.

atugushev:
Thanks @webknjaz and @theryanwalker for the review!

PR #1910, opened 2023-07-13
Validate parsed config against click command params

### LINKED ISSUE #1909: Using negative flag forms (e.g. no-header, no-annotate) in config has no effect
When testing out the new config behavior, if I add entries to `[tool.pip-tools]` like

```toml
no-annotate = true
```
or
```toml
no-header = true
```

then there are two surprises:

- the option does not take effect; it *must* be given in the positive form (`annotate = false` and `header = false`)
- there is no warning or error about unknown options being given, making this problem easy to miss

#### Environment Versions

1. Linux
1. Python version: `3.10.6`
1. pip version: `23.1.2`
1. pip-tools version: `6.14.0`

#### Steps to replicate

```console
$ git clone https://github.com/jazzband/pip-tools
$ cd pip-tools
$ printf '%s\n' '[tool.pip-tools]' 'no-annotate = true' >>pyproject.toml
$ printf '%s\n' httpx >requirements.in
$ python -m venv .venv
$ . ./.venv/bin/activate
$ pip-compile requirements.in
```

#### Expected result

Either 

- there be no annotations in the output, since `--no-annotate` is a valid flag, and `--no-annotate` should also appear in the generated header, or
- a warning or error should be presented due to an unrecognized option being supplied

#### Actual result

```ini
WARNING: the legacy dependency resolver is deprecated and will be removed in future versions of pip-tools. The default resolver will be changed to 'backtracking' in pip-tools 7.0.0. Specify --resolver=backtracking to silence this warning.
#
# This file is autogenerated by pip-compile with Python 3.10
# by the following command:
#
#
anyio==3.7.1
    # via httpcore
certifi==2023.5.7
    # via
    #   httpcore
    #   httpx
exceptiongroup==1.1.2
    # via anyio
h11==0.14.0
    # via httpcore
httpcore==0.17.3
    # via httpx
httpx==0.24.1
    # via -r requirements.in
idna==3.4
    # via
    #   anyio
    #   httpx
sniffio==1.3.0
    # via
    #   anyio
    #   httpcore
    #   httpx
```

atugushev:
Refs: #1909.

#### Config key validation
```
$ cat .pip-tools.toml
[tool.pip-tools]
annotation = false

$ pip-compile
Error: No such config key: annotation. (Possible options: annotate, annotation_style)
```

####  Config value validation
```
$ cat .pip-tools.toml
[tool.pip-tools]
dry-run = ["foo", "bar"]

$ pip-compile
Error: Invalid value for config key 'dry-run': ['foo', 'bar']
```

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -583,15 +583,16 @@ def _validate_config(
     config: dict[str, Any],
 ) -> None:
     """Validate parsed config against click command params."""
-    params = {
+    cli_params = {
         param.name: param
         for param in click_context.command.params
         if param.name is not None
     }
+
     for key, value in config.items():
-        # Validate key
-        if key not in params:
-            possibilities = difflib.get_close_matches(key, params.keys())
+        # Validate unknown keys
+        if key not in cli_params:
+            possibilities = difflib.get_close_matches(key, cli_params.keys())
             raise click.NoSuchOption(
                 option_name=key,
                 message=f"No such config key {key!r}.",
@@ -599,14 +600,14 @@ def _validate_config(
                 ctx=click_context,
             )
 
-        # Validate value
-        param = params[key]
+        # Validate invalid values
+        param = cli_params[key]
         try:
             param.type.convert(value=value, param=param, ctx=click_context)
         except Exception as e:
             raise click.BadOptionUsage(
                 option_name=key,
-                message=f"Invalid value for config key {key!r}: {value!r}",
+                message=f"Invalid value for config key {key!r}: {value!r}.",
                 ctx=click_context,
             ) from e
 
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -582,7 +582,12 @@ def _validate_config(
     click_context: click.Context,
     config: dict[str, Any],
 ) -> None:
-    """Validate parsed config against click command params."""
+    """
+    Validate parsed config against click command params.
+
+    :raises click.NoSuchOption: if config contains unknown keys.
+    :raises click.BadOptionUsage: if config contains invalid values.
+    """
     cli_params = {
         param.name: param
         for param in click_context.command.params
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import collections
 import copy
+import difflib
 import itertools
 import json
 import os
@@ -561,6 +562,7 @@ def override_defaults_from_config_file(
 
     config = parse_config_file(config_file)
     if config:
+        _validate_config(ctx, config)
         _assign_config_to_cli_context(ctx, config)
 
     return config_file
@@ -574,6 +576,39 @@ def _assign_config_to_cli_context(
         click_context.default_map = {}
 
     click_context.default_map.update(cli_config_mapping)
+
+
+def _validate_config(
+    click_context: click.Context,
+    config: dict[str, Any],
+) -> None:
+    """Validate parsed config against click command params."""
+    params = {
+        param.name: param
+        for param in click_context.command.params
+        if param.name is not None
+    }
+    for key, value in config.items():
+        # Validate key
+        if key not in params:
+            possibilities = difflib.get_close_matches(key, params.keys())
+            raise click.NoSuchOption(
+                option_name=key,
+                message=f"No such config key {key!r}.",
+                possibilities=possibilities,
+                ctx=click_context,
+            )
+
+        # Validate value
+        param = params[key]
+        try:
+            param.type.convert(value=value, param=param, ctx=click_context)
+        except Exception as e:
+            raise click.BadOptionUsage(
+                option_name=key,
+                message=f"Invalid value for config key {key!r}: {value!r}",
+                ctx=click_context,
+            ) from e
 
 
 def select_config_file(src_files: tuple[str, ...]) -> Path | None:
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -573,21 +573,16 @@ def test_get_sys_path_for_python_executable():
 @pytest.mark.parametrize(
     ("pyproject_param", "new_default"),
     (
-        # From sync
-        ("ask", True),
```
atugushev, line 577: FYI: I have removed the options for `pip-sync` because we only test the [pip-compile](https://github.com/jazzband/pip-tools/blob/776c3fa3d8b2da155eed7eb9a5b164f78937987a/tests/test_utils.py#L620-L621) CLI. We could add a separate test for pip-sync with its own options, but it seems redundant to me.

piptools/utils.py
```diff
@@ -576,6 +578,40 @@ def _assign_config_to_cli_context(
...
+        # Validate invalid values
+        param = cli_params[key]
+        try:
+            param.type.convert(value=value, param=param, ctx=click_context)
+        except Exception as e:
+            raise click.BadOptionUsage(
+                option_name=key,
+                message=f"Invalid value for config key {key!r}: {value!r}.",
+                ctx=click_context,
+            ) from e
```
chrysle, line 612:
```suggestion
    swap_keys = False
    new_items = {}
    old_items = {}

    for key, value in config.items():
        # Validate unknown keys
        if key not in cli_params:
            # Replace boolean flag contexts like ``no_annotate`` with their equivalents 
            if key.startswith("no_"):
                old_items[key] = value
                key = key[3:]
                value = not value
                swap_keys = True
                new_items[key] = value
            else:
                possibilities = difflib.get_close_matches(key, cli_params.keys())
                raise click.NoSuchOption(
                    option_name=key,
                    message=f"No such config key {key!r}.",
                    possibilities=possibilities,
                    ctx=click_context,
                )

        # Validate invalid values
        param = cli_params[key]
        try:
            param.type.convert(value=value, param=param, ctx=click_context)
        except Exception as e:
            raise click.BadOptionUsage(
                option_name=key,
                message=f"Invalid value for config key {key!r}: {value!r}.",
                ctx=click_context,
            ) from e
    
    if swap_keys:
        for key, value in old_items.items():
            del config[key]
        for key, value in new_items.items():
            config[key] = value
```

chrysle (commented):
Could we add the below condition? It would convert all negating boolean flags to their valid equivalents, e.g. `no-annotate = true` -> `annotate = false`, but error out in case these equivalents or the given values do not exist.

Of course the suggestions for non-valid flags would have to be adapted...

piptools/utils.py
atugushev: That makes sense indeed. I would prefer to implement this in the `parse_config_file()` function though. The `_validate_config()` function does not intend a `config` mutation.

chrysle: In that function, you might be able to implement it like this:

```python
    swap_items = False
    new_items = {}
    old_items = {}

    for key, value in piptools_config.items(): 
        # Replace boolean flag contexts like ``no_annotate`` with their equivalents
        if key.startswith("no_"):
            swap_items = True
            old_items[key] = value
            if isinstance(value, bool):
                value = not value
            new_items[key[3:]] = value

    if swap_items:
        for key, value in old_items.items():
            del piptools_config[key]
        for key, value in new_items.items():
            piptools_config[key] = value


    return piptools_config
```

atugushev: Would you like to address this in a following up PR?

chrysle: Sure!

atugushev: Great, thanks 👍🏻

atugushev: As a reminder to myself: I think we can refactor `parse_config_file` to parse options using this snippet.

<details>

<summary>Snippet</summary>

```python
(Pdb++) opts = {
    opt: option 
    for option in ctx.command.params
    for opt in itertools.chain(option.opts, option.secondary_opts)
}

(Pdb++) pp opts
{'--all-extras': <Option all_extras>,
 '--allow-unsafe': <Option allow_unsafe>,
 '--annotate': <Option annotate>,
 '--no-annotate': <Option annotate >,
 ...
}
 
(Pdb++) opts['--pip-args'].name
'pip_args_str'

(Pdb++) opts['--annotate'].name
'annotate'

(Pdb++) opts['--no-annotate'].name
'annotate'
```


</details>

chrysle: Looks good! I guess this would enable us to rationalise the function `get_click_dest_for_option()` away. Would you like me to address that too, while I'm at it?

atugushev: Sure, feel free to!

AndydeCleyre:
This looks good to me, thanks!

Can you include the exception types to expect, `click.BadOptionUsage` and `click.NoSuchOption`, in the `_validate_config` docstring?

atugushev:
> Can you include the exception types to expect, `click.BadOptionUsage` and `click.NoSuchOption`, in the `_validate_config` docstring?

@AndydeCleyre thanks! Addresses in 278d8df6d9127eba1d6a4a435349d49158988df3.

AndydeCleyre (approved):

atugushev:
Thanks @AndydeCleyre and @chrysle for the review! 🙏🏻

PR #1912, opened 2023-07-15
Don't use setuptools_scm from Sphinx itself

### LINKED ISSUE #1846: The latest and the stable documentation may have been built using the development version
Shouldn't be [there](https://pip-tools.readthedocs.io/) the latest released version?

Screenshot 2023-04-07 at 01 17 17

atugushev:
Perhaps resolves #1846.

> It is discouraged to use `setuptools_scm` from Sphinx itself, instead use `importlib.metadata` after editable/real installation. The underlying reason is, that services like Read the Docs sometimes change the working directory for good reasons and using the installed metadata prevents using needless volatile data there.

https://github.com/pypa/setuptools_scm/#usage-from-sphinx

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,15 +3,12 @@
 
 from __future__ import annotations
 
-from functools import partial
+from importlib.metadata import version as get_version
 from pathlib import Path
-
-from setuptools_scm import get_version
 
 # -- Path setup --------------------------------------------------------------
 
 PROJECT_ROOT_DIR = Path(__file__).parents[1].resolve()
-get_scm_version = partial(get_version, root=PROJECT_ROOT_DIR)
 
 
 # -- Project information -----------------------------------------------------
@@ -20,18 +17,11 @@ project = "pip-tools"
 author = f"{project} Contributors"
 copyright = f"The {author}"
 
-# The short X.Y version
-version = ".".join(
-    get_scm_version(
-        local_scheme="no-local-version",
-    ).split(
-        "."
-    )[:3],
-)
-
 # The full version, including alpha/beta/rc tags
-release = get_scm_version()
+release = get_version(project)
 
+# The short X.Y version
+version = ".".join(release.split(".")[:3])
 
 # -- General configuration ---------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -20,18 +17,14 @@
...
-
 # The full version, including alpha/beta/rc tags
-release = get_scm_version()
+release = get_version(project)
+
+# The short X.Y version
+version = ".".join(release.split(".")[:3])
 
+print("pip-tools version:", version)
+print("pip-tools release:", release)
```
chrysle, line 27: Accidentally left over?

atugushev: Left on purpose to see versions in logs:
```
$ python -m sphinx ...
Running Sphinx v6.2.1
pip-tools version: 7.0.1
pip-tools release: 7.0.1.dev4
making output directory... done
...
```

chrysle: Maybe a comment about that could be useful?

chrysle (approved):
Looks good to me, then! Apologies I didn't get to this sooner.

RonnyPfannschmidt (approved):
lovely, this makes things more robust

i still would recommend to use the default local scheme but i can understand that the current behaviour is a pain for test-pypi

the next major version of setuptools_scm will provide a better behaviour on ci and the primary branches

docs/conf.py
atugushev: Addressed in a49c505c435d3e7e42bd2f0d8ed7ebdbd2701d42.

atugushev:
Thanks @RonnyPfannschmidt for the review and the heads up! Much appreciated 🙏🏻

pyproject.toml
```diff
@@ -110,3 +110,4 @@ markers = ["network: mark tests that require internet access"]
 include = ["piptools*"]
 
 [tool.setuptools_scm]
+local_scheme = "dirty-tag"
```
webknjaz, line 113: I wouldn't make such changes because it looses the context in PR runs and during development.

PR #1913, opened 2023-07-16
Add support for negating CLI boolean flags in config

### LINKED ISSUE #1909: Using negative flag forms (e.g. no-header, no-annotate) in config has no effect

chrysle:
Closes #1909

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -560,7 +560,7 @@ def override_defaults_from_config_file(
     else:
         config_file = Path(value)
 
-    config = parse_config_file(config_file)
+    config = parse_config_file(ctx, config_file)
     if config:
         _validate_config(ctx, config)
         _assign_config_to_cli_context(ctx, config)
@@ -659,17 +659,6 @@ NON_STANDARD_OPTION_DEST_MAP: dict[str, str] = {
 }
 
 
-def get_click_dest_for_option(option_name: str) -> str:
-    """
-    Returns the click ``dest`` value for the given option name.
-    """
-    # Format the keys properly
-    option_name = option_name.lstrip("-").replace("-", "_").lower()
-    # Some options have dest values that are overrides from the click generated default
-    option_name = NON_STANDARD_OPTION_DEST_MAP.get(option_name, option_name)
-    return option_name
-
-
 # Ensure that any default overrides for these click options are lists, supporting multiple values
 MULTIPLE_VALUE_OPTIONS = [
     "extras",
@@ -681,7 +670,9 @@ MULTIPLE_VALUE_OPTIONS = [
 ]
 
 
-def parse_config_file(config_file: Path) -> dict[str, Any]:
+def parse_config_file(
+    click_context: click.Context, config_file: Path
+) -> dict[str, Any]:
     try:
         config = tomllib.loads(config_file.read_text(encoding="utf-8"))
     except OSError as os_err:
@@ -695,11 +686,25 @@ def parse_config_file(config_file: Path) -> dict[str, Any]:
             hint=f"Could not parse '{config_file !s}': {value_err !s}",
         )
 
+    cli_opts = {
+        opt: option
+        for option in click_context.command.params
+        for opt in itertools.chain(option.opts, option.secondary_opts)
+    }
+
     # In a TOML file, we expect the config to be under `[tool.pip-tools]`
     piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})
-    piptools_config = {
-        get_click_dest_for_option(k): v for k, v in piptools_config.items()
-    }
+
+    # Replace boolean flags like ``--no-annotate`` with their equivalents
+    try:
+        piptools_config = {
+            cli_opts["--" + k].name: (
+                not v if k.startswith("no-") and isinstance(v, bool) else v
+            )
+            for k, v in piptools_config.items()
+        }
+    except KeyError:
+        pass
     # Any option with multiple values needs to be a list in the pyproject.toml
     for mv_option in MULTIPLE_VALUE_OPTIONS:
         if not isinstance(piptools_config.get(mv_option), (list, type(None))):
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -3017,3 +3017,31 @@ def test_raise_error_on_invalid_config_option(
 
     assert out.exit_code == 2
     assert "Invalid value for config key 'dry_run': ['invalid', 'value']" in out.stderr
+
+
+def test_cli_boolean_flag_config_option_has_valid_context(
+    pip_conf, runner, tmp_path, make_config_file
+):
+    config_file = make_config_file("no-annotate", True)
+
+    req_in = tmp_path / "requirements.in"
+    req_in.touch()
+
+    out = runner.invoke(cli, [req_in.as_posix(), "--config", config_file.as_posix()])
+
+    assert out.exit_code == 0
+    assert "No such config key 'no_annotate'." not in out.stderr
+
+
+def test_invalid_cli_boolean_flag_config_option_captured(
+    pip_conf, runner, tmp_path, make_config_file
+):
+    config_file = make_config_file("no-annnotate", True)
+
+    req_in = tmp_path / "requirements.in"
+    req_in.touch()
+
+    out = runner.invoke(cli, [req_in.as_posix(), "--config", config_file.as_posix()])
+
+    assert out.exit_code == 2
+    assert "No such config key 'no-annnotate'." in out.stderr


piptools/utils.py
```diff
@@ -695,11 +686,27 @@ def parse_config_file(config_file: Path) -> dict[str, Any]:
...
-    piptools_config = {
-        get_click_dest_for_option(k): v for k, v in piptools_config.items()
-    }
+
+    # Replace boolean flags like ``--no-annotate`` with their equivalents
+    try:
+        piptools_config = {
+            cli_opts["--" + k].name: (
+                not v
+                if k.startswith("no-") and not k == "no-index" and isinstance(v, bool)
```
atugushev, line 703: We can omit `not k == "no-index"` condition, because `cli_opts["--no-index"]` will raise KeyError

piptools/utils.py
```diff
@@ -695,11 +686,27 @@ def parse_config_file(config_file: Path) -> dict[str, Any]:
...
+    }
+
     # In a TOML file, we expect the config to be under `[tool.pip-tools]`
     piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})
-    piptools_config = {
-        get_click_dest_for_option(k): v for k, v in piptools_config.items()
-    }
+
+    # Replace boolean flags like ``--no-annotate`` with their equivalents
+    try:
```
atugushev, line 699: if we run `_validate_config()` function before this line then we won't need the try-except block.

piptools/utils.py
```diff
@@ -695,11 +686,27 @@ def parse_config_file(config_file: Path) -> dict[str, Any]:
             hint=f"Could not parse '{config_file !s}': {value_err !s}",
         )
 
+    cli_opts = {
+        opt: option
+        for option in click_context.command.params
+        for opt in itertools.chain(option.opts, option.secondary_opts)
```
atugushev, line 692:
```suggestion
        for opt in itertools.chain(option.opts, option.secondary_opts)
        if opt.statrswith("--")
```

Better to filter out short option names, because we use long option names in config.

```
 '-P': <Option upgrade_packages>,
 '-U': <Option upgrade>,
 '-f': <Option find_links>,
 '-i': <Option index_url>,
 '-n': <Option dry_run>,
 '-o': <Option output_file>,
 '-p': <Option pre>,
 '-q': <Option quiet>,
 '-r': <Option rebuild>,
 '-v': <Option verbose>,
 ```

File: tests/test_cli_compile.py [hunk 4]
```diff
@@ -3002,13 +3002,13 @@ def test_raise_error_on_unknown_config_option(
...
     assert out.exit_code == 2
-    assert "No such config key 'unknown_option'" in out.stderr
+    assert "No such config key 'unknown-option'" in out.stderr
 
 
 def test_raise_error_on_invalid_config_option(
     pip_conf, runner, tmp_path, make_config_file
 ):
-    config_file = make_config_file("dry-run", ["invalid", "value"])
+    config_file = make_config_file("dry_run", ["invalid", "value"])
```
atugushev, line 3011:
```
dry_run = false 
dry-run = false
```

Shouldn't we support both variants?

tests/test_utils.py
```diff
@@ -615,10 +615,15 @@ def test_callback_config_file_defaults(pyproject_param, new_default, make_config
     # Create a "compile" run example pointing to the config file
     ctx = Context(compile_cli)
     ctx.params["src_files"] = (str(config_file),)
+    cli_opts = {
+        opt: option
+        for option in ctx.command.params
+        for opt in itertools.chain(option.opts, option.secondary_opts)
+    }
```
atugushev, line 622: To DRY this out, could we extract this to a function, like 'get_cli_options(ctx)'?

tests/test_cli_compile.py
```diff
@@ -3017,3 +3017,31 @@ def test_raise_error_on_invalid_config_option(
...
+):
+    config_file = make_config_file("no-annotate", True)
+
+    req_in = tmp_path / "requirements.in"
+    req_in.touch()
+
+    out = runner.invoke(cli, [req_in.as_posix(), "--config", config_file.as_posix()])
+
+    assert out.exit_code == 0
+    assert "No such config key 'no-annotate'." not in out.stderr
```
atugushev, line 3033:
```suggestion
    req_in.write_text("small-fake-a==0.1")

    out = runner.invoke(
        cli,
        [
            req_in.as_posix(),
            "--config",
            config_file.as_posix(),
            "--no-emit-options",
            "--no-header",
            "--output-file",
            "-",
        ],
    )

    assert out.exit_code == 0
    assert out.stdout == "small-fake-a==0.1\n"
```

Asserting for `not in` can be tricky and easily could go wrong in positive test scenarios. I'd suggest checking the whole output instead. This would make us 100% sure that `--no-annotate` is properly handled.

atugushev (commented):
@chrysle thank you very much for fixing this! I've left a few comments above for code improvement.

chrysle:
Thanks for the review! I've tried to apply your suggestions.

tests/test_cli_compile.py
chrysle: Do you have suggestion how to do this with `pytest`? We could also make the `make_config_file()` function accept multiple parameters, BTW.

piptools/utils.py
chrysle: This isn't possible since the boolean flags aren't filtered at this point...

piptools/utils.py
chrysle: Sadly it doesn't, and breaks the tests. I've added this condition after the failures. So far I've not found a way to properly detect a `click` boolean flag.

piptools/utils.py
atugushev: Yeah, this would require to update `_validate_config()`.

atugushev: I've checked and it raises error on `no-index`:

```diff
diff --git piptools/utils.py piptools/utils.py
index f454ce7..470f844 100644
--- piptools/utils.py
+++ piptools/utils.py
@@ -701,6 +701,8 @@ def parse_config_file(
     # In a TOML file, we expect the config to be under `[tool.pip-tools]`
     piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})

+    _validate_config(click_context, piptools_config)
+
     # Replace boolean flags like ``--no-annotate`` with their equivalents
     try:
         piptools_config = {
```

```bash
$ cat .pip-tools.toml
[tool.pip-tools]
no-index = true

$ pip-compile
Usage: pip-compile [OPTIONS] [SRC_FILES]...
Try 'pip-compile -h' for help.

Error: No such config key 'no-index'. Did you mean no_index?
```

piptools/utils.py
atugushev: And with this code https://github.com/jazzband/pip-tools/pull/1913/files#r1264749661 we don't need  `not k == "no-index"` check too.

piptools/utils.py
atugushev: Oh, wait `no-index=True` should be valid flag.

tests/test_cli_compile.py
atugushev: This might take some refactoring. Let's address that in a following-up PR.

atugushev:
@chrysle please look into the https://github.com/jazzband/pip-tools/pull/1913, I've made some changes to clean up the code a bit.

atugushev:
@chrysle should you like to document that feature in the README?

README.md
```diff
@@ -289,19 +289,28 @@ By default, both `pip-compile` and `pip-sync` will look first
...
 
+Since `pip-tools` supports default values vor [all valid command-line flags](/cli/index.md)
+of its subcommands, you can also achieve the functionality in the above example
+by setting a configuration value for the negative flag `--no-annotate`:
+
+```toml
+[tool.pip-tools]
+no-annotate = false
+```
+
```
atugushev, line 313: This may seem like a special case. However, we essentially support any long options (those starting with `--` prefix) listed in [command line reference](https://pip-tools.readthedocs.io/en/latest/cli/) represented in either snake or kebab case (without the double dash). Perhaps it's worth mentioning this instead? What do you think?

README.md
```diff
@@ -289,19 +289,28 @@ By default, both `pip-compile` and `pip-sync` will look first
...
 
-For example, to by default generate `pip` hashes in the resulting
-requirements file output, you can specify in a configuration file
+For example, to by default annotate dependencies with their origins in the resulting
+requirements file output, you can specify in a configuration file:
 
 ```toml
 [tool.pip-tools]
-generate-hashes = true
+annotate = true
```
atugushev, line 297: The `annotate = true` setting is actually the default value which typically doesn't have any extra effect and is rarely used in practice in this way. On the other hand, the 'generate hashes' option is quite good example. Perhaps we could keep it in our documentation?

atugushev (commented):
Thanks for the docs! A few comments to consider below:

chrysle:
Thanks, I agree with your points and reworked the documentation.

README.md
chrysle: Apparently, the Sphinx TOML lexer doesn't seem to concur with that suggestion...

README.md
```diff
@@ -290,18 +290,26 @@ for a `.pip-tools.toml` file and then in your `pyproject.toml`. You can
...
 value.
 
+`pip-tools` supports default values vor [all valid command-line flags](/cli/index.md)
+of its subcommands. Also, it is possible to specify configuration keys in
+command-line manner, for example:
+
+```toml
+[tool.pip-tools]
+--generate-hashes = true
+```
```
atugushev, line 311: Perhaps my [comment](https://github.com/jazzband/pip-tools/pull/1913#discussion_r1272691430) was confusing, I apologise. I meant _options that represented as a snake/kebab case string **without double dash prefix**_. For example:

```toml
# kebab and snake case
generate-hashes = true
generate_hashes = false

# booleans
annotate = false
no-annotate = true

# ... or inverted booleans
annotate = true
no-annotate = false 

```

README.md
atugushev: See https://github.com/jazzband/pip-tools/pull/1913#discussion_r1272724649

chrysle: I've tested that configuration though, and seen it work...

atugushev: It might work, since we do the normalization after the validation for now. However, we should not encourage this, because it will be fixed soon.

chrysle:
Sorry for the misunderstanding, I hope it's better now.

README.md
chrysle: We may be able to call `_validate_config()` directly in `_invert_negative_bool_options_in_config()`.

atugushev:
```
Bug Fixes:

- Add support for negating CLI boolean flags in config (https://github.com/jazzband/pip-tools/pull/1913). Thanks @chrysle
```

@chrysle a side-note about [changelog](https://github.com/jazzband/pip-tools/releases): "Add support" usually goes to "Features" section. Do you think it would make sense to either rephrase the PR title or reconsider its labels?

chrysle:
Thanks for pointing out! How about:

> ```
> Bug Fixes:
> 
> - Fix rejection of negating CLI boolean flags in config (https://github.com/jazzband/pip-tools/pull/1913). Thanks @chrysle
> ```

PR #1919, opened 2023-07-17
Fix a bug where pip-sync would unexpectedly uninstall some packages

atugushev:
- fixes #1917 
- fixes  #1918

##### Contributor checklist

- [x] Provided the tests for the changes.
- [x] Assure PR title is short, clear, and good to be included in the user-oriented changelog

##### Maintainer checklist

- [x] Assure one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [x] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #1927, opened 2023-07-20
Update types using fresh dependencies

atugushev:
Since 21.1 pip has [py.typed](https://github.com/pypa/pip/commit/09e8fa96c7bf821916e9a21135359d1507fa5b68) file, hence many updates on type annotations.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,11 +28,9 @@ repos:
         # Keep exclude in sync with mypy own excludes
         exclude: ^tests/test_data/
         additional_dependencies:
-          - click==8.0.1
-          - pep517==0.10.0
-          - toml==0.10.2
-          - pip==20.3.4
-          - build==0.9.0
+          - click==8.1.6
+          - pip==23.2
+          - build==0.10.0
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.5
     hooks:
--- a/piptools/__main__.py
+++ b/piptools/__main__.py
@@ -5,7 +5,7 @@ import click
 from piptools.scripts import compile, sync
 
 
-@click.group()
+@click.group
 def cli() -> None:
     pass
 
--- a/piptools/_compat/pip_compat.py
+++ b/piptools/_compat/pip_compat.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 import optparse
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Iterable, Iterator, Set, cast
+from typing import Iterable, Iterator, Set, cast
 
 import pip
 from pip._internal.cache import WheelCache
 from pip._internal.index.package_finder import PackageFinder
 from pip._internal.metadata import BaseDistribution
+from pip._internal.metadata.importlib import Distribution as _ImportLibDist
 from pip._internal.metadata.pkg_resources import Distribution as _PkgResourcesDist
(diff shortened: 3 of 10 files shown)


atugushev:
@AndydeCleyre 

FYI: this is the first towards addressing the following comments:
- https://github.com/jazzband/pip-tools/pull/1919#discussion_r1266236409
- https://github.com/jazzband/pip-tools/pull/1919#discussion_r1265535387

webknjaz:
FWIW it might be useful to run mypy against several pip and Python versions (like min + max). Just the last one isn't going to cover a lot of stuff..

atugushev:
> FWIW it might be useful to run mypy against several pip and Python versions (like min + max). Just the last one isn't going to cover a lot of stuff..

@webknjaz I wonder if there are any library/apps in wild that do this?

webknjaz:
Pip itself used to have py3 and py2 runs both in pre-commit, before 2020. I also do so in some places (like Cheroot). One possible strategy is to test every other version or min/max to cover most stuff while still not using too much resources.
I think that ansible-test runs mypy under every supported interpreter version (both for the ansible-core itself and the external collections).

webknjaz:
If you add several check invocations, it's important to add aliases because several checks have the same id and there's no way to select a specific check without this.

atugushev:
> I also do so in some places (like Cheroot)

@webknjaz how do you handle type incompatibilities in code between different versions?

atugushev:
FTR: reopened PR because checks were stuck for some reason.

atugushev:
> FWIW it might be useful to run mypy against several pip and Python versions (like min + max). Just the last one isn't going to cover a lot of stuff..

@webknjaz opened a separate issue https://github.com/jazzband/pip-tools/issues/1928 to address the suggestion.

.pre-commit-config.yaml
```diff
@@ -28,11 +28,9 @@ repos:
...
         exclude: ^tests/test_data/
         additional_dependencies:
-          - click==8.0.1
-          - pep517==0.10.0
-          - toml==0.10.2
-          - pip==20.3.4
-          - build==0.9.0
+          - click==8.1.6
+          - pip==23.2
+          - build==0.10.0
```
chrysle, line 33: We haven't got these requirements pinned in `pyproject.toml` - Should we do this, or instead remove the boundary here?

webknjaz: @chrysle this is kinda a special case. https://pre-commit.ci is heavily cached and the cache virtually never expires. Locally we can use a `pre-commit` subcommand to clean it but no in the CI service. This kinda forces us to have something that would update in the cache key. But also, not having pins will cause difference in behavior on people's machines depending on when they first ran pre-commit and got the env cached.

webknjaz:
> how do you handle [type](https://github.com/jazzband/pip-tools/blob/1ae64ffb2db3b2f80971b865bf6cef0fb14ee773/piptools/_compat/pip_compat.py#L83-L84) [incompatibilities](https://github.com/jazzband/pip-tools/blob/1ae64ffb2db3b2f80971b865bf6cef0fb14ee773/piptools/_compat/pip_compat.py#L90) between different versions in the same code?

MyPy's `assert`-based type narrowing can handle this. Though, there might be the need to have multiple MyPy invocations running against different versions of pip pinned.

PR #1933, opened 2023-07-24
Allow options in config from both `pip-compile` and `pip-sync`

### LINKED ISSUE #1932: `pip-compile` and `pip-sync` have different options, but single config section
#### Environment Versions

1. OS Type: any
1. Python version: any
1. pip version: any
1. pip-tools version: `pip-compile, version 7.1.0`

#### Steps to replicate

```bash
echo pytz > requirements.in
echo -e "[tool.pip-tools]\ngenerate-hashes = true" > .pip-tools.toml
pip-compile
pip-sync
```

#### Expected result

Successful sync.

#### Actual result

```console
$ pip-compile
#
# by the following command:
#
#    pip-compile --generate-hashes
#
pytz==2023.3 \
    --hash=sha256:1d8ce29db189191fb55338ee6d0387d82ab59f3d00eac103412d64e0ebd0c588 \
    --hash=sha256:a151b3abb88eda1d4e34a9814df37de2a80e301e68ba0fd856fb9b46bfbbbffb
    # via -r requirements.in


$ pip-sync
Usage: pip-sync [OPTIONS] [SRC_FILES]...
Try 'pip-sync -h' for help.

Error: No such config key 'generate_hashes'.
```

atugushev:
Fixes #1932.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -588,16 +588,23 @@ def _validate_config(
     :raises click.NoSuchOption: if config contains unknown keys.
     :raises click.BadOptionUsage: if config contains invalid values.
     """
-    cli_params = {
-        param.name: param
-        for param in click_context.command.params
-        if param.name is not None
+    from piptools.scripts.compile import cli as compile_cli
+    from piptools.scripts.sync import cli as sync_cli
+
+    compile_cli_params = {
+        param.name: param for param in compile_cli.params if param.name is not None
     }
 
+    sync_cli_params = {
+        param.name: param for param in sync_cli.params if param.name is not None
+    }
+
+    all_keys = set(compile_cli_params) | set(sync_cli_params)
+
     for key, value in config.items():
-        # Validate unknown keys
-        if key not in cli_params:
-            possibilities = difflib.get_close_matches(key, cli_params.keys())
+        # Validate unknown keys in both compile and sync
+        if key not in all_keys:
+            possibilities = difflib.get_close_matches(key, all_keys)
             raise click.NoSuchOption(
                 option_name=key,
                 message=f"No such config key {key!r}.",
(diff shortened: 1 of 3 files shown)


piptools/utils.py
```diff
@@ -588,33 +588,44 @@ def _validate_config(
     :raises click.NoSuchOption: if config contains unknown keys.
     :raises click.BadOptionUsage: if config contains invalid values.
     """
-    cli_params = {
-        param.name: param
-        for param in click_context.command.params
-        if param.name is not None
+    from piptools.scripts.compile import cli as compile_cli
+    from piptools.scripts.sync import cli as sync_cli
```
chrysle, line 592: Maybe move this to the top?

atugushev: This avoids circular imports.

piptools/utils.py
```diff
@@ -591,33 +591,44 @@ def _validate_config(
...
-            raise click.BadOptionUsage(
-                option_name=key,
-                message=f"Invalid value for config key {key!r}: {value!r}.",
-                ctx=click_context,
-            ) from e
+        # Validate values for both compile and sync
+        for cli_params in (compile_cli_params, sync_cli_params):
+            param = cli_params.get(key)
+            if param is None:
+                continue
```
chrysle, line 622: Do we need this? I think wrong keys are caught above?

atugushev: yeah, but we still checking both cli_params, see 618 line.

chrysle: Sorry if I understood you wrong, but isn't this covered by the `all_keys` variable test? Also, the value can't possibly be `None`?

atugushev: Suppose there is an `ask = true` option in the config. It's included in `all_keys`, but not found within  `compile_cli_params`. This Is why we skip it and proceed to validate it on `sync_cli_params`.

The condition `if key not in all_keys` is True if there is a unknown key that neither `pip-compile` nor `pip-sync` has.

chrysle: Thank you for the explanantion! I actually thought both dicts were combined into one.

piptools/utils.py
```diff
@@ -591,36 +591,47 @@ def _validate_config(
...
+
+            try:
+                param.type.convert(value=value, param=param, ctx=click_context)
```
chrysle, line 625: I think you still need to update that after the merge?

```suggestion
                param.type_cast_value(value=value, ctx=click_context)
```

atugushev: Yeah, I messed up with resolving conflicts :/

piptools/utils.py
atugushev: Considering your questions, I've refactored a bit in 5b029cd. Does it look all right?

piptools/utils.py
atugushev: Addressed in f0236945719081f9c012e052f785b28e8048d514

chrysle (approved):
Thanks for the work!

atugushev:
@chrysle thanks for the review!

AndydeCleyre:
Does this mean any flags between the two tools will have to be the same value, if the config file is used?

atugushev:
@AndydeCleyre correct.

chrysle:
BTW, I wonder, are separate config sections for the commands (for example `[tool.pip-compile]`) out of the question?

atugushev:
@chrysle I wouldn't mind against that.

PR #1936, opened 2023-07-25
Add `-c/--constraint` option to `pip-compile`

### LINKED ISSUE #1092: Improve layered dependency workflow
### What's the problem this feature will solve?
I know this has been discussed a few times (e.g. #398), but the current solution for layered dependencies still doesn't work in all cases. For example the following simple files cannot be compiled:
```
# base.in
requests

# dev.in
-c base.txt
moto
```
Running `pip-compile base.in` results in 
```
certifi==2019.11.28       # via requests
chardet==3.0.4            # via requests
idna==2.9                 # via requests
requests==2.23.0          # via -r base.in
urllib3==1.25.8           # via requests
```
Now `pip-compile dev.in` aborts with an error:
```
Could not find a version that matches idna<2.9,==2.9,>=2.5 (from -c base.txt (line 9))
Tried: 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 2.0, 2.0, 2.1, 2.1, 2.2, 2.2, 2.3, 2.3, 2.4, 2.4, 2.5, 2.5, 2.6, 2.6, 2.7, 2.7, 2.8, 2.8, 2.9, 2.9
There are incompatible versions in the resolved dependencies:
  idna==2.9 (from -c base.txt (line 9))
  idna<2.9,>=2.5 (from moto==1.3.14->-r dev.in (line 2))
```
A solution exists, though: Just use `idna==2.8`.

### Describe the solution you'd like

In my opinion, a solution to the layered dependencies problem requires to compile the outermost layer first: Then the resolver has a chance to find a solution for *all* dependencies. A working solution is the following:
```
# base.in
requests

# dev.in
-r base.in
moto

# constraint.in
-c dev.txt
```
Now I can generate correct dependency sets with
```
pip-compile dev.in
pip-compile base.in constraint.in --output-file base.txt
```
The `constraint.in` is necessary because otherwise the second `pip-compile` will generate versions that differ from those in `dev.txt`.

What do you think of this solution? How about adding a `-c/--constrain` option to `pip-compile`, that acts like I added a `constraint.in` file above. Then I could run
```
pip-compile dev.in
pip-compile -c dev.txt base.in
```
with the same effect as above.

### LINKED ISSUE #1364: --constraint support

There are 2 projects:

+ `lib` is an internal library. It has `setup.py` with dependencies of the library.
+ `srv` is an internal service that uses `lib`. It has a lock file `requirements.txt`.

The goal is to generate `lib/requirements.txt` which has only packages from `lib/setup.py` but the same versions as specified in `srv/requirements.txt`. Motivation:

+ The same version is needed to make sure that if the tests have passed for `lib`, it will work as expected in the environment of `srv`.
+ Running tests for `lib` in the environment of `srv` is complicated on CI: the env of `srv` is too big, and we want to avoid making it for pipelines in `lib`.

The idea is the same as in pip's [Constraints Files](https://pip.pypa.io/en/stable/user_guide/?highlight=--constraint#constraints-files) (`-c/--constraint` option). So, it makes sense to introduce the same key for pip-tools.

#### Alternative Solutions

pip-tools already uses the output file as the constraint. So, a workaround I found is to specify the constraint file as the output file and then restore it. The PoC:

```python
import sys
import shutil
import subprocess
from argparse import ArgumentParser


def main():
    parser = ArgumentParser()
    parser.add_argument('-c', '--constraint', required=True)
    parser.add_argument('--output-file', default='requirements.txt')
    args, rest = parser.parse_known_args()
    shutil.copy(args.constraint, '/tmp/c.txt')
    cmd = [sys.executable, '-m', 'piptools', 'compile', '--output-file', args.constraint]
    try:
        code = subprocess.call(cmd + rest)
        shutil.copy(args.constraint, args.output_file)
    finally:
        shutil.copy('/tmp/c.txt', args.constraint)
    sys.exit(code)

if __name__ == '__main__':
    main()
```

#### Additional context

-

### LINKED ISSUE #1891: pip-compile --extra doesn't take into account already pinned versions
I'm trying to move away from `*.in` files and manage everything from pyproject.toml, but I'm facing a problem where `pip-compile --extra` doesn't take into account the already pinned versions from `pip-compile` (without --extra).

#### Environment Versions

1. OS Type: Arch Linux
1. Python version: 3.11.3
1. pip version: pip 23.1.2 from /usr/lib/python3.11/site-packages/pip (python 3.11)
1. pip-tools version: 6.13.0

#### Steps to replicate

1. Create the following `pyproject.toml`:

```toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"

[project]
name = "foobar"
version = "1"
dependencies = ["django==4.1.7"]

dev = ["pytest"]
```

3. Execute `pip-compile --no-header --no-annotate --resolver backtracking` to obtain:

```
asgiref==3.7.2
django==4.1.7
sqlparse==0.4.4
```

5. Remove the restriction from the `django` dependency from the `pyproject.toml`, and verify that `pip-compile` doesn't update any package.
6. Execute `pip-compile --no-header --no-annotate --resolver backtracking --extra dev -o dev-requirements.txt`, and see that the `django` dependency has been updated in the `dev-requirements.txt`:

```
asgiref==3.7.2
django==4.2.2
iniconfig==2.0.0
packaging==23.1
pluggy==1.2.0
pytest==7.4.0
sqlparse==0.4.4
```

I also tried to pass `--pip-args "-c requirements.txt"`, but nothing changed.

#### Expected result

Actually, I would expect a way to pin *only* the dev dependencies, since the main dependencies were already pinned. Some workflow similar to the [Workflow for layered requirements](https://pip-tools.readthedocs.io/en/latest/#workflow-for-layered-requirements) but with `pyproject.toml` instead of `*.in` files.

### LINKED ISSUE #1907: Adding a nesting structure so that pip-compile can pin the same versions across nested extras.

I am using `setup.cfg` to manage my dependencies and extra dependencies, and I am nesting some extra dependencies in each other. Currently, pip-compile does not guarantee resolving to the same version when I compile the requirements for nested extras.

```cfg
[options.extras_require]
server =
    pydantic~=1.10.7, >1.9.0

dev =
    %(server)s
    wget
```

I would like to constrain the extended extras to the versions pinned by the `nested` extra. Here `server` should pin a pydantic version that `dev` must follow, and all should be generated into a single independent `requirements.txt` for each extra.

This is especially important in building smaller containers, since I can guarantee version pinning and smaller diffs between layers of nested extras.

```Dockerfile
FROM base as builder-1
pip install --no-cache-dir server-requirements.txt

FROM builder
pip install --no-cache-dir dev-requirements.txt
```
#### Alternative Solutions
I can use multiple `requirements.txt` and nest them in each other and introduce dynamic constraints by symlinking the outputs of base extra.

#### Additional context

I use a single `setup.cfg` so that I can nest `extra` pip dependencies easily using `setuptools` magic. This gives me great flexibility to add requirements and does not make my repo much more crowded. If I had to migrate to having multiple `requirments.in` and dynamic constrains, it would add a lot of intermediary files, and it is harder to maintain the repository. I think having a single file in which I can define all packages is way more superior to any alternative IMO.

atugushev:
- resolves #1364
- resolves #1092
- resolves #1891
- resolves #1907
- https://stackoverflow.com/q/76055688/1529164.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -338,10 +338,7 @@ def _determine_linesep(
         path_type=str,
     ),
     multiple=True,
-    help=(
-        "Constrain versions using the given constraints file. "
-        "This option can be used multiple times."
-    ),
+    help="Constrain versions using the given constraints file; may be used more than once.",
 )
 def cli(
     ctx: click.Context,
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -326,6 +326,23 @@ def _determine_linesep(
     help="Do not read any config file.",
     is_eager=True,
 )
+@click.option(
+    "-c",
+    "--constraint",
+    type=click.Path(
+        exists=True,
+        file_okay=True,
+        dir_okay=False,
+        readable=True,
+        allow_dash=False,
+        path_type=str,
+    ),
+    multiple=True,
+    help=(
+        "Constrain versions using the given constraints file. "
+        "This option can be used multiple times."
+    ),
+)
 def cli(
     ctx: click.Context,
     verbose: int,
@@ -366,6 +383,7 @@ def cli(
     unsafe_package: tuple[str, ...],
     config: Path | None,
     no_config: bool,
+    constraint: tuple[str, ...],
 ) -> None:
     """
     Compiles requirements.txt from requirements.in, pyproject.toml, setup.cfg,
@@ -560,6 +578,18 @@ def cli(
                     options=repository.options,
                 )
             )
+
+    # Parse all constraints from `--constraint` files
+    for filename in constraint:
+        constraints.extend(
+            parse_requirements(
+                filename,
+                constraint=True,
+                finder=repository.finder,
+                options=repository.options,
+                session=repository.session,
+            )
+        )
 
     if upgrade_packages:
         constraints_file = tempfile.NamedTemporaryFile(mode="wt", delete=False)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -631,6 +631,7 @@ def test_callback_config_file_defaults(pyproject_param, new_default, make_config
         "find-links",
         "extra-index-url",
         "trusted-host",
+        "constraint",
     ),
 )
 def test_callback_config_file_defaults_multi_value_options(mv_option, make_config_file):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -634,7 +634,7 @@ def test_callback_config_file_defaults(pyproject_param, new_default, make_config
         ("trusted_host", "not-a-list"),
         ("annotate", "not-a-bool"),
         ("max_rounds", "not-an-int"),
-        ("constraint", "not-an-int"),
+        ("constraint", "not-an-list"),
     ),
 )
 def test_callback_config_file_defaults_multi_validate_value(
--- a/README.md
+++ b/README.md
@@ -98,7 +98,6 @@ dependencies = ["django"]
 
 [project.optional-dependencies]
 dev = ["pytest"]
-
 ```
 
 You can produce your pin files as easily as:
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -678,6 +678,7 @@ MULTIPLE_VALUE_OPTIONS = [
     "find_links",
     "extra_index_url",
     "trusted_host",
+    "constraint",
 ]
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass, field
 from functools import partial
 from pathlib import Path
 from textwrap import dedent
-from typing import Any
+from typing import Any, cast
 
 import pytest
 import tomli_w
@@ -234,7 +234,7 @@ def runner():
 @pytest.fixture
 def tmpdir_cwd(tmpdir):
     with tmpdir.as_cwd():
-        yield tmpdir
+        yield Path(tmpdir)
 
 
 @pytest.fixture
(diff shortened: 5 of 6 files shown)


orsinium (approved):

piptools/scripts/compile.py
```diff
@@ -326,6 +326,23 @@ def _determine_linesep(
...
+        file_okay=True,
+        dir_okay=False,
+        readable=True,
+        allow_dash=False,
+        path_type=str,
+    ),
+    multiple=True,
+    help=(
+        "Constrain versions using the given constraints file. "
+        "This option can be used multiple times."
```
AndydeCleyre, line 343: All the other `multiple=True` options are using the exact phrase "may be used more than once" 🤷🏼

tests/test_utils.py
```diff
@@ -634,6 +634,7 @@ def test_callback_config_file_defaults(pyproject_param, new_default, make_config
         ("trusted_host", "not-a-list"),
         ("annotate", "not-a-bool"),
         ("max_rounds", "not-an-int"),
+        ("constraint", "not-an-int"),
```
AndydeCleyre, line 637: int?

AndydeCleyre (approved):

piptools/scripts/compile.py
atugushev: Thanks! Addressed in d5e13f73201aa653835dd35c347b3f82011965fd.

atugushev:
Thanks for the review @AndydeCleyre 🙏🏻

honnix:
Thank you so much for adding this command line option! It seems only local file is supported. Is that correct? pip, however can use constraints file from any URL: https://pip.pypa.io/en/stable/user_guide/#constraints-files. Maybe that could be supported as well by pip-compile?

honnix:
Maybe it could be as simple as removing `click.Path` type? That would skip local file validation unfortunately.

honnix:
I proposed a change in #2038. We could go extra steps to validate local file and remote file, but it doesn't seem to be absolutely necessary.

PR #1941, opened 2023-07-29
Unshallow git clone on RTD build

### LINKED ISSUE #1939: Old 0.1 version on the latest docs
RTD built [latest](https://pip-tools.readthedocs.io/en/latest/) doc on b55abdc31e5a905389da26352a9beab2a6c87bfa commit using `0.1.dev128` version.

#### Expected result

There should be `v7.1.1.devN`

#### Actual result

Screenshot:
image

#### Additional context

Logs on [RTD](https://readthedocs.org/projects/pip-tools/builds/21423718/):
```
$ python -m sphinx -T -E -W --keep-going -b dirhtml ...
Running Sphinx v6.2.1
pip-tools version: 0.1.dev128
pip-tools release: 0.1.dev128
...
```

atugushev:
Fetch full history so that setuptools-scm could detect correct version.

For the context:
- https://github.com/pypa/setuptools_scm/issues/790
- https://docs.readthedocs.io/en/latest/build-customization.html#unshallow-git-clone

Fixes #1939.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -17,6 +17,9 @@ build:
   tools:
     python: >-
       3.11
+  jobs:
+    post_checkout:
+      - git fetch --unshallow || true
 
 python:
   install:


chrysle (approved):

webknjaz:
This is missing fetching the tags. I think there's something in `tox.ini` for that and it'd be wise to put this into a Sphinx extension to reduce duplication...

atugushev:
Apparently, It [fetches](https://readthedocs.org/projects/pip-tools/builds/21446450/) tags from the last 50 commits:
```
git fetch origin --force --prune --prune-tags --depth 50 ...
```

There's a possibility that 50 commits may not be enough, so explicitly fetching all tags might be a good idea.

webknjaz:
Exactly. This is a known issue that I've been facing since the times when Travis CI was still popular, which is why I fetch all of these.

chrysle:
> and it'd be wise to put this into a Sphinx extension to reduce duplication...

Could you tell me more? How and where would that custom extension come to use?

webknjaz:
It can be just a `setup()` function in `conf.py`. Though I realized that someone removed direct use of setuptools-scm, meaning that it must happen before Sphinx runs. So my idea would work only if we were to determine the version independently from packaging.

PR #1942, opened 2023-07-31
Fix non-unique heading anchors on changelog doc page

### LINKED ISSUE #1935: Non-unique heading anchors on pip-tools' changelog documentation

Currently, pip-tools generates non-unique anchors for links in the documentation, as seen for example here: https://pip-tools.readthedocs.io/en/latest/changelog/#id1. This will led to confusion with links for future versions deployed on RTD.

Generate unique anchors, for example https://pip-tools.readthedocs.io/en/latest/changelog/#710-2023-07-18

#### Alternative Solutions

N/A

#### Additional context

- Example on GitHub: https://github.com/jazzband/pip-tools/blob/main/CHANGELOG.md#700-2023-07-14
- https://myst-parser.readthedocs.io/en/stable/syntax/optional.html#anchor-slug-structure

atugushev:
Fixes #1935.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,11 +3,18 @@
 
 from __future__ import annotations
 
+import os
+import sys
 from importlib.metadata import version as get_version
 from pathlib import Path
 
+from sphinx.application import Sphinx
 from sphinx.util import logging
 from sphinx.util.console import bold
+
+# Add the docs/ directory to sys.path to be able to import utils
+docs_dir = os.path.dirname(os.path.dirname(__file__))
+sys.path.insert(0, docs_dir)
 
 logger = logging.getLogger(__name__)
 
@@ -57,3 +64,9 @@ linkcheck_ignore = [
 ]
 
 suppress_warnings = ["myst.xref_missing"]
+
+
+def setup(app: Sphinx) -> None:
+    from docs.utils import TransformSectionIdToName
+
+    app.add_transform(TransformSectionIdToName)
--- a/docs/utils.py
+++ b/docs/utils.py
@@ -1,0 +1,34 @@
+from __future__ import annotations
+
+import re
+from typing import Any
+
+from docutils import nodes
+from sphinx.transforms import SortIds, SphinxTransform
+
+numerical_id = re.compile(r"^id[0-9]+$")
+
+
+class TransformSectionIdToName(SphinxTransform):  # type: ignore[misc]
+    """Transforms section ids from <id1>, <id2>, ... to id <section-name>."""
+
+    default_priority = SortIds.default_priority + 1
+
+    def apply(self, **kwargs: Any) -> Any:
+        for node in self.document.findall(nodes.section):
+            if not self._has_numerical_id(node):
+                continue
+            self._transform_id_to_name(node)
+
+    def _has_numerical_id(self, node: nodes.section) -> bool:
+        node_ids = node["ids"]
+        if not node_ids:
+            return False
+
+        if len(node_ids) != 1:
+            return False
+
+        return bool(numerical_id.match(node_ids[0]))
+
+    def _transform_id_to_name(self, node: nodes.section) -> None:
+        node["ids"] = [nodes.make_id("id " + name) for name in node["names"]]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,7 @@ repos:
           - toml==0.10.2
           - pip==20.3.4
           - build==0.9.0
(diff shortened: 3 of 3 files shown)


docs/utils.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import annotations
```
webknjaz, line 1: Move this to the `_ext/` subir per convention.

docs/conf.py
```diff
@@ -57,3 +64,9 @@
 ]
 
 suppress_warnings = ["myst.xref_missing"]
+
+
+def setup(app: Sphinx) -> None:
```
webknjaz, line 69: Don't add `setup()` here but put it into an extension under `_ext/`.

Also, it should return a dict.

docs/conf.py
```diff
@@ -57,3 +64,9 @@
...
+    from docs.utils import TransformSectionIdToName
```
webknjaz, line 70: This is not needed, especially upon moving to the extension. Just add the extension to the list above and Sphinx will load it through the standard means.

docs/conf.py
```diff
@@ -3,12 +3,19 @@
...
+import sys
 from importlib.metadata import version as get_version
 from pathlib import Path
 
+from sphinx.application import Sphinx
 from sphinx.util import logging
 from sphinx.util.console import bold
 
+# Add the docs/ directory to sys.path to be able to import utils
+docs_dir = os.path.dirname(os.path.dirname(__file__))
```
webknjaz, line 16: Plz use pathlib and only convert to str on insert.

docs/conf.py
```diff
@@ -3,12 +3,19 @@
 
 from __future__ import annotations
 
+import os
+import sys
 from importlib.metadata import version as get_version
 from pathlib import Path
 
+from sphinx.application import Sphinx
```
webknjaz, line 11: Obviously, this won't be necessary upon moving to a proper extension.

docs/utils.py
```diff
@@ -0,0 +1,34 @@
...
+import re
+from typing import Any
+
+from docutils import nodes
+from sphinx.transforms import SortIds, SphinxTransform
+
+numerical_id = re.compile(r"^id[0-9]+$")
+
+
+class TransformSectionIdToName(SphinxTransform):  # type: ignore[misc]
```
webknjaz, line 12: What does this ignore?

docs/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    """Transforms section ids from <id1>, <id2>, ... to id <section-name>."""
+
+    default_priority = SortIds.default_priority + 1
+
+    def apply(self, **kwargs: Any) -> Any:
```
webknjaz, line 17: This returns none, not any.

docs/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def _has_numerical_id(self, node: nodes.section) -> bool:
+        node_ids = node["ids"]
+        if not node_ids:
+            return False
+
+        if len(node_ids) != 1:
+            return False
+
+        return bool(numerical_id.match(node_ids[0]))
```
webknjaz, line 31: This can be simplified:
```suggestion
        node_has_single_id = len(node_ids) == 1
        return node_has_single_id and
numerical_id.match(node_ids[0]) is not None
```

docs/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def _transform_id_to_name(self, node: nodes.section) -> None:
+        node["ids"] = [nodes.make_id("id " + name) for name in node["names"]]
```
webknjaz, line 34: String concatenation is slow.
```suggestion
        node["ids"] = [nodes.make_id(f"id {node_name}") for node_name in node["names"]]
```

docs/__init__.py
webknjaz, line 1: Don't add Python modules among the normal documents.

webknjaz (changes requested):
The idea is good, but the implementation needs polishing:
1) The proper thing to do here is to use the extension interface and not throw arbitrarily loaded modules in the docs dir. Use `_ext/` and make sphinx load it as an extension with it's own `setup()`.
2) There's several maintenance/functional improvements in the comments below.
3) Finally, let's try to drop the dates from anchors, as it'd be cleaner and more predictable to link releases. Links like https://cheroot.cherrypy.dev/en/latest/history/#v11-0-0 and https://setuptools.pypa.io/en/latest/history.html#v68-0-0 are nicer and can be constructed by guessing from the release versions/repo tags.

Issue #1937, opened 2023-07-25
Update compatibility matrix


astrojuanlu:
#### What's the problem this feature will solve?

After gh-1879, I don't think https://github.com/jazzband/pip-tools#versions-and-compatibility was updated.

PR #1950, opened 2023-08-04
Remove outdated compatibility matrix from `README`

jeffwidman:
As noted in https://github.com/jazzband/pip-tools/issues/1937, this matrix is out of date now that https://github.com/jazzband/pip-tools/releases/tag/7.0.0 has shipped.

Technically it could be kept around and add some more rows, but I don't think it's worth the effort... it will always be going out of date. Instead, let the internal dep constraints and classifiers within `setup.py`/`pyproject.toml` constrain what versions of `python` / `pip` are compatible with that version of `pip-tools`. Most users won't care, but those who will can access that data manually. And otherwise it'll all be smoothly resolved by package managers.

So let's just remove it altogether.

Fix: https://github.com/jazzband/pip-tools/issues/1937

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -576,23 +576,6 @@ backtracking resolver is more robust, but can take longer to run in general.
 You can continue using the legacy resolver with `--resolver=legacy` although
 note that it is deprecated and will be removed in a future release.
 
-### Versions and compatibility
-
-The table below summarizes the latest `pip-tools` versions with the required
-`pip` and Python versions. Generally, `pip-tools` supports the same Python
-versions as the required `pip` versions.
-
-| pip-tools      | pip            | Python         |
-| -------------- | -------------- | -------------- |
-| 4.5.\*         | 8.1.3 - 20.0.2 | 2.7, 3.5 - 3.8 |
-| 5.0.0 - 5.3.0  | 20.0 - 20.1.1  | 2.7, 3.5 - 3.8 |
-| 5.4.0          | 20.1 - 20.3.\* | 2.7, 3.5 - 3.8 |
-| 5.5.0          | 20.1 - 20.3.\* | 2.7, 3.5 - 3.9 |
-| 6.0.0 - 6.3.1  | 20.3 - 21.2.\* | 3.6 - 3.9      |
-| 6.4.0          | 21.2 - 21.3.\* | 3.6 - 3.10     |
-| 6.5.0 - 6.10.0 | 21.2 - 22.3.\* | 3.7 - 3.11     |
-| 6.11.0+        | 22.2+          | 3.7 - 3.11     |
-
 [jazzband]: https://jazzband.co/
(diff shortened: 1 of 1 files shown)


chrysle:
> Instead, let the internal dep constraints within `setup.py`/`pyproject.toml` constrain what versions of `python` / `pip` are compatible with that version of `pip-tools`. Most users won't care, but those who will can access that data manually.

Though this only provides constraints for the current version, throwing away the carefully curated information about historical versions.

At least there should be a general note about compatibility.

jeffwidman:
I don't follow... Were these dep / classifier versions not pinned in previous releases? It's not hard to look at the contents of a file for a particular release tag...

My point was that very few users care about this information and for those who do it's not hard to retrieve that data.

If they weren't pinned in the past, well, then yeah might be tough to find without trial and error.

Regardless, now that 3.6 (and now even 3.7) is eol'd and I see more and more packages dropping support for them, won't be too long until there won't be much demand for a package manager that supports older python versions simply because there aren't package updates available.

chrysle:
> I don't follow... Were these dep / classifier versions not pinned in previous releases? It's not hard to look at the contents of a file for a particular release tag...

Well yes, that's true. Hadn't thought of the possibility of digging through the tags.

Then maybe could you keep the introduction of that section, but remove the table and say that required `pip` versions can be determined via `pyproject.toml`?

jeffwidman:
Cool, yeah, some days it feels like I spend hours doing `git bisect` + `git blame` followed by 5 mins to make a 1 line change. 😁 

I personally don't think that section has enough value to merit being on the `README`... 

Users won't care until they run into a problem, in which case they'll google / stackoverflow / start looking at `setup.py` / `pyproject.toml` etc. And then they'll probably be told they need to bump to a newer version of `python` or `pip`, which IMO in most situations is the better solution.

And when it's not, generally they (or whoever is answering their question on StackOverflow) will tell them to look at `pyproject.toml`.

But I just don't see them at any point in this process going to the readme to say "How do I find the supported versions?"

Just my $0.02, no problem if you disagree and prefer to solve differently, I'm a drive-by contributor, not a maintainer here.

chrysle (approved):
Let's do it then! Always happy to find an agreement.

atugushev (approved):
For the last 3 years I've never seen someone use this table.

atugushev:
Thanks @jeffwidman and @chrysle!

chrysle:
> For the last 3 years I've never seen someone use this table.

How can you possibly _see_ that? ;-)

atugushev:
> How can you possibly _see_ that? ;-)

Assumption is based on the number of times I've referred users to this section.

jeffwidman:
>> For the last 3 years I've never seen someone use this table.

> How can you possibly see that? ;-)

It's @atugushev. He sees things that are impossible for normal people to see:

image

PR #1953, opened 2023-08-05
Remove extras from compiled files

### LINKED ISSUE #1613: Keep or remove extras in compiled files?
# Issue

I am opening this issue to discuss if the extras should be kept or removed in the compiled files.

@atugushev made a [good summary](https://github.com/jazzband/pip-tools/pull/1608#issuecomment-1089920834) of the situation:

> Currently, we have direct references without extras and pinned packages with extras in requirements.txt, which looks wrong and should be synced in some single way.

We would like to get feedback from the community, do you think we should keep them or remove them and why?

My opinion is that we should remove them, since `pip-compile` is already listing all packages needed for a project, it seems redundant to me to specify it twice (one time in the extra and one time as a top-level dependency). Also, it's in theory possible to install more packages than those specified in the requirements.txt via the extras. In my opinion the generated .txt file of `pip-compile` should act as a lock file. The only advantage I can see is that we can easily inspect which dependencies are using extras.

@AndydeCleyre said that [the order of installation could matter in some cases](https://github.com/jazzband/pip-tools/pull/1608#issuecomment-1087044018) such as `GDAL`  which requires `numpy` to be installed first. I checked if having the extra (`gdal[numpy]`) in the .txt file was making a difference, and I found that it was not working. You can read [this gist](https://gist.github.com/FlorentJeannot/3f1badadb609efabe567d8848c28d53b) if you want to have a look at the tests I've done (there's a conclusion at the end if you don't want to read it all).

# Links

Some links about the discussion around this:

- [PR to remove extras](https://github.com/jazzband/pip-tools/pull/1608)
- [PR to add extras in direct references](https://github.com/jazzband/pip-tools/pull/1582)
- [PR about the direct reference without extras](https://github.com/jazzband/pip-tools/pull/1453#discussion_r671595549)

# Samples from dependency management tools

The goal is to show you the output of different management tools when the project specifies extras. This may help you make a decision on the issue.

For each tool, I installed `gdal[numpy]==3.2.2`.

## pip

Command: `pip freeze > requirements.txt`

The file contains:

``` 
GDAL==3.2.2
numpy==1.22.3
```

## pip-tools

Command: `pip-compile`

The file contains:

```
gdal[numpy]==3.2.2
    # via -r requirements.in
numpy==1.22.3
    # via gdal
```

## Pipenv

Command: `pipenv lock -r > requirements.txt`

The file contains:

```
-i https://pypi.org/simple
gdal[numpy]==3.2.2
numpy==1.22.3
```

## Poetry

Command: `poetry export -o requirements.txt`

The file contains:

```
gdal==3.2.2
numpy==1.22.3; python_version >= "3.8"
```

# Pros and cons

I'll try to collect all your feedbacks to update these lists.

Pros

- We can clearly see which dependencies are using extras (FlorentJeannot) 

Cons

- In theory, it's possible that `pip-sync` or `pip` could install more packages than what is listed in a .txt file because of extras. I think the output of `pip-compile` should act as a lock file, so it should only install what's specified in the .txt file. (FlorentJeannot) 
- It's redundant. Packages specified in the extras are also in the top-level dependencies. (FlorentJeannot)

atugushev:
Resolves #1613.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -567,6 +567,7 @@ This section lists `pip-tools` features that are currently deprecated.
   to pass the `--allow-unsafe` now to adapt to the upcoming change.
 - The legacy resolver is deprecated and will be removed in future versions.
   The new default is `--resolver=backtracking`.
+- `--strip-extras` is deprecated. It has no effect and will be removed in future versions.
 
 ### A Note on Resolvers
 
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -441,6 +441,12 @@ def cli(
             " in future versions of pip-tools."
         )
 
+    if strip_extras:
+        log.warning(
+            "WARNING: --strip-extras is deprecated. It has no effect and "
+            "will be removed in future versions of pip-tools."
+        )
+
     ###
     # Setup
     ###
@@ -689,7 +695,6 @@ def cli(
         emit_trusted_host=emit_trusted_host,
         annotate=annotate,
         annotation_style=annotation_style,
-        strip_extras=strip_extras,
         generate_hashes=generate_hashes,
         default_index_url=repository.DEFAULT_INDEX_URL,
         index_urls=repository.finder.index_urls,
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -84,7 +84,6 @@ class OutputWriter:
         emit_trusted_host: bool,
         annotate: bool,
         annotation_style: str,
-        strip_extras: bool,
         generate_hashes: bool,
         default_index_url: str,
         index_urls: Iterable[str],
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -567,6 +567,7 @@ This section lists `pip-tools` features that are currently deprecated.
   to pass the `--allow-unsafe` now to adapt to the upcoming change.
 - The legacy resolver is deprecated and will be removed in future versions.
   The new default is `--resolver=backtracking`.
+- `--strip-extras` is deprecated. It has no effect and will be removed in future versions.
```
ryanhiebert, line 570: As noted on the relevant issue, `--strip-extras` does have an effect, even with the new default resolver.

ryanhiebert (changes requested):
It will be a breaking change to change the default. We should do it at some point, but if we're going to add `--include-extras`, and I think we should based on prior feedback, we should add that argument to a minor version bump to anticipate this breaking change, and to allow people to silence the warning of the breaking change of the default.

atugushev:
@ryanhiebert, thanks for the review! I'll close the PR based on this https://github.com/jazzband/pip-tools/issues/1613#issuecomment-1666601099.

Issue #1957, opened 2023-08-08
Generate all requirements files only using pyproject.toml as input


aaronsgithub:
#### What's the problem this feature will solve?

Perhaps this is already possible with the existing functionality of pip-tools and I need to go to timeout.
Or if there are workarounds which would allow me to get close to the situation described below, I would be grateful if someone could share details with me.

I want to use `pyproject.toml` as the sole configuration file in a python project, and use `pip-compile` to generate all the necessary pinned requirements files with unpinned dependencies specified in `pyproject.toml` instead of `requirements.in`.

For argument sake, lets say those requirements files we need are:

- `requirements.txt` all requirements needed to run the application
- `requirements-test.txt` contains **only** dependencies needed for integration tests
- `requirements-dev.txt` superset of the both above files, with additional development dependencies.

Then `pip install` should be able to infer from the `pyproject.toml` file that it has use:

- `requirements.txt` if `pip install .`
- `requirements-dev.txt` if `pip install .[dev]`
- `requirements-test.txt` if `pip install .[test]`

Here is an example `pyproject.toml` with hypothetical `tool.pip-compile` sections:

```
[build-system]
requires = ["setuptools>=40.8.0", "wheel"]
build-backend = "setuptools.build_meta"

[project]
requires-python = ">=3.10"
name = "myproject"
version = "0.1.0dev1"
dynamic = ["dependencies", "optional-dependencies"]

[tool.pip-compile]
dependencies = [
    "alembic",
    "fastapi",
    "pydantic[email]",
    "sqlalchemy",
]

[tool.pip-compile.optional-dependencies]
dev = [
        "black",
        "pip-tools",
]
test = [
        "pytest",
        "pytest-playwright",
]   

[tool.setuptools.dynamic]
dependencies = { file = ["./requirements/requirements.txt"] }
optional-dependencies.dev = { file = ["./requirements/requirements-dev.txt"] }
optional-dependencies.test = { file = ["./requirements/requirements-test.txt"] }
```

And here is how the requirements files could be generated using `pip-compile`:
```
    pip-compile -o requirements/requirements.txt; # includes base deps only
    pip-compile --deps=test -o requirements/requirements-test.txt; # includes test deps only
    pip-compile --extra=dev --extra=test -o requirements/requirements-dev.txt; # includes base + test + dev deps
```
[...]

PR #1681, opened 2022-09-18
Add options for including build dependencies in compiled output

apljungquist:
Fixes jazzband/pip-tools#1396. Add options for including build dependencies in the output from pip-compile. This allows users to improve the reproducibility of their CI since they can ensure that the same versions of build dependencies are used to build their package every time.

Note that build dependencies of build dependencies are not included in the output. This is one reason why builds will still not be completely reproducible.

##### Contributor checklist

- [x] Provided the tests for the changes.
- [x] Assure PR title is short, clear, and good to be included in the user-oriented changelog

##### Maintainer checklist

- [x] Assure one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #1960, opened 2023-08-09
Add `--only-extra` option to `pip-compile`

### LINKED ISSUE #1959: Add support for `--only-extra`
Currently, pip-tools can't compile only extra dependencies. Wouls be nice to have a new option `--only-extra`, for example:

```bash
# includes test deps only
pip-compile --only-extra=test -c requirements.txt -o test-requirements.txt
```

Note `-c requirements.txt` so that common sub-dependencies are in sync with `requirements.txt`.

_Originally posted by @atugushev in https://github.com/jazzband/pip-tools/issues/1957#issuecomment-1669985803_

atugushev:
Resolves #1959.
Refs #1957.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import itertools
 import os
 import shlex
 import sys
@@ -82,6 +81,7 @@ def _determine_linesep(
 @options.rebuild
 @options.extra
 @options.all_extras
+@options.only_extra
 @options.find_links
 @options.index_url
 @options.no_index
@@ -123,6 +123,7 @@ def cli(
     rebuild: bool,
     extras: tuple[str, ...],
     all_extras: bool,
+    only_extras: tuple[str, ...],
     find_links: tuple[str, ...],
     index_url: str,
     no_index: bool,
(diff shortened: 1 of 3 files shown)


webknjaz:
I'm not sure if it is reasonable to support this. It's out of standard and encourages people to misuse [`Provides-Extra`](https://packaging.python.org/en/latest/specifications/core-metadata/#provides-extra-multiple-use) by exposing them to their end-users.

Extras are a public API for installing additional runtime dependencies together with the mandatory ones. All the installers, like pip, always include normal deps. Advertising this to people would cause confusion, especially, when used with `pip install thing[extra] -c constraints.txt` and `--generate-hashes`.

This PR should at least have a huge disclaimer explaining https://github.com/jazzband/pip-tools/pull/1681#issuecomment-1666966545 bundled in the docs and maybe the help string. And maybe don't call it `--only-extra` — having something more explicit like `--only-extract-extra` to signify that it's extracted as a standalone dependency source.

webknjaz:
Oh, and maybe hold this until #1681 is merged? I don't want to keep introducing conflicts for the contributor all the time…

atugushev:
I don't want to encourage bad practices, so I'll close this. If someone finds this feature sensible, please feel free to reopen the issue. Thanks for the feedback!

webknjaz:
The dependency groups PEP is going to be the correct solution so it should be implemented instead. I think it'd okay to do this even before it's fully accepted.

PR #1963, opened 2023-08-10
Speed up tests execution

atugushev:
#### Session-scoped fixtures

The `fake_dists` fixture is one of the most time-consuming fixtures, so it makes sense to make it and all related fixtures session-scoped.

#### Finetune tests distribution over the run

Initially, tests are distributed evenly across all worker processes. But during the execution of tests, some processes may finish running tests faster than others. In these cases, the `--dist=worksteal` option allows these "free" workers to "steal" tests from the queue of other, still working, processes.

This is particularly useful if the run times of different tests vary greatly, as this distribution method ensures more efficient resource usage, improving the performance of testing.

#### Refactored slow tests

`test_generate_hashes_with_split_style_annotations` and `test_generate_hashes_with_line_style_annotations` were utilizing too many external packages. We should always prefer local packages as they are faster and pip-compile doesn't need network to compile them.

#### Stats

Locally, it increases speed by about 30%. Here are CI runs:

Screenshot 2023-08-10 at 21 47 56

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,6 +99,7 @@ disallow_untyped_defs = false
 addopts = [
   # `pytest-xdist`:
   "--numprocesses=auto",
+  "--dist=worksteal",
 
   # Show 20 slowest invocations:
   "--durations=20",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -286,8 +286,8 @@ def pip_with_index_conf(make_pip_conf):
     )
 
 
-@pytest.fixture
-def make_package(tmp_path):
+@pytest.fixture(scope="session")
+def make_package(tmp_path_factory):
     """
     Make a package from a given name, version and list of required packages.
     """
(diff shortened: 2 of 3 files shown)


theryanwalker (approved):
Looks great!

atugushev:
@theryanwalker https://github.com/jazzband/pip-tools/pull/1967 should fix failing CI.

pyproject.toml
```diff
@@ -99,6 +99,7 @@ disallow_untyped_defs = false
 addopts = [
   # `pytest-xdist`:
   "--numprocesses=auto",
+  "--dist=worksteal",
```
webknjaz, line 102: Such settings would benefit from having an inline code comment with the motivation.

webknjaz:
@atugushev one other thing I used to do in the past, was splitting the tests across multiple CI jobs so that it adds parallelism in an extra dimension. We could consider doing the same here.

pyproject.toml
atugushev: Thanks for the suggestion! Addressed in https://github.com/jazzband/pip-tools/pull/1973

atugushev:
> one other thing I used to do in the past, was splitting the tests across multiple CI jobs so that it adds parallelism in an extra dimension. We could consider doing the same here.

@webknjaz, that's an interesting idea. Do you have any suggestions in mind regarding pip-tools?

webknjaz:
Just a thought. It can be implemented as a pytest option in a few lines of code. But it might be not worth it, depending on the overhead of spawning more VMs. This really needs testing to see if it makes things better or worse.

PR #1966, opened 2023-08-14
Add support for command-specific configuration sections

chrysle:
Addresses https://github.com/jazzband/pip-tools/pull/1933#issuecomment-1653068755

This introduces configuration sections specific to `pip-compile` and `pip-sync`.

I'd like to refactor the `make_config_file()` pytest fixture in a follow-up PR, to introduce tests for this feature using multiple configuration sections.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,7 @@ By default, both `pip-compile` and `pip-sync` will look first
 for a `.pip-tools.toml` file and then in your `pyproject.toml`. You can
 also specify an alternate TOML configuration file with the `--config` option.
 
+It is possible to specify configuration values both globally and tool-specific.
 For example, to by default generate `pip` hashes in the resulting
 requirements file output, you can specify in a configuration file:
 
@@ -309,6 +310,27 @@ so the above could also be specified in this format:
 ```toml
 [tool.pip-tools]
 generate_hashes = true
+```
+
+Configuration defaults specific to `pip-compile` and `pip-sync` can be put beneath
+separate sections. For example, to by default perform a dry-run with `pip-compile`:
+
+```toml
+[tool.pip-compile]
+dry-run = true
+```
+
+This does not affect the `pip-sync` command, which also has a `--dry-run` option.
+Note that tool-specific configuration overrides global settings, thus this would
+also make `pip-compile` generate hashes, but discard the global dry-run setting:
+
+```toml
+[tool.pip-tools]
+generate-hashes = true
+dry-run = true
+
+[tool.pip-compile]
+dry-run = false
 ```
 
 You might be wrapping the `pip-compile` command in another script. To avoid
--- a/README.md
+++ b/README.md
@@ -321,8 +321,8 @@ dry-run = true
 ```
 
 This does not affect the `pip-sync` command, which also has a `--dry-run` option.
-Note that local settings take preference over the global ones of the same name, 
-whenever both are declared, thus this would also make `pip-compile` generate hashes, 
+Note that local settings take preference over the global ones of the same name,
+whenever both are declared, thus this would also make `pip-compile` generate hashes,
 but discard the global dry-run setting:
 
 ```toml
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ By default, both `pip-compile` and `pip-sync` will look first
 for a `.pip-tools.toml` file and then in your `pyproject.toml`. You can
 also specify an alternate TOML configuration file with the `--config` option.
 
-It is possible to specify configuration values both globally and tool-specific.
+It is possible to specify configuration values both globally and command-specific.
 For example, to by default generate `pip` hashes in the resulting
 requirements file output, you can specify in a configuration file:
 
@@ -316,20 +316,21 @@ Configuration defaults specific to `pip-compile` and `pip-sync` can be put benea
 separate sections. For example, to by default perform a dry-run with `pip-compile`:
 
 ```toml
-[tool.pip-compile]
+[tool.pip-tools.compile] # "sync" for pip-sync
 dry-run = true
 ```
 
 This does not affect the `pip-sync` command, which also has a `--dry-run` option.
-Note that tool-specific configuration overrides global settings, thus this would
-also make `pip-compile` generate hashes, but discard the global dry-run setting:
+Note that local settings take preference over the global ones of the same name, 
+whenever both are declared, thus this would also make `pip-compile` generate hashes, 
+but discard the global dry-run setting:
 
 ```toml
 [tool.pip-tools]
 generate-hashes = true
 dry-run = true
 
-[tool.pip-compile]
+[tool.pip-tools.compile]
 dry-run = false
 ```
 
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -682,8 +682,8 @@ def parse_config_file(
     # `[tool.pip-tools.compile]` or `[tool.pip-tools.sync]`
     piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})
 
-    # TODO: Replace with `str.removeprefix()` once dropped 3.8
     assert click_context.command.name is not None
+    # TODO: Replace with `str.removeprefix()` once dropped 3.8
     config_section_name = click_context.command.name[len("pip-") :]
 
     piptools_config.update(piptools_config.pop(config_section_name, {}))
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -546,9 +546,7 @@ def override_defaults_from_config_file(
     else:
         config_file = Path(value)
 
-    config = parse_config_file(
-        ctx, config_file
-    )
+    config = parse_config_file(ctx, config_file)
 
     _validate_config(ctx, config)
     _assign_config_to_cli_context(ctx, config)
@@ -683,26 +681,29 @@ def parse_config_file(
     # In a TOML file, we expect the config to be under `[tool.pip-tools]`,
     # `[tool.pip-compile]` or `[tool.pip-sync]`
     piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})
-    pipcompile_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {}).get("compile", {})
-    pipsync_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {}).get("sync", {})
+    pipcompile_config: dict[str, Any] = (
+        config.get("tool", {}).get("pip-tools", {}).get("compile", {})
+    )
+    pipsync_config: dict[str, Any] = (
+        config.get("tool", {}).get("pip-tools", {}).get("sync", {})
+    )
 
     config = piptools_config
-    
+
     if click_context.command.name == "pip-compile":
-        config.pop("compile")
-        config.update(pipcompile_config)
+        if pipcompile_config:
+            config.pop("compile")
+            config.update(pipcompile_config)
     elif click_context.command.name == "pip-sync":
-        config.pop("sync")
-        config.update(pipsync_config)
+        if pipsync_config:
+            config.pop("sync")
+            config.update(pipsync_config)
 
-    print(config)
-
-    if config:
-        config = _normalize_keys_in_config(config)
-        config = _invert_negative_bool_options_in_config(
-            ctx=click_context,
-            config=config,
-        )
+    config = _normalize_keys_in_config(config)
+    config = _invert_negative_bool_options_in_config(
+        ctx=click_context,
+        config=config,
+    )
 
     return config
 
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -684,7 +684,7 @@ def parse_config_file(
 
     # TODO: Replace with `str.removeprefix()` once dropped 3.8
     assert click_context.command.name is not None
-    config_section_name = click_context.command.name[len("pip-"):]
+    config_section_name = click_context.command.name[len("pip-") :]
 
     piptools_config.update(piptools_config.pop(config_section_name, {}))
     piptools_config.pop("compile", {})
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -679,33 +679,24 @@ def parse_config_file(
         )
 
     # In a TOML file, we expect the config to be under `[tool.pip-tools]`,
-    # `[tool.pip-compile]` or `[tool.pip-sync]`
+    # `[tool.pip-tools.compile]` or `[tool.pip-tools.sync]`
     piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})
-    pipcompile_config: dict[str, Any] = (
-        config.get("tool", {}).get("pip-tools", {}).get("compile", {})
-    )
-    pipsync_config: dict[str, Any] = (
-        config.get("tool", {}).get("pip-tools", {}).get("sync", {})
-    )
 
-    config = piptools_config
+    # TODO: Replace with `str.removeprefix()` once dropped 3.8
+    assert click_context.command.name is not None
+    config_section_name = click_context.command.name[len("pip-"):]
 
-    if click_context.command.name == "pip-compile":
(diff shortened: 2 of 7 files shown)


webknjaz:
Could you adjust the PR title to match the description? Currently it reads as if you're adding features but the text suggests that it's tests.

chrysle:
Thanks, it's actually **this** pull request I'd like to add more tests for (codecov is failing), but that needs some refactoring and doesn't really fit in. I've adjusted the description.

webknjaz:
Ah, though I still don't understand what's tool-specific configuration. That title is just to generic to mean anything to me. It's quite unspecific wrt what it refers to.

chrysle:
>  We already have pip-tools specific config, after all. That's the tool, right?

Yeah, but that is the _global_ configuration (for both `pip-compile` and `pip-sync`). I added the possibility to specify configuration values _only for_ those commands. I added some docs to the README.

Is the title better now?

webknjaz:
> Is the title better now?

Much better, thanks! Titles shows up in email notifications among other places and tend to be a factor in deciding whether to even look inside. This is especially important when inbox is overflowing and people need to prioritize + figure out which notifications to discard. An accurate title helps draw attention of people who might be affected by the patch.

README.md
```diff
@@ -290,6 +290,7 @@ By default, both `pip-compile` and `pip-sync` will look first
 for a `.pip-tools.toml` file and then in your `pyproject.toml`. You can
 also specify an alternate TOML configuration file with the `--config` option.
 
+It is possible to specify configuration values both globally and tool-specific.
```
webknjaz, line 293: Let's avoid saying tool-specific since the document doesn't record what is a "tool" in different contexts and people may be getting different incorrect ideas.
"command" is probably fine. Could also be a sub-command, I suppose. Let's see if others have better ideas, though...

README.md
```diff
@@ -311,6 +312,27 @@ so the above could also be specified in this format:
...
+Configuration defaults specific to `pip-compile` and `pip-sync` can be put beneath
+separate sections. For example, to by default perform a dry-run with `pip-compile`:
+
+```toml
+[tool.pip-compile]
+dry-run = true
+```
+
+This does not affect the `pip-sync` command, which also has a `--dry-run` option.
+Note that tool-specific configuration overrides global settings, thus this would
```
webknjaz, line 324: Let's be more specific than "overrides". When I read this, I thought that the entire global section is discarded. We need to choose wording to highlight that the sections are merged and individual specific settings supersede / take precedence over the global ones.

piptools/utils.py
```diff
@@ -570,10 +570,21 @@ def override_defaults_from_config_file(
     else:
         config_file = Path(value)
 
-    config = parse_config_file(ctx, config_file)
-    if config:
-        _validate_config(ctx, config)
-        _assign_config_to_cli_context(ctx, config)
+    piptools_config, pipcompile_config, pipsync_config = parse_config_file(
```
webknjaz, line 573: I don't think it's a good idea to return a tuple. If it is really necessary to have all these bits of data, wrap them into an object (dataclass?) and return that.

Though, I'd argue it should return a single config instead. Maybe, we need to pass an argument telling it for which command that config is to be extracted..

piptools/utils.py
```diff
@@ -702,14 +713,24 @@ def parse_config_file(
             hint=f"Could not parse '{config_file !s}': {value_err !s}",
         )
 
-    # In a TOML file, we expect the config to be under `[tool.pip-tools]`
+    # In a TOML file, we expect the config to be under `[tool.pip-tools]`,
+    # `[tool.pip-compile]` or `[tool.pip-sync]`
```
webknjaz, line 717: Traditionally, everything concerning a project, should be a part of the same namespaces. We can use subsections for this:

```suggestion
    # `[tool.pip-tools.pip-compile]` or `[tool.pip-tools.pip-sync]`
```

tests/test_cli_sync.py
```diff
@@ -450,3 +450,16 @@ def test_allow_in_config_pip_compile_option(run, runner, tmp_path, make_config_f
 
     assert out.exit_code == 0
     assert "Using pip-tools configuration defaults found" in out.stderr
+
+
+@mock.patch("piptools.sync.run")
```
webknjaz, line 455: Don't use `mock` directly. In pytest, use the `mocker` fixture or `monkeypatch` if possible.

webknjaz (commented):
Initial review. I foresee the need to refactor things here and avoid complicating things.

tests/test_cli_sync.py
chrysle: I've adapted the test style to the rest of the file.

piptools/utils.py
chrysle: Though it's pretty tedious for the user to type this..

piptools/utils.py
chrysle: Yes, I can move the test below to `parse_config_file()`.

README.md
chrysle: How about `conflicting settings in the global section get overriden by the local ones`?

piptools/utils.py
theryanwalker: "Namespaces are one honking great idea". Wouldn't it be more appropriate to use `[tool.pip-tools.compile]`, similar to `python -m piptools.compile`?

piptools/utils.py
theryanwalker:
```suggestion
    config = Config.parse_file(config_file, ctx)
```

Given that you're in the process of refactoring this, how about introducing a `Config` class in `piptools/config.py` and moving the related logic there? This is an example of what the Config might look like:

```python
@dataclass
class Config:
    pip_tools: dict[str, Any]
    pip_compile: dict[str, Any]
    pip_sync: dict[str, Any]
    
    @classmethod
    def parse_file(filename):
        ...

    def validate(self):
        ...
        
    def normalize_keys(self):
        ...
    
    def invert_negative_bool_options(self, ctx):
        ...
        
    def get_default_map() -> dict[str, Any]:
        ...
     
    def assign_to_cli(ctx):
        ...    
```

chrysle: @atugushev What do you think, should we dedicate an own file to that?

atugushev: Sounds great 👍🏻

piptools/utils.py
webknjaz: @theryanwalker fair. That's a better suggestion.

```suggestion
    # `[tool.pip-tools.compile]` or `[tool.pip-tools.sync]`
```

README.md
webknjaz: "The local settings take preference over the global ones of the same name, whenever both are declared"?

piptools/utils.py
webknjaz: @theryanwalker I like it, though I'd probably go for a different construction/classmethod name. How about `Config.from_file_path()`?

piptools/utils.py
```diff
@@ -570,10 +570,21 @@ def override_defaults_from_config_file(
...
-    if config:
-        _validate_config(ctx, config)
-        _assign_config_to_cli_context(ctx, config)
+    piptools_config, pipcompile_config, pipsync_config = parse_config_file(
+        ctx, config_file
+    )
+
+    configs = [piptools_config]
+
+    if ctx.command.name == "pip-compile":
```
webknjaz, line 579: Is it still `"pip-compile"` when invoked using `python -m piptools compile`?

piptools/utils.py
```diff
@@ -570,10 +570,21 @@ def override_defaults_from_config_file(
...
+        configs.append(pipcompile_config)
+    elif ctx.command.name == "pip-sync":
+        configs.append(pipsync_config)
+
+    for config in configs:
+        if config:
+            _validate_config(ctx, config)
+            _assign_config_to_cli_context(ctx, config)
```
webknjaz, line 587: Let's dedent this a bit:
```suggestion
    for config in filter(bool, configs):
        _validate_config(ctx, config)
        _assign_config_to_cli_context(ctx, config)
```

webknjaz:
@chrysle looks like this also needs to be rebased due to conflicts in tests.

tests/test_cli_sync.py
webknjaz: Okay, then this is out of the scope of this PR. Marking the thread as resolved.

piptools/utils.py
chrysle: Yes, verified.

README.md
chrysle: Reworded to "command".

piptools/utils.py
webknjaz: @chrysle looks like you forgot to apply the suggested change to the comment line, so I'm reopening this.

piptools/utils.py
```diff
@@ -678,14 +678,34 @@ def parse_config_file(
...
+    # `[tool.pip-compile]` or `[tool.pip-sync]`
     piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})
-    piptools_config = _normalize_keys_in_config(piptools_config)
-    piptools_config = _invert_negative_bool_options_in_config(
+    pipcompile_config: dict[str, Any] = (
+        config.get("tool", {}).get("pip-tools", {}).get("compile", {})
+    )
+    pipsync_config: dict[str, Any] = (
+        config.get("tool", {}).get("pip-tools", {}).get("sync", {})
+    )
```
webknjaz, line 689: No need to double-lookup the dicts. The entire section is already in the var:
```suggestion
    pipcompile_config: dict[str, Any] = piptools_config.get("compile", {})
    pipsync_config: dict[str, Any] = piptools_config.get("sync", {})
```

piptools/utils.py
```diff
@@ -678,14 +678,34 @@ def parse_config_file(
...
+
+    config = piptools_config
+
+    if click_context.command.name == "pip-compile":
+        if pipcompile_config:
+            config.pop("compile")
```
webknjaz, line 695: I would think you'd want to remove the other one…

piptools/utils.py
```diff
@@ -678,14 +678,34 @@ def parse_config_file(
...
+            config.update(pipcompile_config)
+    elif click_context.command.name == "pip-sync":
+        if pipsync_config:
+            config.pop("sync")
+            config.update(pipsync_config)
```
webknjaz, line 700: This can probably be simplified further:
```suggestion
    cmd_to_section_map = {
        "pip-compile": "compile",
        "pip-sync": "sync",
    }
    config_section_name = cmd_to_section_map[click_context.command.name]
    piptools_config.update(piptools_config.pop(config_section_name, {}))
    piptools_config.pop("compile", {})
    piptools_config.pop("sync", {})
```
The above assumes we don't want those subsections in the object we'll be passing around...

webknjaz: The map isn't really needed, though:
```suggestion
    config_section_name = click_context.command.name.removeprefix("pip-")
    piptools_config.update(piptools_config.pop(config_section_name, {}))
    piptools_config.pop("compile", {})
    piptools_config.pop("sync", {})
```

webknjaz: This might be easier to grasp:
```suggestion
    config_section_name = click_context.command.name.removeprefix("pip-")
    piptools_config = {
        **piptools_config,
        "compile": {},
        "sync": {},
        **piptools_config.pop(config_section_name, {}),
    }

```

piptools/utils.py
```diff
@@ -678,14 +678,34 @@ def parse_config_file(
...
+
+    config = _normalize_keys_in_config(config)
```
webknjaz, line 702: No need to rename the variable really.

tests/conftest.py
```diff
@@ -501,14 +501,22 @@ def _maker(
...
         # Make a config file with this one config default override
         config_file = tmpdir_cwd / config_file_name
-        config_to_dump = {"tool": {"pip-tools": {pyproject_param: new_default}}}
+
+        if subsection:
+            config_to_dump = {
+                "tool": {section: {subsection: {pyproject_param: new_default}}}
+            }
+        else:
+            config_to_dump = {"tool": {section: {pyproject_param: new_default}}}
```
webknjaz, line 519: How about
```suggestion
        nested_config = {pyproject_param: new_default}
        if subsection:
            nested_config = {subsection: nested_config}
        config_to_dump = {"tool": {section: nested_config}}
```

piptools/utils.py
```diff
@@ -678,13 +678,24 @@ def parse_config_file(
...
         )
 
-    # In a TOML file, we expect the config to be under `[tool.pip-tools]`
+    # In a TOML file, we expect the config to be under `[tool.pip-tools]`,
+    # `[tool.pip-tools.compile]` or `[tool.pip-tools.sync]`
     piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})
+
+    # TODO: Replace with `str.removeprefix()` once dropped 3.8
+    assert click_context.command.name is not None
+    config_section_name = click_context.command.name.replace("pip-", "")
```
webknjaz, line 687: I think this
```suggestion
    config_section_name = click_context.command.name[len("pip-"):]
```
would be a more accurate representation of what `str.removeprefix()` does.

tests/test_cli_compile.py
```diff
@@ -3562,3 +3562,20 @@ def test_origin_of_extra_requirement_not_written_to_annotations(
...
     )
+
+
+def test_tool_specific_config_option(pip_conf, runner, tmp_path, make_config_file):
+    config_file = make_config_file(
+        "dry-run", True, section="pip-tools", subsection="compile"
+    )
+
+    with open(config_file.as_posix()) as f:
+        print(f.read())
```
webknjaz, line 3573: Is this some leftover debug output?

chrysle:
Thank you for your suggestions and sorry for the long delay.
My code style is simply terrible... My new year's resolution definitely includes paying more attention in computer science class.

piptools/utils.py
```diff
@@ -678,13 +678,24 @@ def parse_config_file(
             hint=f"Could not parse '{config_file !s}': {value_err !s}",
         )
 
-    # In a TOML file, we expect the config to be under `[tool.pip-tools]`
+    # In a TOML file, we expect the config to be under `[tool.pip-tools]`,
+    # `[tool.pip-tools.compile]` or `[tool.pip-tools.sync]`
     piptools_config: dict[str, Any] = config.get("tool", {}).get("pip-tools", {})
+
+    # TODO: Replace with `str.removeprefix()` once dropped 3.8
+    assert click_context.command.name is not None
```
webknjaz, line 686: The comment is disconnected from the line it refers to
```suggestion
    assert click_context.command.name is not None
    # TODO: Replace with `str.removeprefix()` once dropped 3.8
```

webknjaz (approved):

webknjaz:
@chrysle hey, no problem! This is open source and nobody _expects_ free labor to be a top priority of every contributor.

Also, the code style is something that evolves and is subjective most of the time. People are only expected to unconditionally adhere what automatic checks say. Otherwise, code reviews is a multi-party discussion that exists to agree on things and receive insights from other people that aren't related to code style.

At times, it may feel like some comments are about code style at glance, but usually, there's more important underlying issues that motivate people to make suggestions. It's mostly a perception thing. I know that my comments are often seen as style nitpicks while I write them with maintainability in mind, informed by architectural considerations, usually.

Here's a few links that I collected about code reviews, by the way: https://gist.github.com/webknjaz/2aa9c7a43b51c1385260ff87e0cbb9e3.

P.S. The `linkcheck` job is failing because `jazzband.co` has a TLS certificate that expired on Dec 30 last year and wasn't auto-renewed for some reason. A tagged @jezdez on matrix so he could take a look.
This blocks merging the PR, but once fixed, restarting that one failing job should unblock us.

chrysle:
Thank you for your encouraging words and the time you invest into reviewing my code!
I'll also still have to solve the mystery how people get the right ideas for founding successful open source projects ;-)

Issue #826, opened 2019-05-27
Merge constraint contents for different envs


webknjaz:
I want to run pip-compile against the same set of requirements files under different envs (Python interpreters) and get a combined output.
Currently, it doesn't seem possible. Am I overlooking something?

##### Environment Versions

N/A

##### Steps to replicate

N/A

##### Expected result

resuling file is extended with extra entries

##### Actual result

resulting file is overridden

atugushev:
Hello @webknjaz,

Thank for the report. See #651 for the detailed info how to deal with different envs.

webknjaz:
@atugushev thanks. I usually have `-c constraints.txt` at the top of my requirements file. Which means that it's also shared leading to missing pins in constraints in some envs. Currently I have to edit things manually.

atugushev:
> Currently, it doesn't seem possible. Am I overlooking something?

@webknjaz does the https://github.com/jazzband/pip-tools/issues/826#issuecomment-496242086 resolved the issue?

webknjaz:
No, it's not what I was asking about.

atugushev:
I see. Unfortunately, it's not possible. Currently, pip-tools compiles requirements, evaluates markers and outputs results without ones. The official stance of pip-tools regarding cross-environment usage is that `pip-compile` "must be executed for each environment" (as I pointed out before). I believe this decision has been made deliberately to keep things simple.

But anyways, "merging markers" doesn't seem so impossible. At least `pipenv` and `poetry` solved this issue (IIRC) somehow. This might be challenging, but the feature worth it. Let's mark it as feature-request then.

AndydeCleyre:
@webknjaz With the new `5.0.0` release comes some new possibilities for working with multi-environment markers, as `pip-sync` will now happily ignore non-env-matching reqs.

I'm not sure what the best solutions and workflows might be, but here's something I played around with. Maybe it can give you some ideas, and then you can give back some ideas!

`platform-compile.sh`:
```bash
#!/bin/sh -e
# platform-compile.sh [<reqsin>] [<pip-compile-arg>...]

python_version="$(python -c 'from __future__ import print_function; import platform; print(*platform.python_version_tuple()[:2], sep=".")')"
sys_platform="$(python -c 'from __future__ import print_function; import sys; print(sys.platform)')"
machine="$(python -c 'from __future__ import print_function; import platform; print(platform.machine())')"

reqsin="${1:-requirements.in}"
[[ -r "$1" ]] && shift

[...]

PR #1969, opened 2023-08-18
Setup reproducible docs builds using pip-tools

atugushev:
Since the last Sphinx update, the docs build has started [failing](https://readthedocs.org/projects/pip-tools/builds/21650620/). It turns out pip-tools doesn't have reproducible builds for docs. Hence, the PR.

Refs: https://github.com/pradyunsg/furo/discussions/693.


webknjaz:
Somebody should start dogfooding pip-tools...

atugushev:
> Somebody should start dogfooding pip-tools...

@webknjaz I'm kinda lost, what do you mean?

atugushev:
Ah, I wasn't aware of the word "dogfooding". You're right. It would be nice to have `docs/requirements.in` and perhaps dependabot integration. You're reading my mind.

.readthedocs.yaml
```diff
@@ -23,7 +23,7 @@ build:
 
 python:
   install:
-    - requirements: docs/requirements.txt
+    - requirements: requirements/docs.txt
     - method: pip
       path: .
```
theryanwalker, line 28: I'm wondering if there is a way to use `pip-sync` here?

.readthedocs.yaml
```diff
@@ -23,7 +23,7 @@ build:
 
 python:
   install:
-    - requirements: docs/requirements.txt
+    - requirements: requirements/docs.txt
```
webknjaz, line 26: Since we don't already have generic requirements, I wouldn't move this to a separate folder. The location under `docs/` is closer to where it's actually being used + this idiom is widely copied over the ecosystem. Let's not reinvent the wheel.

.readthedocs.yaml
webknjaz: In fact, I don't like installing from a constraints file.
Here https://github.com/aio-libs/aiohttp/pull/7471/files#diff-cde814ef2f549dc093f5b8fc533b7e8f47e7b32a8081e0760e57d5c25a1139d9R18-R44 I've been playing with a new mechanism for RTD build customization. Try reproducing the same here.

webknjaz:
> Found nice docs over [here](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#pin-your-transitive-dependencies).

Yeah, that's a relatively new doc.

.readthedocs.yaml
atugushev: Indeed, currently we don't have any generic requirements. I propose, if we're looking to fully embrace pip-tools and create reproducible builds, that we should consider compiling dependencies for the tests as well, which are currently declared in `pyproject.toml` under `extra=testing`. I'm under the impression that current approach is less favoured based on [this](https://github.com/jazzband/pip-tools/pull/1681#issuecomment-1666966545) and [this](https://github.com/jazzband/pip-tools/pull/1960#issuecomment-1670580284) comments.

So, the next step – to move the test requirements to `requirements/test.{in|txt}` from `extra=testing` in `pyproject.toml` – seems logical. With this in mind, wouldn't it also be logical to relocate preemptively the docs requirements to a common requirements folder eliminating the need to open another folder later on? Would that make sense?

.readthedocs.yaml
atugushev: > I'm wondering if there is a way to use `pip-sync` here?

I guess we can figure it out. I'll try to come up with a solution in a follow-up PR. Thanks for the example, @webknjaz. That's useful.

.readthedocs.yaml
webknjaz: Yes to no extra. But it's a bit more complicated since different envs have different trees. So it's not just one file pair but a matrix of them.
I think that the idea that works is to have tox-env based collections of lockfiles, like the one I implemented in https://github.com/jazzband/pip-tools/issues/826#issuecomment-1324127670.

As for the docs, I'm torn. If we were to have a constraints file following the same principle, it'd be tox-specific and would probably need several env-specific incarnations. Whereas, tox is not the only place where we call sphinx — in RTD we do so as well. The common denominator is the `docs/` dir, so it kinda makes sense to bundle everything under it. OTOH, even in my cheroot PoC, I experimented with having the lockfiles under `requirements/` and I'm not sure how I feel about it. That was my attempt to come up with something generic/reusable, and it seems to need more polishing still. I think I have requirements under `docs/` in _all_ (or most) the other projects.

I guess it's hard for me to decide on one way or the other. Having the `docs/` dir self-contained ain't bad.

atugushev: Thank you for the detailed reply. Your insights certainly provided a lot to consider. Following the principle of "not putting the cart before the horse" let's fix the broken docs build first and then make a decision on how to handle the generic requirements. The commit has been reverted in 9beca35f16ee02cf3a544b696e555594524214d6 so that we can progress further.

webknjaz (approved):

PR #1973, opened 2023-08-20
Comment on `pytest --dist=worksteal` usage

atugushev:
Address https://github.com/jazzband/pip-tools/pull/1963#discussion_r1299439764

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,7 @@ addopts = [
   # `pytest-xdist`:
   "--numprocesses=auto",
 
-  # This distribution method is useful if the run times of different tests vary greatly,
+  # The `worksteal` distribution method is useful if the run times of different tests vary greatly,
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -99,6 +99,9 @@ disallow_untyped_defs = false
 addopts = [
   # `pytest-xdist`:
   "--numprocesses=auto",
+
+  # This distribution method is useful if the run times of different tests vary greatly,
```
webknjaz, line 103: Code comments that are not on the same line tend to drift away from the lines they refer to over time. Referencing the exact context may help notice this:
```suggestion
  # The `worksteal` distribution method is useful if the run times of different tests vary greatly,
```

atugushev: Thanks, applied the fix.

webknjaz (approved):

atugushev:
Thanks @webknjaz!

PR #1981, opened 2023-09-04
🐛 Fix collecting deps for all extras in multiple input packages

### LINKED ISSUE #1980: Running pip-tools compile on multiple sources fails with --all-extras
Running pip-tools compile with `--all-extras` on multiple sources containing extras always fails with the following error:

> ```console
> Error: Invalid value: --extra has no effect when used with --all-extras
> ```

#### Environment Versions

1. Linux
1. Python version: `python 3.10.8`
1. pip version: `pip 22.2.2`
1. pip-tools version: `pip-compile, version 7.3.1.dev21`

#### Steps to replicate

1. Create two small projects with `pyproject.toml` containing
```toml
[project]
name = "package-a"
version = "0.1.0"

test = []
```
2. Run `python -m piptools compile --all-extras package-a/pyproject.toml package-b/pyproject.toml`

#### Expected result

A `requirements.txt` file for the combined projects including all extras should be generated.

#### Actual result

> ```console
> Error: Invalid value: --extra has no effect when used with --all-extras
> ```

dragly:
Previously, an error would always be thrown when running `compile` with `--all-extras` on multiple source files containing extras. The reason was that the first iteration over the first source file would fill the `extras` variable, which in the second iteration would trigger an error since both `all_extras` and `extras` would be set.

This change moves the check outside the loop and early in the script to avoid unnecessary processing before throwing the error.

Further, only moving the check outside the loop would currently leave us with only the extras from the last `src_file` in the iteration over `src_files`.

This change also ensures all extras are in fact collected from all packages when `--all-extras` is passed as an argument.

This fixes #1980

##### Contributor checklist

- [ ] Included tests for the changes. (Happy to add a test, but not sure which type is preferred).

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -205,6 +205,10 @@ def cli(
                 ).format(", ".join(DEFAULT_REQUIREMENTS_FILES))
             )
 
+    if all_extras and extras:
+        msg = "--extra has no effect when used with --all-extras"
+        raise click.BadParameter(msg)
+
     if not output_file:
         # An output file must be provided for stdin
         if src_files == ("-",):
@@ -372,10 +376,7 @@ def cli(
             if not only_build_deps:
                 constraints.extend(metadata.requirements)
                 if all_extras:
-                    if extras:
-                        msg = "--extra has no effect when used with --all-extras"
-                        raise click.BadParameter(msg)
-                    extras = metadata.extras
+                    extras += metadata.extras
             if build_deps_targets:
                 constraints.extend(metadata.build_requirements)
         else:
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -3030,6 +3030,9 @@ def test_cli_compile_all_extras_with_multiple_packages(
     )
 
     assert out.exit_code == 0, out
+    assert "--all-extras" in out.stderr
+    assert "test_package_1/0.1/setup.py" in out.stderr
+    assert "test_package_2/0.1/setup.py" in out.stderr
 
 
 @pytest.mark.parametrize(
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -3001,6 +3001,37 @@ def test_cli_compile_strip_extras(runner, make_package, make_sdist, tmpdir):
     assert "[more]" not in out.stderr
 
 
+def test_cli_compile_all_extras_with_multiple_packages(
+    runner, make_package, make_sdist, tmpdir
+):
+    """
+    Assures that ``--all-extras`` works when multiple sources are specified.
+    """
+    test_package_1 = make_package(
+        "test_package_1",
+        version="0.1",
+        extras_require={"more": []},
+    )
+    test_package_2 = make_package(
+        "test_package_2",
+        version="0.1",
+        extras_require={"more": []},
+    )
+
+    out = runner.invoke(
+        cli,
+        [
+            "--all-extras",
+            "--output-file",
+            "requirements.txt",
+            str(test_package_1 / "setup.py"),
+            str(test_package_2 / "setup.py"),
+        ],
+    )
+
+    assert out.exit_code == 0, out
+
+
 @pytest.mark.parametrize(
     ("package_specs", "constraints", "existing_reqs", "expected_reqs"),
     (
@@ -3132,7 +3163,6 @@ def test_resolver_drops_existing_conflicting_constraint(
     with open("requirements.txt") as req_txt:
         req_txt_content = req_txt.read()
         assert expected_requirements.issubset(req_txt_content.splitlines())
-
 
 def test_resolution_failure(runner):
     """Test resolution impossible for unknown package."""
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -3031,8 +3031,8 @@ def test_cli_compile_all_extras_with_multiple_packages(
 
     assert out.exit_code == 0, out
     assert "--all-extras" in out.stderr
-    assert "test_package_1/0.1/setup.py" in out.stderr
-    assert "test_package_2/0.1/setup.py" in out.stderr
+    assert f"test_package_1{os.path.sep}0.1{os.path.sep}setup.py" in out.stderr
+    assert f"test_package_2{os.path.sep}0.1{os.path.sep}setup.py" in out.stderr
 
 
 @pytest.mark.parametrize(
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -3164,6 +3164,7 @@ def test_resolver_drops_existing_conflicting_constraint(
         req_txt_content = req_txt.read()
         assert expected_requirements.issubset(req_txt_content.splitlines())
 
+
 def test_resolution_failure(runner):
     """Test resolution impossible for unknown package."""
     with open("requirements.in", "w") as reqs_out:


atugushev (commented):
Thanks for the fix! Could you please add a test that confirms the bug?

piptools/scripts/compile.py
```diff
@@ -336,10 +340,8 @@ def cli(
             )
 
             if all_extras:
-                if extras:
-                    msg = "--extra has no effect when used with --all-extras"
-                    raise click.BadParameter(msg)
-                extras = tuple(metadata.get_all("Provides-Extra"))
+                if provides_extra := metadata.get_all("Provides-Extra"):
```
webknjaz, line 343: FTR this will conflict with https://github.com/jazzband/pip-tools/pull/1681.

dragly:
> Thanks for the fix! Could you please add a test that confirms the bug?

Yes 👍

How do you prefer the test to be implemented? I already have an example setup that fails when invoking from the command line. Should I add a test that runs from the command line or rather implement an internal test?

atugushev:
@dragly I'd prefer a test in `test_cli_compile.py`. Use `test_cli_compile_strip_extras` as a reference.

dragly:
I have added a test in a fixup commit now. The succeeds now and will fail if you revert the first commit.

By the way, I struggled a bit while developing this test because pytest consumed all output written to stdout and stderr. The usual `pytest -s` and `pytest -rx` flags did not work. How do you usually debug any errors when working on the tests?

dragly:
I also see that there is a merge conflict now with tests/test_cli_compile.py. I can resolve that. Do you prefer that I do it in a merge commit or that I rebase this branch on main right away?

chrysle:
> I have added a test in a fixup commit now. The succeeds now and will fail if you revert the first commit.
> 
> By the way, I struggled a bit while developing this test because pytest consumed all output written to stdout and stderr. The usual `pytest -s` and `pytest -rx` flags did not work. How do you usually debug any errors when working on the tests?

`pytest --verbose` doesn't work?

> I also see that there is a merge conflict now with tests/test_cli_compile.py. I can resolve that. Do you prefer that I do it in a merge commit or that I rebase this branch on main right away?

I guess that's unimportant.

tests/test_cli_compile.py
```diff
@@ -2779,6 +2779,38 @@ def test_cli_compile_strip_extras(runner, make_package, make_sdist, tmpdir):
...
+        "test_package_1",
+        version="0.1",
+        extras_require={"more": []},
+    )
+    test_package_2 = make_package(
+        "test_package_2",
+        version="0.1",
+        extras_require={"more": []},
+    )
+    dists_dir = tmpdir / "dists"
```
chrysle, line 2798: Guess you have left this over when copying the preceding test.

```suggestion
```

tests/test_cli_compile.py
```diff
@@ -2779,6 +2779,38 @@ def test_cli_compile_strip_extras(runner, make_package, make_sdist, tmpdir):
     assert "[more]" not in out.stderr
 
 
+def test_cli_compile_all_extras_with_multiple_packages(
+    runner, make_package, make_sdist, tmpdir
+):
+    """
+    Assures that --all-extras works when multiple sources are specified.
```
chrysle, line 2786:
```suggestion
    Assures that `--all-extras` works when multiple sources are specified.
```

webknjaz: We should use RST, not Markdown in the docstrings..

piptools/scripts/compile.py
```diff
@@ -175,6 +175,10 @@ def cli(
                 ).format(", ".join(DEFAULT_REQUIREMENTS_FILES))
             )
 
+    if all_extras and extras:
+        msg = "--extra has no effect when used with --all-extras"
+        raise click.BadParameter(msg)
```
chrysle, line 180: Please add an additional test to cover this.

tests/test_cli_compile.py
chrysle: Well, almost the whole test file uses that (and we don't use autodoc).

webknjaz: We should start doing that, then. I usually integrate `sphinxcontrib-apidoc` for automatic generation like this: https://github.com/cherrypy/cheroot/commit/cb9d3f46aa755d63e85d539a2efb7d1fa8f53935. Later on, I remembered to mark the ToC section as **private**: https://cheroot.cherrypy.dev/en/latest/pkg/modules/.

dragly: Updated now according to the suggestion. I will keep it in line with the rest of the file, if that is okay.

dragly:
> > By the way, I struggled a bit while developing this test because pytest consumed all output written to stdout and stderr. The usual `pytest -s` and `pytest -rx` flags did not work. How do you usually debug any errors when working on the tests?
> 
> `pytest --verbose` doesn't work?

Weirdly enough I see the output now. Not sure why it did not work earlier :thinking:

tests/test_cli_compile.py
```diff
@@ -3001,6 +3001,37 @@ def test_cli_compile_strip_extras(runner, make_package, make_sdist, tmpdir):
...
+        version="0.1",
+        extras_require={"more": []},
+    )
+
+    out = runner.invoke(
+        cli,
+        [
+            "--all-extras",
+            "--output-file",
+            "requirements.txt",
```
chrysle, line 3026: Could you redirect this to standard output and check it, to make sure?

dragly: Sorry about the late reply. I printed stdout and stderr now and got:
```
WARNING: the legacy dependency resolver is deprecated and will be removed in future versions of pip-tools.
WARNING: --strip-extras is becoming the default in version 8.0.0. To silence this warning, either use --strip-extras to opt into the new default or use --no-strip-extras to retain the existing behavior.
#
# by the following command:
#
#    pip-compile --all-extras --output-file=requirements.txt /tmp/pytest-of-user/pytest-8/popen-gw16/packages0/test_package_1/0.1/setup.py /tmp/pytest-of-user/pytest-8/popen-gw16/packages1/test_package_2/0.1/setup.py
#
```

chrysle: Could you write an assert statement to ensure the header contains `--all-extras` and both of the files?

dragly: Yes, no problem. I have pushed a fixup with the asserts now. I also tested that they fail if I assert on a different string, such as `test_package_3/0.1/setup.py`.

chrysle (approved):

tests/test_cli_compile.py
```diff
@@ -3001,6 +3001,40 @@ def test_cli_compile_strip_extras(runner, make_package, make_sdist, tmpdir):
...
+            str(test_package_1 / "setup.py"),
+            str(test_package_2 / "setup.py"),
+        ],
+    )
+
+    assert out.exit_code == 0, out
+    assert "--all-extras" in out.stderr
+    assert "test_package_1/0.1/setup.py" in out.stderr
+    assert "test_package_2/0.1/setup.py" in out.stderr
```
chrysle, line 3035: Obviously, Windows isn't happy with POSIX file path syntax.

```suggestion
    assert str(test_package_1 / "setup.py") in out.stderr
    assert str(test_package_2 / "setup.py") in out.stderr
```

webknjaz: Alternatively, just replace the path separators in the strings:
```suggestion
    assert f"test_package_1{os.path.sep}0.1{os.path.sep}setup.py" in out.stderr
    assert f"test_package_2{os.path.sep}0.1{os.path.sep}setup.py" in out.stderr
```

But the pathlib approach is fine. Though, I wouldn't copy-paste the path construction as the same thing is present in code above — it's reasonable to put that into vars and reuse across the test function.

piptools/scripts/compile.py
```diff
@@ -359,10 +363,7 @@ def cli(
             if not only_build_deps:
                 constraints.extend(metadata.requirements)
                 if all_extras:
-                    if extras:
-                        msg = "--extra has no effect when used with --all-extras"
-                        raise click.BadParameter(msg)
-                    extras = metadata.extras
+                    extras += metadata.extras
```
webknjaz, line 366: Why does this have `+=`? How is it related to just moving the check to another location as the PR title advertises?

chrysle: Actually, it doesn't make a difference, does it? It's verified that `extras` and `all_extras` aren't combined beforehand, so we could as well remove it.

```suggestion
                    extras = metadata.extras
```

dragly: It is required. Otherwise, only the extras from the last iteration (the last `src_file`) will be used.

I agree this detail is missing from the description and title. I will add it now.

tests/test_cli_compile.py
dragly: I went with `os.path.sep` since the other alternative uses the absolute path and would require calling `relative_to` or similar, which I think obscures the readability of what is actually tested.

Pushed in fixup now.

tests/test_cli_compile.py
```diff
@@ -3001,6 +3001,40 @@ def test_cli_compile_strip_extras(runner, make_package, make_sdist, tmpdir):
...
+            "requirements.txt",
+            str(test_package_1 / "setup.py"),
+            str(test_package_2 / "setup.py"),
+        ],
+    )
+
+    assert out.exit_code == 0, out
+    assert "--all-extras" in out.stderr
+    assert f"test_package_1{os.path.sep}0.1{os.path.sep}setup.py" in out.stderr
+    assert f"test_package_2{os.path.sep}0.1{os.path.sep}setup.py" in out.stderr
```
chrysle, line 3035: I think you overlooked @webknjaz's refactoring suggestion:

```suggestion
    test_package_1_setup_py =  f"test_package_1{os.path.sep}0.1{os.path.sep}setup.py"
    test_package_2_setup_py =  f"test_package_2{os.path.sep}0.1{os.path.sep}setup.py"

    out = runner.invoke(
        cli,
        [
            "--all-extras",
            "--output-file",
            "requirements.txt",
            test_package_1_setup_py,
            test_package_2_setup_py,
        ],
    )

    assert out.exit_code == 0, out
    assert "--all-extras" in out.stderr
    assert test_package_1_setup_py in out.stderr
    assert test_package_2_setup_py in out.stderr
```

webknjaz: @chrysle only I wanted the pathlib+str for making those vars. OTOH the current state is acceptable, I guess.

webknjaz:
@dragly I took a stab at updating the title — it should refer to one high-level effect, not the implementation details. The details/motivation are to be in the description if needed. Otherwise, it just looks like a non-atomic collection of loosely related changes being enumerated.

Do you think this is accurate enough?

dragly:
> @dragly I took a stab at updating the title — it should refer to one high-level effect, not the implementation details. The details/motivation are to be in the description if needed. Otherwise, it just looks like a non-atomic collection of loosely related changes being enumerated.
> 
> Do you think this is accurate enough?

Yes, that looks perfect! Thanks for updating it :) I agree with the rationale too :+1:

webknjaz (approved):

webknjaz:
@dragly thanks! I noticed, there's a coverage drop in this PR. Could you check it out and compensate for the lost lines? Especially, pay attention to the indirect changes tab in Codecov: https://app.codecov.io/gh/jazzband/pip-tools/pull/1981/indirect-changes.

chrysle:
Can't explain the test failure. It's about a missing `PKG-INFO` file in the build directory, which is obviously present:

```  installing to build/bdist.linux-x86_64/wheel
  running install
  running install_egg_info
  running egg_info
  writing fake_with_deps.egg-info/PKG-INFO
  writing dependency_links to fake_with_deps.egg-info/dependency_links.txt
  writing requirements to fake_with_deps.egg-info/requires.txt
  writing top-level names to fake_with_deps.egg-info/top_level.txt
  reading manifest file 'fake_with_deps.egg-info/SOURCES.txt'
  writing manifest file 'fake_with_deps.egg-info/SOURCES.txt'
  Copying fake_with_deps.egg-info to build/bdist.linux-x86_64/wheel/fake_with_deps-0.1-py3.8.egg-info
  running install_scripts
  error: [Errno 2] No such file or directory: 'build/bdist.linux-x86_64/wheel/fake_with_deps-0.1-py3.8.egg-info/PKG-INFO'
  error: subprocess-exited-with-error
```

webknjaz:
> Can't explain the test failure. It's about a missing `PKG-INFO` file in the build directory, which is obviously present:

Could be an incomplete build deps pin (as in setuptools updated). I suppose using what #1681 addresses in our own tests could make things more stable. Dogfooding FTW!

dragly:
> @dragly thanks! I noticed, there's a coverage drop in this PR. Could you check it out and compensate for the lost lines? Especially, pay attention to the indirect changes tab in Codecov: https://app.codecov.io/gh/jazzband/pip-tools/pull/1981/indirect-changes.

I will try, but have to admit I am not too familiar with reading Codecov reports. 

I looked at the page you linked and the documentation for [indirect changes](https://docs.codecov.com/docs/unexpected-coverage-changes). It seems like it is not hitting a few conditional imports based on Python version.

Is it possible that Python or pip versions change between this PR and the previous run that codecov compares to?

Do you have other pointers on how you typically dig into a codecov change like this?

chrysle:
Thank you!

webknjaz:
> Is it possible that Python or pip versions change between this PR and the previous run that codecov compares to?
> 
> Do you have other pointers on how you typically dig into a codecov change like this?

Sorry, this slipped my attention. If it's conditional imports, depending on the Python version, maybe some of the CI jobs failed to upload coverage (the codecov service can be flaky like that).

PR #1984, opened 2023-09-07
Make BacktrackingResolver ignore extras when dropping existing constraints

### LINKED ISSUE #1977: Resolver error on upgrade
When upgrading the (pinned) dependencies in requirements.in of a project with an existing requirements.txt, pip-compile fails with a resolver error. It claims that an indirect dependency (here: `cachecontrol[filecache]<0.14.0,>=0.13.0`) cannot be resolved even though a version (0.13.1) matching the constraint is listed among the considered versions.

pip-compile successfully creates a requirements.txt from the same requirements.in file (with the same pinned project dependencies), if I first delete the existing requirements.txt. However, I came across this error when [renovateBot](https://docs.renovatebot.com/) failed to update requirements.txt – that's a scenario where I cannot delete requirements.txt first.

#### Environment Versions

1. MacOS 13.5 (also reproduced under Ubuntu jammy)
1. Python version: 3.11.2 (also observed under Python 3.10 and 3.9)
1. pip version: 23.2.1
1. pip-tools version: 7.3.0

#### Steps to replicate

~~~shell
(pip-tools-venv) piptools  $ rm -f requirements.in requirements.txt
(pip-tools-venv) piptools  $ echo "poetry==1.5.1\n" > requirements.in
(pip-tools-venv) piptools  $ pip-compile --generate-hashes -q requirements.in
WARNING: --strip-extras is becoming the default in version 8.0.0. To silence this warning, either use --strip-extras to opt into the new default or use --no-strip-extras to retain the existing behavior.
(pip-tools-venv) piptools  $ grep '^cachecontrol' requirements.txt
cachecontrol[filecache]==0.12.14 \
(pip-tools-venv) piptools  $ echo "poetry==1.6.1\n" > requirements.in
(pip-tools-venv) piptools  $ pip-compile --generate-hashes -q requirements.in
  ERROR: Could not find a version that satisfies the requirement cachecontrol[filecache]<0.14.0,>=0.13.0 (from poetry) (from versions: 0.5, 0.6, 0.7.1, 0.7.2, 0.7.3, 0.8.0, 0.8.1, 0.8.2, 0.8.3, 0.9.0, 0.9.1, 0.9.2, 0.9.3, 0.9.4, 0.10.0, 0.10.1, 0.10.2, 0.10.3, 0.10.4, 0.10.6, 0.10.7, 0.11.0, 0.11.1, 0.11.2, 0.11.3, 0.11.4, 0.11.5, 0.11.6, 0.11.7, 0.12.0, 0.12.1, 0.12.2, 0.12.3, 0.12.4, 0.12.5, 0.12.6, 0.12.7, 0.12.8, 0.12.9, 0.12.10, 0.12.11, 0.12.12, 0.12.13, 0.12.14, 0.13.0, 0.13.1rc0, 0.13.1)
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 92, in resolve
    result = self._result = resolver.resolve(
                            ^^^^^^^^^^^^^^^^^
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/pip/_vendor/resolvelib/resolvers.py", line 546, in resolve
    state = resolution.resolve(requirements, max_rounds=max_rounds)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/pip/_vendor/resolvelib/resolvers.py", line 439, in resolve
    raise ResolutionImpossible(self.state.backtrack_causes)
pip._vendor.resolvelib.resolvers.ResolutionImpossible: [RequirementInformation(requirement=SpecifierRequirement('cachecontrol[filecache]<0.14.0,>=0.13.0'), parent=LinkCandidate('https://files.pythonhosted.org/packages/7d/25/f3bfda3c458d114005af99441d009936b85a34a730aeb9cf57fb2630d9f7/poetry-1.6.1-py3-none-any.whl (from https://pypi.org/simple/poetry/) (requires-python:>=3.8,<4.0)'))]

The above exception was the direct cause of the following exception:

  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/click/decorators.py", line 33, in new_func
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/piptools/scripts/compile.py", line 659, in cli
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/piptools/resolver.py", line 604, in resolve
    is_resolved = self._do_resolve(
                  ^^^^^^^^^^^^^^^^^
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/piptools/resolver.py", line 636, in _do_resolve
    resolver.resolve(
  File "/Users/christoph.ludwig/tmp/piptools/pip-tools-venv/lib/python3.11/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 101, in resolve
    raise error from e
pip._internal.exceptions.DistributionNotFound: No matching distribution found for cachecontrol[filecache]<0.14.0,>=0.13.0
(pip-tools-venv) piptools  $ rm requirements.txt
(pip-tools-venv) piptools  $ pip-compile --generate-hashes -q requirements.in
WARNING: --strip-extras is becoming the default in version 8.0.0. To silence this warning, either use --strip-extras to opt into the new default or use --no-strip-extras to retain the existing behavior.
(pip-tools-venv) piptools  $ grep '^cachecontrol' requirements.txt
cachecontrol[filecache]==0.13.1 \
~~~

#### Expected result

requirements.txt is updated, no need to delete requirements.txt first

#### Actual result

The  resolver error shown if requirements.txt exists already generated for an older version of the project dependency.

chludwig-haufe:
This PR resolves #1977. 

When the backtracking resolver collected the names of incompatible install requirements, it used to apply `utils.key_from_req` to the requirements listed in the exception causes. According to the type hints, `utils.key_from_req` expects either a `pip._internal.req.req_install.InstallRequirement` or a `pip._vendor.packaging.requirements.Requirement`. In some cases, though, the object passed is a  `pip._internal.resolution.resolvelib.requirements.SpecifierRequirement` (i.e., a specialization of `pip._internal.resolution.resolvelib.base.Requirement`).

All mentioned classes define an attribute `name`. In case of `InstallRequirement` and the `Requirement` from the vendor package, the name holds the bare package name. In case of the `Requirement` from the resolvelib package, the name includes the extras (if any).

`key_from_req` is used in other places as well, I don't know whether they rely on the existing behavior. Therefore, I added a new function `utils.key_no_extra_from_req` that also accepts instances of `pip._internal.resolution.resolvelib.base.Requirement` and keeps only the part of the name up to the first opening square bracket (if any).

I was surprised I did not find more tests of the backtracking resolver. In the end, I followed the example of `test_catch_distribution_not_found_error` and directly called the backtracking resolver's method `_do_resolve` with a mocked pip resolver object.

Furthermore, I had to add pytest to the config of the mypy-mirror pre-commit hook.

I still got "untyped def" errors from `test_resolver.py`, even though there's an override for test modules in pyproject.toml that is supposed to disable this check. I therefore had the impression that mypy-mirror ignores the mypy config in pyproject.toml. I didn't investigate this any further, though, and simply annotated the all parameters of my test.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
           - pip==20.3.4
           - build==1.0.0
           - pyproject_hooks==1.0.0
-          - pytest>=7.2.0
+          - pytest==7.4.2
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.5
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,7 @@ repos:
           - pip==20.3.4
           - build==1.0.0
           - pyproject_hooks==1.0.0
+          - pytest>=7.2.0
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.5
     hooks:
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -39,7 +39,7 @@ from .utils import (
     is_pinned_requirement,
     is_url_requirement,
     key_from_ireq,
-    key_from_req,
+    key_no_extra_from_req,
     omit_list_value,
     strip_extras,
 )
@@ -648,7 +648,7 @@ class BacktrackingResolver(BaseResolver):
 
             # Collect all incompatible install requirement names
             cause_ireq_names = {
-                key_from_req(cause.requirement) for cause in cause_exc.causes
+                key_no_extra_from_req(cause.requirement) for cause in cause_exc.causes
             }
 
             # Looks like resolution is impossible, try to fix
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -648,7 +648,7 @@ class BacktrackingResolver(BaseResolver):
 
             # Collect all incompatible install requirement names
             cause_ireq_names = {
-                key_from_req(cause.requirement, remove_extras=True)
+                strip_extras(key_from_req(cause.requirement))
                 for cause in cause_exc.causes
             }
 
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -39,7 +39,7 @@ from .utils import (
     is_pinned_requirement,
     is_url_requirement,
     key_from_ireq,
-    key_no_extra_from_req,
+    key_from_req,
     omit_list_value,
     strip_extras,
 )
@@ -648,7 +648,8 @@ class BacktrackingResolver(BaseResolver):
 
             # Collect all incompatible install requirement names
             cause_ireq_names = {
-                key_no_extra_from_req(cause.requirement) for cause in cause_exc.causes
+                key_from_req(cause.requirement, remove_extras=True)
+                for cause in cause_exc.causes
             }
 
             # Looks like resolution is impossible, try to fix
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -620,16 +620,15 @@ def test_backtracking_resolver_drops_existing_conflicting_constraints(
         Raise a ``DistributionNotFound`` exception that has a ``ResolutionImpossible``
         exception as its cause.
         """
-        try:
-            causes = [
-                NonCallableMock(
-                    requirement=SpecifierRequirement(existing_constraints[ireq.name])
-                )
-                for ireq in conflicting_ireqs
-            ]
-            raise ResolutionImpossible(causes)
-        except ResolutionImpossible as e:
-            raise DistributionNotFound("resolution impossible") from e
+        causes = [
+            NonCallableMock(
+                requirement=SpecifierRequirement(existing_constraints[ireq.name])
+            )
+            for ireq in conflicting_ireqs
+        ]
+        raise DistributionNotFound("resolution impossible") from ResolutionImpossible(
+            causes
+        )
 
     constraint_ireqs = [from_line(req, constraint=False) for req in constraints]
     conflicting_ireqs = [
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -651,7 +651,7 @@ def test_backtracking_resolver_drops_existing_conflicting_constraints(
     # resolver has been rigged to raise a DistributionNotFound exception with
     # a cause that refers to the entries of conflicting_ireqs.
     # We expect _do_resolve() to handle this exception by dropping
-    # the these entries from existing_constraints and returning False.
+    # these entries from existing_constraints and returning False.
     # It should _not_ drop any compatible constraint or re-raise
     # the DistributionNotFound exception.
     result = bt_resolver._do_resolve(resolver, existing_constraints)
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -1,23 +1,11 @@
 from __future__ import annotations
 
-from itertools import chain
-from typing import Callable, Sequence
-from unittest.mock import Mock, NonCallableMock
-
 import pytest
 from pip._internal.exceptions import DistributionNotFound
-from pip._internal.req.req_install import InstallRequirement
-from pip._internal.resolution.resolvelib.requirements import SpecifierRequirement
-from pip._internal.resolution.resolvelib.resolver import Resolver
 from pip._internal.utils.urls import path_to_url
-from pip._vendor.resolvelib.resolvers import ResolutionImpossible
 
 from piptools.exceptions import NoCandidateFound
-from piptools.resolver import (
-    BacktrackingResolver,
-    RequirementSummary,
-    combine_install_requirements,
-)
+from piptools.resolver import RequirementSummary, combine_install_requirements
 
 
 @pytest.mark.parametrize(
@@ -587,73 +575,3 @@ def test_catch_distribution_not_found_error(backtracking_resolver, exception, ca
             resolver=FakePipResolver(),
             compatible_existing_constraints={},
         )
-
-
-@pytest.mark.parametrize(
-    ("constraints", "conflicting_existing", "compatible_existing"),
-    (
-        (
-            [
-                "poetry==1.6.1",
-            ],
-            ["cachecontrol[filecache]==0.12.14"],
-            ["doesnotexist==1.0.0"],
-        ),
-        (
-            [
-                "poetry==1.6.1",
-            ],
-            ["keyring==22.1.0", "pkginfo==1.7.2"],
-            ["platformdirs==3.0.0", "doesnotexist==1.0.0"],
-        ),
-    ),
-)
-def test_backtracking_resolver_drops_existing_conflicting_constraints(
-    backtracking_resolver: Callable[..., BacktrackingResolver],
-    from_line: Callable[..., InstallRequirement],
-    constraints: Sequence[str],
-    conflicting_existing: Sequence[str],
-    compatible_existing: Sequence[str],
-) -> None:
-    def wrap_resolution_impossible(*args, **kwargs):
-        """
-        Raise a ``DistributionNotFound`` exception that has a ``ResolutionImpossible``
-        exception as its cause.
-        """
-        causes = [
-            NonCallableMock(
-                requirement=SpecifierRequirement(existing_constraints[ireq.name])
-            )
-            for ireq in conflicting_ireqs
-        ]
-        raise DistributionNotFound("resolution impossible") from ResolutionImpossible(
-            causes
-        )
-
-    constraint_ireqs = [from_line(req, constraint=False) for req in constraints]
-    conflicting_ireqs = [
-        from_line(req, constraint=True) for req in conflicting_existing
-    ]
-    compatible_ireqs = [from_line(req, constraint=True) for req in compatible_existing]
-    existing_constraints = {
-        ireq.name: ireq for ireq in chain(compatible_ireqs, conflicting_ireqs)
-    }
-
-    bt_resolver = backtracking_resolver(
-        constraint_ireqs, existing_constraints=existing_constraints
-    )
-    resolver = NonCallableMock(
-        spec=Resolver,
-        resolve=Mock(side_effect=wrap_resolution_impossible),
-    )
-
-    # resolver has been rigged to raise a DistributionNotFound exception with
-    # a cause that refers to the entries of conflicting_ireqs.
-    # We expect _do_resolve() to handle this exception by dropping
-    # these entries from existing_constraints and returning False.
-    # It should _not_ drop any compatible constraint or re-raise
-    # the DistributionNotFound exception.
-    result = bt_resolver._do_resolve(resolver, existing_constraints)
-
-    assert result is False
-    assert set(existing_constraints.keys()) == {ireq.name for ireq in compatible_ireqs}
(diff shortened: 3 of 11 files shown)


tests/test_resolver.py
```diff
@@ -575,3 +587,74 @@ def resolve(self, *args, **kwargs):
...
+    constraints: Sequence[str],
+    conflicting_existing: Sequence[str],
+    compatible_existing: Sequence[str],
+) -> None:
+    def wrap_resolution_impossible(*args, **kwargs):
+        """
+        Raise a ``DistributionNotFound`` exception that has a ``ResolutionImpossible``
+        exception as its cause.
+        """
+        try:
```
chrysle, line 623: Since you raise `ResolutionImpossible` anyway, I'd drop this `try` case.

chludwig-haufe: It took me a moment but I think I now understood what you meant. 😄 Changed as well.

piptools/resolver.py
```diff
@@ -648,7 +648,7 @@ def _do_resolve(
 
             # Collect all incompatible install requirement names
             cause_ireq_names = {
-                key_from_req(cause.requirement) for cause in cause_exc.causes
+                key_no_extra_from_req(cause.requirement) for cause in cause_exc.causes
```
chrysle, line 651: Since this is the only place where the new function is used, could you probably merge it with `key_from_req` creating an optional `strip_extras` parameter?

I don't think that would differ with other usages.

chludwig-haufe: In my own projects, I am somewhat wary of such "flag" parameters. But sure, no problem.

chludwig-haufe:
I found that `piptools.utils` already contained a function `strip_extras()` (line 478). Therefore, I switched to using the existing function. I don't see any reason to duplicate this functionality.

chrysle:
I wasn't aware of that function. Say, also since it's already used in _resolver.py_, wouldn't it be easier to just use it on the output of `key_from_req`? What do you think?

Sorry for pushing you around.

chludwig-haufe:
No problem, sometimes implementations need multiple iterations before the best approach becomes apparent.

I will take a look in the evening how to simplify the PR this way.

chrysle (approved):

atugushev (commented):
Thanks for the fix! While I don't mind unit tests, I'd prefer having an integration test instead. Mocking around with private methods like `Resolver._do_resolve()` could lead to extra work during future internal refactoring.

Could you please add at least one "success" integration test in `test_cli_compile.py` to confirm that the issue is resolved? Use `test_cli_compile_strip_extras` as a reference.

chludwig-haufe:
@atugushev: I replaced the unit test with an integration test as you proposed.

I gather from [`pyproject.toml` (line 95ff)](https://github.com/jazzband/pip-tools/blob/4c7bc26541ae11b18f8b9260104eede32cbc114a/pyproject.toml#L95) that your intention is to ignore untyped definitions in the tests. For some reason, mypy called from the pre-commit hook insists on typed definitions on all functions I touched in the tests, including the functions passed in as fixtures.

These functions have "complex" signatures (too complex for `Callable`, at least), whence I had to define protocol classes for them. Furthermore, I cannot reference these protocols if I simply define them next to the fixtures in `conftest.py`. Therefore, I placed them into `tests/utils.py`.

chrysle:
> I gather from [`pyproject.toml` (line 95ff)](https://github.com/jazzband/pip-tools/blob/4c7bc26541ae11b18f8b9260104eede32cbc114a/pyproject.toml#L95) that your intention is to ignore untyped definitions in the tests. For some reason, mypy called from the pre-commit hook insists on typed definitions on all functions I touched in the tests, including the functions passed in as fixtures.

This however sounds like an issue that should be investigated. Haven't found something loadable yet, though.

chludwig-haufe:
@chrysle: For demonstration's sake, I added `--verbose` to the mypy arguments in .pre-commit-config.yaml and then removed the type hint for the parameter `make_package` of `test_resolver_drops_existing_conflicting_constraint()` in `test_cli_compile.py`.

pre-commit rejected this commit, here's the relevant output:

~~~
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1
LOG:  Mypy Version:           1.5.1
LOG:  Config File:            pyproject.toml
LOG:  Configured Executable:  /Users/christoph.ludwig/.cache/pre-commit/repo3zt33lob/py_env-python3.11/bin/python
LOG:  Current Executable:     /Users/christoph.ludwig/.cache/pre-commit/repo3zt33lob/py_env-python3.11/bin/python
LOG:  Cache Dir:              .mypy_cache
LOG:  Compiled:               True
LOG:  Exclude:                ['^tests/test_data/']
LOG:  Found source:           BuildSource(path='tests/test_cli_compile.py', module='tests.test_cli_compile', has_text=False, base_dir='/Users/christoph.ludwig/Python/pip-tools', followed=False)
LOG:  Could not load cache for tests.test_cli_compile: tests/test_cli_compile.meta.json
LOG:  Metadata not found for tests.test_cli_compile
LOG:  Parsing tests/test_cli_compile.py (tests.test_cli_compile)
LOG:  Metadata fresh for tests: file /Users/christoph.ludwig/Python/pip-tools/tests/__init__.py
LOG:  Metadata fresh for piptools.scripts.compile: file /Users/christoph.ludwig/Python/pip-tools/piptools/scripts/compile.py
LOG:  Metadata fresh for unittest.mock: file /Users/christoph.ludwig/.cache/pre-commit/repo3zt33lob/py_env-python3.11/lib/python3.11/site-packages/mypy/typeshed/stdlib/unittest/mock.pyi
LOG:  Metadata fresh for click.testing: file /Users/christoph.ludwig/.cache/pre-commit/repo3zt33lob/py_env-python3.11/lib/python3.11/site-packages/click/testing.py
LOG:  Metadata fresh for piptools.utils: file /Users/christoph.ludwig/Python/pip-tools/piptools/utils.py
LOG:  Metadata fresh for tests.constants: file /Users/christoph.ludwig/Python/pip-tools/tests/constants.py
LOG:  Metadata fresh for tests.utils: file /Users/christoph.ludwig/Python/pip-tools/tests/utils.py

[ ... many more metadata refresh messages  ... ]

LOG:  Metadata fresh for packaging._elffile: file /Users/christoph.ludwig/.cache/pre-commit/repo3zt33lob/py_env-python3.11/lib/python3.11/site-packages/packaging/_elffile.py
LOG:  Loaded graph with 249 nodes (0.058 sec)
LOG:  Found 149 SCCs; largest has 39 nodes
LOG:  Processing 148 queued fresh SCCs
LOG:  Processing SCC singleton (tests.test_cli_compile) as inherently stale
tests/test_cli_compile.py:2874: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
LOG:  Deleting tests.test_cli_compile tests/test_cli_compile.py tests/test_cli_compile.meta.json tests/test_cli_compile.data.json
LOG:  No fresh SCCs left in queue
LOG:  Build finished in 0.253 seconds with 249 modules, and 1 errors
Found 1 error in 1 file (checked 1 source file)
~~~

chrysle:
> pre-commit rejected this commit, here's the relevant output:

I guess that's unrelated; it's just rebuilding its cache.
Could you try if setting [`disallow_incomplete_defs`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_incomplete_defs)  to _false_ in the `mypy` override works for you? I have the strong feeling that could be the culprit.

chludwig-haufe:
> Could you try if setting [`disallow_incomplete_defs`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_incomplete_defs) to _false_ in the `mypy` override works for you? I have the strong feeling that could be the culprit.

Right, with `disallow_incomplete_defs = false` in the test override, mypy / pre-commit let the commit pass.

What do you want me to do about the annotations (and, in particular, the protocol classes) I added to the tests? Should I take them out again (or, maybe, move the protocol classes into a dedicated module, because they don't really fit into `utils`)?

chrysle:
> What do you want me to do about the annotations (and, in particular, the protocol classes) I added to the tests? Should I take them out again

I think that would be the better solution, since it's otherwise unnecessary complex.

chludwig-haufe:
> I think that would be the better solution, since it's otherwise unnecessary complex.

Done.

.pre-commit-config.yaml
```diff
@@ -34,6 +34,7 @@ repos:
           - pip==20.3.4
           - build==1.0.0
           - pyproject_hooks==1.0.0
+          - pytest>=7.2.0
```
chrysle, line 37: Generally requirements should be pinned to ensure stability. With several issues fixed in 7.4.2, I think we should use that as a mark.

```suggestion
          - pytest==7.4.2
```

chludwig-haufe: I agree, since I let RenovateBot or similar tools keep the dependencies up-to-date in my own projects.

But I simply copied the constraint from the `project.optional-dependencies` section in `pyproject.toml`. If pytest is pinned in the pre-commit configuration, then it should be pinned to the same version in `pyproject.toml`.

chludwig-haufe: So far, none of the dependencies in `pyproject.toml` are pinned to a fixed version. Of course, I could pin them, but that's a change in how your project handles dependencies that's unrelated to my bugfix. Do you want this change mixed into this PR?

chrysle: I guess it wouldn't hurt if you just pinned it in the pre-commit configuration right now. There is a proposal to update to newer dependencies in #1927, but it was held back by a refactoring idea. We'll sort it out in that context.

chrysle:
Thank you!

PR #1989, opened 2023-09-20
Fix ignorance of inverted CLI options in config for `pip-sync`

### LINKED ISSUE #1988: Options in pyproject.toml break usage of `pip-sync`
I found it handy to add some often used options to the `[tools.pip-tools]`-section of my `pyproject.toml`. In particular I added:
```
[tool.pip-tools]
strip-extras = true
no-emit-index-url = true
no-emit-trusted-host = true
```
But now, calling `pip-sync requirements.txt` breaks with:
```
Error: No such config key 'no_emit_index_url'. (Possible options: emit_index_url, extra_index_url, index_url)
```
I can see, that using instead `emit-index-url = false` would have the same effect and, actually, is probably the correct solution to this. But given the poor documentation of the `pyproject.toml`-behaviour, we are left guessing and this is tedious and time-consuming. Moreover, `pip-compile` does not complain about the `pyproject`-options and happily uses them. Only `pip-sync` spoils the experience ;-)
So, a bit more consistency would be nice.

#### Environment Versions

1. OS Type: MS Windows 10
1. Python version: Python 3.11.5
1. pip version: pip 23.2.1
1. pip-tools version: pip-compile, version 7.3.0

#### Steps to replicate

1. Put any of the `no-*` options in `pyproject.toml`, e.g. `no-emit-index-url = true`
2. Call `pip-compile` and enjoy.
3. Call `pip-sync` and scratch your head, sort out your frustration and try to think, what could possibly be the issue here.

#### Expected result

The behaviour of `pip-compile` and `pip-sync` when using options from `pyproject.toml` is consistent, well documented and intuitive.

#### Actual result
The subset of options that work with `pip-sync` is unexpectedlty smaller than that of `pip-compile`.

chrysle:
Fixes #1988 

I can say I feel quite embarrassed.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -702,8 +702,9 @@ def _invert_negative_bool_options_in_config(
         # Transform config key to its equivalent in the CLI
         long_option = _convert_to_long_option(key)
         new_key = cli_opts[long_option].name if long_option in cli_opts else key
-        if new_key.startswith("no_"):
-            new_key = new_key[3:]
+        negative_option_prefix = "no_"
+        if new_key.startswith(negative_option_prefix):
+            new_key = new_key[len(negative_option_prefix):]
         assert new_key is not None
 
         # Invert negative boolean according to the CLI
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -702,7 +702,7 @@ def _invert_negative_bool_options_in_config(
         # Transform config key to its equivalent in the CLI
         long_option = _convert_to_long_option(key)
         new_key = cli_opts[long_option].name if long_option in cli_opts else key
-        if new_key.startswith('no_'):
+        if new_key.startswith("no_"):
             new_key = new_key[3:]
         assert new_key is not None
 
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -701,7 +701,7 @@ def _invert_negative_bool_options_in_config(
     for key, value in config.items():
         # Transform config key to its equivalent in the CLI
         long_option = _convert_to_long_option(key)
-        new_key = cli_opts[long_option].name if long_option in cli_opts else key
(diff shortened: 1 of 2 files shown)


piptools/utils.py
```diff
@@ -725,7 +725,11 @@ def _invert_negative_bool_options_in_config(
     for key, value in config.items():
         # Transform config key to its equivalent in the CLI
         long_option = _convert_to_long_option(key)
-        new_key = cli_opts[long_option].name if long_option in cli_opts else key
+        new_key = (
+            cli_opts[long_option].name
+            if long_option in cli_opts
+            else key.replace("no_", "")
+        )
```
atugushev, line 732:
```suggestion
        new_key = cli_opts[long_option].name if long_option in cli_opts else key
        if new_key.startswith('no_'):
            new_key = new_key[3:]
```

This is more explicit. This issue with `key.replace("no_", "")` is that it can potentially break options like `--no-code-no-errors`.

piptools/utils.py
```diff
@@ -726,6 +726,8 @@ def _invert_negative_bool_options_in_config(
         # Transform config key to its equivalent in the CLI
         long_option = _convert_to_long_option(key)
         new_key = cli_opts[long_option].name if long_option in cli_opts else key
+        if new_key.startswith("no_"):
+            new_key = new_key[3:]
```
webknjaz, line 730: I'd probably be more verbose about what that arbitrary constant `3` means:
```suggestion
        negative_option_preffix = "no_"
        if new_key.startswith(negative_option_preffix):
            new_key = new_key[len(negative_option_preffix):]
```

webknjaz (approved):

atugushev (approved):

PR #1998, opened 2023-10-07
Run pip-compile in env python rather than pip-tools python

### LINKED ISSUE #1087: Support for installation via pipx

georgek:
Resolves #1087 

Usually pip-tools is installed inside a project env and run from there. So it gets run in the same version of Python that the project is developed with. This doesn't work if pip-tools is installed somewhere else, like when installed with pipx.

The change here detects whether the currently running Python is different to the one on the PATH. Such a mismatch would probably always lead to unexpected behaviour. So in that case it creates a temporary venv with the Env Python, installs pip-tools, and runs it again there.

This approach could also be modified to allow a `python_executable` to be passed explicitly (and falling back to PATH by default), similar to how `pip-sync` is now. One could generated requirements file for multiple versions of Python (of course, you could already do it with tox, but this gives another option).

There are still some TODOs:

- Add similar behaviour to `pip-sync` (it already takes `python_executable` option, but can use env Python if not given),
- Improve the tests. It currently mocks the subprocess calls but doesn't check the calls are correct (in particular the passed env should be checked probably).

Finally, this is technically a breaking change. It's intended to resolve the `pipx` scenario (or even installed with OS package manager). I can't think of any reason why anyone would want pip-tools to run on an interpreter other than the one in the current env, but it will probably still break someone's workflow...

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -3,19 +3,23 @@ from __future__ import annotations
 import itertools
 import os
 import shlex
+import shutil
 import sys
 import tempfile
+import typing
 from pathlib import Path
+from subprocess import run  # nosec
 from typing import IO, Any, BinaryIO, cast
 
 import click
-from build import BuildBackendException
 from build.util import project_wheel_metadata
 from click.utils import LazyFile, safecall
 from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
 from pip._internal.utils.misc import redact_auth_from_url
 from pyproject_hooks import default_subprocess_runner, quiet_subprocess_runner
+
+from build import BuildBackendException
 
 from .._compat import parse_requirements
 from ..cache import DependencyCache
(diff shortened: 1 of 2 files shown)


piptools/scripts/compile.py
```diff
@@ -115,7 +118,65 @@ def _determine_linesep(
...
+        click.echo("Installing pip-tools in temporary venv...", err=True)
+        with tempfile.TemporaryDirectory(prefix="pip-tools-env-") as venv_path:
+            log.debug(f"Installing venv at {venv_path}")
+            run(  # nosec
+                [
+                    env_python,
+                    "-m",
+                    "venv",
+                    venv_path,
+                ],
```
chrysle, line 149: I do wonder: Could we just run `env_python -m piptools compile *sys.argv[1:]` instead of creating a nested virtualenv?

georgek:
OK, so a few things here:

It's not possible to run `env_python -m piptools ...` or something like that because piptools is not (necessarily) installed `env_python`. There is, of course, `PYTHONPATH` fuckery, but that causes all kinds of problems usually (as `current_python` and `env_python` are potentially different versions of Python)

We just essentially need pip-tools installed in `env_python`, and a venv is the best way to do that. So that's why I'm making my own venv.

I did consider leveraging `build` as this also deals with creating temporary venvs (in fact, the code in the PR is essentially what build does). The venv build makes is for its own purposes, though (it installed build tools and stuff). But of course we could still call `build` to create our own venv, however it doesn't seem to support creating a venv for another interpreter. The only way seems to be to launch a subprocess with the other interpreter (but I might have missed something). It would certainly be nicer to not have to call a subprocess to create the venv.

chrysle:
> Or perhaps the `--python` option can be used instead to avoid having to query `env_python` for its version at all?

Yes, sounds good! I'll try to get that pull request merged soon so you can rebase.

georgek:
One question is what to do with the version in the output header. With the change in #1898 we could get a header like:
```
# by the following command:
#
#    pip-compile --pip-args='--python=python3.8'
```
Which is OK I suppose because the python version is explicit (even if the Python 3.11 bit is now irrelevant).

But if we change it automatically I suppose we should change that header to say it was run with the `env_python` version?

chrysle:
Well, the header is generated in the `writer` package's following passage:

https://github.com/jazzband/pip-tools/blob/65b0d3672ec737db9d9a9dfef737f88b8344129d/piptools/writer.py#L122-L128

We could pass a reference to the target python to the output writer in the compile script by a new parameter, instead of requesting the `sys` version info (though this'd require making `cli()` and `cli_runner()` one again, which I think is the better solution).

georgek:
Yes, I think given that pip can essentially do what we need now (since last year it seems) the solution in this PR is no longer necessary. I will rewrite it soon. I do wonder if we can get the Python version without resorting to a subprocess call ourselves, though. Perhaps it can be queried via `pip`.

georgek:
Me neither. I don't have much time to work on this right now, unfortunately, but it's on my todo list. If nobody else can figure it out first I'll try to find a Windows box to test it on at some point.

Issue #2004, opened 2023-10-16
pip==23.3 breaking "extras" behavior


mattoberle:
Using `pip==23.3.x` can result in invalid `requirements.txt` for projects that declare extras.
A change in `pip==23.3.0` introduced the "canonicalization" of extras names:
- https://github.com/pypa/pip/pull/12002/files

The canonicalization leads to this outcome:
- ` requirements.in`: `sqlalchemy[postgresql_psycopg2binary]`
- `requirements.txt`: `sqlalchemy[postgresql-psycopg2binary]`

For many cases, this probably isn't an issue.
`psycopg2-binary` still makes it into `requirements.txt` as an individual entry.

For more "niche" cases, the extras still carry some significance.
eg. When using [rules_python](https://github.com/bazelbuild/rules_python) under the Bazel build system, the extras must properly resolve to build out a dependency graph.

#### Environment Versions

1. OS Type: Linux
1. Python version: `3.10.9`
1. pip version: `23.3.0`
1. pip-tools version: `7.3.0`

#### Steps to replicate

1. Create a simple `requirements.in` file.
```
sqlalchemy[postgresql_psycopg2binary]==1.4.47
```

2. Create a virtual environment, install `pip-tools` and the previous major `pip` version, run `pip-compile`.

```sh
python -m venv .venv
. ./.venv/bin/activate
pip install pip-tools==7.3.0 pip==22.3.1
pip-compile --no-strip-extras
```

3. Observe that the resulting `requirement.txt` file lists `sqlalchemy[postgresql_psycopg2binary]` (with an underscore).
```
greenlet==3.0.0
    # via sqlalchemy
psycopg2-binary==2.9.9
    # via sqlalchemy
sqlalchemy[postgresql_psycopg2binary]==1.4.47
    # via -r requirements.in
```

4. On the same virtual environment, upgrade to `pip>=23.3`, run `pip-compile`.

```
pip install pip==23.3.0
pip-compile --no-strip-extras
```

5. Observe that the resulting `requirement.txt` file lists `sqlalchemy[postgresql-psycopg2binary]` (with a hyphen).

```
  WARNING: sqlalchemy 1.4.47 does not provide the extra 'postgresql-psycopg2binary'

greenlet==3.0.0
    # via sqlalchemy
psycopg2-binary==2.9.9
    # via sqlalchemy
sqlalchemy[postgresql-psycopg2binary]==1.4.47
    # via
    #   -r requirements.in
    #   sqlalchemy
[...]

PR #2013, opened 2023-11-02
Ensure consistent extras formatting in output

AndydeCleyre:
This aims to resolve the discussion in #2004, ~~and is a WIP~~ (all contributions or alternatives are welcome).

~~`_compat`~~ **`utils`** now exports `install_req_from_line` ~~and `canonicalize_ireq`~~ for other modules to use. **This uses our `copy_install_requirement` which is updated to canonicalize extra strings.**

~~As of now, `canonicalize_ireq` only affects the extras attributes, as that's all we need. So it may be poorly named 🤷🏼~~

~~No tests yet.~~

To avoid circular imports, this relocates `PIP_VERSION` from `_compat` to `utils`, as IMO `utils` should be importable by as many other piptools modules as possible.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/_compat/pip_compat.py
+++ b/piptools/_compat/pip_compat.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import optparse
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Iterable, Iterator, Set, cast
+from typing import TYPE_CHECKING, Any, Iterable, Iterator, Set, cast
 
 import pip
 from pip._internal.cache import WheelCache
@@ -14,7 +14,11 @@ from pip._internal.models.link import Link
 from pip._internal.network.session import PipSession
 from pip._internal.req import InstallRequirement
 from pip._internal.req import parse_requirements as _parse_requirements
+from pip._internal.req.constructors import (
+    install_req_from_line as _install_req_from_line,
+)
 from pip._internal.req.constructors import install_req_from_parsed_requirement
+from pip._vendor.packaging.utils import canonicalize_name
 from pip._vendor.packaging.version import parse as parse_version
 from pip._vendor.pkg_resources import Requirement
 
@@ -73,6 +77,19 @@ class FileLink(Link):  # type: ignore[misc]
         return self._url
 
 
+def canonicalize_ireq(ireq: InstallRequirement) -> None:
+    if hasattr(ireq.req, "extras") and ireq.req.extras:
+        ireq.req.extras = set(map(canonicalize_name, ireq.req.extras))
+    if hasattr(ireq, "extras") and ireq.extras:
+        ireq.extras = set(map(canonicalize_name, ireq.extras))
+
+
+def install_req_from_line(*args: Any, **kwargs: Any) -> InstallRequirement:
+    ireq = _install_req_from_line(*args, **kwargs)
+    canonicalize_ireq(ireq)
+    return ireq
+
+
 def parse_requirements(
     filename: str,
     session: PipSession,
@@ -91,6 +108,7 @@ def parse_requirements(
             file_link = FileLink(install_req.link.url)
             file_link._url = parsed_req.requirement
             install_req.link = file_link
+        canonicalize_ireq(install_req)
         yield install_req
 
 
--- a/piptools/_compat/pip_compat.py
+++ b/piptools/_compat/pip_compat.py
@@ -2,9 +2,8 @@ from __future__ import annotations
 
 import optparse
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Iterable, Iterator, Set, cast
+from typing import TYPE_CHECKING, Iterable, Iterator, Set, cast
 
-import pip
 from pip._internal.cache import WheelCache
 from pip._internal.index.package_finder import PackageFinder
 from pip._internal.metadata import BaseDistribution
@@ -14,15 +13,8 @@ from pip._internal.models.link import Link
 from pip._internal.network.session import PipSession
 from pip._internal.req import InstallRequirement
 from pip._internal.req import parse_requirements as _parse_requirements
-from pip._internal.req.constructors import (
-    install_req_from_line as _install_req_from_line,
-)
 from pip._internal.req.constructors import install_req_from_parsed_requirement
-from pip._vendor.packaging.utils import canonicalize_name
-from pip._vendor.packaging.version import parse as parse_version
 from pip._vendor.pkg_resources import Requirement
-
-PIP_VERSION = tuple(map(int, parse_version(pip.__version__).base_version.split(".")))
 
 # The Distribution interface has changed between pkg_resources and
 # importlib.metadata, so this compat layer allows for a consistent access
@@ -30,6 +22,8 @@ PIP_VERSION = tuple(map(int, parse_version(pip.__version__).base_version.split("
 # (and later), but is overridable. `select_backend` returns what's being used.
 if TYPE_CHECKING:
     from pip._internal.metadata.importlib import Distribution as _ImportLibDist
+
+from ..utils import PIP_VERSION, canonicalize_ireq
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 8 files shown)


piptools/_compat/pip_compat.py
```diff
@@ -63,6 +67,19 @@ def _from_importlib(cls, dist: _ImportLibDist) -> Distribution:
         return cls(dist._dist.name, dist._dist.version, requires, dist.direct_url)
 
 
+def canonicalize_ireq(ireq: InstallRequirement) -> None:
```
webknjaz, line 70: Instead of mutating the incoming object, can we make a new one and return it? Changing things passed into function is going to be confusing, reducing maintainability/readability on the calling side.

AndydeCleyre: Sure, will change it today.

AndydeCleyre: So I'd like to use `utils.copy_install_requirement`, and `utils` currently imports `PIP_VERSION` from `_compat`. 

To avoid circular imports, I suggest to move the new stuff into `utils`, import it where needed from there (including into `_compat`), and removing `utils`'s current `PIP_VERSION` import from `_compat`.

I could inline the `PIP_VERSION` logic over in `utils`, but I suggest just moving `PIP_VERSION` into `utils` altogether.

AndydeCleyre: I'm noticing we have a mix of relative and absolute imports for piptools modules, so I'm not sure which is preferred, though I like relative imports.

AndydeCleyre: Alright, please have a look at the changes and let me know what you think. Thanks!

AndydeCleyre:
I wonder if this logic ought to be shoved into `copy_install_requirement`?

piptools/_compat/pip_compat.py
webknjaz: > I'm noticing we have a mix of relative and absolute imports for piptools modules, so I'm not sure which is preferred, though I like relative imports.

In general, I like relative imports. Though, there's cases where you want to patch things in other modules, and that might dictate how you import. Also, the context side of this is answered by https://youtu.be/x-kB2o8sd5c?si=yRWpqG3YmZRe_3AY&t=1038, I believe.

webknjaz: As for the circular imports, we should probably implement https://github.com/sanitizers/octomachinery/blob/master/tests/circular_imports_test.py in pip-tools, which is obviously outside of the scope of this PR.

webknjaz: By the way, one of the strategies I adopted that helps avoid surprise import-time side effects is not putting any logic into `__init__.py` files since they are evaluated when you import things levels deeper too. Of course, it's something to consider separately.

piptools/utils.py
```diff
@@ -88,6 +93,22 @@ def comment(text: str) -> str:
...
 
+def canonicalize_ireq(ireq: InstallRequirement) -> InstallRequirement:
+    """
+    Return a copy of ireq with canonicalized extras strings
+    """
+    ireq = copy_install_requirement(
+        ireq, extras=set(map(canonicalize_name, ireq.extras))
+    )
+    if ireq.req:
+        ireq.req.extras = set(ireq.extras)
```
webknjaz, line 104: Did you mean
```suggestion
        ireq.req.extras = set(map(canonicalize_name, ireq.extras))
```
?

AndydeCleyre: No, because they get normalized on line 101.

webknjaz:
> `copy_install_requirement`

No preference here.

AndydeCleyre:
> > `copy_install_requirement`
> 
> No preference here.

@atugushev Do you think this logic belongs in or out of `copy_install_requirement`?

piptools/utils.py
```diff
@@ -88,6 +93,22 @@ def comment(text: str) -> str:
     return click.style(text, fg="green")
 
 
+def canonicalize_ireq(ireq: InstallRequirement) -> InstallRequirement:
+    """
+    Return a copy of ireq with canonicalized extras strings
+    """
+    ireq = copy_install_requirement(
+        ireq, extras=set(map(canonicalize_name, ireq.extras))
```
atugushev, line 101: > > `copy_install_requirement`
>
> Do you think this logic belongs in or out of `copy_install_requirement`?

That sounds reasonable. Let's move this normalization to copy_install_requirement and use this function in all our InstallRequirement constructors.

piptools/utils.py
```diff
@@ -88,6 +93,22 @@ def comment(text: str) -> str:
...
+    )
+    if ireq.req:
+        ireq.req.extras = set(ireq.extras)
+    return ireq
+
+
+def install_req_from_line(*args: Any, **kwargs: Any) -> InstallRequirement:
+    return canonicalize_ireq(_install_req_from_line(*args, **kwargs))
```
atugushev, line 109: It might be worthwhile to normalize extras that could be passed to `pip-compile --extra` as well. What do you think @AndydeCleyre?

AndydeCleyre: I'd like to first see the current display of the extras in that case, but I haven't been able to get those extras-names to show up at all in the output.

piptools/_compat/pip_compat.py
AndydeCleyre: Looking good in this regard?

piptools/utils.py
AndydeCleyre: I'm not inclined to do it unless someone can demonstrate how or if it would make any difference.

AndydeCleyre:
I've just been rebasing this periodically. Are any changes needed? Is the PR still wanted?

@atugushev @webknjaz

tboddyspargo:
> Is the PR still wanted?

Thanks for keeping this updated, @AndydeCleyre. I, for one, am still interested in this PR!

AndydeCleyre:
I *thought* that pip was handling "canonicalized" extras, but:

```console
$ pip-sync /dev/null
$ pip install --dry-run --quiet --report=- 'sentry_sdk[pure_eval]' | jq -r '.install[].metadata.name'
```

```
urllib3
asttokens
certifi
executing
pure-eval
sentry-sdk
six
```

```console
$ pip install --dry-run --quiet --report=- 'sentry_sdk[pure-eval]' | jq -r '.install[].metadata.name'
```

```
urllib3
certifi
sentry-sdk
```

So I need to step back and request a sanity check about what we should be doing here. @atugushev ?

piptools/_compat/pip_compat.py
webknjaz: Marking this as resolved.

webknjaz:
Sorry, this fell off of my radar. Sounds like a bug in pip to fix as well?

AndydeCleyre:
> Sounds like a bug in pip to fix as well?

Maybe these:

- https://github.com/pypa/pip/issues/11445
- https://github.com/pypa/pip/issues/11715

AndydeCleyre:
> > Sounds like a bug in pip to fix as well?
> 
> Maybe these:
> 
> - https://github.com/pypa/pip/issues/11445
> - https://github.com/pypa/pip/issues/11715
> 

These were just marked as resolved! I haven't retested with pip from git yet but I'm optimistic.

chrysle (approved):
This looks good to me. I think we should move it forward? We can still respond to any issues this might cause afterwards (though I'm optimistic).

chrysle:
CI is fixed on `main` now. Could you rebase?

AndydeCleyre:
I'm getting a failure with `test_diff_should_not_uninstall`, but on main as well.

AndydeCleyre:
I rebased onto main, but it looks like the recently merged PR #2087 introduced flake8 and mypy violations in `test_pip_compat.py` and `pip_compat.py`, so those are now replicated here.

chrysle:
Hmm yes, I also wondered whether the one or other import was necessary, but trusted that `pre-commit` would catch that. I think that check should be marked as required to pass before merging, seems like it isn't.

chrysle:
The linters ought to be satisfied, now.

AndydeCleyre:
Thanks. I won't have access to a computer for a day but feel free to rebase again.

chrysle:
I'll wait for you. Happy time out!

chrysle:
Thanks!

PR #2022, opened 2023-11-20
Makes `compile` and `sync` directly runnable as modules

jamesbraza:
Before:

```bash
> python -m piptools.scripts.compile --help
```

After:

```bash
> python -m piptools.scripts.compile --help
Usage: python -m piptools.scripts.compile [OPTIONS] [SRC_FILES]...

  Compiles requirements.txt from requirements.in, pyproject.toml, setup.cfg,
  or setup.py specs.

Options:
  --version                       Show the version and exit.
...
```

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -521,3 +521,7 @@ def cli(
 
     if dry_run:
         log.info("Dry-run, so nothing updated.")
+
+
+if __name__ == "__main__":
+    cli()
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -251,3 +251,7 @@ def _get_installed_distributions(
         skip=[],
     )
     return [Distribution.from_pip_distribution(dist) for dist in dists]
+
+
+if __name__ == "__main__":
+    cli()


chrysle:
Thanks! Could you add a functional test, please?

jamesbraza:
@chrysle trying to do something [like this](https://stackoverflow.com/a/19011259), I am getting defeated by all the relative imports in `compile.py`.  So I opened https://github.com/jazzband/pip-tools/pull/2024, then I can come back to this the test of `__main__` behaviors

webknjaz:
Is that instead of `python -m piptools compile`? Why?

jamesbraza:
Oh yeah `python -m piptools compile` can work too actually, didn't know about that.  I am not attached to this PR if there's an alternate path to the same solution.  Feel free to close this out

webknjaz:
> Oh yeah `python -m piptools compile` can work too actually, didn't know about that. I am not attached to this PR if there's an alternate path to the same solution. Feel free to close this out

Yep. I think I discovered it by reading the code. It should probably be documented better.

webknjaz:
Oh, it's actually mentioned here https://pip-tools.rtfd.io/en/latest/#example-usage-for-pip-compile.

PR #2024, opened 2023-11-23
Absolute imports over relative imports

jamesbraza:
Trying to test `__main__` code per [this comment](https://github.com/jazzband/pip-tools/pull/2022#issuecomment-1821103433), it seems the best way is to use `importlib` as [instructed here](https://stackoverflow.com/a/19011259).

However, `importlib.machinery` doesn't play nice with relative imports.  So this PR moves the internals to use absolute imports.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -15,18 +15,18 @@ from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
 from pip._internal.utils.misc import redact_auth_from_url
 
-from .._compat import parse_requirements
-from ..build import build_project_metadata
-from ..cache import DependencyCache
-from ..exceptions import NoCandidateFound, PipToolsError
-from ..logging import log
-from ..repositories import LocalRequirementsRepository, PyPIRepository
-from ..repositories.base import BaseRepository
-from ..resolver import BacktrackingResolver, LegacyResolver
-from ..utils import dedup, drop_extras, is_pinned_requirement, key_from_ireq
-from ..writer import OutputWriter
-from . import options
-from .options import BuildTargetT
+import piptools.scripts.options as options
+from piptools._compat import parse_requirements
+from piptools.build import build_project_metadata
+from piptools.cache import DependencyCache
+from piptools.exceptions import NoCandidateFound, PipToolsError
+from piptools.logging import log
+from piptools.repositories import LocalRequirementsRepository, PyPIRepository
+from piptools.repositories.base import BaseRepository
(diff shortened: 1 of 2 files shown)


webknjaz (changes requested):
Plz no. If something doesn't implement imports correctly, fix that on the import side, instead of applying inconvenient hacks. FWIW, relative imports are more straightforward to follow.

jamesbraza:
Out of curiosity, why do you say absolute imports are an inconvenient hack?

In my own style, I find absolute imports easier to follow, and [PEP 8](https://peps.python.org/pep-0008/) is a proponent of absolute imports too:

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved

PEP 8 used to discourage relative imports more aggressively ([source](https://stackoverflow.com/a/4209771)) too:

> Relative imports for intra-package imports are highly discouraged. Always use the absolute package path for all imports.

I've also hit situations where relative imports can cause problems, depending on how one invokes `python` and configures `PYTHONPATH`. From my local testing tonight, absolute imports work with `importlib.machinery`, whereas relative imports don't.

---

Anyways, feel free to close this out if `pip-tools`'s style prefers relative imports, my need for this PR is no more because an alternate approach was pointed out in the source PR

webknjaz:
It's not currently enforced, so is probably up to each person.
https://mail.python.org/pipermail/python-dev/2010-October/104476.html suggests that in the times of Python 2.5, things were less stable/obvious maybe, more complicated. So I'm not surprised that people realized that it wasn't relevant anymore.
Plus, Python 3 changed a lot of stuff for the better.

> I've also hit situations where relative imports can cause problems, depending on how one invokes `python` and configures `PYTHONPATH`. From my local testing tonight, absolute imports work with `importlib.machinery`, whereas relative imports don't.

That's exactly why I don't want absolute imports: it encourages people to keep misconfiguring their environments, unfortunately. I don't think that it's a tolerable side effect.

We actually have an intentional relative import in ansible-core that is also a sanity check that we haven't broken our import machinery level hooks. A few months ago, colleagues from another team tripped over that and also proposed a hack that was removing that relative import, which of course was rejected.
I helped them realize that they were calling the machinery incorrectly, and that caused an actual fix to be found. Apparently, some people miss/forget that there's a `package` argument in the APIs of importlib, and it's meant to be used.
I have a hunch that if you pass what the package is to that API, then it'll magically work for you.

webknjaz:
@jamesbraza FWIW, I looked into that other PR and realized that what you suggest would probably mess with the import machinery. So I would still use subprocesses to test it.

PR #2041, opened 2024-01-03
Allow force-enabling or force-disabling colorized output

aneeshusa:
`click` allows emitted colored text via `click.style`,
which is used by pip-compile to emit colorized e.g. comments to stderr.
click uses auto-detection to only enable colors when writing to a TTY.

At Lyft, we operate pip-compile as a service:
pip-compile invocations are forwarded to a set of remote machines,
which leverage a shared cache for much faster compiles.
However, they run pip-compile as a subprocess without a TTY attached
meaning the output we stream back to the user is not colorized.

Therefore, add a `--color` and matching `--no-color` argument
to force-enable as well as force-disable color
(the default remains click's auto-detection).
This could also be used for e.g. running pip-compile in CI.

AFAIK pip-sync does not emit colored output,
so I did not add the `--color/--no-color` options there.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -74,6 +74,7 @@ def _determine_linesep(
 )
 @click.pass_context
 @options.version
+@options.color
 @options.verbose
 @options.quiet
 @options.dry_run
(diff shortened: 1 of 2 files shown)


chrysle (approved):
Thanks!

aneeshusa:
Thank you for taking a look @chrysle!

PR #2048, opened 2024-01-20
Fix config file search and inclusion in header

### LINKED ISSUE #1902: annotation --config is added even there is no configuration

chrysle:
Just stumbled over this issue again, so I thought I might as well fix it.

Closes #1902

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/options.py
+++ b/piptools/scripts/options.py
@@ -245,6 +245,7 @@ max_rounds = click.option(
 src_files = click.argument(
     "src_files",
     nargs=-1,
+    is_eager=True,
     type=click.Path(exists=True, allow_dash=True),
 )
 
@@ -328,7 +329,6 @@ config = click.option(
         f"Read configuration from TOML file. By default, looks for the following "
         f"files in the given order: {', '.join(DEFAULT_CONFIG_FILE_NAMES)}."
     ),
-    is_eager=True,
     callback=override_defaults_from_config_file,
 )
 
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -630,7 +630,7 @@ def select_config_file(src_files: tuple[str, ...]) -> Path | None:
     # NOTE: input.
     working_directory = Path.cwd()
     src_files_as_paths = (
-        (working_directory / src_file).resolve() for src_file in src_files or (".",)
+        (working_directory / src_file).resolve() for src_file in src_files + (".",)
     )
     candidate_dirs = (src if src.is_dir() else src.parent for src in src_files_as_paths)
     config_file_path = next(


piptools/utils.py
```diff
@@ -630,7 +630,7 @@ def select_config_file(src_files: tuple[str, ...]) -> Path | None:
     # NOTE: input.
     working_directory = Path.cwd()
     src_files_as_paths = (
-        (working_directory / src_file).resolve() for src_file in src_files or (".",)
+        (working_directory / src_file).resolve() for src_file in src_files + (".",)
```
webknjaz, line 633: Could you explain this? Obviously, there's some unspecified side effect in external logic, but I don't get it.

piptools/scripts/options.py
```diff
@@ -245,6 +245,7 @@ def _get_default_option(option_name: str) -> Any:
 src_files = click.argument(
     "src_files",
     nargs=-1,
+    is_eager=True,
```
webknjaz, line 248: Why is this needed?

chrysle: See https://github.com/jazzband/pip-tools/issues/1902#issuecomment-1624421303. Making only `src_files` eager doesn't work though, as it won't pick up the config if not specified, then.

sirosen:
I'm going to pick this up and try to get it into shape to include in the 7.4.2 release. I think this is a basically good fix and just needs some cleanup.

If I understand #1902, this is going down one of the viable paths. `--config` is marked eager and sets a callback, and that determines order of execution within `click`. If we want to establish better control over the ordering, a refactor may be appropriate -- I have ideas but that would inflate the size of the fix, and they can wait.

At a minimum, this needs
- a rebase
- better comments

I'm also going to assess to see if we can do more in terms of the tests.

sirosen:
After discussion, I'm moving this out of 7.4.2 to try to be very narrow in what extra work we include for a bugfix release. But I've assigned myself and plan to come back to it.

PR #2066, opened 2024-03-05
Fix `pip-compile` to properly exclude non-relevant options from output header

### LINKED ISSUE #2065: flag in COMPILE_EXCLUDE_OPTIONS showing up in header
#### Environment Versions

1. OS Type
  Windows mingw
1. Python version: `$ python -V`
  Python 3.11.8
1. pip version: `$ pip --version`
  pip 24.0
1. pip-tools version: `$ pip-compile --version`
  pip-compile, version 7.4.1.dev14 (`main` as of 03/04/2024)

#### Steps to replicate

1. `pip-compile --no-reuse-hashes -o requirements.txt`

#### Expected result

`--no-reuse-hashes` doesn't show up in the generated header comment.

#### Actual result

```
#
# by the following command:
#
#    pip-compile --generate-hashes --index-url=.. --no-reuse-hashes --output-file=requirements.txt --strip-extras --trusted-host=..
#
```

mjsir911:
Fixes #2065

Click context only had positive --reuse-hashes to loop over, inverse is automatically deduced.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -371,9 +371,14 @@ def get_compile_command(click_ctx: click.Context) -> str:
         # Get the latest option name (usually it'll be a long name)
         option_long_name = option.opts[-1]
 
+        negative_option = None
+        if option.is_flag and option.secondary_opts:
+            # get inverse flag --no-{option_long_name}
+            negative_option = option.secondary_opts[-1]
+
         # Exclude one-off options (--upgrade/--upgrade-package/--rebuild/...)
         # or options that don't change compile behaviour (--verbose/--dry-run/...)
-        if option_long_name in COMPILE_EXCLUDE_OPTIONS:
+        if {option_long_name, negative_option} & COMPILE_EXCLUDE_OPTIONS:
             continue
 
         # Exclude config option if it's the default one
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -374,6 +374,8 @@ def test_is_url_requirement_filename(caplog, from_line, line):
         (["--upgrade"], "pip-compile"),
         (["-P", "django"], "pip-compile"),
         (["--upgrade-package", "django"], "pip-compile"),
+        (["--reuse-hashes"], "pip-compile"),
+        (["--no-reuse-hashes"], "pip-compile"),
         # Check options
         (["--max-rounds", "42"], "pip-compile --max-rounds=42"),
         (["--index-url", "https://foo"], "pip-compile --index-url=https://foo"),


chrysle:
Thanks! As it should be worthwhile to include `--reuse-hashes` in the output header, could you try the following logic to test for the negative forms of boolean flags in `COMPILE_EXCLUDE_OPTIONS`?

```patch
diff --git a/piptools/utils.py b/piptools/utils.py
index 3b2061f..fc03846 100644
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -371,9 +371,13 @@ def get_compile_command(click_ctx: click.Context) -> str:
         # Get the latest option name (usually it'll be a long name)
         option_long_name = option.opts[-1]
 
+        negative_option = None
+        if option.is_flag:
+            negative_option = f"--no-{option_long_name.lstrip('--')}"
+
         # Exclude one-off options (--upgrade/--upgrade-package/--rebuild/...)
         # or options that don't change compile behaviour (--verbose/--dry-run/...)
-        if option_long_name in COMPILE_EXCLUDE_OPTIONS:
+        if option_long_name or negative_option in COMPILE_EXCLUDE_OPTIONS:
             continue
```
Moreover, since the pull request title will be included in the user-facing changelog, and I categorise this as a bugfix rather than a feature, as the title suggests, would you be fine with changing it to something like "Fix superfluous inclusion of `--no-reuse-hashes` flag in `pip-compile` output header"?

mjsir911:
> As it should be worthwhile to include --reuse-hashes in the output header

As it currently stands, `pip-compile` doesn't include options passed with default values:

https://github.com/jazzband/pip-tools/blob/11971512a541f914869603d0d2abcffd8c659fc4/piptools/utils.py#L341-L343

https://github.com/jazzband/pip-tools/blob/11971512a541f914869603d0d2abcffd8c659fc4/piptools/utils.py#L392-L394

& Yeah reuse-hashes is the default:
https://github.com/jazzband/pip-tools/blob/11971512a541f914869603d0d2abcffd8c659fc4/piptools/scripts/options.py#L229-L237

I can still apply your suggested changes, looks to be more clear and allows for embedding the inverse flags within the `COMPILE_EXCLUDE_OPTION`.

I do have to push back on the suggested commit summary change though, especially with your changes it does become solely a "fix `get_compile_command` to properly filter out specified inverse command arguments"

chrysle:
> Even if the above wasn't the case, your suggested changes wouldn't allow through a `--reuse-hashes` because `option_long_name` is still being filtered against. 

Ah, you're correct, I didn't realise that.

> especially with your changes it does become solely a "fix `get_compile_command` to properly filter out specified inverse command arguments"

How about "Fix `pip-compile` to properly exclude non-relevant options from output header"? The internal API should not be relevant to (typical) end users.

mjsir911:
> How about "Fix pip-compile to properly exclude non-relevant options from output header"? The internal API should not be relevant to (typical) end users.

Works for me, thanks for helping me along with this!

mjsir911:
OH I did misunderstand @chrysle I was focusing on the commit message. does that get added to the changelog too or is it purely the PR title?

chrysle:
> or is it purely the PR title?

Yes, since we're using the `release-drafter` GH Action, but we're planning to move away from it.

chrysle (approved):

PR #2077, opened 2024-04-15
Allow build deps as long as at least one setup file is in sources list.

### LINKED ISSUE #2076: Using pip-compile with --build-deps-for or --all-build-deps gives error when using multiple source files.
Using `pip-compile --all-build-deps requirments.in pyproject.toml` gives the error `--build-deps-for and --all-build-deps can be used only with the setup.py, setup.cfg and pyproject.toml specs.`.

According to the help, using multiple source files is valid, so any options that apply to at least one source file should be allowed as long as that option does not cause a conflict.

#### Environment Versions

1. pip-tools version: 7.4.0+

#### Steps to replicate

1. `pip-compile --all-build-deps requirments.in pyproject.toml`

#### Expected result

No error.

#### Actual result

`--build-deps-for and --all-build-deps can be used only with the setup.py, setup.cfg and pyproject.toml specs.`

Walter-Gates-Bose:
Resolves [2076](https://github.com/jazzband/pip-tools/issues/2076).

Passing ``--build-deps-for`` or ``--all-build-deps`` should not emit an error as long as at least one file in the sources list is a setup file.

Using `pip-compile --all-build-deps requirments.in pyproject.toml` gives the error `--build-deps-for and --all-build-deps can be used only with the setup.py, setup.cfg and pyproject.toml specs.`.

According to the help, using multiple source files is valid, so any options that apply to at least one source file should be allowed as long as that option does not cause a conflict.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -332,13 +332,6 @@ def cli(
     setup_file_found = False
     for src_file in src_files:
         is_setup_file = os.path.basename(src_file) in METADATA_FILENAMES
-        if not is_setup_file and build_deps_targets:
-            msg = (
-                "--build-deps-for and --all-build-deps can be used only with the "
-                "setup.py, setup.cfg and pyproject.toml specs."
-            )
-            raise click.BadParameter(msg)
-
         if src_file == "-":
             # pip requires filenames and not files. Since we want to support
             # piping from stdin, we need to briefly save the input from stdin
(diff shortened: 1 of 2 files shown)


webknjaz (changes requested):
I'm :-1: on this one since it opens up a can of worms (https://github.com/jazzband/pip-tools/issues/2076#issuecomment-2057957477) — the implication of such an interface would be that the build deps of everything listed in the requirement inputs would be collected which is simply untrue.

webknjaz:
cc @apljungquist ^

PR #2084, opened 2024-05-03
Removed strange (updating-requirements)= from README.md

denisSurkov:
Removed strange `(updating-requirements)=`

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,8 +191,6 @@ sqlparse==0.4.3
 And it will produce your `requirements.txt`, with all the Django dependencies
 (and all underlying dependencies) pinned.
 
-(updating-requirements)=
-
 ### Updating requirements
 
 `pip-compile` generates a `requirements.txt` file using the latest versions


webknjaz (changes requested):
It's MyST, it should be there.

WhyNotHugo:
It doesn't look like the README is rendered with `MyST`. The string `(updating-requirements)=` is just rendered as a regular paragraph on GitHub.

webknjaz:
Yep. GitHub uses its own flavor — GFM. But this document is also used in Sphinx. GH is known for breaking rendering, though (they recently broke RST even more than before..). So I don't really care about this. I think it's better to have nicer docs on RTD where we have full control over the markup. That includes the changelog (which is currently isn't even displayed there, I think).

One way to solve this would be to remove most of the README and link the Sphinx docs but that would remove the ability to see so much content on the GH repo front page.

denisSurkov:
Well, that's true. @webknjaz , sorry I didn't check it beforehand. I'll close PR.

https://pip-tools.readthedocs.io/en/stable/#updating-requirements

no strange string in docs

webknjaz:
FTR this is something to be discussed, clearly. And if someone would like to offer different opinions and implementation suggestions, I won't be in a way. Of course, it'll mean taking on more maintenance responsibility in this specific area of the project.
I just wanted to document why things are as they are today so that it could be taken into account in the future. Not many people were involved in plugging in Sphinx and then MyST so it may not be obvious to an average contributor.

WhyNotHugo:
I wasn't aware the sphinx was including the README. Thanks for your clarity.

PR #2087, opened 2024-05-07
Preserve relative paths for editable dependencies in output

macro1:
Accomplishes preservation of relative dependency paths by reverting back to input requirements definition for editable dependencies.

There are a couple prior attempts of this. The two I read through treated this like a formatting problem, I'm not sure it is. The issue, as has been pointed out in #204 is that paths are converted into urls in the form of absolute paths within pip.

If we maintain knowledge of the original relative paths within pip-tools, the values will be maintained more faithfully, and the complexity should be lower.

This PR accomplishes it by forcing the return value from `parse_requirements()` to give the values we need elsewhere. A better approach may be to return something different than `InstallRequirement` or preserve the relationship with the parsed requirement and provide both where it makes sense.

##### Contributor checklist

##### Maintainer checklist


chrysle:
The change looks good to me (we'll need a unit test, though). The CI failure is unrelated, I'll look into it.

> A better approach may be to return something different than InstallRequirement

Do you have anything in mind?

chrysle:
Ah yes, I love Windows, too. Will review soon.

tests/test_pip_compat.py
chrysle, line 1: Wouldn't this fit in `test_repository_pypi.py`, too?

macro1: I can move it. I didn't look around too much tbh. I generally name the test modules after the modules being tested (personal convention I suppose).

tests/test_pip_compat.py
```diff
@@ -0,0 +1,32 @@
...
+
+from .constants import PACKAGES_RELATIVE_PATH
+
+
+def test_parse_requirements_preserve_editable_relative(repository):
+    test_package_path = str(
+        PurePosixPath(Path(PACKAGES_RELATIVE_PATH)) / "small_fake_a"
+    )
+
+    infile = tempfile.NamedTemporaryFile("w", delete=False)
```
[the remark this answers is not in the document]
macro1: That would likely work, as it's directory-based (so we can close the file for tests running on Windows). I'll update.

tests/test_pip_compat.py
chrysle: Hmm, we could as well leave it where it is, probably it would contribute to clarity.

chrysle (approved):

chrysle:
@macro1 Could you look into fixing the linting errors?

Issue #1326, opened 2021-02-16
multiple pip-compile runs with multiple python versions -> single requirements.txt


RuRo:
#### What's the problem this feature will solve?

I have a single `requirements.in` file, I would like to generate a single `requirements.txt` file, that would work with multiple python versions. 
Afaik, currently, the only way to solve this issue is to run `pip-compile` N times - once per python version.
This however leaves you with N different `requirements.txt` files with mutually unsynchronized versions.

> Disclamer:
> 
> I am aware of the limitations of `pip/PyPI` dependency resolution (#635, #639).
> I am not asking `pip-tools` to guess the dependencies for different python versions.
> 
> **I am willing to run `pip-compile` for each python version** that I want to support.
> What I want is a way to aggregate the results of these multiple runs of `pip-compile`.

#### Example problem

Let's say, that my `requirements.in` contains just `required-package` (this is a hypothetical package, that doesn't actually exist).

`required-package` has the following versions in different versions of python:
- `python3.6` has `required-package 1.0.0` which requires packages `a` and `b`
- `python3.7` has `required-package 1.0.0` which requires packages `a` and `c`
- `python3.8` has `required-package 1.0.0` which requires just the package `c`
  `python3.8` **also** has a newer version - `required-package 2.0.0` which is **only** available in `python >= 3.8` and requires just the package `d`
For simplicity, let's assume that all other packages don't have any further dependencies and have only one version - `1.0.0`.

If I understand it correctly, currently the only thing I can do is create 3 `requirements.txt`, one for each python version:
```
# python3.6 -m piptools compile requirements.in --no-header --no-annotate --output-file=requirements36.txt
required-package==1.0.0
a==1.0.0
b==1.0.0
```
```
# python3.7 -m piptools compile requirements.in --no-header --no-annotate --output-file=requirements37.txt
required-package==1.0.0
a==1.0.0
c==1.0.0
```
```
# python3.8 -m piptools compile requirements.in --no-header --no-annotate --output-file=requirements38.txt
required-package==2.0.0
d==1.0.0
```
There are 2 big problems with the above result:
1) `3.6` and `3.7` require different packages `b` and `c`
2) `3.8` has a different version of `required-package`

#### Desired result

Ideally, I would like to have a `common.requirements.txt` along the lines of
```
required-package==1.0.0
a==1.0.0
b==1.0.0; python_version < "3.7"
c==1.0.0; python_version >= "3.7"
```
More realistically, I would be OK with something like
```
[...]

PR #2098, opened 2024-06-10
Run pre-commit for changes to pyproject.toml

davidwtbuxton:
The pre-commit hook should run when a git change includes edits to a pyproject.toml file, because that file may be used for listing a project's dependencies.

No tests were updated.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,5 @@
   description: Automatically compile requirements.
   entry: pip-compile
   language: python
-  files: ^requirements\.(in|txt)$
+  files: ^pyproject\.toml|requirements\.(in|txt)$
   pass_filenames: false


webknjaz:
Hi there! Thanks for the contrib. My feeling is that we shouldn't attempt to enumerate every possible input file there. The reason is that every end-user has unique configuration, and they should identify which files they use and which ones shouldn't be accessed by `pip-tools`. There's also `setup.cfg` / `setup.py`, and numerous conventions on where to put the `.in` + `.txt` file pairs. For example, I have `requirements/*.in` in projects where I use pip-tools to make a DIY lockfile for the environments.

So my initial feeling is that we shouldn't accept this change, which might even be breaking for a number of the end-users because of the changed defaults. Instead, it would be useful to have a PR explaining the use in the README/documentation.

davidwtbuxton:
Hi @webknjaz thanks for reviewing this pull request and explaining why you are not in favour of accepting the change. I think that pyproject.toml will be increasingly popular for specifying dependencies, so perhaps it will make sense to change this default in future, but I am happy for this change to be closed and not merged right now.

webknjaz:
@davidwtbuxton `pyproject.toml` cannot possibly encapsulate the diversity of use cases, though. People tend to confuse different types of dependencies and environments, and forget what they lock, *actually*: https://github.com/jazzband/pip-tools/issues/1326#issuecomment-1834517252. Until https://discuss.python.org/t/lock-files-again-but-this-time-w-sdists/46593 is a thing, I doubt there's going to be any useful convention to follow.

Issue #2104, opened 2024-06-23
CI failure: `pip freeze` no longer protects `setuptools`, `distribute` or `wheel` on Python 3.12+


ncoghlan:
Some CI tests are failing due to this change in `pip` to no longer protect `setuptools`, `distribute` or `wheel` by default on Python 3.12+: https://github.com/pypa/pip/commit/5dc65eabb75f89d4f4749b6c764042c227f6870a

That has been in the `pip` dev branch for a while, and was published in `pip 24.1` in the past few days.

This change affects the `pip-tools` test suite by way of the `PACKAGES_TO_IGNORE` value, which reads the `pip freeze` internal API iviahttps://github.com/jazzband/pip-tools/blob/53309647980e2a4981db54c0033f98c61142de0b/piptools/_compat/pip_compat.py#L101

Simplest CI fix would likely be to drop those packages from the affected test cases, since their handling depends on the versions of other components.

ncoghlan:
The `test_compile_recursive_extras_build_targets` failures feels like it might be due to https://github.com/pypa/pip/pull/12449 (`pip` used to inject `wheel` as an implicit build dependency for every `setup.py` project)

chrysle:
> The `test_compile_recursive_extras_build_targets` failure feels like it might be due to [pypa/pip#12449](https://github.com/pypa/pip/pull/12449) (`pip` used to inject `wheel` as an implicit build dependency for every `setup.py` project)

Could you remove this line, then?

https://github.com/jazzband/pip-tools/blob/53309647980e2a4981db54c0033f98c61142de0b/tests/test_cli_compile.py#L3458

chrysle:
> Some CI tests are failing due to this change in `pip` to no longer protect `setuptools`, `distribute` or `wheel` by default on Python 3.12+: [pypa/pip@5dc65ea](https://github.com/pypa/pip/commit/5dc65eabb75f89d4f4749b6c764042c227f6870a)

We don't test on 3.12+ yet, though...

chrysle:
> The `test_compile_recursive_extras_build_targets` failure feels like it might be due to [pypa/pip#12449](https://github.com/pypa/pip/pull/12449) (`pip` used to inject `wheel` as an implicit build dependency for every `setup.py` project)

True, I've opened a PR, but there are still the [failing Windows tests](https://github.com/jazzband/pip-tools/actions/runs/9658693126/job/26640434682?pr=2106)... Could you look into them?

[...]

PR #2106, opened 2024-06-25
 Fix build deps compilation for `setuptools < 70.1.0`

chrysle:
See https://github.com/jazzband/pip-tools/pull/1681#issuecomment-2212541289 for context.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/build.py
+++ b/piptools/build.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import collections
 import contextlib
+import os
 import pathlib
 import sys
 import tempfile
@@ -119,6 +120,7 @@ def build_project_metadata(
     src_file: pathlib.Path,
     build_targets: tuple[str, ...],
     *,
+    upgrade_packages: tuple[str, ...] | None = None,
     attempt_static_parse: bool,
     isolated: bool,
     quiet: bool,
@@ -159,7 +161,9 @@ def build_project_metadata(
             return project_metadata
 
     src_dir = src_file.parent
-    with _create_project_builder(src_dir, isolated=isolated, quiet=quiet) as builder:
+    with _create_project_builder(
+        src_dir, upgrade_packages=upgrade_packages, isolated=isolated, quiet=quiet
+    ) as builder:
         metadata = _build_project_wheel_metadata(builder)
         extras = tuple(metadata.get_all("Provides-Extra") or ())
         requirements = tuple(
@@ -182,7 +186,11 @@ def build_project_metadata(
 
 @contextlib.contextmanager
 def _create_project_builder(
-    src_dir: pathlib.Path, *, isolated: bool, quiet: bool
+    src_dir: pathlib.Path,
+    *,
+    upgrade_packages: tuple[str, ...] | None = None,
+    isolated: bool,
+    quiet: bool,
 ) -> Iterator[build.ProjectBuilder]:
     if quiet:
         runner = pyproject_hooks.quiet_subprocess_runner
@@ -192,6 +200,15 @@ def _create_project_builder(
     if not isolated:
         yield build.ProjectBuilder(src_dir, runner=runner)
         return
+
+    if upgrade_packages is not None:
+        # Write packages to upgrade to a temporary file to set as
+        # constraints for the installation to the builder environment,
+        # in case build requirements are among them
+        tmpfile = tempfile.NamedTemporaryFile(mode="w+t", delete=False)
+        tmpfile.write("\n".join(package for package in upgrade_packages))
+        tmpfile.flush()
+        os.environ["PIP_CONSTRAINT"] = tmpfile.name
 
     with build.env.DefaultIsolatedEnv() as env:
         builder = build.ProjectBuilder.from_isolated_env(env, src_dir, runner)
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -3431,7 +3431,6 @@ def test_compile_recursive_extras_build_targets(runner, tmp_path, current_resolv
             """
         )
     )
-    (tmp_path / "constraints.txt").write_text("wheel<0.43")
     out = runner.invoke(
         cli,
         [
@@ -3446,8 +3445,6 @@ def test_compile_recursive_extras_build_targets(runner, tmp_path, current_resolv
             "--find-links",
             os.fspath(MINIMAL_WHEELS_PATH),
             os.fspath(tmp_path / "pyproject.toml"),
-            "--constraint",
-            os.fspath(tmp_path / "constraints.txt"),
             "--output-file",
             "-",
         ],
@@ -3455,6 +3452,60 @@ def test_compile_recursive_extras_build_targets(runner, tmp_path, current_resolv
     expected = rf"""foo[footest] @ {tmp_path.as_uri()}
 small-fake-a==0.2
 small-fake-b==0.3
+
+# The following packages are considered to be unsafe in a requirements file:
+# setuptools
+"""
+    try:
+        assert out.exit_code == 0
+        assert expected == out.stdout
+    except Exception:  # pragma: no cover
+        print(out.stdout)
+        print(out.stderr)
+        raise
+
+
+@backtracking_resolver_only
+def test_compile_build_targets_setuptools_no_wheel_dep(
+    runner, tmp_path, current_resolver
+):
+    """
+    Regression test for https://github.com/jazzband/pip-tools/pull/1681#issuecomment-2212541289
+    """
+    (tmp_path / "pyproject.toml").write_text(
+        dedent(
+            """
+            [project]
+            name = "foo"
+            version = "0.0.1"
+            dependencies = ["small-fake-a"]
+            """
+        )
+    )
+    (tmp_path / "constraints.txt").write_text("wheel<0.43")
+    out = runner.invoke(
+        cli,
+        [
+            "--build-isolation",
+            "--no-header",
+            "--no-annotate",
+            "--no-emit-options",
+            "--extra",
+            "dev",
+            "--build-deps-for",
+            "wheel",
+            "--find-links",
+            os.fspath(MINIMAL_WHEELS_PATH),
+            os.fspath(tmp_path / "pyproject.toml"),
+            "--constraint",
+            os.fspath(tmp_path / "constraints.txt"),
+            "--upgrade-package",
+            "setuptools < 70.1.0",  # setuptools>=70.1.0 doesn't require wheel any more
+            "--output-file",
+            "-",
+        ],
+    )
+    expected = r"""small-fake-a==0.2
 wheel==0.42.0
 
 # The following packages are considered to be unsafe in a requirements file:
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -684,8 +684,8 @@ def test_url_package(runner, line, dependency, generate_hashes):
         ["-n", "--rebuild", "--no-build-isolation"]
         + (["--generate-hashes"] if generate_hashes else []),
     )
-    assert out.exit_code == 0
     assert dependency in out.stderr
+    assert out.exit_code == 0
 
 
 @pytest.mark.parametrize(
@@ -787,13 +787,13 @@ def test_direct_reference_with_extras(runner):
             "pip-tools[testing,coverage] @ git+https://github.com/jazzband/pip-tools@6.2.0"
         )
     out = runner.invoke(cli, ["-n", "--rebuild", "--no-build-isolation"])
-    assert out.exit_code == 0
     assert (
         "pip-tools[coverage,testing] @ git+https://github.com/jazzband/pip-tools@6.2.0"
         in out.stderr
     )
     assert "pytest==" in out.stderr
     assert "pytest-cov==" in out.stderr
+    assert out.exit_code == 0
 
 
 def test_input_file_without_extension(pip_conf, runner):
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -370,6 +370,7 @@ def cli(
                 metadata = build_project_metadata(
                     src_file=Path(src_file),
                     build_targets=build_deps_targets,
+                    upgrade_packages=upgrade_packages,
                     attempt_static_parse=not bool(build_deps_targets),
                     isolated=build_isolation,
                     quiet=log.verbosity <= 0,


webknjaz (approved):

webknjaz (changes requested):
This is not a correct fix. It just shoves the problem under the carpet. We need to address the underlying problem, which is revealed not due to `pip` but because `setuptools` changed. Yet, the bug is in pip-tools: https://github.com/jazzband/pip-tools/issues/2104#issuecomment-2212544824. The correct change in the test is to include something like `-P 'setuptools < 70.1.0'`. But that should be accompanied with a fix for the bug.

webknjaz:
Alternatively, it's even better to emulate such a situation with artificial package stubs so that it doesn't need to hit the network.

chrysle:
@webknjaz I reworked the PR; please have a look. Of course there are still the failing Windows tests, can we just drop that horrible Microsoft product ;-)?

WhyNotHugo:
Slightly changing the order of assertions should help see more details into the errors:

```diff
diff --git a/tests/test_cli_compile.py b/tests/test_cli_compile.py
index c5031fc..8db0038 100644
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -787,13 +787,13 @@ def test_direct_reference_with_extras(runner):
             "pip-tools[testing,coverage] @ git+https://github.com/jazzband/pip-tools@6.2.0"
         )
     out = runner.invoke(cli, ["-n", "--rebuild", "--no-build-isolation"])
-    assert out.exit_code == 0
     assert (
         "pip-tools[coverage,testing] @ git+https://github.com/jazzband/pip-tools@6.2.0"
         in out.stderr
     )
     assert "pytest==" in out.stderr
     assert "pytest-cov==" in out.stderr
+    assert out.exit_code == 0
 
 
 def test_input_file_without_extension(pip_conf, runner):
```

Generally, I prefer to compare `out.stderr` before checking the status code in tests; if `stderr` isn't the expected output, `pytest` will print it when failing the assertion.

webknjaz:
@chrysle the windows failure is because of the deprecation warning in pip, right?

chrysle:
Could you elaborate on that? It seems that the resolver only found some pre-versions of a package, only I don't know much because the output is shortened.

webknjaz:
Yeah, it's weird. Somebody needs to use https://github.com/marketplace/actions/debugging-with-tmate to see what's happening on Windows. The vars in CI seem to suggest that `stderr` is incomplete or something. Can it be some output buffering specific to windows? Do we need `PYTHONUNBUFFERED=1` in tests?

tests/test_cli_compile.py
```diff
@@ -3446,15 +3445,67 @@ def test_compile_recursive_extras_build_targets(runner, tmp_path, current_resolv
...
+# The following packages are considered to be unsafe in a requirements file:
+# setuptools
+"""
+    try:
+        assert out.exit_code == 0
+        assert expected == out.stdout
+    except Exception:  # pragma: no cover
+        print(out.stdout)
+        print(out.stderr)
+        raise
```
webknjaz, line 3465: Let's not include this hack with more complex logic. This should be better:
```suggestion
    assert 0 == out.exit_code and expected == out.stdout, f"\n{out.stdout=}\n\n{out.stderr=}\n\n"
```

tests/test_cli_compile.py
```diff
@@ -3446,15 +3445,67 @@ def test_compile_recursive_extras_build_targets(runner, tmp_path, current_resolv
...
+            "dev",
+            "--build-deps-for",
+            "wheel",
+            "--find-links",
+            os.fspath(MINIMAL_WHEELS_PATH),
+            os.fspath(tmp_path / "pyproject.toml"),
+            "--constraint",
+            os.fspath(tmp_path / "constraints.txt"),
+            "--upgrade-package",
+            "setuptools < 70.1.0",  # setuptools>=70.1.0 doesn't require wheel any more
```
webknjaz, line 3503: FWIW, it'd be better to make it not need internet access. But that's not a blocker, just something to think about for the future.

tests/test_cli_compile.py
```diff
@@ -684,8 +684,8 @@ def test_url_package(runner, line, dependency, generate_hashes):
         ["-n", "--rebuild", "--no-build-isolation"]
         + (["--generate-hashes"] if generate_hashes else []),
     )
-    assert out.exit_code == 0
     assert dependency in out.stderr
+    assert out.exit_code == 0
```
webknjaz, line 688: I'd prefer not to include unrelated changes right in this PR. But if you keep this, it's probably better to merge the instructions into a single `assert` statement. This way, it'd be following the [AAA principle](https://docs.pytest.org/en/stable/explanation/anatomy.html) to the letter: one test == one assertion.
```suggestion
    assert 0 == out.exit_code and dependency in out.stderr
```

webknjaz:
@chrysle could you try rebasing?

webknjaz:
Closed+reopened to re-trigger the CI as the logs got stale and garbage-collected.

webknjaz:
> Of course there are still the failing Windows tests

Interestingly, the “previous” tests pass, even on Windows. But the “latest” crash all over the place because of the relative vs. absolute path differences. “main” fails as well.
“previous” is set to `pip==22.2.*` currently.

Honestly, I'm leaning towards making the tests more stable by pinning pip in regular push/PR runs and only having it hit newer versions in nightlies...

webknjaz:
I've converted this to draft to prevent accidental merging ahead of time. I'll undraft it once I'm done experimenting.

piptools/build.py
```diff
@@ -193,6 +201,15 @@ def _create_project_builder(
...
         return
 
+    if upgrade_packages is not None:
+        # Write packages to upgrade to a temporary file to set as
+        # constraints for the installation to the builder environment,
+        # in case build requirements are among them
+        tmpfile = tempfile.NamedTemporaryFile(mode="w+t", delete=False)
+        tmpfile.write("\n".join(package for package in upgrade_packages))
+        tmpfile.flush()
+        os.environ["PIP_CONSTRAINT"] = tmpfile.name
```
webknjaz, line 211: So apparently this is not being cleaned up: neither the env file, nor the file descriptor behind `tempfile.NamedTemporaryFile()` is closed.

This was mentioned as necessary in https://github.com/jazzband/pip-tools/pull/1681#discussion_r1668032908.

The proper way to handle it would be to use a context manager.

webknjaz: FTR, I'm taking over this. I moved it into another CM locally. Will push shortly.

piptools/build.py
```diff
@@ -159,7 +161,9 @@ def build_project_metadata(
             return project_metadata
 
     src_dir = src_file.parent
-    with _create_project_builder(src_dir, isolated=isolated, quiet=quiet) as builder:
+    with _create_project_builder(
+        src_dir, upgrade_packages=upgrade_packages, isolated=isolated, quiet=quiet
```
webknjaz, line 165:
```suggestion
        src_dir,
        upgrade_packages=upgrade_packages,
        isolated=isolated,
        quiet=quiet,
```

piptools/build.py
```diff
@@ -180,9 +184,41 @@ def build_project_metadata(
         )
 
 
+@contextlib.contextmanager
+def _temporary_constraints_file_set_for_pip(
+    upgrade_packages: tuple[str, ...]
```
webknjaz, line 189:
```suggestion
    upgrade_packages: tuple[str, ...],
```

tests/test_cli_compile.py
```diff
@@ -3446,15 +3444,61 @@ def test_compile_recursive_extras_build_targets(runner, tmp_path, current_resolv
...
+
+# The following packages are considered to be unsafe in a requirements file:
+# setuptools
+"""
+    assert 0 == out.exit_code and expected == out.stdout, f"\n{out.stdout=}\n\n{out.stderr=}\n\n"
+
+
+@backtracking_resolver_only
+def test_compile_build_targets_setuptools_no_wheel_dep(
+    runner, tmp_path, current_resolver
```
webknjaz, line 3463:
```suggestion
    runner,
    tmp_path,
    current_resolver,
```

piptools/build.py
```diff
@@ -193,7 +243,13 @@ def _create_project_builder(
...
-    with build.env.DefaultIsolatedEnv() as env:
+    maybe_pip_constrained_context = (
+        contextlib.nullcontext()
+        if upgrade_packages is None
+        else _temporary_constraints_file_set_for_pip(upgrade_packages)
+    )
+
+    with maybe_pip_constrained_context, build.env.DefaultIsolatedEnv() as env:
         builder = build.ProjectBuilder.from_isolated_env(env, src_dir, runner)
         env.install(builder.build_system_requires)
```
webknjaz, line 254: So, while debugging via tmate, I discovered that on Windows, this raises a `SubprocessError` from somewhere within `pypa/build`. I'm not yet sure what's happening. The command was `pip install --pep517`something.. I didn't get to save it since the SSH session expires in about 40 minutes and I just lost it. I think, it's https://github.com/pypa/build/blob/88ae833/src/build/env.py#L257-L265.

This then causes the visible error we saw in the logs — `[RuntimeError("generator didn't yield")]`(https://github.com/python/cpython/blob/6bc3e83/Lib/contextlib.py#L143). I'm not entirely sure why, though — might be a CPython bug.

Now that I'm thinking about it, I'm curious if the order of the CMs in the `with` statement matters. Perhaps, `build.env.DefaultIsolatedEnv()` gets interrupted and processes the `__exit__()` first, suppressing the error. Then, the control flow of `_create_project_builder()` is interrupted, never reaching the `yield`, tripping the validation in the outer `@contextlib.contextmanager()` decorator.
Though, I don't understand why `SubprocessError` doesn't escape the `with` block… can it be https://github.com/python/cpython/blob/6bc3e83/Lib/contextlib.py#L163-L167?

webknjaz: Calling `_create_project_builder()` or `build_project_metadata()` doesn't seem to be enough to reproduce this. I suspect, Click might be the cause.

piptools/build.py
```diff
@@ -180,9 +187,52 @@ def build_project_metadata(
...
+
+@contextlib.contextmanager
+def _temporary_constraints_file_set_for_pip(
+    upgrade_packages: tuple[str, ...],
+) -> Iterator[None]:
+    with tempfile.NamedTemporaryFile(mode="w+t") as tmpfile:
+        # Write packages to upgrade to a temporary file to set as
+        # constraints for the installation to the builder environment,
+        # in case build requirements are among them
+        tmpfile.write("\n".join(package for package in upgrade_packages))
```
webknjaz, line 222:
```suggestion
        tmpfile.write("\n".join(upgrade_packages))
```

webknjaz:
UPD: it looks like `pyproject.toml` stubs in the tests that don't have a `[build-system]` section cause `build` to crash on Windows trying to access the `setuptools.build_meta:__legacy__` fallback.

webknjaz:
Oh, it looks like specifying it manually makes another module non-importable: `out.stderr="Backend 'setuptools.build_meta' is not available.\nFailed to parse C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\popen-gw2\\test_compile_build_targets_set0\\pyproject.toml\n"`. So that ain't it.
I bet there's some escaping issue with Windows paths and/or `PYTHONPATH` etc.

webknjaz:
*(documenting the troubleshooting env setup procedure)*

Here's what I do upon connection to the windows job over tmate (== tmux over SSH):
```shell
.tox/piplatest-coverage/Scripts/pip install -e .
.tox/piplatest-coverage/Scripts/pytest --no-cov -n0 -svvvvv -l tests/test_cli_compile.py::test_compile_build_targets_setuptools_no_wheel_dep
.tox/piplatest-coverage/Scripts/pip-compile.exe --build-deps-for wheel C:/msys64/tmp/pytest-of-runneradmin/pytest-0/test_compile_build_targets_set0/pyproject.toml --output-file - -vvvvvv
```

This lets me populate the temporary files for inspection (the pytest command). It also allows me to edit files in-place (`Ctrl+b c` to create a new tmux tab; `Ctrl+b n` / `Ctrl+b p` / `Ctrl+b 1` / `Ctrl+b 2` to switch tabs) and rerun pytest or `pip-compile` directly.

Editing files is done via `nano` since `vim` does not play well with that windows/cygwin env, I haven't checked why.

Scrolling the terminal output: `Ctrl+b PgUp`.

webknjaz:
> ```shell
> .tox/piplatest-coverage/Scripts/pip-compile.exe --build-deps-for wheel C:/msys64/tmp/pytest-of-runneradmin/pytest-0/test_compile_build_targets_set0/pyproject.toml --output-file - -vvvvvv
> ```


Better repro without running `pytest` upfront:
```console
$ .tox/piplatest-coverage/Scripts/pip-compile.exe --build-deps-for wheel tests/test_data/packages/small_fake_with_pyproject/pyproject.toml --output-file - -vvvvvv
```

webknjaz:
I verified that setting the `PIP_CONSTRAINT` env var is exactly what breaks on Windows. Commenting it out makes the command pass.

webknjaz:
```console
# test-venv/Scripts/pip-compile --build-deps-for wheel tests/test_data/packages/small_fake_with_pyproject/pyproject.toml --output-file - -vvvvvv
Using pip-tools configuration defaults found in 'pyproject.toml'.
Creating isolated environment: venv+pip...
tmpfile.name='C:\\msys64\\tmp\\tmpti95poux'
Installing packages in isolated environment:
- setuptools < 70
Getting metadata for wheel...
Backend 'setuptools.build_meta' is not available.
Failed to parse D:\a\pip-tools\pip-tools\tests\test_data\packages\small_fake_with_pyproject\pyproject.toml
```

I wonder if reversing the slashes is needed or something.

webknjaz:
Ah-ha!

I think I understand what's happening now! The temporary file is open and flushed, but the file descriptor remains open, which on Windows prevents other processes from accessing it. So when we call `build`, it calls `pip install` that sees `PIP_CONSTRAINT`, attempts to open it, and likely fails to install the build deps because of permission denied.

It seems like calling `.close()` upon `.flush()` (or should it be instead, even?) makes it accessible. We just have to use `delete_on_close=False` so that `.close()` does not also remove the file from disk right away. With `delete=True` (which is the default), the CM will clean it up on exit.

Some of this is mentioned @ https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

webknjaz:
Urgh.. looks like `delete_on_close` got introduced in Python 3.12. At least I now have an idea of a workaround. Will try to complete this PR today.

piptools/build.py
```diff
@@ -193,10 +254,17 @@ def _create_project_builder(
...
+        contextlib.nullcontext()
+        if upgrade_packages is None
+        else _temporary_constraints_file_set_for_pip(upgrade_packages)
+    )
+
     with build.env.DefaultIsolatedEnv() as env:
         builder = build.ProjectBuilder.from_isolated_env(env, src_dir, runner)
-        env.install(builder.build_system_requires)
-        env.install(builder.get_requires_for_build("wheel"))
+        with maybe_pip_constrained_context:
```
webknjaz, line 265: TODO: check if moving this back into the outer CM works.

webknjaz (approved):
This will be a part of the CI fix, I excluded the pip pinning commit as it will be coming from #2142.

Test run with pip pins: https://github.com/jazzband/pip-tools/actions/runs/12292796650.

Please do not add any changes to this PR.

PR #2107, opened 2024-06-26
Support scripts with inline script metadata as input files

### LINKED ISSUE #2027: Implement support for PEP 723 "Inline script metadata"

It's just a new source of deps.

$sbj

#### Alternative Solutions

... not doing this?

#### Additional context

It is provisional and they ask that the user tooling doesn't expose this just yet: https://github.com/pypa/pipx/issues/1128

* https://peps.python.org/pep-0723/
* https://discuss.python.org/t/pep-722-723-decision/36763

chrysle:
Closes #2027.

##### Contributor checklist

##### Maintainer checklist


chrysle:
@webknjaz please review; there isn't any other active maintainer left, as it seems..

WhyNotHugo:
What's the use-case here? Inline-scripts are designed to be a standalone file which contains everything needed to run it. Creating a lock file defeats this purpose.

chrysle:
Hmm, suppose you have a Python script that has dependencies, is probably often in use and not part of a package with corresponding metadata; then setting up an environment fastly would be possible through this functionality.

chrysle:
@WhyNotHugo would you like to review this? We're really short on help ATM.

tests/test_cli_compile.py
```diff
@@ -3771,3 +3772,70 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
 
     assert out.exit_code == 0, out
+
+
+def test_compile_inline_script_metadata(runner, tmp_path, current_resolver):
+    (tmp_path / "script.py").write_text(
+        dedent(
+            """
+            # /// script
+            # dependencies = [
```
webknjaz, line 3782: Should we also have tests for extras?

piptools/scripts/compile.py
```diff
@@ -344,14 +355,50 @@ def cli(
             )
             raise click.BadParameter(msg)
 
-        if src_file == "-":
-            # pip requires filenames and not files. Since we want to support
-            # piping from stdin, we need to briefly save the input from stdin
-            # to a temporary file and have pip read that.  also used for
+        if src_file == "-" or (
+            os.path.basename(src_file).endswith(".py") and not is_setup_file
```
webknjaz, line 359: pathlib is usually nicer
```suggestion
            Path(src_file).suffix == ".py" and not is_setup_file
```

piptools/scripts/compile.py
```diff
@@ -344,14 +355,50 @@ def cli(
...
+        ):
+            # pip requires filenames and not files.  Since we want to support
+            # piping from stdin, and inline script metadadata within Python
+            # scripts, we need to briefly save the input or extracted script
+            # dependencies to a temporary file and have pip read that.  Also used for
             # reading requirements from install_requires in setup.py.
+            if os.path.basename(src_file).endswith(".py"):
```
webknjaz, line 366:
```suggestion
            if Path(src_file).suffix == ".py":
```

tests/test_cli_compile.py
webknjaz: Oh, there's no explicit extras in the spec. But how about `requires-python`?

webknjaz:
@chrysle sorry, I didn't have time lately. I'll try to come back for another review round during/after EuroPython.

tests/test_cli_compile.py
chrysle: I'm not sure how that would influence the build process?

webknjaz: It should reject incompatible runtimes, for example.

webknjaz:
@chrysle I clicked “rebase” and it looks like there's some actionable feedback in the earlier comments.

piptools/scripts/compile.py
merwok: The src path should probably be a local variable (not temporary like in the suggestion), to be used for suffix check here and open / read_text on lines below.

piptools/scripts/compile.py
```diff
@@ -344,14 +355,50 @@ def cli(
...
+                    reqs_str = metadata.get("dependencies", [])
+                    tmpfile = tempfile.NamedTemporaryFile(mode="wt", delete=False)
+                    input_reqs = "\n".join(reqs_str)
+                    comes_from = (
+                        f"{os.path.basename(src_file)} (inline script metadata)"
+                    )
+                else:
+                    raise PipToolsError(
+                        "Input script does not contain valid inline script metadata!"
+                    )
```
webknjaz, line 396: Let's restructure this as follows
```suggestion
                script_metadata_matches = list(
                    filter(
                        lambda m: m.group("type") == name,
                        re.finditer(INLINE_SCRIPT_METADATA_REGEX, script),
                    )
                )

                if not script_metadata_matches:
                    raise PipToolsError(
                        "Input script does not contain valid inline script metadata!"
                    )

                if len(script_metadata_matches) > 1:
                    raise ValueError(f"Multiple {name} blocks found")

                script_metadata_match = script_metadata_matches[0]
                packaging_metadata_toml = "".join(
                    line[2:] if line.startswith("# ") else line[1:]
                    for line in script_metadata_match
                    .group("content")
                    .splitlines(keepends=True)
                )
                metadata = tomllib.loads(packaging_metadata_toml)
                reqs_str = metadata.get("dependencies", [])
                tmpfile = tempfile.NamedTemporaryFile(mode="wt", delete=False)
                input_reqs = "\n".join(reqs_str)
                comes_from = (
                    f"{os.path.basename(src_file)} (inline script metadata)"
                )
```

Issue #2079, opened 2024-04-23
[TODO][CI] Add macOS 14 ARM 64 to the matrix


webknjaz:
Apparently, GitHub has just replaced `macos-latest` with arm64 macOS 14 in GHA. We need to make sure we're running that too.

webknjaz:
UPD: we should explore the same for Ubuntu and Windows, now that they are available.

webknjaz:
> UPD: we should explore the same for Ubuntu and Windows, now that they are available.

@sirosen do you want to keep this issue open for these platforms or file another one?

sirosen:
I think we should close this, and file a fresh one. Both because we have a merged PR linked to this and because I think it deserves a little more discussion.

I can do it!

PR #2132, opened 2024-11-03
feat: Add macOS 14 ARM 64 runners to the matrix

Mr-Sunglasses:
fix: #2079 

##### Contributor checklist

- [X] PR title is short, clear, and ready to be included in the user-facing changelog.

##### Maintainer checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,14 +45,14 @@ jobs:
 
   test:
     name: ${{ matrix.os }} / ${{ matrix.python-version }} / ${{ matrix.pip-version }}
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os:
-          - Ubuntu
-          - Windows
-          - macOS
+          - ubuntu-latest
+          - windows-latest
+          - macos-14
         python-version:
           - "3.11"
           - "3.10"
@@ -67,7 +67,7 @@ jobs:
             )
           }}
         include:
-          - os: Ubuntu
+          - os: ubuntu-latest
             python-version: >-
               ${{
                 github.job_workflow_sha
@@ -139,14 +139,14 @@ jobs:
 
   pypy:
     name: ${{ matrix.os }} / ${{ matrix.python-version }} / ${{ matrix.pip-version }}
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os:
-          - Ubuntu
-          - MacOS
-          - Windows
+          - ubuntu-latest
+          - macos-14
+          - windows-latest
         python-version:
           - pypy-3.8
         pip-version:


webknjaz (approved):
Rebase this once #2145 is in.

webknjaz (changes requested):
FYI `macos-latest` already runs `arm64`. So the current patch doesn't change the runtime.
The idea of #2079 is to have both `arm64` **and** `intel` jobs in CI.

Although, we have to be mindful of the CI resources as this would be computationally expensive. We need to be strategic about this and perhaps have two variants on cron and on merge rather than in PRs, for example.

This needs more research to be implemented correctly.

webknjaz:
@sirosen it should be doable to take over this and merge before the release.

webknjaz:
I wonder if x86-64 this still makes sense with GH taking away the native test env from FOSS: https://github.com/actions/runner-images/issues/12520.

sirosen:
After reading the runner-images thread, I looked at some other CI services to see if GitHub is an outlier here. It seems that there's a broader mass migration off of support for Intel Macs. For example, [this was CircleCI's announcement thread](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718) for removal of x86-64.

My current thinking is that we may have to accept the loss of a testing platform here, and hope that our coverage is still good enough. We're already in a version of this position given that our CI runs only on `macos-latest`.

I'm undecided about what to do with this PR exactly. I would like to be intentional about changing the job names if we're doing that (which this currently does). And I'm overall not sure what my position is around use of `-latest`. When reading CI output, I'd rather see `macos-15` than `macos-latest`, but when maintaining CI config it's nice to have things update automatically.
@webknjaz, do you have a strong preference about use of `latest`? That might help guide my thinking on how to rebase/refactor this.

webknjaz:
Yep, I want static version identifiers, not `-latest`. They are pain to predict sometimes.

We might either lose `x86_64` or try to see if Rosetta works..

webknjaz:
With CPython downgrading intel macs to tier 2 (https://github.com/python/cpython/issues/137638), we shouldn't attempt spending an unjustified amount of resources on it either, I suppose.

sirosen:
I think I'll pick this up, rebase it into a new PR, and switch us over to explicit labels for the latest runners.

If we wanted to add a job to the matrix which runs under Rosetta, we could try, but I don't have relevant experience, so I don't want to block CI cleanup on that -- treat it as a potential CI enhancement instead.

---

Aside: I got the sense that the CPython decision was motivated in part by your investigation. :+1: nice work keeping people informed!

webknjaz:
Ack. Agreed 🤝

webknjaz:
@sirosen looks like GH walked back on taking macos intel from us. Just got an email a couple of hours ago that they've made `macos-15-intel` and the intel support will be dropped in 2027.

sirosen:
This has been on my TODO list, and it looks like there are new macOS runners we can use:
https://github.blog/changelog/2026-02-26-macos-26-is-now-generally-available-for-github-hosted-runners/

`macos-26` is ARM
`macos-26-intel` is Intel

I'll try to make time to go after this, and try these new runner types.

webknjaz:
Resolved by #2348

Issue #2131, opened 2024-10-28
Files included relatively in requirements.in file gets converted to absolute paths


rosen-dash:
Originally reported here - https://github.com/pypa/pip/issues/13055

#### Environment Versions

1. OS Type: MacOS 15.0.1, Ubuntu 24.04
1. Python version: `$ python -V` Python 3.9.6
1. pip version: `$ pip --version` pip 24.3.1
1. pip-tools version: `$ pip-compile --version` 7.4.1

#### Steps to replicate

Sample repo in referred issue.

# With pip 24.2
```
backoff==2.2.1
    # via -r requirements.in
```
# With pip 24.3.1
```
backoff==2.2.1
    # via -r /Users/rosen/Personal/sample/requirements.in
```
#### Expected result

The relative references should be preserved

#### Actual result

Relative paths get replaced with absolute paths

jmac105:
Same happens if specifying a constraint file inside requirements.in, such as `-c requirements.txt`. However if the constraint file is set via command-line argument instead, like `pip-compile -c requirements.txt` then the annotation uses the relative path instead. 

Can replicate with pip 24.3 too, suspect this is a consequence of this change? https://github.com/pypa/pip/pull/12877/files#diff-e9f0a8e3da86978e2042a5a8073078ae9e37ee041dbc943c066e0c622c52d0d4

waqas-saeed:
like @jmac105 said I can confirm this is caused by pip 24.3.1. if I use a bit older version like 23.0.1 then pip-compile shows relative path.

AndydeCleyre:
FWIW this is not an issue in my [feature/relpaths-post-6.8.0](https://github.com/AndydeCleyre/pip-tools/tree/feature/relpaths-post-6.8.0) branch, but there are no plans to integrate those changes here.

AndydeCleyre:
Alternatively, it's not a problem with `uv pip compile`, either.

webknjaz:
@macro1 the CI is now green and working on PRs unblocked.

johnraz:
@webknjaz is there a consensus for a solution that needs to be worked on?

I can definitely help with a PR but I'd need guidance on what needs to be fixed and I'm confused where to look for more infos.

Thanks !

tarasinf:
I'm curious if there's an update on the timeline for addressing this issue, or if there's a recommended workaround in the meantime?

webknjaz:
[...]

PR #2134, opened 2024-11-03
🧪 Bump MyPy Python to 3.9 @ pre-commit config

Mr-Sunglasses:
The `pre-commit.ci` service dropped support for Python 3.8 as it's
gone EOL. This change makes the check runnable again. It is suboptimal
as we haven't yet dropped support for Python 3.8. #2144 will address
the underlying issue more thoroughly.

Resolves #2133.

##### Contributor checklist

- [x] Included tests for the changes.
- [X] PR title is short, clear, and ready to be included in the user-facing changelog.

##### Maintainer checklist

- [x] Verified one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

Issue #2110, opened 2024-07-03
Recursive extra dependencies compiles wrong requirements.txt


maximee-leroyy:
When pyproject.toml references its own local package to recursively include extra dependencies, it outputs a requirements.txt file which references the local package with the absolute path which is a problem as its not portable and should instead just list the dependencies

It was previously possible with setup.py with this configuration:
```python
from setuptools import find_packages, setup

extra_require_tools = ["build"]
extra_require_test = ["pyyaml"]

setup(
    name='package',
    version='1.0.0',
    packages=find_packages(),
    extras_require={
        "dev": extra_require_tools + extra_require_test,
        "tools": extra_require_tools,
        "test": extra_require_test,
    },
)
```

#### Environment Versions

1. OS Type: Ubuntu 22.04.4 LTS
1. Python version: Python 3.11.6
1. pip version: pip 23.2.1
1. pip-tools version: pip-compile, version 7.4.1

#### Steps to replicate

1. Create a pyproject.toml with this content:
```toml
[project]
name = "package"
version = "1.0.0"

[build-system]
requires = ["setuptools>=69.1", ]
build-backend = "setuptools.build_meta"

[project.optional-dependencies]
dev = ["package[test,tools]"]
tools = ["build"]
test = ["pyyaml"]
```
2. Generate the requirements-dev.txt
```bash
pip-compile --extra=dev --no-emit-index-url --output-file=requirements-dev.txt pyproject.toml
```

#### Expected result
```
#
# This file is autogenerated by pip-compile with Python 3.11
# by the following command:
#
#    pip-compile --extra=dev --no-emit-index-url --output-file=requirements-dev.txt pyproject.toml
#
build==1.2.1
    # via package
packaging==24.1
    # via build
pyproject-hooks==1.1.0
    # via build
pyyaml==6.0.1
    # via package
```

#### Actual result
```
#
# This file is autogenerated by pip-compile with Python 3.11
# by the following command:
#
#    pip-compile --extra=dev --no-emit-index-url --output-file=requirements-dev.txt pyproject.toml
#
build==1.2.1
    # via package
package[test,tools] @ file:///tmp/example
    # via package (pyproject.toml)
packaging==24.1
    # via build
pyproject-hooks==1.1.0
    # via build
pyyaml==6.0.1
    # via package
```

#### Linked issues
- https://github.com/jazzband/pip-tools/issues/204#issuecomment-1574051140
- https://github.com/jazzband/pip-tools/issues/1685
[...]

PR #2137, opened 2024-12-02
Mark known test failures

macro1:
Marks known test failures tied to #2131 as we do not have a strategy for remediating.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -10,6 +10,7 @@ from textwrap import dedent
 from unittest import mock
 from unittest.mock import MagicMock
 
+import pip
 import pytest
 from pip._internal.req.constructors import install_req_from_line
 from pip._internal.utils.hashes import FAVORITE_HASH
@@ -1476,6 +1477,10 @@ def test_multiple_input_files_without_output_file(runner):
     assert out.exit_code == 2
 
 
+@pytest.mark.xfail(
+    Version(pip.__version__) >= Version("24.3"),
+    reason="known issue in pip versions 24.3 and later (issue #2131)",
+)
 @pytest.mark.parametrize(
     ("options", "expected"),
     (
@@ -3232,6 +3237,10 @@ def test_resolver_reaches_max_rounds(runner):
     assert out.exit_code != 0, out
 
 
+@pytest.mark.xfail(
+    Version(pip.__version__) >= Version("24.3"),
+    reason="known issue in pip versions 24.3 and later (issue #2131)",
+)
 def test_preserve_via_requirements_constrained_dependencies_when_run_twice(
     pip_conf, runner
 ):


webknjaz:
IIRC, the fix is #2106. I'd rather accept it instead.

tests/test_cli_compile.py
```diff
@@ -3455,7 +3464,6 @@ def test_compile_recursive_extras_build_targets(runner, tmp_path, current_resolv
     expected = rf"""foo[footest] @ {tmp_path.as_uri()}
 small-fake-a==0.2
 small-fake-b==0.3
-wheel==0.42.0
```
webknjaz, line 3458: I don't think it's acceptable to both mark the test as xfail and make changes to it simultaneously.

tests/test_cli_compile.py
```diff
@@ -1476,6 +1477,10 @@ def test_multiple_input_files_without_output_file(runner):
     assert out.exit_code == 2
 
 
+@pytest.mark.xfail(
+    Version(pip.__version__) >= Version("24.3"),
+    reason="known issue in newer versions of pip (#2110)",
```
webknjaz, line 1482: The message should be clear about what versions are bad so that it's more obvious when it can be dropped. This is a part of the test run output to the terminal, saying that some arbitrary pip version that was newer at an arbitrary point in time in the past is unhelpful.

.pre-commit-config.yaml
```diff
@@ -35,7 +35,6 @@ repos:
           - build==1.0.0
           - pyproject_hooks==1.0.0
           - pytest==7.4.2
-        language_version: python3.8
```
webknjaz, line 38: I don't think this should be a part of this PR. If we want to merge a few PRs simultaneously, it's best to make a cumulative PR with a real/natural merge commit combining several branches into `main` so that merging would pick up all the PRs individually.

webknjaz (changes requested):
While I don't think that the tests should be suppressed at all, if we go this way, the places mentioned below must be fixed first.

tests/test_cli_compile.py
macro1: the test above marked with xfail is this one:
`test_preserve_via_requirements_constrained_dependencies_when_run_twice()`
whereas the test being adjusted is this one:
`test_compile_recursive_extras_build_targets()`

i can see how it may look like the same test though

.pre-commit-config.yaml
macro1: that's fine. my goal here was to get CI passing again, which included this change. it is currently brought in with a merge, but i'm happy to rebase once the other pr is in.

macro1:
@webknjaz we either suppress or halt development on the project. there's no proposed solution for #2110

tests/test_cli_compile.py
webknjaz: Oh, that's what I missed. In that case, this change is incorrect, as it masks the problem that #2106 is attempting to fix.

webknjaz (changes requested):
> we either suppress or halt development on the project. there's no proposed solution for #2131

I guess you could work on the solution to that issue with a few temporary commits that unblock the CI for you and could be dropped later.

Meanwhile, the correct solution to the failing tests is #2106. Fixing it is what has to happen first.

I'm blocking this because the change at https://github.com/jazzband/pip-tools/pull/2137/files#r1867006850 is harmful in that it contributes to shoving the problem under the rug, which is a big deal.

Since nobody seems to be attempting to look into completing #2106, I'll try to check it myself later today.

macro1:
alright. so we'll rely on #2106 for the wheel expected in output issue, #2134 for the pre-commit issue (the mypy step no longer supports 3.8 runtime). in this PR, i'll narrow it down to the known issue introduced in pip 24.3

i'm curious about #2106 and the problems we would potentially be shoving beneath the rug, so i may comment there later when i have more time

webknjaz:
I've converted this to draft to prevent accidental merging ahead of time.

macro1:
@webknjaz currently this PR is blocked by your request for changes, and now it is also a draft, so i don't think it's going anywhere

what needs to be merged ahead of this PR? the changes have been reduced to marking specific tests as expected failures when using a pip version of 24.3 or higher, which will remain true till there's a solution for #2131

webknjaz:
@macro1 in #2106, I'm experimenting with making the testing more predictable by pinning pip more aggressively in the test infra (https://github.com/jazzband/pip-tools/pull/2106#issuecomment-2516076142). I recall that PR failing weirdly on Windows, and now that's not exactly the case.

Now that I identified this, I'm going to see if I can change that setup together with #2134 and make pip updates in CI more controlled.

webknjaz:
@macro1 this is not technically necessary anymore. I've still clicked “rebase” to get the CI statuses to me green. Though, perhaps this should be updated together with bumping the pip version that I pinned in another PR. And with the corresponding fixes instead of disabling the tests...

webknjaz:
@macro1 let me know if you think this should be merged anyway.

macro1:
@webknjaz i think i mentioned before there's no current solution for the failing tests this PR is set to mark as known failures. you're currently skipping the tests by limiting the versions of pip we run any tests for, whereas this PR was specifically targeting the known failures.

imo introducing the idea of supported versions of pip runs contrary to how people expect to use this project, but it's a difference of philosophy and probably won't matter in the long-run.

i disagree and do not think it's my choice on whether to merge these changes in at this point. we now have the concept of supported pip versions.. if that is sufficient, that is the approach you have led us to. in my opinion, you've made the choice for us on which approach to take

i do want you to know i'm really happy CI is able to produce green statuses again, and it's awesome there are active contributors here

webknjaz:
@macro1 I asked you if you wanted to merge because I think your opinion is important too. I'm kinda open to it but it might make sense to couple it with bumping pip.

As for "supported", Hugo raised a concern in another PR. We talked about renaming it but postponed, prioritizing making the PR green. It turned out to be confusing but I did not mean to imply that everything else is unsupported. I just wanted to have something that the PR authors would be able to rely on and could be bumped in a controlled and transparent way.
In the metadata, this is not limited.
We should probably rename it to "latest tested".

macro1:
including this change when adding the other pip versions to testing makes sense to me

Issue #2140, opened 2024-12-04
Stability-focused test infra strategy


webknjaz:
#### What's the problem this feature will solve?

The CI has been prone to external influence for a long time. This is due to unpinned deps. Pip specifically. Historically, we've been running the tests against a pip version that was pinned to a minor release, and the latest release, plus the `main` branch of the pip repo.

This turned out to be unsustainable since it causes PR runs to become unpredictable over time. Frankly, this is at odds with the reproducibility promise that pip-tools itself seeks to implement for its users.

It also contributes to other CI problems accumulating over time and blocks progress in many other aspects.

#### Describe the solution you'd like

I want to make sure that the CI runs in PRs and pushes use pinned pip, short term. With the future possibility to have more lock files, long-term.

With that in mind, I want to let the nightly runs keep hitting newer pip versions.

#### Alternative Solutions

…suffering the status quo?

#### Additional context

N/A

webknjaz:
One of the problems found while looking into fixing the CI: https://github.com/jazzband/pip-tools/pull/2141.

N.B. I'm going to prepare a series of separate PRs and then a PR merging-in all those branches into one. The combined one is what will need to be approved and merged.

webknjaz:
Found out that we're hitting a problem with the conditionals relying on `${{ github.job_workflow_sha }}` which would have to be rewritten as a hack: https://github.com/actions/runner/issues/2417#issuecomment-1741105197.

webknjaz:
Another PR: #2142.

webknjaz:
I've found a weird behavior that I thought might be a bug in `contextlib.contextmanager`: https://github.com/jazzband/pip-tools/pull/2106/files#r1870250007. But I can't replicate it outside the `pip-compile` invocation (by importing and making the underlying calls separately). I now suspect that Click might be influencing that.
Here's how the problem manifests itself:
```python-traceback
$ pip-compile --build-deps-for wheel --find-links tests/test_data/minimal_wheels /tmp/pytest-of-${USER}/pytest-current/test_compile_build_targets_setcurrent/pyproject.toml --output-file - -vvvvvv
Using pip-tools configuration defaults found in 'pyproject.toml'.
Traceback (most recent call last):
  File "/home/${USER}/src/github/jazzband/pip-tools/some-venv-py313/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
             ~~~^^
[...]

PR #2142, opened 2024-12-04
🧪 Only run "lowest" & "supported" pip in tox for PR/push @ CI

webknjaz:
This PR implements "supported" and "lowest" (renamed from "previous") envs in tox. It then integrates both into the CI runs triggered by `push` and `pull_request` events.
It drops running the tests against "latest" and "main" envs in regular PRs, keeping them in
the matrix executed on cron.

This improves reproducibility of the CI that will unblock contribution flows.

It additionally fixes reusable workflow conditionals relying on `${{ github.job_workflow_sha }}` that are broken within GitHub.

Refs:
* #2140
* https://github.com/actions/runner/issues/2417#issuecomment-1741105197
* https://github.com/actions/toolkit/issues/1264

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,9 @@
 [tox]
 envlist =
     # NOTE: keep this in sync with the env list in .github/workflows/ci.yml.
-    py{38,39,310,311,312,py3}-pip{lowest,latest,main}-coverage
-    pip{lowest,latest,main}-coverage
-    pip{lowest,latest,main}
+    py{38,39,310,311,312,py3}-pip{supported,lowest,latest,main}-coverage
+    pip{supported,lowest,latest,main}-coverage
+    pip{supported,lowest,latest,main}
     checkqa
     readme
 skip_missing_interpreters = True
@@ -14,6 +14,7 @@ extras =
     testing
     coverage: coverage
 deps =
+    pipsupported: pip==24.2
     piplowest: pip==22.2.*
     piplatest: pip
     pipmain: https://github.com/pypa/pip/archive/main.zip
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,9 @@
 [tox]
 envlist =
     # NOTE: keep this in sync with the env list in .github/workflows/ci.yml.
-    py{38,39,310,311,312,py3}-pip{previous,latest,main}-coverage
-    pip{previous,latest,main}-coverage
-    pip{previous,latest,main}
+    py{38,39,310,311,312,py3}-pip{lowest,latest,main}-coverage
+    pip{lowest,latest,main}-coverage
+    pip{lowest,latest,main}
     checkqa
     readme
 skip_missing_interpreters = True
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -1,9 +1,9 @@
 [tox]
 envlist =
     # NOTE: keep this in sync with the env list in .github/workflows/ci.yml.
-    py{38,39,310,311,312,py3}-pip{previous,latest,main}-coverage
-    pip{previous,latest,main}-coverage
-    pip{previous,latest,main}
+    py{38,39,310,311,312,py3}-pip{supported,lowest,latest,main}-coverage
+    pip{supported,lowest,latest,main}-coverage
+    pip{supported,lowest,latest,main}
```
hugovk, line 6:
```suggestion
    py{38,39,310,311,312,py3}-pip-{supported,lowest,latest,main}-coverage
    pip-{supported,lowest,latest,main}-coverage
    pip-{supported,lowest,latest,main}
```
Could we add dashes (and of course elsewhere where referenced) to improve readability?

tox.ini
```diff
@@ -14,7 +14,8 @@ extras =
     testing
     coverage: coverage
 deps =
-    pipprevious: pip==22.2.*
+    pipsupported: pip==24.2
+    piplowest: pip==22.2.*
```
hugovk, line 18: What does "supported" actually mean here?

`pyproject.toml` requires `pip >= 22.2`, so that maps neatly onto `piplowest`, but the this lowest is also supported.

Is "pipsupported" the highest version we claim to support?

Like, there's a pip 24.3, but we don't claim to support that (yet)?

If so, would `piphighest` be clearer, and mirror `piplowest` better?

(and with dashes for readability)

hugovk (commented):
> This PR implements "supported" and "minimum" (renamed from "previous") envs in tox. It then integrates both into the CI runs triggered by `push` and `pull_request` events.

Looks like "minimum" is actually "lowest" in the code, please could you rename one or the other so the PR title/description matches the code?

tox.ini
webknjaz: > What does "supported" actually mean here?
> 
> `pyproject.toml` requires `pip >= 22.2`, so that maps neatly onto `piplowest`, but the this lowest is also supported.
> 
> Is "pipsupported" the highest version we claim to support?
> 
> Like, there's a pip 24.3, but we don't claim to support that (yet)?
>

Yes, kind of. I just wanted to have something that is pinned in CI. So after this effort, when a PR fails the CI, it's clear that it's not related to something that said PR changed. I don't want to think too much about this. The immediate goal is to unblock development, and it can be refined later.

> If so, would `piphighest` be clearer, and mirror `piplowest` better?

Maybe, I was considering it. But in order to avoid spending a lot of time overthinking, I've gone with "supported" as "something we run in CI".

> (and with dashes for readability)

I was reminded recently that tox makes dash-separated chunks factors, which might not be desired. I thought about it promptly and decided that it's not worthy of being in the scope of this PR and blocking it. But it could be debated separately, once this is complete. I'm currently blocked on getting #2106 to a working state (plus have some work to do in `pypi-publish`, which is mostly what's distracting me from this PR).

tox.ini
webknjaz: Later. See the other comment.

tox.ini
hugovk: Thanks, sounds good. Let's check this again when the other things are in place 👍

PR #2164, opened 2025-02-18
Prefer f-strings instead of str.format

WhyNotHugo:
No user facing changes.

### DIFF
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -266,8 +266,8 @@ class LegacyResolver(BaseResolver):
                     raise RuntimeError(
                         "No stable configuration of concrete packages "
                         "could be found for the given constraints after "
-                        "{max_rounds} rounds of resolving.\n"
-                        "This is likely a bug.".format(max_rounds=max_rounds)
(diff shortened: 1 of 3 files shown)


piptools/resolver.py
```diff
@@ -429,9 +429,9 @@ def get_best_match(self, ireq: InstallRequirement) -> InstallRequirement:
 
         # Format the best match
         log.debug(
```
webknjaz, line 431: Logging is supposed to pass params through dedicated arguments, though. Instead of pre-rendering strings before passing them into logging..

webknjaz (commented):
I wonder if pyupgrade could've handled this.

sirosen:
I think pyupgrade would, indeed, handle a couple of these and is a good idea (original implementation or ruff) regardless for the numerous other fixes it applies. But that's doesn't block this PR.

I've hit the GitHub rebase button since a lot of CI has been fixed significantly at this point.
I was hoping CI would be clean, but there still appear to be errors, which is disappointing.

PR #2175, opened 2025-04-23
Implement a `--group` flag for dependency-group parsing for `pip-compile`

### LINKED ISSUE #2062: Eventually implement support for the PEP 735 "dependency groups"

It's just a new source of deps and an extension to the first round of the `pyproject.toml` runtime deps resolution.

$sbj. If somebody's interested in making a PoC, it's probably fine to start drafting something already. Need to wait for the PEP acceptance to actually merge it in, though.

#### Alternative Solutions

... not doing this?

#### Additional context

Refs:
* https://peps.python.org/pep-0735/
* https://discuss.python.org/t/pep-735-dependency-groups-in-pyproject-toml/39233/237
* POC for pip: https://github.com/pypa/pip/compare/main...sirosen:pip:pep-735-support

sirosen:
Resolves #2062 [^1]

[^1]: Aside: sorry it took so long to submit a PR! This has been on my to-do list for a while.

This is an initial proposed implementation, but even after finishing it, I have two major outstanding questions.

### Q1: Should `--group` be considered a source file?

I didn't merge parsed `--group` inputs into `src_files`, but I did count `len(src_files) + len(groups)` in a couple of contexts to make it behave similarly.
I'm not sure if this was the right approach, or if I missed or misunderstood the goals of the `src_files` checking in any way.

Extra attention in this area during review would be appreciated -- as a new contributor, I did my best to follow the existing code and keep the changes small, but may have made mistakes.

### Q2: What tests are missing?

The initial tests are pretty simple and limited. Probably, more tests would be worthwhile, but it's hard to tell what cases would be useful vs redundant with existing tests.
What additional checks seem important?

### Note: Accepts `pyproject.toml:X` and `pyproject.toml::X`

**Update**: This was removed from the PR!

Although supporting `::` and `:` side-by-side is easy from the perspective of maintenance and support, after plenty of time to think about this, I decided that it provides a worse experience for users. It's different from `pip install --group` and doesn't have a strong enough justification for that difference.

~I decided on this after weighing the options from #2062 and would be willing to adjust if there's a strong objection, but think this is the best approach.~
~`pip-compile` can have its own syntax, but accepting the `pip install` syntax is an easy gain for compatibility.~

~Because the filename part is already optional, the behavior which handles output of options also naturally handles the `:` vs `::` variation.~

### Commits

- Add `--group` to `pip-compile` options
- Implement dependency-groups parsing, based on pip
- Preserve input arg for --group when writing output
- Add tests for [dependency-groups]

##### Contributor checklist

##### Maintainer checklist


reactive-firewall (approved):
> [!IMPORTANT]
> Not a project maintainer, just a fellow project user. My comments are my own and do not reflect the views of the project maintainers.

## :+1: LGTM;

:ballot_box_with_check: Specifically, these changes resolve jazzband/pip-tools#2062
:ballot_box_with_check: Specifically, I agree with @sirosen that development groups should not be considered `src_files` and see that as key to addopting [PEP-735](https://peps.python.org/pep-0735/)
:ballot_box_with_check: Specifically, the introduction of a `--group` option seems like an ideal way to support development groups.

**Additional Comments**:
 * In regards to the question of what is missing, I'm not sure myself, it would be nice to see a more complex example with multiple groups configured (e.g., `dev-all`, `dev-some-ide`, `dev-other-ide`, `dev-ci` in one `pyproject.toml`, seems likely useful :shrug:)

webknjaz:
@reactive-firewall did you generate that with an LLM?

sirosen:
I was wondering the same thing. It has a clear split between what looks like generated junk at the start and a small human comment at the end.
I really don't want to get dragged off topic in this PR. It isn't the time or place for it. However, this is borderline spam.

@reactive-firewall, FYI we are currently working on an LLM usage policy which includes a requirement of disclosure when these tools are used. Please make sure that your comments are making a meaningful contribution. Remember that they take time and effort to read and respond to, and many projects have extremely limited resources -- a few hours per week from one or two people. Spending even half an hour writing a good reply is a significant chunk of my time budget for open source work.

reactive-firewall:
> @reactive-firewall did you generate that with an LLM?

nope

reactive-firewall:
> However, this is borderline spam.

:bow: Opps, sorry, I ment to be helpful; I'll withdraw from this PR now, **no response required** :hear_no_evil:

> FYI we are currently working on an LLM usage policy which includes a requirement of disclosure when these tools are used

**TL;DR** -- You can find my personal policy over here: https://github.com/reactive-firewall-org/multicast/blob/master/.github/AI_USAGE_POLICY.md (feel free to adopt any part of it you find helpful)

sirosen:
Okay, I'm going to wait a day and mark these off-topic.

@reactive-firewall, I know this thread probably doesn't feel great for you, so I want to make sure we part on the best terms possible. One thing I'll note is that the bulleted list, all with emojis, all with the same starting formula, is a bit of a red flag. I also stand by what I said that I don't know that any of those three bullet points add anything. The first recapitulates literally the first line of the PR description, `Resolves #2062`. I recommend giving some attention to how your initial interaction will be perceived, in the present (broader) context. Cheers! :+1:

sirosen:
I've rebased, added some missing tests, and added a changelog entry.

After the rebase, I also moved some things into `piptools._internal` (which is new since this opened!).

I also removed support for `pyproject.toml::groupname` -- only a single colon is now supported. I think it's better, in this case, to be consistent with `pip` and `uv` and not introduce something new. If we use two different syntaxes, we'll be stuck with both forever. (Which is fine, but IMO not the best user interface.)

I'm going to re-evaluate the treatment of `src_files` which I was unsure about at the time of the original PR.

sirosen:
Okay. I've read back over the `src_files` codepaths and I think keeping them separate is the right call for now.

I would like to do some exploration of possible refactors to that part of the codebase. I think it's possible to define some kind of input object which is built from `src_files`, `groups`, `output_file`, and maybe some of the other flags (`--extras`? `--build-deps-for`?) and abstracts away a lot of the combinations and usages which are checked.

---

That more or less concludes my self-review. I think this is ready.

~I note that codecov has dropped -- I'm not sure if that's a direct consequence of this PR. I think I might have accidentally allowed it to drop in `main`.~ Update: a small additional test seems to have made codecov happy!

gaborbernat:
Heads-up that the ``pep751`` branch behind #2380 also lands ``--group`` support, but in the ``pip-lock`` (PEP 751) context. The relevant entry points there are ``piptools/pylock/config.py:load_dependency_groups_table`` (delegates to ``packaging.dependency_groups``) and ``piptools/pylock/config.py:build_group_configs`` (groups + ``[tool.pip-tools].conflicts`` -> per-resolution group sets). They are pylock-specific so they don't conflict with this PR, but the small ``--group`` parsing surface is intentionally close to what you have here so both flows agree on the input shape.

Happy to rebase #2380 onto this once it lands, and to fold the pip-lock flow over to whatever helper module ``--group`` settles on for ``pip-compile``. Not asking for anything from this PR; just flagging that the two are aware of each other so we don't end up with two parsers for the same flag a release later.

PR #2178, opened 2025-04-29
Fix crash with `pip==25.1`, which changed caching for find_all_candidates

### LINKED ISSUE #2176: pip-compile errors when generating hashes with AttributeError 'cache_clear' with pip v25.1
When running with [pip v25.1](https://pip.pypa.io/en/stable/news/#v25-1), `pip-compile --generate-hashes` fails with exception:

> `AttributeError: 'function' object has no attribute 'cache_clear'`

#### Environment Versions

1. OS Type: Linux - Ubuntu 22.04.5 LTS (Jammy Jellyfish)
1. Python version: 3.10.12
1. pip version: 25.1
1. pip-tools version: 7.4.1

#### Steps to replicate

```text
# requirements.in
numpy
```

```shell
pip install -U 'pip >= 25.1'
pip-compile --generate-hashes
```

#### Expected result

Successful lock file generation

#### Actual result

Exception: `AttributeError: 'function' object has no attribute 'cache_clear'`

<details><summary>Traceback (click to expand)</summary>

```python
  File "./venv/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
  File "./venv/lib/python3.10/site-packages/click/core.py", line 1161, in __call__
  File "./venv/lib/python3.10/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
  File "./venv/lib/python3.10/site-packages/click/core.py", line 1443, in invoke
  File "./venv/lib/python3.10/site-packages/click/core.py", line 788, in invoke
  File "./venv/lib/python3.10/site-packages/click/decorators.py", line 33, in new_func
  File "./venv/lib/python3.10/site-packages/piptools/scripts/compile.py", line 471, in cli
    hashes = resolver.resolve_hashes(results) if generate_hashes else None
  File "./venv/lib/python3.10/site-packages/piptools/resolver.py", line 167, in resolve_hashes
    with self.repository.allow_all_wheels(), log.indentation():
  File "/usr/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "./venv/lib/python3.10/site-packages/piptools/repositories/pypi.py", line 452, in allow_all_wheels
    self.finder.find_all_candidates.cache_clear()
AttributeError: 'function' object has no attribute 'cache_clear'
```

</details>

#### Additional notes

Caused by pypa/pip#13306 - you may want to catch `AttributeError` when running `finder.find_all_candidates.cache_clear()`, then attempt to run `finder._all_candidates.clear()`

##### Workaround

Use pip v25.0 or lower

```shell
pip install 'pip < 25.1'
```

gkreitz:
In `pip==25.1`, the implementation of caching for `find_all_candidates` changed, and it no longer exposes any method in its public interface to clear its cache. This causes `pip-compile --generate-hashes` to crash with an `AttributeError`.

In this PR, I replace the cache clearing with simply constructing a new object. I'm not sufficiently familiar with pip/pip-tools internals to know if that's a bad idea (it's unclear to me how much state lies in the finder object, beyond the cache), but I felt it was potentially cleaner than the alternative approach (catch the `AttributeError` and do `finder._all_candidates.clear()`) suggested in #2176.

 Fixes #2176 

##### Contributor checklist

- [X] PR title is short, clear, and ready to be included in the user-facing changelog.

##### Maintainer checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -640,7 +640,6 @@ This will be indicated in the output with one of the following suffixes:
 - [pip-compile-multi](https://pip-compile-multi.readthedocs.io/en/latest/) - pip-compile command wrapper for multiple cross-referencing requirements files.
 - [pipdeptree](https://github.com/tox-dev/pipdeptree) to print the dependency tree of the installed packages.
 - `requirements.in`/`requirements.txt` syntax highlighting:
-
   - [requirements.txt.vim](https://github.com/raimon49/requirements.txt.vim) for Vim.
   - [Python extension for VS Code](https://marketplace.visualstudio.com/items?itemName=ms-python.python) for VS Code.
   - [pip-requirements.el](https://github.com/Wilfred/pip-requirements.el) for Emacs.
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -447,9 +447,13 @@ class PyPIRepository(BaseRepository):
         Wheel.support_index_min = _wheel_support_index_min
         self._available_candidates_cache = {}
 
-        # If we don't clear this cache then it can contain results from an
-        # earlier call when allow_all_wheels wasn't active. See GH-1532
-        self.finder.find_all_candidates.cache_clear()
+        # Finder internally caches results, and there is no public method to
+        # clear the cache, so we re-create the object here. If we don't clear
+        # this cache then it can contain results from an earlier call when
+        # allow_all_wheels wasn't active. See GH-1532
+        self._finder = self.command._build_package_finder(
+            options=self.options, session=self.session
+        )
 
         try:
             yield


sirosen:
I'm going to look into what's going on with all of these failing builds as soon as I have time. I'm pretty sure they have nothing to do with this PR, and are instead caused by some non-reproducible elements in the build envs.

We'll potentially merge this even with CI red; like I said, I'm pretty sure it's right and we can back it out if it turns out to be an issue after we unwind the other CI woes.

AndydeCleyre:
I think we need to remove the keyword option `mix_stderr` here in `conftest.py`:

```python
cli_runner = CliRunner(mix_stderr=False)
```

AFAIU that option has done nothing for a while and is no longer part of Click.

sirosen:
Yes, we will need to handle click 8.2+ changes. Since `pip-tools` still has reason to support Python 3.8, I think the appropriate short-term solution is to pin to `click>=8.1,<8.2`. We can potentially follow the guidance in https://click.palletsprojects.com/en/stable/support-multiple-versions/ , but right now `pip-tools` can be broken unless you pin _it's_ dependencies[^1] -- so the priority is to get it fixed before we start making any additional changes.

[^1]: There is some inherent humor in this, seeing as `pip-tools` is primarily useful for making things -- including CI -- highly reproducible...

Handling will depend on how exactly things go as I wade through the space of unconstrained packages in the tests.
Based on what I was most recently looking at in CI, I don't think `click` is the only or even the primary issue. But pinning it does seem to be helping my local testing exploration, so this might be the right way to start tackling things.

AndydeCleyre:
I'd like to clear the way for fixing #2176 (thanks!) and #2131, so have put up #2183.

sirosen:
This was merged as part of #2195 ! :rocket: 

Thanks @gkreitz for the PR; we made sure to preserve your authorship on the commit here after rebase.

PR #2181, opened 2025-05-22
Restrict dependencies; improve CI reproducibility

sirosen:
`click` v8.2 needs to be handled -- as a follow-up change -- but it's
hard to know if the application itself will work with it yet. Instead
of restricting that in tox.ini, I have chosen to restrict it in the
metadata itself.

`setuptools` on newer versions breaks some of the tests, but
restricting it for package metadata is definitely wrong. Therefore, it
is only restricted for now on the `pipsupported` build.

This should lead to improved CI, but will not result in all tests
passing.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ keywords = ["pip", "requirements", "packaging"]
 dependencies = [
   # direct dependencies
   "build >= 1.0.0",
-  "click >= 8",
+  "click >= 8 , < 8.2",
   "pip >= 22.2",
   "pyproject_hooks",
   "tomli; python_version < '3.11'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ keywords = ["pip", "requirements", "packaging"]
 dependencies = [
   # direct dependencies
   "build >= 1.0.0",
-  "click >= 8 , < 8.2",
+  "click >= 8, < 8.2",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -37,7 +37,7 @@ keywords = ["pip", "requirements", "packaging"]
 dependencies = [
   # direct dependencies
   "build >= 1.0.0",
-  "click >= 8",
+  "click >= 8 , < 8.2",
```
webknjaz, line 40: I assume the whitespace before the comma is accidental?
```suggestion
  "click >= 8, < 8.2",
```

tox.ini
```diff
@@ -15,6 +15,8 @@ extras =
     coverage: coverage
 deps =
     pipsupported: pip==24.2
+    pipsupported: setuptools<=75.8.2
```
webknjaz, line 18: nit: let's make new things sparse.
```suggestion
    pipsupported: setuptools <= 75.8.2
```

webknjaz (approved):
Love the overall change! Here's a few cosmetic updates that would be nice to have.

pyproject.toml
sirosen: I was unsure about having a space there or not. If it looks like a mistake to you; let's apply the change. :+1:

PR #2183, opened 2025-05-24
Fix minor CI issues 

AndydeCleyre:
. . . so we can focus on more important issues.

- Temporarily ignore issues #2131 and #2176 (and any other pip 25.1 issues)
    
    This is in order to see other tests all pass in CI during the PR review.
    I suggest dropping this commit before the PR is merged.

- Stop expecting these to be specially ignored: setuptools, wheel, distribute
    
    See: https://github.com/jazzband/pip-tools/issues/2104

- Use --strip-extras to suppress irrelevant warning for test_all_no_emit_options

- Accept complaint of *any* missing key in a known invalid pyproject.toml
    
    The stderr may stop after noting the missing version,
    without mentioning the missing name

- Introduce compatibility with click>=8.2, maintaining compatibility with <8.2

- Update test pkg: fake_with_deps
    
    - The version range for pyzmq was bumped for compatibility with newer ecosystem
    - The format was changed to a pyproject.toml, addressing this warning:
    
    DEPRECATION: Building 'fake_with_deps' using the legacy setup.py bdist_wheel mechanism,
    which will be removed in a future version.
    pip 25.3 will enforce this behaviour change.
    A possible replacement is to use the standardized build interface
    by setting the `--use-pep517` option,
    (possibly combined with `--no-build-isolation`),
    or adding a `pyproject.toml` file to the source tree of 'fake_with_deps'.
    Discussion can be found at https://github.com/pypa/pip/issues/6334
    
    The rest of the test packages may be updated accordingly in the future.

##### Contributor checklist

- [ ] Included tests for the changes. (**No new tests. The aim is to get existing ones passing in CI.**)

##### Maintainer checklist


AndydeCleyre:
Sure, and thanks for working on this. Maybe this will be useful as a reference for some bit, if not for cannibalization.

sirosen:
Yes to both of the above messages! Your effort and work on this is very much appreciated, and I plan to either cherry-pick directly or create co-authored commits.

But I have another full day of travel and will probably not get to look at this properly until Monday at the earliest.

Everyone's help pitching in on this problem is great to see! 🙂

sirosen:
I just created #2184 with a cherry-pick from here. I'm trying to give credit where it's due and be very transparent.

I think most of these fixes are good, but should be part of adding support for Python 3.12+ or support for newer pip versions. I'd like to handle that next phase by moving forward the `pipsupported` versions and incorporating fixes (from here and possibly elsewhere). And I'd like to add Python 3.12 to the CI matrix. So I intend to inch the infrastructure forward and cherry-pick more from here, assuming that sounds good to everyone.

webknjaz:
@AndydeCleyre could you rebase so we could see what's remaining unmerged?

AndydeCleyre:
Sure, it's now rebased, and I've crossed out the two obsolete items from the original description.

sirosen:
I've just done a reread of the changes here; the only issue is the test ignore commit.

I think we can actually drop that commit from here now (?), since we have `pipsupported` pinned back appropriately in `main`, but I also have #2195 up as a path to resolution, which pulls in some prior work from another PR + my own fix.

`click` is encouraging some slightly different patterns for compatibility (feature detection over version sniffing), but I don't think it matters in this case. And besides, it's the testsuite only, so if it breaks we'll deal with it -- users won't be impacted.

In any case, I see [2960d04](https://github.com/jazzband/pip-tools/pull/2183/commits/2960d0456375576e8ff9ae404f087598fc132ec1) as the current blocker for approval here.

AndydeCleyre:
I've removed the test ignores and rebased. Even though the tests pass in CI, locally these are failing:

- `test_annotate_option`
- `test_preserve_via_requirements_constrained_dependencies_when_run_twice`

Both are suffering from #2131.

sirosen:
I've just hit the rebase button because I think that with #2195 merged, this will now pass and we can merge it.
(Sorry progress on my side has been so slow, but we're getting things moving!)

sirosen (approved):

AndydeCleyre:
Thanks, and no apologies please! I've been away from this for a while and appreciate your work.

PR #1650, opened 2022-06-30
Support relative paths and regularize path handling and formatting

AndydeCleyre:
This replaces #1329, rebased with care on top of pip-tools 6.8.0.

**Changelog-friendly one-liners**:
- Support relative paths in input files, as long as they lead with `file:`, `<vcs>+file:`, or `.`.
- Relative paths in input files become relative paths in output files.
- `pip-compile` will interpret relative paths in an input file as relative to that input file, rather than the current folder, if `--read-relative-to-input` is passed.
- `pip-compile` will reconstruct relative req paths as relative to the output file, rather than the current folder, if `--write-relative-to-output` is passed.
- `pip-sync` will interpret relative paths in an input file as relative to that input file, rather than the current folder, if `--read-relative-to-input` is passed.
- Annotation paths are now relative to the output file.
- Don't overwrite an input file ending in '.txt' when deriving the output file path.
- Don't confuse dots in folder names with file extensions when deriving the output file path.
- Include extras more reliably in output lines, like `pkg[extra1,extra2]`.

---

- Fixes:
  - #204
  - #1107
  - #1165
- Related: 
  - #453 (Not addressed in this PR)
  - #673 
  - #702
  - #1067
  - pypa/pip#1289
  - pypa/pip#6112
  - pypa/pip#6247
  - pypa/pip#6276
  - pypa/pip#6658
  - pypa/pip#6905
  - pypa/pip#7650
  - pypa/pip#8035
  - pypa/pip#8765
  - pypa/pip#8828
  - pypa/packaging#264

##### Contributor checklist

- [x] Provided the tests for the changes.
- [ ] Assure PR title is short, clear, and good to be included in the user-oriented changelog

##### Maintainer checklist

- [ ] Assure one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #2195, opened 2025-06-24
Support pip versions up to 25.1

### LINKED ISSUE #2131: Files included relatively in requirements.in file gets converted to absolute paths

sirosen:
_This PR is based on #2178 and therefore includes those commits._

### Part 1: Relativizing Paths

Resolves #2131

This is a much simpler approach than the one which was attempted in #1650 , and it is imperfect. There is a heuristic being used here to try to produce good outputs, and that heuristic can be tricked. More on this below.

Although I did not make use of that work directly in creating this changeset, it deserves citation as inspiration, and for having demonstrated a path to more complete handling which we won't pursue -- the avenues closed off by that change proposal and review process helped to inform the choice here to do something *ultra simple*.

---

First, this changeset introduces a new conditional behavior for tweaking the `comes_from` annotation data.
`pip-compile` needs to handle the fact that the `comes_from` field on a parsed requirement may be rewritten from a relative path to an absolute one.

The change in `pip` occurred in version 24.3. `comes_from` is populated with absolute paths. I believe this is an effect of https://github.com/pypa/pip/pull/12877 , which was a change included in that release and in which paths are made absolute in an internal pip context.
As far as `pip-compile` is concerned, the main thing which matters is that requirements and constraints files will always appear with absolute paths on newer pip versions.

`pip-compile` targets usage in which a relative path is supplied as input and ought to be preserved in the annotations for its outputs. As such, even though the relative path has been translated to an absolute one, the original (relative) path should be rendered.

Because the data coming back from `pip` are ambiguous to some degree, as a heuristic, this change will use whether or not the original path was absolute to determine behavior. Therefore, the following files will result in correct behavior:

    # reqs.in
    -r other-reqs.in
    # other-reqs.in
    some-pkg

`some-pkg` came from `-r other-reqs.in`.

By contrast, if the CWD is `/opt`, and `-r reqs.in` is used, these data will be *incorrectly* rendered with relative paths:

    # reqs.in
    -r /opt/other-reqs.in
    # /opt/other-reqs.in
    some-pkg

`some-pkg` came from `-r /opt/other-reqs.in`, but will be annotated with `-r other-reqs.in`

Note that the relative path which is used is, in this scenario, correct with respect to the original invocation directory, but not reflective of the inputs to the program.

It is possible, if the absolute paths are really wanted in such a case, to ask for them using `pip-compile -r /opt/reqs.in`.

---

It is possible to enhance this behavior in the future, potentially with a more robust approach, but this changeset tries to make a simple guess which will be right in >80% of realistic usages.

### Part 2: Update the `pipsupported` tox env

With this fix in hand + #2178, it should be possible for tests to pass on newer `pip` versions.

Update `pipsupported` to pick the latest versions of `pip` and `setuptools`.[^1]
For Python 3.8, a lower `pip` version is chosen, as 25.0 dropped support for 3.8 .

[^1]: The `setuptools` bound might not be needed anymore on newer pip versions, but this needs separate assessment.

### Part 3: Normalize Windows Paths in `comes_from`

After finally getting the tests to work consistently with the paths produced by various `pip` versions for `comes_from` data, I've updated this changeset to also add a layer of normalization which ensures that we format paths as posix paths.

The driver for this was the discovery that you can make `pip` produce paths with mixed path separators, e.g.:
```
C:\foo\bar/baz/requirements.in
```

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/_compat/pip_compat.py
+++ b/piptools/_compat/pip_compat.py
@@ -87,12 +87,12 @@ def parse_requirements(
     comes_from_stdin: bool = False,
 ) -> Iterator[InstallRequirement]:
     # the `comes_from` data will be rewritten in different ways in different conditions
-    # each rewrite rule is expressible as a str->str function or as a static replacement
-    rewrite_comes_from: Callable[[str], str] | str
+    # each rewrite rule is expressible as a str->str function
+    rewrite_comes_from: Callable[[str], str]
 
     if comes_from_stdin:
         # if data is coming from stdin, then `comes_from="-r -"`
-        rewrite_comes_from = "-r -"
+        rewrite_comes_from = _rewrite_comes_from_to_hardcoded_stdin_value
     elif pathlib.Path(filename).is_absolute():
         # if the input path is absolute, just normalize paths to posix-style
         rewrite_comes_from = _normalize_comes_from_location
@@ -113,28 +113,32 @@ def parse_requirements(
             install_req.link = file_link
         install_req = copy_install_requirement(install_req)
 
-        if isinstance(rewrite_comes_from, str):
-            install_req.comes_from = rewrite_comes_from
-        else:
-            install_req.comes_from = rewrite_comes_from(install_req.comes_from)
+        install_req.comes_from = rewrite_comes_from(install_req.comes_from)
 
         yield install_req
 
 
+def _rewrite_comes_from_to_hardcoded_stdin_value(_: str, /) -> str:
+    """Produce the hardcoded ``comes_from`` value for stdin."""
+    return "-r -"
+
+
 def _relativize_comes_from_location(original_comes_from: str, /) -> str:
     """
+    Convert a ``comes_from`` path to a relative posix path.
+
     This is the rewrite rule used when ``-r`` or ``-c`` appears in
     ``comes_from`` data with an absolute path.
 
-    The ``-r`` or ``-c`` qualifier is retained, and the path is relativized
-    with respect to the CWD.
+    The ``-r`` or ``-c`` qualifier is retained, the path is relativized
+    with respect to the CWD, and the path is converted to posix style.
     """
     # require `-r` or `-c` as the source
     if not original_comes_from.startswith(("-r ", "-c ")):
         return original_comes_from
 
     # split on the space
-    prefix, _, suffix = original_comes_from.partition(" ")
+    prefix, space_sep, suffix = original_comes_from.partition(" ")
 
     file_path = pathlib.Path(suffix)
 
@@ -144,26 +148,30 @@ def _relativize_comes_from_location(original_comes_from: str, /) -> str:
 
     # make it relative to the current working dir
     suffix = file_path.relative_to(pathlib.Path.cwd()).as_posix()
-    return f"{prefix} {suffix}"
+    return f"{prefix}{space_sep}{suffix}"
 
 
 def _normalize_comes_from_location(original_comes_from: str, /) -> str:
     """
-    This is the rewrite rule when ``-r`` or ``-c`` appears in ``comes_from``
-    data and the input path was absolute, meaning we should not relativize the locations.
+    Convert a ``comes_from`` path to a posix-style path.
 
-    Instead, we apply minimal normalization, ensuring that posix-style paths are used.
+    This is the rewrite rule when ``-r`` or ``-c`` appears in ``comes_from``
+    data and the input path was absolute, meaning we should not relativize the
+    locations.
+
+    The ``-r`` or ``-c`` qualifier is retained, and the path is converted to
+    posix style.
     """
     # require `-r` or `-c` as the source
     if not original_comes_from.startswith(("-r ", "-c ")):
         return original_comes_from
 
     # split on the space
-    prefix, _, suffix = original_comes_from.partition(" ")
+    prefix, space_sep, suffix = original_comes_from.partition(" ")
 
     # convert to a posix-style path
     suffix = pathlib.Path(suffix).as_posix()
-    return f"{prefix} {suffix}"
+    return f"{prefix}{space_sep}{suffix}"
 
 
 def create_wheel_cache(cache_dir: str, format_control: str | None = None) -> WheelCache:
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -84,9 +84,9 @@ class TestFilesCollection:
     def get_path_to(self, filename: str) -> str:
         """Given a filename, find the (first) path to that filename in the contents."""
         return next(
-            k
-            for k in self.contents.keys()
-            if (k == filename) or k.endswith(f"/{filename}")
+            stub_file_path
+            for stub_file_path in self.contents
+            if (stub_file_path == filename) or stub_file_path.endswith(f"/{filename}")
         )
 
 
@@ -3891,7 +3891,9 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
     assert out.exit_code == 0, out
 
 
-@pytest.mark.parametrize("input_path_absolute", (True, False))
+@pytest.mark.parametrize(
+    "input_path_absolute", (True, False), ids=("absolute-input", "relative-input")
+)
 @pytest.mark.parametrize(
     "test_files_collection",
     (
@@ -3922,23 +3924,22 @@ def test_second_order_requirements_path_handling(
     test_files_collection,
 ):
     """
-    Given nested requirements files, the internal requirements will be written
-    in the output, and it will be absolute or relative depending only on
-    whether or not the initial path was absolute or relative.
+    Test normalization of ``-r`` includes in output.
+
+    Given nested requirements files, the internal requirements file path will
+    be written in the output, and it will be absolute or relative depending
+    only on whether or not the initial path was absolute or relative.
     """
     test_files_collection.populate(tmp_path)
 
     # the input path is given on the CLI as absolute or relative
     # and this determines the expected output path as well
-    if input_path_absolute:
-        input_path = (tmp_path / "requirements.in").as_posix()
-        output_path = (tmp_path / "requirements2.in").as_posix()
-    else:
-        input_path = "requirements.in"
-        output_path = "requirements2.in"
+    input_dir_path = tmp_path if input_path_absolute else pathlib.Path(".")
+    input_path = (input_dir_path / "requirements.in").as_posix()
+    output_path = (input_dir_path / "requirements2.in").as_posix()
 
-    with monkeypatch.context() as m:
-        m.chdir(tmp_path)
+    with monkeypatch.context() as revertable_ctx:
+        revertable_ctx.chdir(tmp_path)
 
         out = runner.invoke(
             cli,
@@ -3998,13 +3999,11 @@ def test_second_order_requirements_relative_path_in_separate_dir(
     pip_produces_absolute_paths,
 ):
     """
-    As in the above test, nested requirements file path handling.
-    But in this case, the primary variable is the relative location of the two
-    requirements.
+    Test normalization of ``-r`` includes when the requirements files are in
+    distinct directories.
 
-    The expectation is that the output path will be relative to the current
-    working directory, *not* relative to the dir containing the initial
-    requirements file.
+    Confirm that the output path will be relative to the current working
+    directory.
     """
     test_files_collection.populate(tmp_path)
     # the input is the path to 'requirements.in' relative to the starting dir
@@ -4021,8 +4020,8 @@ def test_second_order_requirements_relative_path_in_separate_dir(
             pathlib.Path(input_path).parent / ("../" * relative_segments) / output_path
         ).as_posix()
 
-    with monkeypatch.context() as m:
-        m.chdir(tmp_path)
+    with monkeypatch.context() as revertable_ctx:
+        revertable_ctx.chdir(tmp_path)
         out = runner.invoke(
             cli,
             [


sirosen:
As an aside: [0be9928](https://github.com/jazzband/pip-tools/pull/2195/commits/0be99285896b3a1b9ef06f1a75a18b19ec22625f) indicates that something is wrong, since `pre-commit` changed that line for me locally. I was surprised but ignored it. And now it's changed back. I bet our isort pre-commit config is wrong.

piptools/_compat/pip_compat.py
```diff
@@ -83,7 +84,24 @@ def parse_requirements(
...
+    rewrite_comes_from: str | Callable[[str], str] | None = None
+
+    if comes_from_stdin:
+        # if data is coming from stdin, then `comes_from="-r -"`
+        rewrite_comes_from = "-r -"
+    else:
+        # if the input was a relative path, set the rewrite rule to rewrite
+        # absolute paths to be relative
+        if not _filename_is_abspath(filename):
+            rewrite_comes_from = _rewrite_absolute_comes_from_location
```
webknjaz: Let's separate the variable initialization from the type declaration. Here's a suggestion that avoids having a nested else-if in favor of elif with a flipped guard check. Flat control flow is easier to follow than branchy.
```suggestion
    rewrite_comes_from: str | Callable[[str], str] | None

    if comes_from_stdin:
        # if data is coming from stdin, then `comes_from="-r -"`
        rewrite_comes_from = "-r -"
    elif _filename_is_abspath(filename):
        rewrite_comes_from = None
    else:
        # if the input was a relative path, set the rewrite rule to rewrite
        # absolute paths to be relative
        rewrite_comes_from = _rewrite_absolute_comes_from_location
```

piptools/_compat/pip_compat.py
```diff
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
 import optparse
+import os
```
webknjaz: If at all possible, it'd be nice to standardize around pathlib objects internally and only convert them to strings when interfacing with external things.

webknjaz:
@sirosen I think this is nice overall. And even if imperfect, we should merge it to unblock other things. But see my inline suggestions first and decide whether they need to be in the scope of this PR or postponed.

piptools/_compat/pip_compat.py
sirosen: I agree that pathlib gives us a nicer model for handling paths, but I didn't want to increase the scope of this change so I tried to stick with "just strings and os.path" to keep my changes more minimal.

Let me take a look at this to see if I can do the work using pathlib and not feel that I'm turning this patch into a refactor. :grin: 
I'll check back in shortly, one way or the other.

piptools/_compat/pip_compat.py
sirosen: Yeah, I like this declaration style a bit better. I'm going to apply the suggestion, but I'll also double-check that I'm happy with how this reads in terms of comments in the broader context.

piptools/_compat/pip_compat.py
sirosen: I made this change and liked the result, but it made me realize that I should add some additional tests to explicitly verify the new behavior. I'm working on that now.

sirosen: @webknjaz, when you have the time, I'd appreciate your review on the new tests.

The tests describe the change in behavior between `pip<24.3` and `pip>=24.3` in greater detail than I did in the PR description. I want to make sure you're okay with the results. (And if not, let's refine the behavior further!)

Also, I'm open to suggestions for ways of improving the tests. I'd like to refactor the testsuite, but I'm trying to stay focused on the release needs right now. So I made some tradeoffs in order to make the tests very self-contained and not introduce new structure or fixtures.

sirosen: I broke all of Windows CI because I didn't think about the path separator while writing the tests. Oops. :joy: 

It will take me a little bit of time to circle back and fix that, but the request for review stands. Just note that the tests are not Windows-friendly and I'm already aware!

piptools/_compat/pip_compat.py
```diff
@@ -94,7 +113,40 @@ def parse_requirements(
...
             file_link._url = parsed_req.requirement
             install_req.link = file_link
-        yield copy_install_requirement(install_req)
+        install_req = copy_install_requirement(install_req)
+        if rewrite_comes_from is None:
+            pass
+        elif isinstance(rewrite_comes_from, str):
+            install_req.comes_from = rewrite_comes_from
+        else:
+            install_req.comes_from = rewrite_comes_from(install_req.comes_from)
```
webknjaz: What if we flip these?
```suggestion
        if isinstance(rewrite_comes_from, str):
            install_req.comes_from = rewrite_comes_from
        elif rewrite_comes_from is not None:
            install_req.comes_from = rewrite_comes_from(install_req.comes_from)
```

tests/test_cli_compile.py
```diff
@@ -3838,3 +3838,211 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+            "--output-file",
+            "-",
+            "--quiet",
+            "--no-header",
+            "--no-emit-options",
+            "-r",
+            input_path,
+        ],
+    )
+    monkeypatch.undo()
```
webknjaz: Would it be cleaner to use https://click.palletsprojects.com/en/stable/api/#click.testing.CliRunner.isolated_filesystem ? or `with monkeypatch.context()`?

tests/test_cli_compile.py
```diff
@@ -3838,3 +3838,211 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+        elif (
+            not first_order_path_absolute and second_order_path_absolute
+        ):  # False, True
+            output_path = "requirements2.in"
+        elif (
+            first_order_path_absolute and not second_order_path_absolute
+        ):  # True, False
+            output_path = str(tmp_path) + "/./requirements2.in"
+        else:  # False, False
+            output_path = "./requirements2.in"
```
webknjaz: This seems to be a part of setup, right? Why not move it closer. Can we abstract the entire preparation into a fixture?

tests/test_cli_compile.py
```diff
@@ -3838,3 +3838,211 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+    `-r` inputs are preserved.
+    """
+    # in pip v24.3, new normalization will occur because `comes_from` started
+    # to be normalized to abspaths
+    pip_current_version = get_pip_version_for_python_executable(sys.executable)
+    expect_new_normalization = pip_current_version >= Version("24.3")
+
+    # setup: there are two files, the second one has an actual package name
+    req_in = tmp_path / "requirements.in"
+    req_in2 = tmp_path / "requirements2.in"
```
webknjaz: try this
```suggestion
    req_in = (tmp_path / "requirements.in").as_posix()
    req_in2 = (tmp_path / "requirements2.in").as_posix()
```

tests/test_cli_compile.py
```diff
@@ -3838,3 +3838,211 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+        if first_order_path_absolute and second_order_path_absolute:  # True, True
+            output_path = str(req_in2)
+        elif (
+            not first_order_path_absolute and second_order_path_absolute
+        ):  # False, True
+            output_path = "requirements2.in"
+        elif (
+            first_order_path_absolute and not second_order_path_absolute
+        ):  # True, False
+            output_path = str(tmp_path) + "/./requirements2.in"
```
webknjaz:
```suggestion
            output_path = str(tmp_path.as_posix()) + "/./requirements2.in"
```

tests/test_cli_compile.py
```diff
@@ -3838,3 +3838,211 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+        if expect_path_normalization:
+            output_path = "subdir2/requirements2.in"
+        else:
+            output_path = "subdir1/../subdir2/requirements2.in"
+    else:
+        # test params didn't match body
+        raise NotImplementedError(relative_requirement_location)
+
+    # the input is given relative to the starting dir
+    input_path = str(req_in.relative_to(tmp_path))
```
webknjaz, line 4025:
```suggestion
    input_path = str(req_in.relative_to(tmp_path).as_posix())
```

tests/test_cli_compile.py
```diff
@@ -3838,3 +3838,211 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+    assert out.stdout == dedent(
+        f"""\
+        small-fake-a==0.2
+            # via -r {output_path}
+        """
+    )
+
+
+@pytest.mark.parametrize(
+    "relative_requirement_location", ("parent", "subdir", "sibling_dir")
```
webknjaz: Instead of having a lot of setup logic in the test and raw strings as params, let's try to move it out.

One cool technique that one of the pytest maintainers once shared is using dataclasses to encapsulate coupled bits of data in params.

The idea is that you define a shape like
```python
from dataclasses import dataclass


@dataclass(frozen=True)
class Thing:
    name: str
    stuff: object

    ...

    def __str__(self) -> str:
        return self.name
```

And then use it like
```python
@pytest.mark.parametrize(
    'thing',
    (
        Thing('case-name'),
        Thing('another-case'),
    ),
    id=str,
)
```

If you can extend that with more attributes and maybe add a method for setting up the test directory, this would remove complex logic from the test function, keeping only testing in.

tests/test_cli_compile.py
```diff
@@ -3838,3 +3838,211 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+
+    Sibling dir layout:
+       - root_dir/
+         - subdir1/
+           - requirements.in  # containing '-r ../subdir2/requirements2.in'
+         - subdir2/
+           - requirements2.in
+    """
+    pip_current_version = get_pip_version_for_python_executable(sys.executable)
+    expect_path_normalization = pip_current_version >= Version("24.3")
```
webknjaz: Can this be in a session-scoped fixture? Maybe even outside this test module?

tests/test_cli_compile.py
```diff
@@ -3838,3 +3838,211 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+    out = runner.invoke(
+        cli,
+        [
+            "--output-file",
+            "-",
+            "--quiet",
+            "--no-header",
+            "--no-emit-options",
+            "-r",
+            input_path,
```
webknjaz: I'd move the `str()` call here so that it's clear that a string is being passed in, which is not possible to judge by just the var name, without having to look for the context above.
```suggestion
            str(input_path),
```

tests/test_cli_compile.py
sirosen: I've used a similar technique in the past -- maybe with named tuples? -- but this looks like a nice way to do it, and `id=str` feels clever and classically pythonic.

I'm happy to do this, and some of the other associated refactoring. I was trying to avoid introducing a bunch of new project infrastructure in the form of fixtures, but it's probably worth it. I'll start working through it!

tests/test_cli_compile.py
sirosen: Oh, this, and the other `as_posix()` fix, is probably what I got wrong for the windows builds! I'm going to apply these right away to see if they pass.

tests/test_cli_compile.py
sirosen: I like the context-manager for `monkeypatch`. `CliRunner.isolated_filesystem` puts another layer of tempdir management in the mix, which is okay but not necessary.

piptools/_compat/pip_compat.py
sirosen: To my eye, it's almost 100% the same. I'll apply because I don't have a strong opinion and this is a couple of lines shorter.

tests/test_cli_compile.py
sirosen: I'll handle this in the course of refactoring towards some fixtures.

tests/test_cli_compile.py
webknjaz: Thanks! It was just that I was reading through the tests, and they seemed to do a lot of stuff, have a lot of logic and branches. It felt like we'd need tests for tests. So in my suggestion I was trying to come up with something that would keep the tests simple as the typical recommendation is to avoid any complexity in tests.

piptools/_compat/pip_compat.py
webknjaz: I feel like having fewer branches contributes to lowering cognitive complexity.

tests/test_cli_compile.py
webknjaz: Yes, that's what I was hoping to fix with these suggestions. Though, it's in-browser coding, so I might've missed things.

piptools/_compat/pip_compat.py
webknjaz: Oh, I missed that self-contained comment, so I started suggesting infra updates :)

webknjaz:
@sirosen apparently, I suggested `.as_posix()` in the wrong places. It doesn't return a `PurePosixPath` as I'd hoped but an instance of `str`. So the `.as_posix()` invocations will have to be applied on conversion, not early.

tests/test_cli_compile.py
```diff
@@ -3838,3 +3838,211 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+        if expect_path_normalization:
+            output_path = "subdir2/requirements2.in"
+        else:
+            output_path = "subdir1/../subdir2/requirements2.in"
+    else:
+        # test params didn't match body
+        raise NotImplementedError(relative_requirement_location)
+
+    # the input is given relative to the starting dir
+    input_path = req_in.relative_to(tmp_path).as_posix()
```
webknjaz: I think the one you pass into CLI shouldn't matter. Only the paths that end up written to output would be important to convert like this.

tests/test_cli_compile.py
sirosen: I made a couple of session scoped fixtures (pip version is one, the result of the comparison is another). Local to the test module for now, but definitely movable in the future.

tests/test_cli_compile.py
sirosen: I took a crack at this, trying to avoid baking in any test execution logic, and am happy with what I produced _for now_. As we've been discussing, I'd like to do a broader test refactor later to see where the things I did apply and where they can/should be modified.

piptools/_compat/pip_compat.py
sirosen: All good! As long as we keep the new infra bits small I don't mind.

I've done as much on that front as I plan to do for now, but let me know if you see a spot where we should refine it further.

sirosen:
Windows CI is still failing on the new tests, and as seen in comments above I did get a little mixed up about applying `as_posix()` in at least one wrong place.

So this is going to take a bit more time to sort out, but I think it's worth the extra time. It gives us clear information about what the pip-compile behavior _really is_, which we won't have otherwise, and there's a chance that I've introduced some bug in the Windows behavior which I need to fix.

sirosen:
I've just finished working out the issues with Windows CI. I made more changes to the way that the paths are normalized/handled and documented it in the original PR comment as "Part 3: Normalize Windows Paths in `comes_from`".

https://github.com/jazzband/pip-tools/pull/2195/commits/1c6da3314082820955d6c5097a97eb15c3fc8ca8 shows what it took to get CI passing on Windows _without_ adding this extra normalization.

The mixed-separator paths (`C:\foo\bar/baz/requirements.in`) are sufficiently "wrong" that I decided to grow the scope of these changes to normalize things to posix-style, even though it's a bit of scope creep. @webknjaz, if you are strongly against this expansion to the PR, I've separated it out and I can revert that change ( https://github.com/jazzband/pip-tools/pull/2195/commits/f2436f2ee6126322881bc4c5fa1c07a90d33f689 ).

piptools/_compat/pip_compat.py
```diff
@@ -83,7 +84,23 @@ def parse_requirements(
...
     isolated: bool = False,
+    comes_from_stdin: bool = False,
 ) -> Iterator[InstallRequirement]:
+    # the `comes_from` data will be rewritten in different ways in different conditions
+    # each rewrite rule is expressible as a str->str function or as a static replacement
+    rewrite_comes_from: Callable[[str], str] | str
+
+    if comes_from_stdin:
+        # if data is coming from stdin, then `comes_from="-r -"`
+        rewrite_comes_from = "-r -"
```
webknjaz:
```suggestion
    rewrite_comes_from: Callable[[str], str]

    if comes_from_stdin:
        # if data is coming from stdin, then `comes_from="-r -"`
        rewrite_comes_from = lambda _: "-r -"
```

piptools/_compat/pip_compat.py
```diff
@@ -94,7 +111,59 @@ def parse_requirements(
             file_link = FileLink(install_req.link.url)
             file_link._url = parsed_req.requirement
             install_req.link = file_link
-        yield copy_install_requirement(install_req)
+        install_req = copy_install_requirement(install_req)
+
+        if isinstance(rewrite_comes_from, str):
+            install_req.comes_from = rewrite_comes_from
+        else:
+            install_req.comes_from = rewrite_comes_from(install_req.comes_from)
```
webknjaz:
```suggestion
        install_req.comes_from = rewrite_comes_from(install_req.comes_from)
```

piptools/_compat/pip_compat.py
```diff
@@ -94,7 +111,59 @@ def parse_requirements(
...
+
+        yield install_req
+
+
+def _relativize_comes_from_location(original_comes_from: str, /) -> str:
+    """
+    This is the rewrite rule used when ``-r`` or ``-c`` appears in
```
webknjaz, line 130: It'd be nice to have docstrings follow PEP 257 and use verbs with imperative mood in the summary lines for functions. Functions correspond to doing something, so they shouldn't be described as static objects.

OTOH, this looks more like a code comment than a docstring. So perhaps just mark it as one?

piptools/_compat/pip_compat.py
```diff
@@ -94,7 +111,59 @@ def parse_requirements(
...
+        return f"{prefix} {file_path.as_posix()}"
+
+    # make it relative to the current working dir
+    suffix = file_path.relative_to(pathlib.Path.cwd()).as_posix()
+    return f"{prefix} {suffix}"
+
+
+def _normalize_comes_from_location(original_comes_from: str, /) -> str:
+    """
+    This is the rewrite rule when ``-r`` or ``-c`` appears in ``comes_from``
```
webknjaz, line 158: This also looks like a code comment, similarly.

piptools/_compat/pip_compat.py
```diff
@@ -94,7 +111,59 @@ def parse_requirements(
...
+    # require `-r` or `-c` as the source
+    if not original_comes_from.startswith(("-r ", "-c ")):
+        return original_comes_from
+
+    # split on the space
+    prefix, _, suffix = original_comes_from.partition(" ")
+
+    # convert to a posix-style path
+    suffix = pathlib.Path(suffix).as_posix()
+    return f"{prefix} {suffix}"
```
webknjaz: Should we reconstruct using exactly the same separator?
```suggestion
    # split on the space
    prefix, whitespace_sep, suffix = original_comes_from.partition(" ")

    # convert to a posix-style path
    suffix = pathlib.Path(suffix).as_posix()
    return whitespace_sep.join((prefix, suffix))
```

tests/test_cli_compile.py
```diff
@@ -3838,3 +3889,157 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
     out = runner.invoke(cli, [req_in.as_posix(), "--output-file", fifo.as_posix()])
 
     assert out.exit_code == 0, out
+
+
+@pytest.mark.parametrize("input_path_absolute", (True, False))
```
webknjaz: Let's label these booleans in the test report output.
```suggestion
@pytest.mark.parametrize("input_path_absolute", (True, False), ids=("absolute-input", "relative-input")),
```

tests/test_cli_compile.py
```diff
@@ -3838,3 +3889,157 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+    runner,
+    tmp_path,
+    monkeypatch,
+    test_files_collection,
+    pip_produces_absolute_paths,
+):
+    """
+    As in the above test, nested requirements file path handling.
+    But in this case, the primary variable is the relative location of the two
+    requirements.
```
webknjaz: This part seems like a code comment. Also, a summary line would be nice.

tests/test_cli_compile.py
```diff
@@ -3838,3 +3889,157 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+    test_files_collection.populate(tmp_path)
+
+    # the input path is given on the CLI as absolute or relative
+    # and this determines the expected output path as well
+    if input_path_absolute:
+        input_path = (tmp_path / "requirements.in").as_posix()
+        output_path = (tmp_path / "requirements2.in").as_posix()
+    else:
+        input_path = "requirements.in"
+        output_path = "requirements2.in"
```
webknjaz: Let's try being more linear?
```suggestion
    input_dir_path = tmp_path if input_path_absolute else pathlib.Path('.')
    input_path = (input_dir_path / "requirements.in").as_posix()
    output_path = (input_dir_path / "requirements2.in").as_posix()
```

tests/test_cli_compile.py
```diff
@@ -39,6 +41,55 @@
...
+                path.write_text(content)
+            else:
+                path.write_text(content(tmp_path))
+
+    def get_path_to(self, filename: str) -> str:
+        """Given a filename, find the (first) path to that filename in the contents."""
+        return next(
+            k
+            for k in self.contents.keys()
+            if (k == filename) or k.endswith(f"/{filename}")
```
webknjaz: `.keys()` is redundant in `dict`s. Also, it's not immediately obvious what the semantics of `k` is. I'll try to guess, though:
```suggestion
            stub_file_path
            for stub_file_path in self.contents
            if (stub_file_path == filename) or stub_file_path.endswith(f"/{filename}")
```

tests/test_cli_compile.py
```diff
@@ -3838,3 +3889,157 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+    # and this determines the expected output path as well
+    if input_path_absolute:
+        input_path = (tmp_path / "requirements.in").as_posix()
+        output_path = (tmp_path / "requirements2.in").as_posix()
+    else:
+        input_path = "requirements.in"
+        output_path = "requirements2.in"
+
+    with monkeypatch.context() as m:
+        m.chdir(tmp_path)
```
webknjaz: Let's try to avoid single-letter vars. 
```suggestion
    with monkeypatch.context() as revertable_ctx:
        revertable_ctx.chdir(tmp_path)
```

tests/test_cli_compile.py
```diff
@@ -3838,3 +3889,157 @@ def test_stdout_should_not_be_read_when_stdin_is_not_a_plain_file(
...
+    if not pip_produces_absolute_paths:
+        # traverse upwards to the root tmp dir, and append the output path to that
+        # similar to pathlib.Path.relative_to(..., walk_up=True)
+        relative_segments = len(pathlib.Path(input_path).parents) - 1
+        output_path = (
+            pathlib.Path(input_path).parent / ("../" * relative_segments) / output_path
+        ).as_posix()
+
+    with monkeypatch.context() as m:
+        m.chdir(tmp_path)
```
webknjaz:
```suggestion
    with monkeypatch.context() as revertable_ctx:
        revertable_ctx.chdir(tmp_path)
```

piptools/_compat/pip_compat.py
sirosen: Haha! I had the exact same idea!

I tried this, and got a flake8 warning, `E731: do not assign a lambda expression, use def`
So I decided "letting it be `str -> str | str` isn't so bad".

Would you rather see a `# noqa: E731` and write it this way? It seems pretty natural to me and I think we'd be ignoring the rule with good reason.
I didn't like the result when I tried writing
```python
def _rewrite_comes_from_stdin(_: str) -> str:
    return "-r -"
```
and trying to place that somewhere nice. It felt very verbose and it lost the locality of this information within `parse_requirements`.

piptools/_compat/pip_compat.py
sirosen: Let me take a pass at rewriting the docstrings. I'm not always that consistent about the style and this is a good opportunity for me to try to be better in that regard.

piptools/_compat/pip_compat.py
sirosen: I don't think it makes a difference in this case, since we're always handling an ASCII space.
Were you thinking about robustness to future refactors?

I'm inclined against this change but because I don't know the reason for it, I want to check before dismissing it.

tests/test_cli_compile.py
sirosen: Yeah, I think you came up with a good name for it. :+1:

tests/test_cli_compile.py
sirosen: I hadn't thought to use `Path(".")` in here; thanks for this!

piptools/_compat/pip_compat.py
webknjaz: Thanks! Perhaps, we don't have a `flake8-docstrings` or similar enabled… It'd be nice to have, too. I wonder if it'd be possible to make https://vale.sh check these more thoroughly.

tests/test_cli_compile.py
webknjaz: Yeah, I had to first check in REPL if that'd work as I was worried about it maybe adding a `./` in front of the string, but it worked like a charm :)

piptools/_compat/pip_compat.py
webknjaz: Marking this thread as resolved.

piptools/_compat/pip_compat.py
webknjaz: Marking this thread as resolved.

piptools/_compat/pip_compat.py
webknjaz: Fair. I was thinking that I didn't like this piece of data being hardcoded in two places, so I wanted to link these places through a var.
I know that `str.join()` is rather ugly, would you feel different if we were to keep an f-string?
```suggestion
    # split on the space
    prefix, whitespace_sep, suffix = original_comes_from.partition(" ")

    # convert to a posix-style path
    suffix = pathlib.Path(suffix).as_posix()
    return f"{prefix}{whitespace_sep}{suffix}"
```

But I won't insist. If you still think that we shouldn't change this bit, feel free to dismiss.

piptools/_compat/pip_compat.py
webknjaz: Yes, I think that a `noqa` is totally justified. I'd only request for such a suppression to be accompanied by a code comment with an explanation. I'm observing a log of noqas in old codebases that nobody knows why they are there and if they should remain. The only way to figure that out would be Git paleontology (if you're lucky and the commit message is useful at all) or removing it and rerunning the linters. Both are time-consuming. So I consider it a good tone to always have justifications in places with suppressions. It's a bit better with pylint, because it has human-readable identifiers. But most other linters just have letters and numbers. And XXX666 wouldn't really tell the reader anything unless they memorized every linting violation in existence. And even with that, understanding the violation doesn't always reveal the original motivation.

```suggestion
    rewrite_comes_from: Callable[[str], str]

    if comes_from_stdin:
        # if data is coming from stdin, then `comes_from="-r -"`
        # "E731: do not assign a lambda expression, use def"
        # is suppressed because this is more elegant.
        rewrite_comes_from = lambda _: "-r -"  # noqa: E731
```

Thinking about it, MyPy will probably see the lambda as `(Any) -> Any`. And so using a cast may shield us from the linting violation as a side effect:
```suggestion
    rewrite_comes_from: Callable[[str], str]

    if comes_from_stdin:
        # if data is coming from stdin, then `comes_from="-r -"`
        rewrite_comes_from = cast(Callable[[str], str], lambda _: "-r -")
```

Another hack I can think of would be `functools.partial()`, which will probably work but is an overkill. And it'll probably make MyPy go crazy…
```suggestion
    rewrite_comes_from: Callable[[str], str]

    if comes_from_stdin:
        # if data is coming from stdin, then `comes_from="-r -"`
        rewrite_comes_from = partial(lambda _: "-r -")
```

------

Thinking more about it, placing a `def` right before `_relativize_comes_from_location` isn't so bad either, as it'd play well with MyPy. Yes, locality might be an issue if you call it `_rewrite_comes_from_stdin`. However, if the name reflects what it actually does (which is not replacing but returning a static thing), it might be better. Something like:
```python
def _produce_hardcoded_comes_from_stdin_string(_: str, /) -> str:
    return "-r -"
```

------

Anyway… I'll leave it for you to decide which path to take.

piptools/_compat/pip_compat.py
sirosen: I get it if it's about capturing that `" "` from the `partition()` call as data. I didn't follow the rationale before.

I do prefer the f-string in this case over `str.join`. I'll go with the version you just suggested!

piptools/_compat/pip_compat.py
sirosen: I think that lambda+noqa will work here, since `mypy` will first see the annotation of `Callable[[str], str]`.

But I'm starting to like the idea of just creating a globally defined function and trying to name it well. It leaves us with the least extra explaining to do in our comments here, and fewer codepaths, so it becomes easier to follow what's actually happening in this function.

I'll make the change and see how it looks when I have it on the screen in front of me. :grin:

sirosen: Yep, once the function is well-named, I like the result of a top-level `def` and assignment here. I'm marking this resolved.

webknjaz (approved):
Wonderful! Please, see if you can improve Git history before merging.

mering:
Thanks for fixing this! When do you have the next release including this planned?

Issue #1886, opened 2023-06-16
Switch to `ruff` for code linting


chrysle:
`ruff`'s [implementation of the most common code formatting tools](https://github.com/astral-sh/ruff#rules) would enable us dropping them in exchange for one dependency.

webknjaz:
flake8-pytest would still have to run as it is.

chrysle:
> Though, maybe not anymore. You could make a demo PR for comparison.

What and how should that demonstrate?

> It'd also request the ruff config to be in a dedicated file.

Not necessarily. `ruff` can also be managed via `pyproject.toml`.

AndydeCleyre:
@chrysle I think @webknjaz meant to type "_I'd_ also request..." because of a preference for managing separate tools in separate files, not because it's necessary on ruff's side.

webknjaz:
> ruff can also be managed via pyproject.toml.

I know and I hate when multiple configs are dumped into the same file which is why I asked for another way.

webknjaz:
> What and how should that demonstrate?

Integrate it in the pre-commit config so it runs in the CI and we could compare the output. Also, if there are any rules from flake8 or its plugins, we'd need to keep it but change `.flake8` to use `select =`, excludes would not be necessary. Make sure to migrate the plugin configuration too and check that all the rules are implemented and can be configured similarly.

webknjaz:
> meant to type "I'd also request..."

Yes, thank you! Sometimes autocorrect is the worst...

PR #2197, opened 2025-06-26
Make 'isort' consistent about 'build'

sirosen:
This does not seem to reproduce consistently -- `isort` is executing
correctly in pre-commit.ci -- but in some environments and
invocations, `isort` is incorrectly reordering imports of `build`
because it gets confused by the `piptools.build` module.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,10 @@ pip-sync = "piptools.scripts.sync:cli"
 
 [tool.isort]
 profile = "black"
+# explicitly mark 'build' as a third-party package
+# otherwise, in some executions, `isort` can mistake `piptools.build` for
+# `build` and treat it as a first-party module name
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
We could also just use ruff for managing imports. See: https://github.com/jazzband/pip-tools/issues/1886

WhyNotHugo:
(it's probably best to merge this as a short-term fix anyway)

sirosen:
I wouldn't mind using `ruff`, although I have been perfectly happy with the tooling which predates it.
I'm not in a hurry about it though.

Right now the major thing I see in pre-commit that needs attention is the mypy configuration. It doesn't even belong in pre-commit (I have some strong opinions on this one), so fixing that will be a body of work in its own right.

PR #2202, opened 2025-07-13
Normalize the changelog style and introduce sphinx extlink roles

sirosen:
This is an early preparatory move for using towncrier.
Configuring it well (e.g., as in `cheroot`) will probably involve integrating these link styles into our templates, which we can only do if we have these link styles available.

---

This change is broken into two commits.

The first is the addition of `sphinx.ext.extlinks` + config + an automated rewrite of the changelog to normalize a lot of
stylistic elements and use <code>{issue}`123`</code> style and similar.

The second is a "manual fixup" (mostly done with editor commands and a bit of manual review) and is separated out to make the changes more visible in review.
The goal here is just to get some stylistic uniformity, especially for the really old changelog entries, so that editor highlighting won't be confused and we'll have a clearer claim to a consistent style.

In this phase, I'm more interested in having a consistent style than in exactly what that style is.
I can make small adjustments if anything is objectionable, but I'd rather save further rewrites (e.g., "should we replace those `Thanks $USER` lines with an `AUTHORS` file?") for future discussion and debate.

##### Contributor checklist

- [x] ~Included tests for the changes.~ N/A

##### Maintainer checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -39,6 +39,7 @@ logger.info(bold("%s release: %s"), project, release)
 extensions = [
     # Stdlib extensions:
     "sphinx.ext.intersphinx",
+    "sphinx.ext.extlinks",
     # Third-party extensions:
     "myst_parser",
     "sphinxcontrib.apidoc",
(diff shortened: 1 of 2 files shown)


webknjaz:
@sirosen could you integrate `sphinx-issues` instead? I'm replacing the `extlinks` copy-paste everywhere but sometimes, there's leftovers.

sirosen:
Yep, happily done! I like `sphinx-issues` but I saw you had done the same with `extlinks` and was just trying to keep things in sync. :smile: 

Looks like the linkcheck lint is bothered by routing through the sponsors redirects, but that should be easy to handle with `linkcheck_allowed_redirects`. I'm doing that right now.

sirosen:
Ah, I tried fixing the redirects and now I get rate limited by GitHub (which is fair!); I see other projects are just skipping validation of the GitHub links and I'll configure the same for us.

webknjaz:
Yeah, I've actually put sphinx-issues into the towncrier issue, but it's the last item in there 😂

docs/conf.py
```diff
@@ -62,6 +63,8 @@
     "python": ("https://docs.python.org/3", None),
 }
 
+issues_github_path = "jazzband/pip-tools"
+issues_user_uri = "https://github.com/sponsors/{user}"
```
webknjaz: This is the default already, since v4: https://github.com/sloria/sphinx-issues?tab=readme-ov-file#400-2024-01-19

```suggestion
```

sirosen: Oh, awesome!

webknjaz (approved):

docs/conf.py
webknjaz: Yeah, I specifically suggested it upstream to reduce copy-paste when migrating to this plugin + spread this practice among the others..

PR #2203, opened 2025-07-16
Setup Towncrier for changelog management, remove release-drafter

sirosen:
These changes are broken down into several commits for simpler review.

## What

- Convert to Towncrier and prepare notes for the upcoming release.

## How

- Remove release-drafter.
- Setup Towncrier configuration.
- Write Towncrier newsfiles to document all changes since v7.4.1 .
- Setup integrations in git, sphinx, pre-commit, and tox.

##### Contributor checklist

- [x] A newsfile is created in `changelog.d/` (see [`changelog.d/README.md`](https://github.com/jazzband/pip-tools/blob/main/changelog.d/README.md)) for instructions) or the PR text says "no changelog needed".

##### Maintainer checklist


webknjaz:
@sirosen I was thinking about having the change log fragments dir nested under the docs dir to reduce the number of the high-level things. Any objections?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,9 +3,9 @@
 ##### Contributor checklist
 
 - [ ] Included tests for the changes.
-- [ ] PR title is short, clear, and ready to be included in the user-facing changelog.
+- [ ] A newsfile is created in `changelog.d/` (see `changelog.d/README.md` for instructions) or the PR text says "no changelog needed".
 
 ##### Maintainer checklist
 
-- [ ] Verified one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
+- [ ] If no changelog is needed, apply the `skip-changelog` label.
```
webknjaz, line 10: We might need to record some guidelines for this (maybe, in a follow-up).

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,9 +3,9 @@
 ##### Contributor checklist
 
 - [ ] Included tests for the changes.
-- [ ] PR title is short, clear, and ready to be included in the user-facing changelog.
+- [ ] A newsfile is created in `changelog.d/` (see `changelog.d/README.md` for instructions) or the PR text says "no changelog needed".
```
webknjaz: Do we want to call this newsfiles or change notes / fragments for consistency?

changelog.d/.towncrier_template.md.jinja
```diff
@@ -0,0 +1,57 @@
+*{{ versiondata["date"] }}*
```
webknjaz: Don't we have to render the actual version title? Looks like it's missing. Never mind the `sphinxcontrib-towncrier` preview. It may be quirky.

changelog.d/2178.bugfix.md
webknjaz, line 1: Could you use symlinks instead of having copies of fragments?

tox.ini
```diff
@@ -168,3 +168,11 @@ passenv =
...
 allowlist_externals =
   git
+
+[testenv:changelog-draft]
+deps =
+    towncrier
+    -r{toxinidir}/docs/requirements.txt
+commands =
+    towncrier --version
+    {envpython} -c 'import sys, subprocess; subprocess.run([sys.executable, "-m", "towncrier", "build", "--version", "main", "--draft"], stderr=subprocess.PIPE)'
```
webknjaz: Is this meant to eat stderr? Why wrap it in this state? Also, I usually have posargs IIRC, that allow me to pass args for experiments.

tox.ini
```diff
@@ -168,3 +168,11 @@ passenv =
 skip_install = false
 allowlist_externals =
   git
+
+[testenv:changelog-draft]
+deps =
+    towncrier
+    -r{toxinidir}/docs/requirements.txt
+commands =
+    towncrier --version
```
webknjaz: We should probably separate these.
```suggestion
commands_pre =
    towncrier --version
commands =
```

changelog.d/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+  - The number is the PR number or issue number which your PR addresses.
+
+  - The category is `bugfix`, `feature`, `deprecation`, `breaking`, `doc`,
+    `packaging`, `contrib`, or `misc`.
+
+  - For example, if your PR fixes bug #404, the newsfile should be named
+    `changelog.d/404.bugfix.md`.
+
+- If multiple issues are addressed, make an identical copy of the newsfile with
```
webknjaz: I'd prefer a symlink, though.

changelog.d/README.md
```diff
@@ -0,0 +1,33 @@
...
+  another issue number in the name.
+  Towncrier will automatically merge identical files into one entry with multiple links.
+
+- Prefer the simple past or constructions with "now".
+
+You can preview the changelog by running `tox run -e changelog-draft`.
```
webknjaz: I would probably tell the users to build the Sphinx docs and see it rendered in the surrounding context.

changelog.d/README.md
webknjaz, line 1: Did you decide not to follow my boilerplate? Any thoughts? Concrete fragment examples to show? These could be a few of the fragments you're adding in this PR copied into this doc.
Also, let's mention the possibility of bylines.

changelog.d/2202.contrib.md
```diff
@@ -0,0 +1 @@
+`pip-tools` now uses `sphinx-issues` to link to issues, PRs, commits, and user accounts.
```
webknjaz: Where possible, I'd add links to external projects + bylines.

changelog.d/2195.feature.2.md
```diff
@@ -0,0 +1,3 @@
+Requirements file paths in `pip-compile` output are now normalized to
+posix-style, even when `pip-compile` is run on Windows.
```
webknjaz:
```suggestion
POSIX-style, even when `pip-compile` is run on Windows.
```

.github/PULL_REQUEST_TEMPLATE.md
sirosen: Yes, I think we can refine this more in the contributing docs? For now, I wanted to make some kind of update here and keep it pretty simple.

.github/PULL_REQUEST_TEMPLATE.md
sirosen: I was trying to be consistent in calling them "newsfiles" since that seems to be the terminology in Towncrier.
"Change notes" may be more contributor-friendly though, for those who aren't familiar with the tool.

I have no strong opinion on this. Since you've raised it, I think I'll switch it to "change notes" and apply that elsewhere as well.

changelog.d/.towncrier_template.md.jinja
sirosen: Oh, yep, this looks like a mistake. I noticed the version was not visible in the `cheroot` template and didn't understand quite what might be happening there. But for this case we should have it.

changelog.d/2178.bugfix.md
sirosen: Happy to oblige! I'll do a cleanup pass for that.

changelog.d/2202.contrib.md
sirosen: Sure! Since there's no docs site to link, I'll point to the sphinx-issues repo.

changelog.d/README.md
sirosen: Yep, consistent with the note on it above, I'll not just fix my own copies, but also this bit of contributor doc.

changelog.d/README.md
sirosen: First, regarding bylines: yes, I'll add that for sure. More thoughts on that topic to follow (I want to rewrite the past entries too).

I was trying to balance concerns between trying to keep things similar to what you have in other projects, but also produce a document which gets new contributors to the instructions quickly. I'm a little concerned about the example from `cheroot` -- I think that for a new contributor that doc is a bit intimidating and gives too much detail about the rationale for Towncrier and changelogs up front.

My preference is for this doc to read primarily as a "how to", and for the motivation behind it to be separated out. (Similar to diataxis, but simply as sections within a doc.)

I think we can reintroduce all of the elments I removed in a sub-section, keeping the start of the doc short and recipe-like.
Would you be okay with the content flowing as follows?

```markdown
## Adding Change Notes with PRs

<mostly the currently proposed "how to" content + mention of adding a byline>

### Categories

The change categories are defined as follows:

- `bugfix:` A bug fix for something we deemed an improper undesired behavior that got corrected in the release to match pre-agreed expectations.
- ... <list follows>

Sometimes it's not clear which category to use for a change. Do your best and a maintainer can discuss this with you during review.

### Example Change Notes

<copy from cheroot example>

### Rationale

When somebody makes a change, they must record the bits that would affect end-users only including information that would be useful to them. ... <content from cheroot example follows>

We write change notes in the past tense because this suits the users who will be reading these notes.
Combined with others, the notes will be a part of the "news digest" telling the readers what **changed** in a specific version of the library since the previous version.

<anything else from the cheroot doc which I've missed probably goes here>
```

changelog.d/README.md
sirosen: Good idea! I prefer this as well, since it will also help them spot malformed content.

I still think this is handy for maintainers, but we can direct users to `tox run -e build-docs`.

tox.ini
sirosen: You had done something similar:
https://github.com/cherrypy/cheroot/blob/fdaa24d8470d68341a84d977f32436e1344e8230/tox.ini#L336-L346

You used `sh`, and I slightly prefer this subprocess usage but don't feel strongly.

At first, I tried without this, thinking it wasn't strictly necessary. But running several times, I found that stderr sometimes was interleaved with stdout. I definitely should add a comment on it though.

If we're wrapping it, I feel like `{posargs}` gets a little messy. I guess it could be done with...
```suggestion
    {envpython} -c 'import sys, subprocess; subprocess.run([sys.executable, "-m", "towncrier", "build", "--version", "main", "--draft"] + sys.argv[1:], stderr=subprocess.PIPE)' {posargs}
```
that? I think that works. I'll try it.

.github/PULL_REQUEST_TEMPLATE.md
webknjaz: Yep, that's the reason I've been using that terminology.

.github/PULL_REQUEST_TEMPLATE.md
webknjaz: Yes, I was imagining it'd be just linked here.

changelog.d/.towncrier_template.md.jinja
webknjaz: I think that old versions of Towncrier would prepend the version title outside Jinja and than it changed. The recent Sphinx plugin version doesn't support those old versions of Towncrier anymore.
Also, the plugin adds a title IIRC, needs checking but I wouldn't worry about the preview. Making sure that the regular release run puts it in place is what's most important.

changelog.d/2178.bugfix.md
webknjaz: Thanks!

tox.ini
webknjaz: Right. I forgot about it. Your way is cleaner. Although, I'd probably pass [`subprocess.DEVNULL`](https://docs.python.org/3/library/subprocess.html#subprocess.DEVNULL) there instead. [`subprocess.PIPE`](https://docs.python.org/3/library/subprocess.html#subprocess.PIPE) confused me because what it exposes ended up being unused. Also, you probably need `check=True` to replicate the same behavior so that Python wouldn't eat the underlying return code. I'd go for [`subprocess.check_call()`](https://docs.python.org/3/library/subprocess.html#subprocess.check_call), though, as it encapsulates it. And my Cheroot variant was also adding `-bb -I` that would be nice to keep them too.

changelog.d/README.md
webknjaz: Perfect!

changelog.d/README.md
webknjaz: Yes. Though, that tox env was originally intended for certain GHA integrations IIRC. I haven't asked to implement that in the issue since that would extend the scope and complexity. Perhaps in the future..

webknjaz:
@sirosen my bot for change note fragments is now plugged into this repo and reports via Checks API as follows: https://github.com/jazzband/pip-tools/pull/2203/checks?check_run_id=46466229160. As you get to adding the config (possibly in a separate PR), the reported check name may change (pip has that).

changelog.d/2202.contrib.md
webknjaz: Yeah, either repo or PyPI is fine with me. I just like having more context available where possible.

tox.ini
webknjaz: Something like this
```suggestion
    {envpython} -bb -E -s -I -c 'import sys, subprocess; subprocess.check_call([sys.executable, "-bb", "-E", "-s", "-m", "towncrier", "build", "--version", "main", "--draft"] + sys.argv[1:], stderr=subprocess.DEVNULL)' {posargs}
```

changelog.d/README.md
webknjaz: I like this idea. I was mostly writing those explanations in isolation with no feedback loop, having been tired of re-explaining how people should author their notes. And so I've been dragging this document around across projects.
I think it's imperfect, and I'd like to improve it. I've been thinking of how to integrate some ideas from https://redhat-documentation.github.io/supplementary-style-guide/#release-notes and lurking into the release notes channel in the Write The Docs Slack, but haven't gotten to it yet.

So if you manage to make it more friendly, I'd love to copy it back into other projects. Making `sphinxcontrib-towncrier` was one of the sides of me attempting to improve authoring change logs.


> ```
> ### Example Change Notes
> 
> <copy from cheroot example>
> ```

I think you should copy some of the actual change notes you're adding in this PR. I've been doing that, and it feels more immersive when yarl fragment examples talk about yarl and aiohttp examples talk about aiohttp.

I'd be nice to improve my bot eventually to be able to help with change note authoring, similar to CPython's `blurb-it`. But that's a separate endeavor.

webknjaz:
> > Could you make sure the contributors are credited in the notes, too?
> 
> On review, I don't like my decision not to include the `Thanks <user>` lines, as with the past changelog entries.
> 
> I think I was struggling with the fact that there's no trivial way to do this in the format that I converted the past changelog entries to use. (Where the "Thanks" comes after the issue reference).
> 
> I'll include it here and I think I'll also do a separate change to make the changelog match the "new" format, of
> 
> ```
> - <text>. Thanks <user>!
>   (#NNN, #NNN)
> ```
> 
> I want things to read uniformly if possible, from old to new. :)
>

Haha! I'm getting flashbacks of one time I've gone deep down the rabbit hole to do something similar post Towncrier reconfig in setuptools: https://github.com/pypa/setuptools/pull/2436.

I'm not sure how I feel about “Thanks” vs. “by”. This is something a contributor would be adding, so some may feel weird thanking themselves, while regular bylines might not trigger the same reaction.

I've been following the style over the past few years, where a byline would be either a part of a sentence (with no period before it but one after), or as a dedicated line (no periods). The standalone byline can be placed either at the end, or after the first paragraph. The in-sentence one could be integrated in one of the sentences wherever it feels right.

So my style would be like one of these:
```rst
A phrase -- by :user:`gh-username`.
```
```rst
A phrase -- by :user:`gh-username`.

Another paragraph.
```
```rst
A phrase -- by :user:`gh-username`. Extra detail.

Another paragraph.
```
```rst
A phrase. More detail.

-- by :user:`gh-username`

Another paragraph.

Or more.

.. code-block:: python

   <usage-example>

Recommendations.
```
```rst
A phrase. More detail.

Another paragraph.

-- by :user:`gh-username`
```
And with that, references would always come after since they're outside the fragment, in the template + have labels that wouldn't work before the note.

> > I was thinking about having the change log fragments dir nested under the docs dir to reduce the number of the high-level things. Any objections?
> 
> I somewhat prefer to have it in the root of the repo, but that's weakly held. If your preference here is a strong one, I can move it.
>
> My rationale is that this is a touch-point for any contributor, so it's simpler (especially for a new contributor) to only traverse one dir. And I think changelogs are important enough to get "two slots" (the file and the dir) in the repo root.
>

I just know that many people dislike having many things in the root, so I've been moving the fragments deeper (and closer to the docs). But we can keep it at the top if you think it's best. I won't insist.

sirosen:
I've just tinkered through what I think is "good enough to merge". I've tried to address all of the minor things like switching over from `subprocess.PIPE` to `subprocess.DEVNULL`, 100% better, or using symlinks for matching newsfiles.

The contributor notes are fleshed out more, but in the structure that I prefer where we put prescriptive "do this!" guidance up front, and explain/elaborate on it below. I feel like it's pretty good, but of course I'm open to feedback/improvements. It's cribbed somewhat from the `cheroot` example (... mid-post, I'm editing to make that a co-authored commit :grin: ), but I've applied some minor tweaks. Notably, the category descriptions are adjusted a bit to fit a CLI tool.

The bylines have been introduced, and integrated into the new fragments.
I did a minimal edit to the existing changelog to make every existing "Thanks" note into a byline for some level of uniformity, and then I started trying to make the style uniform... which I then abandoned and backed out. I think I need to do a bit more tweaking for that, and I'm not wanting to block the release on getting that done. So the ordering will be a little inconsistent in terms of where the PR reference goes, where the byline goes, and where punctuation goes. But all of the new items should be in "good style".

Overall, I think this is ready for a (final?) review. I'm going to scan over the comments and resolve the ones which I don't think we need to discuss, to make sure I didn't forget anything. I'm leaving the thread on the contributor doc, since that's the most likely thing to need more feedback.

changelog.d/.towncrier_template.md.jinja
sirosen: Fixed this up, and it's now where an `## Unreleased Changes` header gets injected too, for drafts.

.github/PULL_REQUEST_TEMPLATE.md
sirosen: I think this is okay for now? But if you have some idea on where/how to link, I'm happy to include a small change.

webknjaz:
> then I started trying to make the style uniform... which I then abandoned and backed out. I think I need to do a bit more tweaking for that, and I'm not wanting to block the release on getting that done. So the ordering will be a little inconsistent in terms of where the PR reference goes, where the byline goes, and where punctuation goes. But all of the new items should be in "good style".

That's something I wouldn't even put into this PR. It's nice to have in the future, but I personally wouldn't invest time into it. I wonder, however, if throwing it into something like Claude Chat or Claude Code would work and reduce your workload. This is something I'd like to explore in the interest of self-education. One other idea would be to try integrating prose linting tooling like https://vale.sh (I'm tracking some of it @ https://github.com/aio-libs/multidict/issues/276) but that's definitely something out of the scope *right now*.

> I'm going to scan over the comments and resolve the ones which I don't think we need to discuss, to make sure I didn't forget anything.

It's usually inconvenient for people who started such inline threads to go back in a following review and try to understand what's actually solved. I often have to spend some time tracking down all the collapsed discussions to make sure they are *really* solved (since in that hidden state it's unclear if there's something I wanted to follow up on or double-check). So I tend to recommend people to either let the topic starters resolve threads (since they would be the ones to judge if something is resolved) or at least add a comment “marking this thread as resolved” before clicking the button. Without that comment, the button click doesn't even generate any notifications, and it becomes difficult to track what's happening and when.

docs/changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+````{only} not is_release
+```{towncrier-draft-entries}
+main
```
webknjaz: Since we're hardcoding this, should we use some more unique sentinel? Like
```suggestion
$$sphinx-draft-preview$$
```
or something along the lines of making it clear why the Jinja template checks against a seemingly arbitrary string `"main"`.

webknjaz:
>     * [x]  Included tests for the changes.

I marked this for you since I consider invoking Towncrier through Sphinx a very good smoke test already.

changelog.d/.towncrier_template.md.jinja
```diff
@@ -0,0 +1,63 @@
+{%- if versiondata["version"] == "main" -%}
+## Unreleased Changes
+{% else -%}
+## {{ versiondata["version"] }}
+{% endif %}
+
+*{{ versiondata["date"] }}*
```
webknjaz: This pattern is a bit new for me as I typically use RST and the title is coming from the Sphinx's directive entirely (and the version is `setuptools-scm`-made in those cases), but perhaps, we could take this opportunity to prefix the date with clearer text:
```suggestion
{%- set is_draft_preview = versiondata["version"] == "$$sphinx-draft-preview$$" -%}
{%- if is_draft_preview -%}
## Unreleased Changes
{% else -%}
## {{ versiondata["version"] }}
{% endif %}

*{%- if is_draft_preview -%}Generated for preview on {%- endif -%}{{ versiondata["date"] }}*
```
(we probably don't need an else-branch with an `"actually released on "` wording, right?)

---

By the way, have you tested a regular non-`--draft` run locally? Does it produce satisfactory output?

changelog.d/.towncrier_template.md.jinja
```diff
@@ -0,0 +1,63 @@
...
+
+{% for section, _ in sections.items() %}
+{% if sections[section] %}
+{% for category, val in definitions.items() if category in sections[section] -%}
+
+{{ definitions[category]['name'] }}:
```
webknjaz: I think this deserves to be a title:
```suggestion
### {{ definitions[category]['name'] }}:
```
Alternatively, at least make it more distinct:
```suggestion
**{{ definitions[category]['name'] }}**:
```
```suggestion
*{{ definitions[category]['name'] }}*:
```

webknjaz: Having it as plain text doesn't help people notice that there are multiple categories rather than just a wall of text, in my view: https://pip-tools--2203.org.readthedocs.build/en/2203/changelog/#unreleased-changes.

changelog.d/.towncrier_template.md.jinja
```diff
@@ -0,0 +1,63 @@
...
+        {%- set _ = commit_refs.append(cf) -%}
+      {%- else -%}
+        {%- set _ = arbitrary_refs.append(cf) -%}
+      {%- endif -%}
+    {%- endfor -%}
+  {%- endwith -%}
+
+- {{ text }} {%
+  if pr_issue_numbers -%}
+  ({issue}`{{ pr_issue_numbers|join('`, {issue}`') }}`)
```
webknjaz: I'm uneasy about dropping the "PRs and issues" prefix. Having this start with a paren enables weird sequences when the change notes end with a period.

You can see multiple sequences like `. (#` @ https://pip-tools--2203.org.readthedocs.build/en/2203/changelog/#unreleased-changes which is what I was trying to avoid by (1) placing these refs on separate lines (so weird things in notes wouldn't make it look odd), and (2) would work with both single-liners and multi-paragraph blurbs.

Having no prefix only works when somebody forgets to end their phrase with a period.

Besides, commit and arbitrary refs have the same level of importance and I think they should follow the same style.

.github/PULL_REQUEST_TEMPLATE.md
webknjaz: In aiohttp, we have a bit of explanation right in the PR template: https://github.com/aio-libs/aiohttp/pull/11320. However, it's also possible to add a Markdown link to RTD (or GH README preview), just like in the following checkbox one line below.

docs/conf.py
```diff
@@ -14,6 +15,14 @@
 # -- Path setup --------------------------------------------------------------
 
 PROJECT_ROOT_DIR = Path(__file__).parents[1].resolve()
+IS_RELEASE_ON_RTD = (
+    os.getenv("READTHEDOCS", "False") == "True"
+    and os.environ["READTHEDOCS_VERSION_TYPE"] == "tag"
+)
+if IS_RELEASE_ON_RTD:
+    tags: set[str]
+    # pylint: disable-next=used-before-assignment
```
webknjaz: I've since learned that it's possible to have something like
```python
def setup(app: Sphinx) -> dict[str, bool | str]:
    """Register project-local Sphinx extension-API customizations.

    :param app: Initialized Sphinx app instance.
    :returns: Extension metadata.
    """
    if IS_RELEASE_ON_RTD:
        app.tags.add("is_release")

    return {
        'parallel_read_safe': True,
        'parallel_write_safe': True,
        'version': release,
    }
```
at the very bottom.

I think this can work better with typing, plus would let us drop linting suppressions.

tox.ini
```diff
@@ -173,6 +173,10 @@ allowlist_externals =
...
     -r{toxinidir}/docs/requirements.txt
-commands =
+commands_pre =
     towncrier --version
-    {envpython} -c 'import sys, subprocess; subprocess.run([sys.executable, "-m", "towncrier", "build", "--version", "main", "--draft"], stderr=subprocess.PIPE)'
+# wrap the invocation of `towncrier build --version main --draft` to discard stderr
+# unfortunately, stderr gets interleaved with stdout, often mixing towncrier's draft build output
+# with user messaging about the invocation
+commands =
+    {envpython} -bb -I -c 'import sys, subprocess; subprocess.run([sys.executable, "-bb", "-I", "-m", "towncrier", "build", "--version", "main", "--draft"] + sys.argv[1:], stderr=subprocess.DEVNULL)' {posargs}
```
webknjaz: Did the `-E -s` bits not work for some reason? https://github.com/jazzband/pip-tools/pull/2203#discussion_r2222207208

changelog.d/README.md
```diff
@@ -0,0 +1,96 @@
...
+  - For example, if your PR fixes bug #404, the change notes should be named
+    `changelog.d/404.bugfix.md`.
+
+- If multiple issues are addressed, create a symlink to the change notes with
+  another issue number in the name.
+  Towncrier will automatically merge files into one entry with multiple links.
+
+- Prefer the simple past or constructions with "now".
+
+- Include a byline, `` `-- by {user}`github-username` ``
```
webknjaz: Looks like that first backtick leaks into the render @ https://pip-tools--2203.org.readthedocs.build/en/2203/contributing/#adding-change-notes-with-prs. This should correct the mistake:
```suggestion
- Include a byline, `` -- by {user}`github-username` ``
```

changelog.d/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+  - The category is `bugfix`, `feature`, `deprecation`, `breaking`, `doc`,
+    `packaging`, `contrib`, or `misc`.
+
+  - For example, if your PR fixes bug #404, the change notes should be named
+    `changelog.d/404.bugfix.md`.
+
+- If multiple issues are addressed, create a symlink to the change notes with
+  another issue number in the name.
+  Towncrier will automatically merge files into one entry with multiple links.
```
webknjaz, line 29: Should we also mention that if a change set addresses multiple aspects, it's possible to add change notes with multiple categories and the same number? Symlinking also works.

changelog.d/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+The categories for change notes are defined as follows.
+
+- `bugfix`: A fix for something we deemed improper or undesired behavior.
+
+- `feature`: A new behavior, such as a new flag or environment variable.
+
+- `deprecation`: A declaration of future removals and breaking changes in behavior.
+
+- `breaking`: A change in behavior which changes or violates established user expectations. e.g., Removing a flag or changing output formatting.
```
webknjaz: Should this be capitalized differently?
```suggestion
- `breaking`: A change in behavior which changes or violates established user expectations.
  E.g., removing a flag or changing output formatting.
```

changelog.d/README.md
```diff
@@ -0,0 +1,96 @@
...
+-- by {user}`sirosen`.
+```
+
+### Rationale
+
+When somebody makes a change, they must record the bits that would affect end-users only including information that would be useful to them.
+Then, when the maintainers publish a new release, they'll automatically use these records to compose a change log for the respective version.
+
+We write change notes in the past tense because this suits the users who will be reading these notes.
+Combined with others, the notes will be a part of the "news digest" telling the readers what **changed** in a specific version of the library since the previous version.
```
webknjaz: Let's use one of the following:
```suggestion
Combined with others, the notes will be a part of the "news digest" telling the readers what **changed** in a specific version of the app since the previous version.
```
```suggestion
Combined with others, the notes will be a part of the "news digest" telling the readers what **changed** in a specific version of the project since the previous version.
```
```suggestion
Combined with others, the notes will be a part of the "news digest" telling the readers what **changed** in a specific version of `pip-tools` since the previous version.
```

Technically, `pip-tools` isn't meant to be used as an importable library with a public Python API. Saying “library” here would be inconsistent with our vision and might also confuse people into assuming things that aren't necessarily true.

webknjaz:
I like this overall, just added a few notes where I thought there were minor inconsistencies.

.pre-commit-config.yaml
```diff
@@ -42,3 +42,45 @@ repos:
...
+        ####.(
+        bugfix
+        | feature
+        | deprecation
+        | breaking
+        | doc
+        | packaging
+        | contrib
+        | misc
+        )(.#)?(.md)?
```
jayaddison, line 62: What is the optional single-digit number before the `md` prefix here provided for?

(if I parse the expression correctly, it allows for, for example, `1234.misc.0.md`?)

changelog.d/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+It is important to maintain a changelog to explain to users what changed
+between versions.
+
+To avoid merge conflicts, we use
+[Towncrier](https://towncrier.readthedocs.io/en/stable/) to maintain our
+changelog.
+
+Towncrier uses separate files, "news fragments", for each pull request.
+On release, those fragments are compiled into the changelog.
```
jayaddison, line 11:
```suggestion
On release, those fragments are collated into the changelog.
```

changelog.d/README.md
jayaddison: I would perhaps suggest parenthesizing the example:

```suggestion
- `breaking`: A change in behavior which changes or violates established user expectations (e.g. removing a flag or changing output formatting).
```

changelog.d/README.md
```diff
@@ -0,0 +1,96 @@
...
+### Examples
+
+Example bugfix, `2082.bugfix.md`:
+
+```md
+Fixed the ordering of format controls to preserve underlying `pip` behavior
+-- by {user}`sethmlarson`.
+```
+
+Example contributor update, `2202.contrib.md`:
```
jayaddison, line 72: I wonder whether we could/should add fully-qualified GitHub hyperlinks to the numbered pull requests mentioned in these two examples?

Someone attempting to follow the guidance closely might find that helpful to confirm how the number part of the filename corresponds to the origin of the changeset.

changelog.d/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+```md
+`pip-tools` now uses [`sphinx-issues`](https://github.com/sloria/sphinx-issues)
+to link to issues, PRs, commits, and user accounts
+-- by {user}`sirosen`.
+```
+
+### Rationale
+
+When somebody makes a change, they must record the bits that would affect end-users only including information that would be useful to them.
```
jayaddison:
```suggestion
When making a change to `pip-tools`, it is important to communicate the differences that end-users will experience in a manner that they can understand.

Details of the change that are primarily of interest only to `pip-tools` developers may be irrelevant to most users, and if so, then those details can be omitted from the news fragment.
```

changelog.d/README.md
```diff
@@ -0,0 +1,96 @@
...
+Then, when the maintainers publish a new release, they'll automatically use these records to compose a change log for the respective version.
+
+We write change notes in the past tense because this suits the users who will be reading these notes.
+Combined with others, the notes will be a part of the "news digest" telling the readers what **changed** in a specific version of the library since the previous version.
+
+This methodology has several benefits, including those covered by the
+[Towncrier Philosophy](https://towncrier.readthedocs.io/en/stable/#philosophy):
+
+- Change notes separate the user-facing description of changes from the implementation details.
```
jayaddison, line 93:
```suggestion
- News fragments separate the user-facing description of changes from the implementation details.
```

changelog.d/README.md
```diff
@@ -0,0 +1,96 @@
...
+  Details go into the git history, but users aren't expected to care about them.
```
jayaddison, line 94:
```suggestion
  Code changes are recorded and can be recovered from the `git` history, but users aren't expected to care about them.
```

jayaddison: Also: is this line missing a hyphen prefix?

sirosen:
@jayaddison, I feel that a drive-by review here is unhelpful. To be direct, I experienced your review negatively. I understand that you are probably trying to help us get this done, and at the highest possible quality level, but unfortunately now I need to spend time on this message and on figuring out what to do with your review. I do this work, which includes taking time to be clear in this message, at the expense of working on the actual PR.

We've been working on this slowly but methodically for a while at this point, and have relatively few remaining issues to address. You seem to be missing some of the relevant context of the prior discussion (e.g., the agreement we reached to consistently call the fragments "change notes") and also are including suggestions which are at-best neutral ("collated" vs "compiled").

I also find that some of these suggestions violate the stylistic rules of brevity and directness which are invaluable for technical writing. I suggest, to anyone who will listen, "On Writing Well" (Zinsser) and "Politics and the English Language" (Orwell). Both are freely available from various sources. Orwell's essay is short, but I will quote the most pertinent passage, his rules of writing:

> (i) Never use a metaphor, simile, or other figure of speech which you are used to seeing in print.
> (ii) Never us a long word where a short one will do.
> (iii) If it is possible to cut a word out, always cut it out.
> (iv) Never use the passive where you can use the active.
> (v) Never use a foreign phrase, a scientific word, or a jargon word if you can think of an everyday English equivalent.
> (vi) Break any of these rules sooner than say anything outright barbarous.

Rules (ii) and (iii) do not state that we should make our writing worse in pursuit of brevity. But we should make it brief in pursuit of quality.

---

EDIT: On review, only one of the suggested changes is one I object to on the aforementioned stylistic grounds. I was a little misled by the way GitHub presented one of the diffs.

changelog.d/README.md
sirosen: I do not find the extra verbiage helpful here.

We could add backticks to `git`, but as it is part of the phrase, "git history", and not a command name, I consider that suggestion neutral.

It is a continuation on the previous thought, and therefore should not be made into a separate list item.

changelog.d/README.md
sirosen: This was discussed above, and we are using the terminology "change notes" intentionally, and trying to be consistent in always calling the newsfiles "change notes".

If you spot us using other terms, that would be appreciated as feedback -- it is possible we have missed some usages.

.pre-commit-config.yaml
webknjaz: That's how Towncrier allows having multiple fragments of the same category for the same issue. It's rarely used in practice. I imagine it might be useful if we want to separate a few change notes and control their order.

changelog.d/README.md
sirosen: When I first read this, I thought it was introducing the lines below; hence my long-form comment including notes about brevity. But that's a trick of how GitHub shows the diff, which is rather unfortunate.

I think we can improve this phrasing -- and I notice now `affect` vs `effect` -- and at first glance this suggestion seems good to me. I'll circle back on this later and quite possibly accept; thanks!

changelog.d/README.md
sirosen: I don't find this distinction important here. In context, both words are effectively the same.

changelog.d/README.md
webknjaz: Do you mean in this doc or during the contribution cycle? If it's the later, the bot validates that the file name pattern is correct. It doesn't check the number (because it could be a PR or an issue, or an old PR being augmented, and there's no PR number prior to creation so one may choose to use the issue number before filing a PR).

As for linking these from the changelog, they are linked from the Sphinx docs. You can see it in the RTD preview for this PR.

.pre-commit-config.yaml
sirosen: Notably, this is not the regex, since this is a `fail` hook.

You can have multiple distinct newsfiles in Towncrier of the same type and for the same issue.

changelog.d/README.md
webknjaz: I think a bit of confusion might be coming from not understanding that it's a paragraph continuation in Markdown.

webknjaz:
@sirosen FTR, I posted on bluesky hoping to get some feedback on the change note writing instructions (that specific document), in terms of being easy to understand from the PoV of less experienced contributors. Perhaps, I wasn't clear enough on the shape of the feedback I wanted, though..

changelog.d/README.md
jayaddison: I mean in the README for the contributing section, as displayed as a webpage in the PR preview at: https://pip-tools--2203.org.readthedocs.build/en/2203/contributing/#examples

The idea would be to link to two genuine (presumably `pip-tools`) pull requests that contain change notes (thank you for the terminology correction, @sirosen - and yep, I admit it is often suboptimal to enter a code review without reading prior history).

The idea would be that a potential contributor could inspect those actual pull requests when constructing their own.

There is a cart-before-horse problem here, though: we can't link to real `pip-tools` Towncrier change notes before we have the functionality in place.

I could offer a follow-up pull request at some point for consideration and to demonstrate precisely what I mean -- but only if that seems useful to you both.

changelog.d/README.md
jayaddison: Yep, I should perhaps have added an explicit comment/note to explain the suggested subdivision of that paragraph.  Another learning that I'll try to remember for future.

I do think that the existing usage of `affect` is grammatically correct in this context (we are interested in changes that cause an influence - an affect - on the userbase).

jayaddison:
For transparency: I arrived here not from the bluesky feedback request, but because an updated release of `pip-tools` may unblock a `pip` usage exception I've encountered (#2176).

When something like that appears as a blocker in an upstream project -- even if a workaround is possible -- my usual inclination is to inspect the remaining tasks/issues, and to attempt to understand and/or contribute to those.

I do recognise that responding to code review commentary -- especially uninformed commentary -- involves effort.  My intent is to provide contributions that offset that cost with a longer-term quality benefit -- but that's hard to quantify.

In any case: I'll step aside for now; thank you for considering and responding to my feedback.

changelog.d/README.md
sirosen: Oh, yep, it is being used as a verb. "affect end-users" not "an effect on end-users". :facepalm: 
The dangers of reading while overtired!

Anyway, I like the new phrasing, thanks again!

docs/changelog.md
sirosen: Since it gets passed as the version, I'll make it `sphinx-draft-preview-version`. I'd rather stick to non-punctuation chars, so that it doesn't get confused with markdown syntax or anything like that.

changelog.d/.towncrier_template.md.jinja
sirosen: I like the "Generated for preview on..." style; adding that to my current changeset!

I did a non-draft run pretty early in this work, so it easily could have drifted. I'll double-check it after this next batch of changes.

changelog.d/.towncrier_template.md.jinja
sirosen: I wanted these to be titles, but I resisted in order to stay consistent with the past changelogs.
I'll update them all to match, but I'll do it in a dedicated commit so that it's easy to revert if we'd rather save this for later.

changelog.d/.towncrier_template.md.jinja
sirosen: Yeah, I'm easily convinced on this. I don't mind the `. (#` formatting, but it _is_ a little weird looking.

I'll adjust (back) to "PRs and issues".

tox.ini
sirosen: Are they necessary? I had to lookup these flags as I've not used all of them in the past, and [the docs for -I](https://docs.python.org/3/using/cmdline.html#cmdoption-I), at least on the current doc version, say that `-I` implies `-E -s`.

changelog.d/README.md
sirosen: Ach! Good catch! I was fiddling with it to get the right rendering and slipped in an extra backtick by mistake.

docs/conf.py
sirosen: This is pretty neat; I never knew you could put this hook into `conf.py`. I wonder if some of the custom plugin cases I've bumped into could be replaced with this... :thinking:

changelog.d/README.md
webknjaz: I think it's a good idea. And it aligns with my earlier suggestion to use real change pip-tools PR notes in examples.

changelog.d/2201.contrib.md
webknjaz, line 1: Could you add a symlink from 2203 to this note?

changelog.d/2201.contrib.md
```diff
@@ -0,0 +1,2 @@
+`pip-tools` now uses Towncrier to manage the changelog
+-- by {user}`sirosen` and {user}`webknjaz`.
```
webknjaz:
```suggestion
-- by {user}`sirosen` and {user}`webknjaz`,
with suggestions from {user}`jayaddison`.
```

changelog.d/README.md
webknjaz: Maybe. I'll let Stephen decide on the final writing style here.

docs/changelog.md
webknjaz: Ack. I mostly wanted to communicate the idea, not necessarily insist on the implementation.

docs/conf.py
webknjaz: Oh yeah, I've been doing this for years. Just keep in mind that `conf.py` is a special case and there be dragons (I was once very confused when I placed some directive classes into it and couldn't understand why that was exploding — this was when I was first writing that `sphinxcontrib-towncrier` extension in-tree). But most of the time, it works just fine. For more complex things, I still tend to move code into dedicated modules in-project.

Some examples include https://github.com/cherrypy/cheroot/blob/587d1fa5e19eb1337d2b9c22995783761628c4b1/docs/spelling_stub_ext.py vs. https://github.com/ansible/awx-plugins/blob/c4180cd22bc6b83750d3256892c38f342eb1e9c5/docs/conf.py#L236-L272.

tox.ini
webknjaz: Ah, okay. This was copied from older projects that used to support huge ranges of Python versions (down to Python 2). I should probably update this in my projects, then. I think that some of the strictness options used to break within Click but couldn't remember which ones. I think, this was `-bb`.

That said, we forgot to start catching warnings early. Let's try this
```suggestion
    {envpython} -bb -I -Werror -c 'import sys, subprocess; subprocess.run([sys.executable, "-bb", "-I", "-Werror", "-m", "towncrier", "build", "--version", "main", "--draft", *sys.argv[1:]], stderr=subprocess.DEVNULL)' {posargs}
```

sirosen:
I'm going to drop a quick comment on this meta-discussion, but I think we're all acting well here and there's no "problem" to resolve. I want to make sure we all leave this discussion feeling positive about the project and our interactions.

> I do recognise that responding to code review commentary -- especially uninformed commentary -- involves effort. My intent is to provide contributions that offset that cost with a longer-term quality benefit -- but that's hard to quantify.

@jayaddison, I think your review was in general of good quality, but missing some context in particular cases. And we are allowed to disagree about some of the micro-edits.

The trouble is primarily that an unsolicited review is a _surprise_ for the people working on the PR. Many maintainers, myself included, have had some bad experiences with users trying to push them to work more or harder on projects where their time is volunteered. When I saw the unexpected review, and it wasn't from someone with a history on the project, it put me on my back foot.

I felt the need to be honest about feeling taxed by the additional feedback. But now I want to be clear that you and your contributions are welcome on `pip-tools` and in particular on this PR.

---

Sorry for the verbosity! I said "quick comment", not "short comment", didn't I? :stuck_out_tongue_winking_eye: 

Let's get this over the finish line. And, to @webknjaz's comment above, let's credit @jayaddison for feedback in the changelog!

changelog.d/2201.contrib.md
sirosen: Yep, I was just thinking something similar!

changelog.d/.towncrier_template.md.jinja
webknjaz: I'm not too worried about temporary inconsistency. My preference is to just have the final template in right away. But I'll let you decide if you want to postpone this.

jayaddison:
Thank you, @sirosen :)  Next time I will do some adjacent discussion thread / etc canvassing before peppering the place with review comments.

changelog.d/.towncrier_template.md.jinja
webknjaz: Thanks!

changelog.d/README.md
sirosen: Yeah, 100% a good callout. I was trying to swap to `pip-tools` in a lot of contexts.

I know it makes these sorts of docs slightly harder to template/copy/compare across projects, but I think it's the best way for it to read. Including this in my current batch of changes.

changelog.d/README.md
sirosen: Well, it's definitely wrong the way I had it! :grin: 

I have a coworker who catches me out every time I write `e.g. foo` instead of `e.g., foo`, and now I can't un-see the missing commas, so I'll insist on that.

I think I'll mix these slightly. I'll try a semantic line-break + parentheses and then look at the render to see how I like it.

changelog.d/README.md
sirosen: If it's okay with you, let's omit that for now? It's something which would arise naturally during PR review and I don't want to try to figure out how to document some of the more nuanced Towncrier behaviors.

docs/changelog.md
sirosen: I realized I needed this also in the `tox r -e changelog-draft` invocation, so I ended up just going with `DRAFT_VERSION`, which I think is a pretty unambiguous special-string.

changelog.d/README.md
```diff
@@ -0,0 +1,98 @@
...
+`pip-tools` now uses [`sphinx-issues`](https://github.com/sloria/sphinx-issues)
+to link to issues, PRs, commits, and user accounts
+-- by {user}`sirosen`.
+```
+
+### Rationale
+
+When making a change to `pip-tools`, it is important to communicate the differences that end-users will experience in a manner that they can understand.
+
+Details of the change that are primarily of interest only to `pip-tools` developers may be irrelevant to most users, and if so, then those details can be omitted from the news fragment.
```
webknjaz: Do we still want to adjust the terminology here?
```suggestion
Details of the change that are primarily of interest only to `pip-tools` developers may be irrelevant to most users, and if so, then those details can be omitted from the change note.
```

sirosen: Should be covered by the latest round of tweaks, good eye as always!

.github/PULL_REQUEST_TEMPLATE.md
sirosen: I was overthinking this one. The PR template now has a link to the `changelog.d/README.md`. Maybe there's a better way to do it, but I linked directly to https://github.com/jazzband/pip-tools/blob/main/changelog.d/README.md

My thought is that we may tweak the doc over time, so linking to the published RTD site could be out of date.

changelog.d/.towncrier_template.md.jinja
```diff
@@ -0,0 +1,66 @@
...
+  {%- endwith -%}
+
+- {{ text }}
+
+  {% if pr_issue_numbers -%}
+  *PRs and issues:* {issue}`{{ pr_issue_numbers|join('`, {issue}`') }}`
+  {{- '\n' -}}
+  {%- endif -%}
+
+  {% if commit_refs -%}
```
webknjaz: @sirosen are you sure about these `-%}` here? I've spent countless hours strugging to make rendering consistent when different combinations of a PR/issue number / commit hash / arbitrary ref are present at various points. And what I got in Cheroot is tested to work with those cases. Prior to that, it'd subtly break bits of markup (sometimes, in very bad ways because RST is more delicate to work with than Markdown).

Here's the commit with all the tweaks I had to apply back when I was deep down that rabbit hole: https://github.com/aio-libs/multidict/commit/a6a6c0cec2ce83b7cd4c3c80a8ca6c865272fa67.

So I'd recommend either keeping that combination of line breaks and hyphens or testing each of these cases locally (a change note with just one reference type, with two reference types/symlinks and with all three).

.github/PULL_REQUEST_TEMPLATE.md
webknjaz: Sure, no rush to make it perfect in this PR.

changelog.d/.towncrier_template.md.jinja
sirosen: I made a change where I made the old sections into headers. So the names of sections may be inconsistent but the styling will be consistent.

```md
<!-- new stuff -->
### Features
...
### Contributor-facing changes
...
<!-- old stuff -->
### Major Changes
...
### Other Changes
...
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,9 +3,9 @@
 ##### Contributor checklist
 
 - [ ] Included tests for the changes.
-- [ ] PR title is short, clear, and ready to be included in the user-facing changelog.
+- [ ] A change note is created in `changelog.d/` (see [`changelog.d/README.md`](https://github.com/jazzband/pip-tools/blob/main/changelog.d/README.md) for instructions) or the PR text says "no changelog needed".
```
webknjaz: How about
```suggestion
- [ ] A change note is created in `changelog.d/` (see [`changelog.d/README.md`](https://github.com/jazzband/pip-tools/blob/main/changelog.d/#readme) for instructions) or the PR text says "no changelog needed".
```

changelog.d/.towncrier_template.md.jinja
sirosen: I had some improper rendering in at least one case while I was working on this all, and I tweaked these to fix.

I'd rather re-test right now than try to go with existing formatting. I'll put this in the queue to work through.

.github/PULL_REQUEST_TEMPLATE.md
sirosen: I like it; let's do that.

CHANGELOG.md
```diff
@@ -1,1178 +1,1180 @@
+<!-- towncrier release notes start -->
+
 ## v7.4.1
 
 *05 Mar 2024*
 
-Bug Fixes:
+### Bug Fixes
 
 - Skip constraint path check ({pr}`2038`).
```
webknjaz: With bylines, now the old entries have this `). --` thing. Can it be mass-replaced to move the period past those bylines.

sirosen: :+1: That should be a pretty easy find-and-replace in vim.

changelog.d/.towncrier_template.md.jinja
```diff
@@ -0,0 +1,66 @@
...
+{% for text, change_note_refs in sections[section][category].items() %}
+
+  {%-
+    set pr_issue_numbers = change_note_refs
+    | map('lower')
+    | map('int', default=None)
+    | select('integer')
+    | map('string')
+    | list
+  -%}
```
webknjaz, line 26: I only now noticed that this doesn't have my hex replacement hack. Did some digging and realized that it's been fixed in Jinja at the end of last year: https://github.com/pallets/jinja/issues/1921.

changelog.d/.towncrier_template.md.jinja
```diff
@@ -0,0 +1,66 @@
...
+      {%- else -%}
+        {%- set _ = arbitrary_refs.append(cf) -%}
+      {%- endif -%}
+    {%- endfor -%}
+  {%- endwith -%}
+
+- {{ text }}
+
+  {% if pr_issue_numbers -%}
+  *PRs and issues:* {issue}`{{ pr_issue_numbers|join('`, {issue}`') }}`
```
webknjaz: _(consistency)_
```suggestion
  *PRs and issues:* {issue}`{{ pr_issue_numbers | join('`, {issue}`') }}`
```

CHANGELOG.md
webknjaz: Ack. Marking this thread as resolved.

changelog.d/.towncrier_template.md.jinja
webknjaz: Thanks! This is one of those things that is difficult to review and I'm afraid to touch :)

changelog.d/README.md
webknjaz: Ack. Marking as resolved.

changelog.d/2201.contrib.md
webknjaz: Marking as resolved.

changelog.d/2201.contrib.md
webknjaz: Marking as resolved.

changelog.d/README.md
webknjaz: Ack, marking this as resolved.

docs/changelog.md
webknjaz: Ack. Marking as resolved.

changelog.d/.towncrier_template.md.jinja
sirosen: Oh, I forgot to call that out explicitly for you, sorry! Yeah, I traced it back and saw that it was fixed, so we didn't need it anymore. But I should have said something.

sirosen:
I just re-tested a release build of the docs and was disappointed to find that MyST is not evaluating `{only}` directives properly. I need to make a MRE later so that I can submit this as a bug to upstream. For now, I've found a workaround which I put together in
https://github.com/jazzband/pip-tools/pull/2203/commits/9a14a538f36365b2517d18e6d2b7c714fd753aa6

It's not the prettiest thing I've ever come up with, but it does solve the problem.

changelog.d/.towncrier_template.md.jinja
webknjaz: This is fine. The old categories are different and it's okay to skip normalizing them.

changelog.d/.towncrier_template.md.jinja
sirosen: I found a bad case for what I had while re-testing. A commit reference with no issue number wasn't spaced the way I expected.

I redid the leading/trailing whitespace chomping indicators to be aligned to the front of each item type + tested all of the combinations. I think this is all set. I hope! :sweat_smile:

changelog.d/README.md
sirosen: I'd definitely welcome a follow-up improvement for this. I don't think I want to do it right now since it's already taking so much more effort than I expected to get all of the details right in the PR.

This seems like something which is coherent to come back to after the release and I'm itching to get that done first. :smile:

changelog.d/README.md
sirosen: I'm marking this resolved. I moved both of the examples given to this style and it reads nicely to me.

changelog.d/README.md
sirosen: We've got this fix in, so I'm marking resolved.

tox.ini
sirosen: I've incorporated those `-Werror` flags + the use of *-expansion. Marking resolved. :slightly_smiling_face:

changelog.d/.towncrier_template.md.jinja
webknjaz: Well, now I know :) Marking this as resolved.

sirosen:
> > I'm going to scan over the comments and resolve the ones which I don't think we need to discuss, to make sure I didn't forget anything.
>
> It's usually inconvenient for people who started such inline threads to go back in a following review and try to understand what's actually solved. I often have to spend some time tracking down all the collapsed discussions to make sure they are really solved (since in that hidden state it's unclear if there's something I wanted to follow up on or double-check).

Most of the time I agree with this, but this has been a tough PR for me to manage in terms of making sure all of the comments have been addressed from my side. So I have marked off a few of the things which are more "minor bugfix" items -- more conservatively on this last pass.

I think I've addressed all of the current feedback, but that tweak I mentioned above to make the `only` doc directive work definitely needs fresh review, since it's a bit weird.

webknjaz:
Ack. So https://github.com/executablebooks/MyST-Parser/issues/262#issuecomment-1449474949 doesn't help here?

changelog.d/.towncrier_template.md.jinja
webknjaz: Thanks! Marking this as resolved.

sirosen:
If there's a way to do it, I haven't found it yet. :sweat_smile: 

The issue with trying to use it on the current work that it's a non-MyST include directive with the myst parser set.
So we get
```
/home/sirosen/projects/pip-tools/docs/changelog.md:19: ERROR: Error in "include" directive:
unknown option: "offset-heading".
```

Trying to jump back and forth between MyST and RST is very strange!

webknjaz:
Ah, that's fair. We should ping @chrisjsewell to see if there's better workarounds.. But maybe in a follow-up.

webknjaz:
@sirosen looks like the upstream issue is https://github.com/executablebooks/MyST-Parser/issues/607

tox.ini
```diff
@@ -102,6 +102,8 @@ changedir = {toxinidir}/docs
 isolated_build = true
 passenv =
   SSH_AUTH_SOCK
+  READTHEDOCS
+  READTHEDOCS_VERSION_TYPE
```
webknjaz: Let's order these alphabetically:
```suggestion
  READTHEDOCS
  READTHEDOCS_VERSION_TYPE
  SSH_AUTH_SOCK
```

tox.ini
```diff
@@ -168,3 +170,16 @@ passenv =
 skip_install = false
 allowlist_externals =
   git
+
+[testenv:changelog-draft]
```
webknjaz, line 174: Should we add a job for running this to `reusable-qa.yml`?

tox.ini
sirosen: I'll amend the commit with this.

tox.ini
sirosen: Good idea, should be easy enough... :crossed_fingers:

changelog.d/.towncrier_template.md.jinja
```diff
@@ -0,0 +1,70 @@
+{%- set is_draft_preview = versiondata["version"] == "DRAFT_VERSION" -%}
+
+{%- if is_draft_preview -%}
+## Unreleased Changes
+{% else -%}
+## {{ versiondata["version"] }}
+{% endif %}
+
+*{%- if is_draft_preview -%}Generated for preview on {% endif -%}{{ versiondata["date"] }}*
```
webknjaz, line 9: I just now realized that Hynek even added a warning admonition for this: https://github.com/python-attrs/attrs/blob/199cd548e11197147b9e7438e13c11c621b7b173/changelog.d/towncrier_template.md.jinja#L4-L6. And I remembered he also wrapped the version titles as links to GitHub (I saw this during the initial integration in there since he didn't know it was possible to use my extension with Markdown initially so I ended up watching the progress but forgot that by now). Pretty neat! Perhaps, we need to look into doing the same later.

tox.ini
webknjaz: Marking as resolved.

tox.ini
```diff
@@ -168,3 +170,16 @@ passenv =
...
+deps =
+    towncrier
+    -r{toxinidir}/docs/requirements.txt
```
webknjaz: Is this really needed to just call Towncrier? Or are you using this as a hack for "constraints"? In that case, we should do
```suggestion
    -c{toxinidir}/docs/requirements.txt
```
instead (it won't explode because there's no hashes).

tox.ini
```diff
@@ -168,3 +170,16 @@ passenv =
...
+commands_pre =
+    towncrier --version
+# wrap the invocation of `towncrier build --version main --draft` to discard stderr
+# unfortunately, stderr gets interleaved with stdout, often mixing towncrier's draft build output
+# with user messaging about the invocation
+commands =
+    {envpython} -bb -I -Werror \
+      -c 'import sys, subprocess; subprocess.run([sys.executable, "-bb", "-I", "-Werror", "-m", "towncrier", "build", "--version", "DRAFT_VERSION", "--draft", *sys.argv[1:]], stderr=subprocess.DEVNULL)' {posargs}
```
webknjaz:
```suggestion
  {envpython} -bb -I -Werror \
    -c 'import sys, subprocess; subprocess.run([sys.executable, "-bb", "-I", "-Werror", "-m", "towncrier", "build", "--version", "DRAFT_VERSION", "--draft", *sys.argv[1:]], stderr=subprocess.DEVNULL)' {posargs}
```

tox.ini
webknjaz: I just noticed that this has inconsistent indents. I standardized on two-space indents for ini-files.
```suggestion
  towncrier
  -c{toxinidir}/docs/requirements.txt
```

tox.ini
webknjaz: Ack. Marking as resolved.

webknjaz:
@sirosen I've spent some time trying to come up with a better hack for MyST-Parser and failed. So I'm giving up for now.

I've looked through the current diff and it seems like once the final `tox.ini` things are sorted, this should be good to go in. You might want to check how Git looks before merging but nothing beyond that.

tox.ini
sirosen: I can make sure to use 2-ist spacing. The file is already a little inconsistent, I think, so it wasn't clear which way to go. We'll do 2-ist for the new stuff and I'll standardize it in a PR sometime later.

sirosen: I wanted to pin the `towncrier` version to match the doc build.

> it won't explode because there's no hashes

Ah! Right, okay, I'll switch it to `-c`.

changelog.d/.towncrier_template.md.jinja
sirosen: It might be possible to do more by following his example, since I see that he's unconditionally including the towncrier content ( https://github.com/python-attrs/attrs/blob/199cd548e11197147b9e7438e13c11c621b7b173/docs/changelog.md?plain=1#L5 ).

Let's circle back on this after the release. I think his setup may have multiple things we can try out.

sirosen:
> You might want to check how Git looks before merging but nothing beyond that.

I'll do a quick pass, but I'm not sure it's going to be easy to cleanup, since we've done a few rounds of iteration. But I'll squash together anything obvious.

sirosen:
Okay, I think I've reached the point of diminishing returns after a couple of interactive rebase passes. I could probably spend hours rebuilding a perfect history, with only the "useful" variations on this work included, but I don't think that's a good use of our time.

PR #2209, opened 2025-07-30
Drop release attestations for Jazzband upload

webknjaz:
This patch cleans up the attestation files that the first `pypi-publish` invocation leaves on disk so that the following one would not attempt uploading them to the private package index of Jazzband as it might not support such uploads just yet.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -139,6 +139,8 @@ jobs:
           Publish v${{ github.ref_name }} to PyPI
           🔏
         uses: pypa/gh-action-pypi-publish@release/v1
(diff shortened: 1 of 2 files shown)


webknjaz:
I'm 99% sure it doesn't. I think it's not even devpi but a flask app or something like that.
Though, the thing is that internal validation in `pypi-publish` will fail if attestation files are on disk already. It's meant to be called from a dedicated job, not twice. This hack will unblock us. But maybe eventually, when I get to making a reusable workflow (and it's supported), we'll just migrate everything to that.

PR #2211, opened 2025-07-30
Release version 7.5.0

sirosen:
These changes were applied via `towncrier build --version 7.5.0`.

##### Contributor checklist

##### Maintainer checklist


sirosen:
I'm looking at the new CI setup, and I think we will
- merge this
- tag as `7.5.0` (matches past tags)
- create a GitHub release for 7.5.0

and that should take care of the publishing side.

And then I'll share in places like DPO that we've finally got this out.

---

I'm not familiar with the Jazzband infrastructure, but I don't think there's anything I need to do after the publish-to-jazzband step runs in GitHub.

CHANGELOG.md
```diff
@@ -1,5 +1,110 @@
+
+
```
webknjaz, line 2: @sirosen were these lines added by mistake?
```suggestion
```

CHANGELOG.md
```diff
@@ -1,5 +1,110 @@
...
 <!-- towncrier release notes start -->
 
+## 7.5.0
```
webknjaz, line 5: This should've been `towncrier build --version v7.5.0`
```suggestion
## v7.5.0
```

webknjaz:
> I'm looking at the new CI setup, and I think we will
> 
> * merge this

I'd like to release from the top commit in this branch and only merge when we know the release succeeded.

> * tag as `7.5.0` (matches past tags)

I'd like to standardize around `v`-prefixed tags as it's more common with Git. Unprefixed variants are often used for unstable branches I believe it's confusing to have no prefix. It's an easy visual distintion between tags and branches. Additionally, the change log is also prefixed, usually (and it was in the past). `v7.5.0` would be making everything more consistent across the board.

> * create a GitHub release for 7.5.0
> 
> and that should take care of the publishing side.

This is what I want to do right away, once the PR is ready but unmerged. Doing this will automatically create a tag so that does not need to be a separate step.

Though, I noticed you made a branch in your fork which would present a challenge following the process I intended. Could you push the release branch upstream instead?

> And then I'll share in places like DPO that we've finally got this out.

Also, on bsky / mastodon / gh issues that people are watching. We might want to look into (partially) automating this later on.

> I'm not familiar with the Jazzband infrastructure, but I don't think there's anything I need to do after the publish-to-jazzband step runs in GitHub.

Yes, that doesn't need anything past running the CD job.

CHANGELOG.md
sirosen: I wasn't clear on why those got injected, so I left them. They weren't a manual addition, but got added by the `towncrier` run.

I can happily remove them. If `towncrier` adds them again, I'll try to figure out why.

CHANGELOG.md
sirosen: Ah, yeah, that's the format used in the past.
I was basing it off of the past tags (without the `v`).

sirosen:
> I'd like to release from the top commit in this branch and only merge when we know the release succeeded.

Okay, let's do it that way.

> I'd like to standardize around `v`-prefixed tags as it's more common with Git.

Sure, I'm happy with or without. 

> > * create a GitHub release for 7.5.0
> > 
> > and that should take care of the publishing side.
> 
> This is what I want to do right away, once the PR is ready but unmerged. Doing this will automatically create a tag so that does not need to be a separate step.
> 
> Though, I noticed you made a branch in your fork which would present a challenge following the process I intended. Could you push the release branch upstream instead?

Yeah, why don't I replace this PR with a fresh one from a branch on the upstream repo?
I can incorporate the minor fixes as well.

> > And then I'll share in places like DPO that we've finally got this out.
>
> Also, on bsky / mastodon / gh issues that people are watching. We might want to look into (partially) automating this later on.

:100: I would like us to make "announcing the release" a part of our release workflow. We'll get there!

webknjaz:
> Yeah, why don't I replace this PR with a fresh one from a branch on the upstream repo? I can incorporate the minor fixes as well.
> 

Yep, I don't think GH allows changing the incoming branch/fork in PRs, only the base.

sirosen:
Replaced this with #2212 !

PR #2214, opened 2025-07-31
Update the 'pipsupported' version to 25.2

sirosen:
Update the 'pipsupported' build, now that there's a new release.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/changelog.d/2214.contrib.md
+++ b/changelog.d/2214.contrib.md
@@ -1,0 +1,1 @@
+`pip-tools` now tests on and officially supports `pip` version 25.2 .
--- a/changelog.d/2214.contrib.md
+++ b/changelog.d/2214.contrib.md
@@ -1,1 +1,1 @@
-`pip-tools` now tests on and officially supports `pip` version 25.2 .
+`pip-tools` now tests on and officially supports `pip` version 25.2 -- by :user:`sirosen`.
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,7 @@ extras =
     coverage: coverage
 deps =
     pipsupported: pip == 24.3.1 ; python_version < "3.9"
-    pipsupported: pip == 25.1.1 ; python_version >= "3.9"
+    pipsupported: pip == 25.2 ; python_version >= "3.9"
     pipsupported: setuptools <= 80.0
 
     piplowest: pip == 22.2.* ; python_version < "3.12"


webknjaz:
Do we have pip unpinned in RTD / Sphinx land?

webknjaz:
By the way, there was talk about rephrasing the tox factors. I think @hugovk mentioned it being confusing.. But that would be a separate PR, of course.

sirosen:
> Do we have pip unpinned in RTD / Sphinx land?

Yeah, it looks like it. I'm surprised by it, but I'll take a look. Maybe just fold it into this fix, since it's "also pinning `pip`".

webknjaz:
Oh.. RTD is still configured with their control.

Could you convert it to use this config instead https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/27bdb93/.readthedocs.yaml, in a separate PR?

This also requires a small change in tox: https://github.com/ansible/awx-plugins/blob/c4180cd/tox.ini#L373-L394

webknjaz:
> I'm surprised by it

I bet the GHA job doesn't fail only because the wheels are cached and the installer doesn't bother to look for newer ones. Looking at `docs/requirements.txt`, `pip` isn't pinned in there because `docs/requirements.in` doesn't actually include `.`. But since `sphinx.ext.autodoc` would attempt importing various bits of `piptools`, it makes sense that it hit a change related to the pip 25.2 release.

changelog.d/2214.contrib.md
```diff
@@ -0,0 +1 @@
+`pip-tools` now tests on and officially supports `pip` version 25.2 .
```
jayaddison: NB: I think a byline may be missing here?  (noticed during preparation of #2217)

```suggestion
`pip-tools` now tests on and officially supports `pip` version 25.2 .
 -- by {user}`sirosen`
```

webknjaz: Strictly speaking, it's not mandatory. But yes, it could be added if Stephen wants. Another bit that caught my eye is a stray whitespace before the trailing period..

Note that if byline is standalone, it doesn't need a leading whitespace.

jayaddison: Perhaps we could avoid the padding space by applying some rephrasing:

```suggestion
`pip-tools` now tests on and officially supports version 25.2 of `pip`.
```

(I don't like placing decimal-point-delimited version numbers next to a full-stop, either)

sirosen: I like the idea (yes, I find it awkward to put punctuation right next to an unadorned version number), but it turned out not to be necessary after adding a byline.

sirosen: I don't mind always including the byline if it helps us keep the style consistent for contributor PRs.

I have a couple of projects in which I give "`Thanks @user`" for 3rd party contributions and not for maintainer-driven changes, which is probably the habit that led to not having one here.

sirosen:
I've given this a rebase to try to get things in order for us to do a 7.5.1 release.

Also addressed the feedback by adding a byline. (And if you look at the PR history, you'll see that I got the co-authorship wrong at first and amended it. :sweat_smile: )

webknjaz (approved):

PR #2219, opened 2025-08-04
Pin pip-tools' dependencies for "build-docs"

sirosen:
_This is split off from #2218. This part will have no changelog, #2218 will have a contrib changelog entry._

In order to get consistent doc builds, we need the dependencies of
`pip-tools` (in particular, right now we need `pip`) pinned to
versions where the build works.

In order to provide a `requirements.txt` file with these pins, we add
`.` to `docs/requirements.in`. This requires a couple of tweaks:

- Use `-e .` so that `pip-compile` preserves it as a relative path
- Remove the `changedir` tox directive, so that doc building and
  pinning happens from the tox ini dir (repo root)

Additionally, `pip` itself is downgraded from 25.2 (where we have the
doc build issue today) to 25.1 .

Other updates to `docs/requirements.txt` occur as side-effects of this
change. New packages are added and paths are updated.

##### Contributor checklist

##### Maintainer checklist


webknjaz:
I'm so glad I asked to add the changelog draft job into the CI 😂

sirosen:
Oh, I forgot to post this comment! It's been sitting in a tab for a few hours! :sweat_smile: 

CI failure on the changelog-draft job:
```
DEPRECATION: Constraints are only allowed to take the form of a package name and a version specifier. Other forms were originally permitted as an accident of the implementation, but were undocumented. The new implementation of the resolver no longer supports these forms. A possible replacement is replacing the constraint with a requirement. Discussion can be found at https://github.com/pypa/pip/issues/8210
ERROR: Unnamed requirements are not allowed as constraints
```

So I don't think I can do this the way that I wanted.

That error makes it look like I could hack it by using `pip-tools @ .`, but that's seems fragile. Not sure what the best path is yet.

webknjaz:
This is something I've been annoyed about for quite some time. We should really work on a better story for `.` vs.`-e .` and all the bits surrounding this in the context of constraints...

sirosen:
I've made an additional tweak here to use `./pyproject.toml` as a CLI input to the `pip-compile` command, which looks to me like it gives us our desired result. Since that removes `pip-tools` from `docs/requirements.txt`, I've reverted the change to `skip_install`.

webknjaz:
Oh, that's a clever workaround. Nice!

Issue #2215, opened 2025-08-01
`pip-compile` fails when `Click` is a dependency


merelcht:
With the latest `pip-tools` version `7.5.0` the `pip-compile` command fails when `Click` is one of the project dependencies. 

Whereas with the previous versions it just runs fine.

#### Environment Versions

1. OS Type: Both Unix and Windows (see broken pip-compile builds here https://github.com/kedro-org/kedro/actions/runs/16662774430/job/47182583460)
1. Python version: `$ python -V`: all versions from 3.9-3.12
1. pip version: `$ pip --version`: pip 25.2
1. pip-tools version: `$ pip-compile --version` 7.5.0

#### Steps to replicate

1. Clone out the `kedro` repository: https://github.com/kedro-org/kedro
2. Install the kedro dependencies  : `make install-test-requirements`
`make install-pre-commit`, `uv pip install "pip-tools>=6.5"`
3. Run `pip-compile`

#### Expected result

A successful `pip-compile` e.g.:

```
# This file is autogenerated by pip-compile with Python 3.11
# by the following command:
#
#    pip-compile
#
antlr4-python3-runtime==4.9.3
    # via omegaconf
appdirs==1.4.4
    # via kedro-telemetry
arrow==1.3.0
    # via cookiecutter
attrs==25.3.0
    # via
    #   kedro
    #   kedro (pyproject.toml)
binaryornot==0.4.4
    # via cookiecutter
build==1.2.2.post1
    # via
    #   kedro
    #   kedro (pyproject.toml)
cachetools==6.1.0
    # via
    #   kedro
    #   kedro (pyproject.toml)
certifi==2025.7.14
    # via requests
chardet==5.2.0
    # via binaryornot
charset-normalizer==3.4.2
    # via requests
click==8.1.8
    # via
    #   cookiecutter
    #   kedro
    #   kedro (pyproject.toml)
cookiecutter==2.6.0
    # via
    #   kedro
    #   kedro (pyproject.toml)
dynaconf==3.2.11
    # via
    #   kedro
    #   kedro (pyproject.toml)
fsspec==2025.7.0
    # via
    #   kedro
    #   kedro (pyproject.toml)
gitdb==4.0.12
    # via gitpython
gitpython==3.1.45
    # via
    #   kedro
    #   kedro (pyproject.toml)
idna==3.10
    # via requests
importlib-metadata==8.7.0
    # via
    #   kedro
    #   kedro (pyproject.toml)
[...]

PR #2221, opened 2025-08-04
When statically parsing pyproject.toml, convert relative paths to absolute for use as file URIs

sirosen:
When statically parsing a `pyproject.toml`, we can encounter self-referential package usage. For example, an extra which names another extra on the same package.

The source for these requirements is the current package, but as a concrete requirement, not an abstract one. To make it concrete, we use a URI dependency link pointed at the package directory as a file-URI.

When the input path to `pyproject.toml` is relative, it needs to be made absolute for use as a URI. Failing to do so results in errors.

Because the `pip-compile` output of an such an extra depends concretely on the specific version of the package found in-tree, and not abstractly on the package name (without the URI), tests confirm that we see the rendered package directory as a file URI in `pip-compile` output.

##### Contributor checklist

##### Maintainer checklist


changelog.d/2215.bugfix.md
```diff
@@ -0,0 +1,2 @@
+Fix static parsing of `pyproject.toml` data when the `pyproject.toml` is
```
webknjaz:
```suggestion
Fix static parsing of {file}`pyproject.toml` data when the {file}`pyproject.toml` is
```

tests/test_build.py
```diff
@@ -95,6 +96,44 @@ def test_static_parse_valid(tmp_path):
     assert metadata.extras == ("baz",)
 
 
+@pytest.mark.parametrize("input_path_is_absolute", (True, False))
```
webknjaz: Could you add IDs here?

tests/test_cli_compile.py
```diff
@@ -4044,3 +4044,63 @@ def test_second_order_requirements_relative_path_in_separate_dir(
             # via -r {output_path}
         """
     )
+
+
+@pytest.mark.parametrize("pyproject_path_is_absolute", (True, False))
```
webknjaz: IDs plz?

tests/test_cli_compile.py
```diff
@@ -4044,3 +4044,63 @@ def test_second_order_requirements_relative_path_in_separate_dir(
...
+def test_that_self_referential_pyproject_toml_extra_can_be_compiled(
+    pip_conf, runner, tmp_path, monkeypatch, pyproject_path_is_absolute
+):
+    """
+    Test that a `pyproject.toml` source file can use self-referential extras
```
webknjaz: Docstrings are usually parsed as RST:
```suggestion
    Test that a :file:`pyproject.toml` source file can use self-referential extras
```

tests/test_cli_compile.py
```diff
@@ -4044,3 +4044,63 @@ def test_second_order_requirements_relative_path_in_separate_dir(
...
+                "--extra",
+                "ext2",
+                input_path,
+            ],
+        )
+
+    assert out.exit_code == 0
+    assert out.stdout == dedent(
+        f"""\
+        foo[ext1] @ {src_file.parent.absolute().as_uri()}
```
webknjaz, line 4182: Is this really what we want? I don't think people would expect absolute paths from someone else's computer to end up in their lock files..

sirosen: I agree, and it's why I was pretty hesitant at first. But also, that's the behavior from 7.4.1 -- it's just not tested.

In theory, someone could be using this in some controlled context like an image build and _wanting_ the concrete dependency path that this provides. So although I'm uncertain that it's right for the long term, I don't want to treat it as blocking the fix.

webknjaz: Wait, did it always use absolute paths in 7.4.1? Not relative? I thought it'd conform to the shape of input that the 7.5.0 changelog promises.. Can we use `file://./rel/path/`?

sirosen: The path in this case is part of the `InstallRequirement` itself -- the "link" attribute is populated with this file URI.
The normalization in 7.5.0 only applies to the `comes_from` data ("via").

We could reach down into this as well and convert to `file://./rel/path/` as you say, but I'm not sure if we should. It would be a separate bit of data transformation -- we could share code but it's a different attribute of the parsed requirements data that we'd be manipulating.

webknjaz: I suppose, we could postpone that for now. There might be a use case coverage missing for stripping the `@ http://..source` in some contexts.

webknjaz:
Oh, looks like Windows is unhappy about the delimiters again.

sirosen:
The thing I broke in the tests is treatment of `comes_from` data in this scenario. We're stringifying the path to the package dir here:
https://github.com/jazzband/pip-tools/blob/6e0ef336c9cad949753b6e1cc8bc3fbfecabd860/piptools/build.py#L91

It's easy to "fix" by just adding an `as_posix()` call there, but I can imagine this becoming inconsistent with the behavior when that information is rendered via some other path. I'm going to run some experiments later this week to try to figure out what the variations are on this.

maurerle:
Since v7.5.0 my `pip-compile -o requirements.txt pyproject.toml` does not work.
It throws 

```
  File "/home/user/.conda/envs/assume/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/user/.local/lib/python3.12/site-packages/click/core.py", line 1442, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.local/lib/python3.12/site-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/user/.local/lib/python3.12/site-packages/click/core.py", line 1226, in invoke
  File "/home/user/.local/lib/python3.12/site-packages/click/core.py", line 794, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.local/lib/python3.12/site-packages/click/decorators.py", line 34, in new_func
  File "/home/user/.conda/envs/assume/lib/python3.12/site-packages/piptools/scripts/compile.py", line 371, in cli
    metadata = build_project_metadata(
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/assume/lib/python3.12/site-packages/piptools/build.py", line 159, in build_project_metadata
    project_metadata = maybe_statically_parse_project_metadata(src_file)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/assume/lib/python3.12/site-packages/piptools/build.py", line 106, in maybe_statically_parse_project_metadata
    requirement.url = src_file.parent.as_uri()
                      ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/assume/lib/python3.12/pathlib.py", line 467, in as_uri
    raise ValueError("relative path can't be expressed as a file URI")
ValueError: relative path can't be expressed as a file URI
```

the same error is thrown when just running `pip-compile`.
Is this fixed with this PR?

sirosen:
It might or it might not. There's not enough information to say. You should look at the initial issue, which is #2215 and see if you match the conditions described there (a self-referential extra).

maurerle:
Yes, this is exactly my issue. Would be very good to have this.
For now I either have to pin "pip-compile<7.5" or run pip-compile with "$PWD/pyproject.toml" to fix this.

sirosen:
Please use the issues for any open ended discussion or questions. I'd like to keep the PR discussion focused on the actual implementation.

sirosen:
I've just gotten back to this, taken a look at the open threads and marked the ones which were applied as suggestions as resolved.

I looked at what happens in the case of dynamic metadata, and we have a similar bit of path formatting in there.

I'd like to resolve this by adding calls to `as_posix()` to both locations in `piptools/build.py`.
One is [the place mentioned above](https://github.com/jazzband/pip-tools/blob/78f18e11386c3c696da4ede645bf082925f8aae5/piptools/build.py#L91), where a static build is happening and we're writing the `comes_from` data.
The other is [this line in this function](https://github.com/jazzband/pip-tools/blob/78f18e11386c3c696da4ede645bf082925f8aae5/piptools/build.py#L305) which does the same work on requirements from dynamic metadata.

@webknjaz, are you okay with that increase in scope? I'll need to add some new tests to cover it and another changelog note.
As an overall project direction, I think it's best to try to make all of our outputs posix paths, to make the outputs consistent.

webknjaz:
@sirosen agreed. Increasing the scope sounds justified.

sirosen:
I've rebased to keep this fresh with the latest fixes in `main` and added a new changelog note `changelog.d/2221.bugfix.md` to cover the new path normalization which happens.

With the `click`-related issue starting to impact users -- and now that I'm clear that it's not a `click` bug but an intentional change in behavior -- I'm wanting to push this forward and try to narrow what we put into 7.5.1 .

sirosen:
I think something is wrong with the codecov integration.
The report on uncovered lines doesn't make much sense:
https://app.codecov.io/gh/jazzband/pip-tools/pull/2221?dropdown=coverage

One thing I notice is that it seems to mark everything related to `piplowest` testing as uncovered, which suggests that somehow we're not producing distinct coverage data for those builds in a way that codecov understands.

I don't think it's at all related to this PR.

webknjaz:
@sirosen will this address #2215 / #2233 ?

changelog.d/2215.bugfix.md
webknjaz, line 1: Should we symlink 2233 (upon confirmation), also 2221 (or `2221.bugfix.1.md` rather)?

webknjaz (approved):
Looks good, but maybe add those change note refs before merging..

changelog.d/2215.bugfix.md
sirosen: Yeah, #2233 is getting fixed here as well, so I added the symlink. And I did the `2221.bugfix.1.md` one as well, since "why not?" :smile: 

I'm going to put this in the merge queue, along with any other changes I can round up right now.

webknjaz: @sirosen the issues didn't end up auto-closed. Next time, plz try to remember to mark the PR as closing them (via web UI or closes/resolves/fixes keywords).

PR #2224, opened 2025-08-04
Bugfix: path normalization applied to URLs

### LINKED ISSUE #2223: pip-tools 7.5.0 strips a slash from constraint URLs
After upgrading from pip-tools 7.4.1 to 7.5.0, `pip-compile` outputs URLs in "via" lines with a slash missing. 

#### Environment Versions

1. Linux
1. Python version: 3.12.3
1. pip version: 24.0
1. pip-tools version: 7.5.0

#### Steps to replicate

Create file `demo.in`, referencing an HTTPS constraint URL that constrains one of the dependencies:

```
django
-c https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt
```

Run `pip-compile -o demo.txt demo.in` and observe the via-comments.

#### Expected result

Constraint URL is preserved intact in comment

#### Actual result

Constraint URL is missing a slash from the authority delimiter:

```
...
django==4.2.23
    # via
    #   -c https:/raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt
    #   -r demo.in
...
```

sirosen:
Resolves #2223

The `-r` and `-c` path normalization logic did not check if the
requested requirements or constraint files were remote files being
loaded, e.g., via HTTPS. As a result, a usage like
`-c https://example.com/constraints.txt` incorrectly got converted to
a path and normalized, stripping one of the two slashes which
separates the scheme from the netloc.

A regression test is added which reproduces the bug, and the gap in
detection is closed by checking explicitly if inputs can parse as
URIs. Any given URI (including file URIs) will be exempted from the
path rewrites.

The check is implemented by calling `urllib.parse.urlparse()` and
checking the `scheme` attribute of the parse result.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/_compat/pip_compat.py
+++ b/piptools/_compat/pip_compat.py
@@ -191,7 +191,7 @@ def _is_remote_pip_uri(value: str) -> bool:
     The test is performed by trying a URL parse and reading the scheme.
     """
     scheme = urllib.parse.urlsplit(value).scheme
-    return scheme in ("http", "https", "file")
+    return scheme in {"file", "http", "https"}
 
 
 def create_wheel_cache(cache_dir: str, format_control: str | None = None) -> WheelCache:
--- a/piptools/_compat/pip_compat.py
+++ b/piptools/_compat/pip_compat.py
@@ -141,8 +141,8 @@ def _relativize_comes_from_location(original_comes_from: str, /) -> str:
     # split on the space
     prefix, space_sep, suffix = original_comes_from.partition(" ")
 
-    # if the value part is a URI, return the original
-    if _is_uri(suffix):
+    # if the value part is a remote URI for pip, return the original
+    if _is_remote_pip_uri(suffix):
         return original_comes_from
 
     file_path = pathlib.Path(suffix)
@@ -174,8 +174,8 @@ def _normalize_comes_from_location(original_comes_from: str, /) -> str:
     # split on the space
     prefix, space_sep, suffix = original_comes_from.partition(" ")
 
-    # if the value part is a URI, return the original
-    if _is_uri(suffix):
+    # if the value part is a remote URI for pip, return the original
+    if _is_remote_pip_uri(suffix):
         return original_comes_from
 
     # convert to a posix-style path
@@ -183,23 +183,15 @@ def _normalize_comes_from_location(original_comes_from: str, /) -> str:
     return f"{prefix}{space_sep}{suffix}"
 
 
-def _is_uri(value: str) -> bool:
+def _is_remote_pip_uri(value: str) -> bool:
     """
-    Test a string to see if it is a URI.
+    Test a string to see if it is a URI treated as a remote file in ``pip``.
+    Specifically this means that it's a 'file', 'http', or 'https' URI.
 
-    The test is performed by trying a URL parse and seeing is a scheme is populated.
-
-    This means that according to this rule, valid URLs such as
-    ``example.com/data/my_pip_constraints.txt`` may fail to count as URIs.
-    However, we cannot safely distinguish such strings from real filesystem paths.
-    e.g., ``./example.com/`` may be a directory.
-
-    Importantly, realistic usage such as
-    ``-c https://example.com/constraints.txt``
-    is properly detected by this technique.
+    The test is performed by trying a URL parse and reading the scheme.
     """
-    parse_result = urllib.parse.urlparse(value)
-    return parse_result.scheme != ""
+    scheme = urllib.parse.urlsplit(value).scheme
+    return scheme in ("http", "https", "file")
 
 
 def create_wheel_cache(cache_dir: str, format_control: str | None = None) -> WheelCache:
--- a/piptools/_compat/pip_compat.py
+++ b/piptools/_compat/pip_compat.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import optparse
 import pathlib
+import urllib.parse
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Callable, Iterable, Iterator, Set, cast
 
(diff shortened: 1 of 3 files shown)


sirosen:
This fix was not thought all the way through by me for the Windows context. `urllib.parse.urlparse` isn't going to cut it, since `C:/foo` parses to a "url" with a scheme of "c". I think the rest of the framework for the fix is right, but more sophisticated detection is needed.

sirosen:
Thanks to @ichard26 for a useful pointer when I asked in pypa discord about how `pip` treats these URIs!

Specifically:
https://github.com/pypa/pip/blob/20b39ec104b94181c01731192d03cbe0d1f9f3a7/src/pip/_internal/req/req_file.py#L564-L591

I've updated to simply mirror that set of known schemes here, which nicely solves the problem of `C:/foo/requirements.txt` looking like a URL with `scheme = "C"`.

piptools/_compat/pip_compat.py
```diff
@@ -169,11 +174,26 @@ def _normalize_comes_from_location(original_comes_from: str, /) -> str:
...
 
+def _is_remote_pip_uri(value: str) -> bool:
+    """
+    Test a string to see if it is a URI treated as a remote file in ``pip``.
+    Specifically this means that it's a 'file', 'http', or 'https' URI.
+
+    The test is performed by trying a URL parse and reading the scheme.
+    """
+    scheme = urllib.parse.urlsplit(value).scheme
+    return scheme in ("http", "https", "file")
```
webknjaz: A set reads slightly better semantically.
```suggestion
    return scheme in {"file", "http", "https"}
```

piptools/_compat/pip_compat.py
```diff
@@ -169,11 +174,26 @@ def _normalize_comes_from_location(original_comes_from: str, /) -> str:
...
 
 
+def _is_remote_pip_uri(value: str) -> bool:
+    """
+    Test a string to see if it is a URI treated as a remote file in ``pip``.
+    Specifically this means that it's a 'file', 'http', or 'https' URI.
+
+    The test is performed by trying a URL parse and reading the scheme.
+    """
+    scheme = urllib.parse.urlsplit(value).scheme
```
webknjaz, line 193: Are there any cases where `urlsplit()` would raise an exception?

tests/test_cli_compile.py
```diff
@@ -4044,3 +4045,78 @@ def test_second_order_requirements_relative_path_in_separate_dir(
...
+):
+    """
+    Test normalization of ``-c`` constraints when the constraints are HTTPS URLs.
+    The constraints should be preserved verbatim.
+
+    This is a regression test for
+    https://github.com/jazzband/pip-tools/issues/2223
+    """
+    reqs_in = tmp_path / "requirements.in"
+    constraints_url = "https://example.com/files/common_constraints.txt"
```
webknjaz: I'd swap these so that `reqs_in` initialization and use are close to each other:
```suggestion
    constraints_url = "https://example.com/files/common_constraints.txt"

    reqs_in = tmp_path / "requirements.in"
```

tests/test_cli_compile.py
```diff
@@ -4044,3 +4045,78 @@ def test_second_order_requirements_relative_path_in_separate_dir(
...
+    # although this is fragile, it can be adapted if pip changes
+    def fake_url_get(url):
+        response = mock.Mock()
+        response.reason = "Ok"
+        response.status_code = 200
+        response.url = url
+        response.text = "small-fake-a==0.2"
+        return response
+
+    mock_get = mock.Mock(side_effect=fake_url_get)
```
webknjaz, line 4094: Normally, I'd reach the mocking API via `mocker` (`pytest-mock`). But it looks like it's not in the deps so we can postpone that.

tests/test_cli_compile.py
```diff
@@ -4044,3 +4045,78 @@ def test_second_order_requirements_relative_path_in_separate_dir(
...
+                [
+                    "--output-file",
+                    "-",
+                    "--quiet",
+                    "--no-header",
+                    "--no-emit-options",
+                    "-r",
+                    input_path,
+                ],
+            )
```
webknjaz: Would this be more elegant?
```suggestion
        revertable_ctx.setattr(PipSession, "get", mock_get)
        out = runner.invoke(
            cli,
            [
                "--output-file",
                "-",
                "--quiet",
                "--no-header",
                "--no-emit-options",
                "-r",
                input_path,
            ],
        )
```

piptools/_compat/pip_compat.py
sirosen: I don't believe there are any (reasonably reachable) cases. `pip`'s usage doesn't take any effort to sanitize the inputs or check for errors. I looked at [the definition](https://github.com/python/cpython/blob/30194914c29d852143b930f007f6a663cfe4bf8f/Lib/urllib/parse.py#L470) and tried some strange inputs and didn't find any error cases (including obvious things like `""`) either by reading/deduction or by experimentation.

tests/test_cli_compile.py
sirosen: Yeah, I'm not very attached to this manner of mocking. It's not ideal but I don't think it's too terribly fragile or awkward.

I'd like to see if we can use [`responses`](https://github.com/getsentry/responses), since `pip` is using a modified `requests` session object for its requests. My experiences with `responses` have been excellent, so I think that if it works, that would be great.

tests/test_cli_compile.py
sirosen: Yes, I think so! Means the same thing and saves a layer of nesting. Mixing `monkeypatch` and `unittest.mock` leaves me all confused about which one I ought to be using. :sweat_smile:

PR #2236, opened 2025-09-19
Update to support click >= 8.3

### LINKED ISSUE #2235: Support click v8.3+
`click` version 8.3 just released, and causes failures for `pip-tools`.

I suspect the failures are related to the new `UNSET` sentinel, which is the default for flags and unset options rather than `None`.

sirosen:
Resolves #2235

The latest `click` release makes two changes which are visible in `pip-tools`, one of which is only seen in the testsuite.

1.  The default for `Parameter.default` is now an internal `UNSET` sentinel.

    Because `UNSET` isn't in the public API, checking `Option.default` is not the right way to check if an option has no explicit default value. We could use `Option.to_info_dict()`, but we're *also* checking for a falsy value right now -- the simple fix is to check for a parsed value of `None`.

2.  Changes to EOF handling in `CliRunner.isolation()` result in the stream being closed on exit.

    Between pallets/click#2934 and pallets/click#2940, we now get the intended behavior for `CliRunner.isolation()`, in that it outputs an EOF when the context manager exits. To solve, update a test to read stderr before exiting the context manager.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/changelog.d/2235.bugfix.md
+++ b/changelog.d/2235.bugfix.md
@@ -1,2 +1,2 @@
-Fix an incompatibility with `click>=8.3` which made `pip-compile` display incorrect
+Fix an incompatibility with ``click >= 8.3`` which made ``pip-compile`` display incorrect
 options in the compile command in output headers -- by :user:`sirosen`.
--- a/changelog.d/2235.bugfix.md
+++ b/changelog.d/2235.bugfix.md
@@ -1,0 +1,2 @@
+Fix an incompatibility with `click>=8.3` which made `pip-compile` display incorrect
+options in the compile command in output headers -- by :user:`sirosen`.
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -399,7 +399,7 @@ def get_compile_command(click_ctx: click.Context) -> str:
                 continue
 
         # Skip options without a value
-        if option.default is None and not value:
+        if value is None:
             continue
 
         # Skip options with a default value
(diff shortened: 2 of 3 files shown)


webknjaz:
Oh, so that's why I started getting `pip-compile --allow-unsafe --cert=None --client-cert=None --index-url=None --output-file=<filename>.txt --pip-args=None --strip-extras <filename>.in` in headers!

changelog.d/2235.bugfix.md
```diff
@@ -0,0 +1,2 @@
+Fix an incompatibility with `click>=8.3` which made `pip-compile` display incorrect
```
webknjaz:
```suggestion
Fix an incompatibility with ``click >= 8.3`` which made ``pip-compile`` display incorrect
```

webknjaz (approved):
LGTM, thanks!

webknjaz:
@sirosen any idea why the CI is failing?

sirosen:
Unfortunately, no, I'm not clear on what's up. We can see from PR that the prior test passed... but now it fails.

I can't devote my full attention to it right now, but I'll continue to probe slowly. I'm just rerunning a couple of times first to see if it's flaky.

sirosen:
The failures appear to have been related to tests which are marked as requiring external network access. I _think_ that means that those tests are subject to potential influence by the state of pypi.org, etc, and I'm not sure that we have cause to block on them.

I'm putting this into the merge queue.

changelog.d/2235.bugfix.md
anasanas89647, line 1: Le

PR #2242, opened 2025-09-26
Separately control CPython versions in nightly CI

### LINKED ISSUE #2226: Upgrade Python 3.8 jobs against pip's main branch in nightly CI to Python 3.9
Pip dropped Python 3.8 and now those jobs are failing every day. We could still suport Python 3.8 with older pips until it goes EOL, I suppose.

sirosen:
The pip versions being tested are controlled via a JSON array passed as
a string. Use the same mechanism to pass CPython versions to test.

This allows PR builds to continue to test 3.8 (EOL) and 3.9 (EOL soon)
while nightly builds will only run on 3.10+ .

Resolves #2226

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/changelog.d/2226.misc.md
+++ b/changelog.d/2226.misc.md
@@ -1,0 +1,2 @@
+The CPython versions tested in nightly CI runs are now separate from
+branch and PR CI, and don't include very old versions -- by {user}`sirosen`.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,12 @@ on:
           to test against under CPython.
         required: true
         type: string
+      cpython-versions:
+        description: >-
+          A JSON string with CPython versions
+          to test against.
+        required: true
+        type: string
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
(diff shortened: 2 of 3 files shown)


webknjaz (approved):
Alright.. This is fine short-term but what I *really* want is to push matrix definitions *out* of the reusable workflows so the reusable workflow would only have one job with one python/os/arch combo. This worked really well for `reusable-tox.yml` and we should work towards moving in the direction of the same style.

When this is done, we'd be able to control pairs of pip+python more granularly, on the calling side.

changelog.d/2226.misc.md
```diff
@@ -0,0 +1,2 @@
+Narrow the CPython versions tested in nightly CI runs and separate these from
```
webknjaz: non-imperative plz?

sirosen: :+1: I'll fix it up.

sirosen:
> This is fine short-term but what I really want is to push matrix definitions out of the reusable workflows so the reusable workflow would only have one job with one python/os/arch combo.

This sounds like a wonderful longer term solve. Let's queue it up as something to work on.

I'm not a huge fan of the current solution, but I saw it as the shortest path towards healthier CI. (And this was listed for 7.5.1 so I wanted to choose either "put in a relatively easy solution" or "explicitly move it to 7.5.2".)

PR #2254, opened 2025-10-28
🧪 Set up CI to rerun failed tests verbosely

webknjaz:
This is useful for getting detailed troubleshooting clues. The snippet is adopted from my other projects.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/changelog.d/2254.contrib.md
+++ b/changelog.d/2254.contrib.md
@@ -1,0 +1,5 @@
+The CI is now set up to invoke failed tests again with
+maximum level of detail -- by {user}`webknjaz`.
+
+The change is aimed at helping troubleshoot failures
+that might be difficult to reproduce locally.


webknjaz:
Demo output: https://github.com/jazzband/pip-tools/actions/runs/18878312355?pr=2254

AndydeCleyre:
Great!

- Do we really want *maximum* verbosity here, or is  there a more practical signal:noise ratio?
- Would it be worthwhile to place any limits, like only re-running up to `n` failed tests?

webknjaz:
Dunno. I usually go for a bit less verbosity but here I wanted to see what I can squeeze out of it.

I think it's fine to have it maximum, though, because it's only supposed to run/be viewed when things fail. If one wants no verbosity, they can look into the previous step's output.

I'm not sure about limiting the number of tests — what if there's clues that are more obvious in the tests that are deselected?

Let's get this in as is and observe if it annoys people, then adjust.

webknjaz:
One other idea I had is printing out more instructions about what this step is. But I want to solve it in a more generic way for my `reusable-tox.yml` effort. So I'm trying not to make individual projects with copies of this thing diverge too much.

AndydeCleyre (approved):
Great, thanks!

PR #2253, opened 2025-10-26
Implement compatibility with pip 25.3

shifqu:
Removed `use_pep517` reference to enable compatibility with pip 25.3+

Resolves #2252.

##### Contributor checklist

- [x] Included tests for the changes.
- [X] A change note is created in `changelog.d/` (see [`changelog.d/README.md`](https://github.com/jazzband/pip-tools/blob/main/changelog.d/#readme) for instructions) or the PR text says "no changelog needed".

##### Maintainer checklist

- [ ] If no changelog is needed, apply the `skip-changelog` label.
- [x] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

Issue #2252, opened 2025-10-25
VERY URGENT pip-compile breaks with latest version of pip==25.3


vkruglik-aka:
#### Environment Versions

1. OS Type: linux
1. Python version: Python 3.11.14
1. pip version: pip 25.3 from /usr/local/lib/python3.11/site-packages/pip (python 3.11)
1. pip-tools version: pip-compile, version 7.4.1

#### Steps to replicate

1. Latest version of pip - 25.3 - released on Oct 24, 2025 officially removed the use-pep-517 option, so the type `InstallRequirement` no longer has the `use_pep517` member.
2. When I run `pip-compile` with the latest pip (25.3), pip compile suffers an exception upon attempting to access `template.use_pep517` in piptools/utils.py", line 492: `AttributeError: 'InstallRequirement' object has no attribute 'use_pep517'`

#### Expected result

`pip-compile` is expected not to be broken

#### Actual result
`pip-compile` is broken when running with the latest version of pip - 25.3
```
Traceback (most recent call last):
  File "/usr/local/bin/pip-compile", line 7, in <module>
    sys.exit(cli())
             ^^^^^
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1226, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 794, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/piptools/scripts/compile.py", line 460, in cli
    resolver = resolver_cls(
               ^^^^^^^^^^^^^
[...]

PR #2148, opened 2024-12-16
Add Python 3.12 to the CI matrix

webknjaz:
##### Contributor checklist

- [x] Included tests for the changes.
- [x] PR title is short, clear, and ready to be included in the user-facing changelog.

##### Maintainer checklist

- [x] Verified one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #2257, opened 2025-10-30
📦 Add a `stable` extra for `pip` upper bound

webknjaz:
It is supposed to correspond to the range of what's tested in our CI.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/changelog.d/2257.packaging.md
+++ b/changelog.d/2257.packaging.md
@@ -1,0 +1,10 @@
+The core packaging metadata started exposing a new `stable` extra
+-- by {user}`beruic`, {user}`sirosen` and {user}`webknjaz`.
+
+This new opt-in install-level feature limits the `pip` dependency to
+the latest version we're testing in our CI at the time of release.
+
+<!-- pyml disable-num-lines 2 commands-show-output -->
+```console
+$ pip install 'pip-tools [stable]'
+```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,9 @@ repository = "https://github.com/jazzband/pip-tools"
 changelog = "https://github.com/jazzband/pip-tools/releases"
 
 [project.optional-dependencies]
+stable = [
+  "pip <= 25.3",
+]
 testing = [
   "pytest >= 7.2.0",
   "pytest-rerunfailures",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,9 @@ repository = "https://github.com/jazzband/pip-tools"
 changelog = "https://github.com/jazzband/pip-tools/releases"
 
 [project.optional-dependencies]
+stable = [
+  "pip <= 26.2",
+]
 testing = [
   "pytest >= 7.2.0",
   "pytest-mock >= 3.15.1",
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Similar to `pip`, `pip-tools` must be installed in each of your project's
 
 ```console
 $ source /path/to/venv/bin/activate
-(venv) $ python -m pip install pip-tools
+(venv) $ python -m pip install 'pip-tools [stable]'
 ```
 
 **Note**: all of the remaining example commands assume you've activated your
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Similar to `pip`, `pip-tools` must be installed in each of your project's
 
 ```console
 $ source /path/to/venv/bin/activate
-(venv) $ python -m pip install pip-tools
+(venv) $ python -m pip install 'pip-tools [stable]'
 ```
 
 **Note**: all of the remaining example commands assume you've activated your
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,8 @@ skip_missing_interpreters = True
 [testenv]
 description = run the tests with pytest
 extras =
+    pipsupported: stable
+    piplowest: stable
     testing
     coverage: coverage
 deps =
--- a/tox.ini
+++ b/tox.ini
@@ -21,6 +21,8 @@ warnings-to-errors = -Werror
 [testenv]
 description = run the tests with pytest
 extras =
+    pipsupported: stable
+    piplowest: stable
     testing
     coverage: coverage
 deps =


webknjaz (commented):
We'll need some sort of linting for keeping these in sync w/ CI and perhaps docs.

pyproject.toml
```diff
@@ -56,6 +56,9 @@ repository = "https://github.com/jazzband/pip-tools"
 changelog = "https://github.com/jazzband/pip-tools/releases"
 
 [project.optional-dependencies]
+supported-pip = [
+  "pip < 25.3",
```
webknjaz: @sirosen any thoughts on whether this should be a range or whether we should include more deps in this extra?

sirosen: By "whether this should be a range", I understand you to mean "include a lower bound".

Because an extra is purely opt-in, and `pip-tools` will always be installable without it, I like the idea of publishing our full range. I find it unlikely to be problematic for anyone, but if it is they'll have a trivial workaround.

I don't think we should include other packages. Although we recently had some breakage with `click`, the demonstrated value with `pip` is the real motivation here.

sirosen (commented):
I am trying to decide what we should name this extra -- not taking it as a given that the proposed name, `supported-pip`, is best.

I think `supported-pip` may be confusing if we start pinning anything other than `pip` in the future. And I think we should at least start with `pip`-only, which makes the possibility of such confusion stronger in the future. So I think we should remove `pip` from the name of the extra.

That would leave us with `pip-tools[supported]`, which I think is good. I can't come up with anything better, at least!

pyproject.toml
webknjaz: @sirosen the lower bound of `"pip >= 22.2"` is already present in mandatory deps and will be merged with the upper bound here. I just thought it might be confusing to some. But yeah, it'll work w/o the lower limit.

I'm marking this thread as resolved, then.

webknjaz:
> I think `supported-pip` may be confusing if we start pinning anything other than `pip` in the future. And I think we should at least start with `pip`-only, which makes the possibility of such confusion stronger in the future. So I think we should remove `pip` from the name of the extra.
> 
> That would leave us with `pip-tools[supported]`, which I think is good. I can't come up with anything better, at least!

That's a good point. Alternatively, we could have multiple composable extras and the users could use `pip-tools [supported-click, supported-pip]`.

webknjaz:
@sirosen should we include this in 7.5.2?

sirosen:
I still feel uncertain about it, at least as far as naming is concerned.
I'm going to put it into the 7.5.3 milestone so that we're not feeling pressured at all to rush discussion here.

webknjaz:
Okay, fair.

pyproject.toml
beruic: Since #2253 merged, this should probably be:

```suggestion
  "pip <= 25.3",
```

beruic: Should this include a lower bound as well?

webknjaz: The lower bound is already present among mandatory runtime deps. See https://github.com/jazzband/pip-tools/pull/2257#discussion_r2492535863.

sirosen:
I have been thinking about the name here even more. The `supported` name is still "wrong" for me in a way that I'm having trouble ignoring, even though I want us to add this extra.

My main issue is that the opposite of "supported" is "unsupported", so I don't like the possible implication that `pip install 'pip-tools[supported]'` means "you are supported" and `pip install 'pip-tools'` means "you are unsupported".
We would obviously not behave that way, but it may be how users read it.

What if we used `stable` instead?
We could have this suite of extras:
```
stable-pip: pip <= {tested-version}
stable-click: click <= {tested-version}
stable: pip-tools[stable-pip], pip-tools[stable-click]
```
(We could start with `stable` alone, and refactor to this if we feel the need to provide more control.)

Stable/unstable is common terminology, and for Python users, familiar with RTD sites, stable/latest is also common.
Both of those seem to properly communicate "there's a chance that this could break which we can't fully control".

@webknjaz, WDYT? `pip-tools[stable]`? And the more I think about it, the more I'm okay with it capping the `click` version.

webknjaz:
`stable` sounds good. Could also be `tested` / `confirmed-working`.

pyproject.toml
beruic: Ok. It's only the updated `pip` dependency above then.

beruic: @webknjaz This thread has been resolved, but the suggestion had not been applied, and the version dependency has not been changed in any other way.

webknjaz: Looks like @sirosen marked it as resolved too early. I'll unresolve.

sirosen: Ah, my mistake, sorry! I thought the conversation was done and didn't realize the initial suggestion was still pending.

webknjaz: Yeah, I didn't want to rush with discussions ongoing + other efforts being prioritized.

webknjaz: Yeah, I didn't want to rush with discussions ongoing + other efforts being prioritized.

webknjaz: UPD: I've absorbed this suggestion into the rebased commit now.

webknjaz:
@sirosen I've made all the requested changes but I'm not sure if it's time to undraft this yet. Perhaps, we should have some linting/automation for keeping things in sync as a prerequisite for merging the patch.

changelog.d/2257.packaging.md
webknjaz, line 1: :books: https://pip-tools--2257.org.readthedocs.build/en/2257/changelog/#unreleased-changes

wyattscarpenter:
I'm quite looking forward to this extra as I've been requiring pip-tools as a dev dependency and have been stymied lately by https://github.com/jazzband/pip-tools/issues/2252 and a previous pip–pip-tools compatibility problem along the same lines. So it will be quite useful to me to be able to require this extra at some point, to easily give users a known-good configuration. (Although I may cease maintaining my project before this PR lands.)

With that in mind, I +0.5 the idea that this PR should be undrafted and incorporated as soon as possible, and the linting/automation to keep it in sync (which is also important) should be a subsequent PR. I mean, the stable requirement is hardly going to get less stable fast if it isn't updated, right? I'd like to be able to require [stable] as soon as the next release of pip-tools, if possible, which will give me compatibility now and effortless forwards-compatibility later; likely a strict improvement on the current situation where I pin on my end.

There also doesn't seem to be anywhere where the current compatibility with pip is listed in the docs, although I may have missed something. There used to be one with https://github.com/jazzband/pip-tools/issues/1089 but that seems gone now. This is a mixed blessing because it means I, personally, currently have nowhere to look for that, but also that's one fewer thing to ensure compatibility with using the aforementioned "automation for keeping things in sync".

This extra should probably be mentioned in the readme/documentation, and not just the changelog. Its name is not fully self-explanatory (and can't be; its name would have to be a complicated sentence in its own right to be fully explanatory) yet it is a feature that will be useful to end-users.

It seems to me like you should just require known-good dependencies by default and then rely on the user to pass `--no-dependencies` if they want to override that? But I'm no python packaging guru; I'm sure you have reasons for doing it the opt-in way.

webknjaz:
> With that in mind, I +0.5 the idea that this PR should be undrafted and incorporated as soon as possible, and the linting/automation to keep it in sync (which is also important) should be a subsequent PR. I mean, the stable requirement is hardly going to get less stable fast if it isn't updated, right? I'd like to be able to require [stable] as soon as the next release of pip-tools, if possible, which will give me compatibility now and effortless forwards-compatibility later; likely a strict improvement on the current situation where I pin on my end.

I'm a bit uncertain, because this increases the risk of having multiple places out-of-sync. But if @sirosen @beruic have different opinions, we could probably make such a split.

@wyattscarpenter FTR, you could technically start depending on this extra already — Python ecosystem usually ignores those that don't exist.

> There also doesn't seem to be anywhere where the current compatibility with pip is listed in the docs, although I may have missed something. There used to be one with #1089 but that seems gone now. This is a mixed blessing because it means I, personally, currently have nowhere to look for that, but also that's one fewer thing to ensure compatibility with using the aforementioned "automation for keeping things in sync".

Yeah, people wanted to get rid of the manually maintained table some time ago. I don't recall being too happy about it but I was also less involved, I suppose. Somebody from GH removed it: #1950.

It's been on my mind for quite a while too: https://github.com/jazzband/pip-tools/pull/2148#issuecomment-2546994236.

Actually, @sirosen and I also had the exact same conversation in the fall about needing to ressurect that. And I want this to be provided by a Sphinx extension instead of being hardcoded in README.

> This extra should probably be mentioned in the readme/documentation, and not just the changelog. Its name is not fully self-explanatory (and can't be; its name would have to be a complicated sentence in its own right to be fully explanatory) yet it is a feature that will be useful to end-users.

Agreed.

> It seems to me like you should just require known-good dependencies by default and then rely on the user to pass `--no-dependencies` if they want to override that? But I'm no python packaging guru; I'm sure you have reasons for doing it the opt-in way.

There's been a lot of interest in this over many years but we still don't have a metadata standard to express this. And so I don't think that any installer will support such feature until there's an accepted PEP.
If you're interested, there's PEP 771 which is a draft: https://discuss.python.org/t/pre-publish-pep-771-default-extras-for-python-software-packages/77892 / https://wheelnext.dev/proposals/pep771_default_extras/ / https://peps.python.org/pep-0771/ /

sirosen:
Yeah, I think we have to figure out what we want a little, which is why I've been avoiding putting any pressure on this PR to move faster.

> Actually, @sirosen and I also had the exact same conversation in the fall about needing to ressurect that. And I want this to be provided by a Sphinx extension instead of being hardcoded in README.

I'm still not totally sure what I think we should do about _older_ versions of `pip-tools`, which is to say "not the latest". But we definitely can keep a documented statement of "`pip-tools` is tested against `pip` versions `19.2` through `27.1`" (obviously fake numbers :wink: ).

For the current version, it's easier to decide to document it, but for older versions... we enter more murky territory, since we don't actively support old versions. I'd have trouble picking out how many we document.

---

I'm considering the merits of using the `pip-tools[stable]` extra as the source of truth for this information. If we have `pip-tools[stable]`, I think it should be possible to wire up
- a script which extracts the `pip` version bounds from project metadata (lower from `project.dependencies`, upper for `project.optional-dependencies.stable`)
- templating with [`cog`](https://pypi.org/project/cogapp/) or similar (I've been using it a lot lately and enjoying it) to inject those bounds into `tox.ini` and docs
- a CI job to check for drift

I think that's sufficiently separate from the goal of this PR that I wouldn't want to fold it in.

@webknjaz, what do you think of opening an issue for using `pip-tools[stable]` as the canonical source of truth, and unblocking this? I strongly agree that we want to keep things in sync, and we'd like that to be automated, but we have the ability to do it manually today.

webknjaz:
I wouldn't care about the older versions until somebody asks. That's busy work that we don't have a proof anybody will ever want to make use of. I also like Sphinx over `cog` because this allows not having extra commits to display stuff.

An issue sounds good.

webknjaz:
*I didn't get into detail in my previous response b/c I had a health incident and wasn't feeling well.* Here's a few more notes I had in mind.


---

> I'm considering the merits of using the `pip-tools[stable]` extra as the source of truth for this information

Either this, or the oposite direction would be an in-tree PEP 517 build backend wrapper. Having a standard backend is probably less maintenance overhead. And PEP 621 is good as the source of truth.


> * to inject those bounds into `tox.ini` and docs

Both have plugin systems in place. It's quite easy to use `toxfile.py` and `docs/conf.py` to inject centrally sourced info both runtimes. We'd probably make use of a shared utility module for parsing it.

For tox, this would be the [`tox_add_env_config() hook`](https://tox.wiki/en/latest/plugins.html#tox.plugin.spec.tox_add_env_config).
And for Sphinx, it'd be an RST directive — it's pretty easy to generate an RST table and feed it to the helpers, having it injected into the document (I've got some examples of prior art in `sphinxcontrib-towncrier` and old versions of `ansible-lint` or `ansible-navigator`, I think).


> * a CI job to check for drift

Ideally, the CI matrix would be generated from the same info w/o any need to have additional checks. OTOH, with the current mapping to tox factors with min/max pip, this isn't even needed in the current form.

PR #2249, opened 2025-10-09
Fix SyntaxWarning in Python 3.14

tusharsadhwani:
Seeing this warning on 3.14:

```
/Users/tushar/code/foo/venv/lib/python3.14/site-packages/piptools/build.py:207: SyntaxWarning: 'return' in a 'finally' block
```

This fixes that.

PR #2258, opened 2025-11-03
Replace return in finally in metadata build

EpicWink:
Fixes `AttributeError` when using `--upgrade-package` (and dynamically building project metadata) and pip internals change (causing an internal error):

```python
  File ".venv/lib/python3.9/site-packages/piptools/scripts/compile.py", line 385, in cli
    constraints.extend(metadata.requirements)
AttributeError: 'NoneType' object has no attribute 'requirements'
```

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/changelog.d/2258.bugfix.md
+++ b/changelog.d/2258.bugfix.md
@@ -1,0 +1,1 @@
+Using `--upgrade-package` (and dynamically building project metadata) no longer causes an {exc}`AttributeError` when a pip internals change causes an error -- by {user}`Epic_Wink` and {user}`tusharsadhwani`.


changelog.d/2258.bugfix.md
```diff
@@ -0,0 +1 @@
+Fix `AttributeError` when using `--upgrade-package` (and dynamically building project metadata) and a pip internals change causes an internal error -- by {user}`Epic_Wink`.
```
webknjaz: Let's rephrase this not to sound like a TODO-list item due to imperative mood.
```suggestion
Using `--upgrade-package` (and dynamically building project metadata) 
no longer causes an `AttributeError` in pip 25.3 internals -- by {user}`Epic_Wink` and {user}`tusharsadhwani`.
```

piptools/build.py
webknjaz, line 1: @EpicWink could you extract a `Co-Authored-By` from https://github.com/jazzband/pip-tools/pull/2249 to give them credit?

tests/test_build.py
```diff
@@ -78,6 +79,32 @@ def test_build_project_metadata_raises_error(tmp_path):
...
 
+def test_build_project_metadata_upgrading_raises_error(tmp_path):
+    """Test build_project_metadata doesn't swallow error."""
+    src_file = tmp_path / "pyproject.toml"
+    src_file.write_text(textwrap.dedent("""
+        [project]
+        # missing name
+        version = "0.1"
+        dependencies=["test_dep"]
+    """))  # fmt: skip
```
webknjaz: What happens w/o this? Suppressions should have a code comment with reasoning.

sirosen: I'd prefer we just didn't turn off the formatter here. I recognize that `black` isn't to everyone's taste, but the project-level choice to use it is exactly about not having this decision to make.

piptools/build.py
EpicWink: @sirosen is right: this was developed independently (@tusharsadhwani discovered the issue by running in Python 3.14 and named the pull request accordingly, whereas I got an indecipherable error after upgrading to pip 25.3 (on Python 3.9).

I don't know how to "extract a `Co-Authored-By`" anyway: I assume you mean for me to update the first commit's message with a footer, splitting out the new test case into a new commit. If you instead mean to include it in the merge commit, that's up to you maintainers.

I've at least included attribution in the changelog entry.

tests/test_build.py
EpicWink: I've removed the disablement and applied the formatting, but IMO my original formatting looks better and more readable.

piptools/build.py
webknjaz: @EpicWink here, I helped Hynek figure out some bits around this a few years back: https://hynek.me/til/easier-crediting-contributors-github/.

Yeah, I mostly wanted it to be somewhere in the commit since GH would then display multiple authors. It could be split, but that's up to you. I just wanted to make a suggestion to make sure that another person is included and their contribution isn't ignored.

webknjaz:
> Regarding the citation/co-authored line: it's possible that this was developed without reference to #2249 , so I don't think we need to ask for the co-authored-by marker. But I would like to share credit in the changelog (which is more visible anyway), even if it was written totally independently.

@sirosen yep, I had a feeling that it was developed independently. But since it's a de-facto identical patch, I wanted to make sure we remain inclusive even if we accept this patch as a part of somebody else's PR. Hence the suggestion.

changelog.d/2258.bugfix.md
```diff
@@ -0,0 +1 @@
+Using `--upgrade-package` (and dynamically building project metadata) no longer causes an `AttributeError` when a pip internals change causes an error -- by {user}`Epic_Wink` and {user}`tusharsadhwani`.
```
webknjaz:
```suggestion
Using `--upgrade-package` (and dynamically building project metadata) no longer causes an {exc}`AttributeError` when a pip internals change causes an error -- by {user}`Epic_Wink` and {user}`tusharsadhwani`.
```

webknjaz:
The Sphinx error at https://github.com/jazzband/pip-tools/actions/runs/19064778125/job/54492382188?pr=2258#step:10:158 does not immediately appear related:
```console
/home/runner/work/pip-tools/pip-tools/.tox/build-docs/lib/python3.14/site-packages/piptools/_compat/pip_compat.py:docstring of piptools._compat.pip_compat.FileLink:1: WARNING: py:class reference target not found: str | ForwardRef('IndexContent') | None
```

I wonder what changed..

webknjaz:
Looks like because that class inherits something from pip's private API, the types from pip start being included. Specifically, this thing: https://github.com/pypa/pip/blob/2b8b99522f2fc686ede57b0392cf2e3a50e27284/src/pip/_internal/models/link.py#L217C21-L217C46.

Interestingly, our released docs display that just fine: https://pip-tools.readthedocs.io/en/stable/pkg/piptools._compat.pip_compat/#piptools._compat.pip_compat.FileLink.

Perhaps, this is because pip added `from __future__ import annotations`: https://github.com/pypa/pip/commit/65fe65b5f517c289362e1c96fedefc852f111a91#diff-3919ca53335487395177edeffff8b60bd360f1ad18c95593287f6814b9ecefb8R1.

@sirosen do we just pin pip in sphinx envs for now? Perhaps, https://github.com/tox-dev/sphinx-autodoc-typehints will help, though.

OTOH, I'm not sure why this CI job started showing this warning so suddenly — the docs env is practically the only one where we pin the dep tree and it's stuck with older pip and sphinx.

sirosen:
I'm pretty sure the failures are not related to this PR -- I opened #2259 with what I think is the appropriate resolution to unblock this (and other PRs). Let's discuss over there if there's more nuance to this, but I think it's really just "CI picked up 3.14 before we were ready".

piptools/build.py
EpicWink: I've edited the commits to add the attribution, splitting the test into a new commit. I took the opportunity to clean up the commits, rebasing `main` and squashing edits.

changelog.d/2258.bugfix.md
```diff
@@ -0,0 +1 @@
+Using `--upgrade-package` (and dynamically building project metadata) no longer causes an {exc}`AttributeError` when a pip internals change causes an error -- by {user}`Epic_Wink` and {user}`tusharsadhwani`.
```
webknjaz, line 1: @EpicWink @sirosen looks like "causes an error" should be dropped from the phrase..

sirosen: Good catch. I'll submit a small follow up to correct it in just a bit.

PR #2108, opened 2024-06-26
Add compat modules for `tomllib` and `importlib.metadata`

chrysle:
The conditional backports increased, thus I added two compatibility modules.

##### Contributor checklist

- [ ] Included tests for the changes.
- [ ] PR title is short, clear, and ready to be included in the user-facing changelog.

##### Maintainer checklist

- [x] Verified one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #2260, opened 2025-11-04
Fix pip-compile handling of `-r` paths which are not subpaths of the cwd

### LINKED ISSUE #2231: pip-compile breaks when using paths outside of the current directory
In commit `809239a` the logic for relative/absolute path detection was reworked, however there seem to be some issues with it. 

When running pip compile with `pip-compile requirements.in` the following scenarios will cause errors:
- Having a -r with a relative path (`-r ../otherProject/requirements.in`)
- Having a -r with an absolute path (`-r `D:\code\otherProject\requirements.in)

Both of these break with the same error:
```
suffix = file_path.relative_to(pathlib.Path.cwd()).as_posix()
  File "C:\Python310\lib\pathlib.py", line 818, in relative_to
    raise ValueError("{!r} is not in the subpath of {!r}"
```

I understand why these specificity changes were necessary to prevent possible ambiguity, however there should be a fallback when the preferred path does not exist vs an error.

#### Environment Versions

1. OS Type: - Win 11
1. Python version: Python 3.10.11
1. pip version: pip 25.2
1. pip-tools version: pip-compile, version 7.5.0

#### Steps to replicate

Run pip compile with `pip-compile requirements.in` with the following scenarios:
- Having a -r with a relative path (`-r ../otherProject/requirements.in`)
- Having a -r with an absolute path (`-r `D:\code\otherProject\requirements.in)

#### Expected result
pip-compile would use the files specified since there were not other options which were more preferred.

#### Actual result

```
suffix = file_path.relative_to(pathlib.Path.cwd()).as_posix()
  File "C:\Python310\lib\pathlib.py", line 818, in relative_to
    raise ValueError("{!r} is not in the subpath of {!r}"
```

sirosen:
resolves #2231

When `-r` is used to pass a path, we are normalizing with `pathlib.Path.relative_to`. This fails when the input is not a subpath of the current working directory.

In Python 3.12+ pathlib supports this usage (`walk_up=True`), but on older Pythons we need to fallback to using `os.path.relpath`.

In order to support this usage and give a clear indication of how we should upgrade when older versions are dropped, a small compat wrapper is here defined, `_relative_to_walk_up` which uses `relative_to(..., walk_up=True)` when it is available, and only uses `os.path.relpath` when necessary. After we drop Python 3.11 in a few years, the helper can be removed.

A new regression test reproduces #2231 and is fixed with this patch.
As part of writing that test, I wanted to leverage the `TestFilesCollection` helper, so this was extended in a small way to support usage without an explicit name.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/piptools/_compat/path_compat.py
+++ b/piptools/_compat/path_compat.py
@@ -1,5 +1,5 @@
 """
-Compatibility helpers for working with paths and ``pathlib`` across platforms
+Compatibility helpers for working with paths and :mod:`pathlib` across platforms
 and Python versions.
 """
 
@@ -8,6 +8,8 @@ from __future__ import annotations
 import os.path
 import pathlib
 import sys
+
+__all__ = ("relative_to_walk_up",)
 
 
 def relative_to_walk_up(path: pathlib.Path, start: pathlib.Path) -> pathlib.Path:


changelog.d/2231.bugfix.md
```diff
@@ -0,0 +1,2 @@
+Fix `pip-compile` handling of relative path includes which are not subpaths of
```
webknjaz: non-imperative mood plz

changelog.d/2231.bugfix.md
webknjaz, line 1: symlink the PR?

piptools/_compat/pip_compat.py
```diff
@@ -210,3 +212,18 @@ def get_dev_pkgs() -> set[str]:
...
+
+    This is a compatibility helper for ``pathlib.Path.relative_to(..., walk_up=True)``
+    on all Python versions. (``walk_up: bool`` is Python 3.12+)
+    """
+    # prefer `pathlib.Path.relative_to` where available
+    if sys.version_info >= (3, 12):
+        return path.relative_to(start, walk_up=True)
+
+    str_result = os.path.relpath(path, start=start)
+    return pathlib.Path(str_result)
```
webknjaz: Perhaps this should go into a separate `pathlib_compat.py`?

changelog.d/2231.bugfix.md
sirosen: I need to break this habit of writing changelogs like commit messages. I even updated docs at work to remind people to use past tense for changelogs, and then I go around forgetting it myself! :laughing: 

Thanks for the save.

piptools/_compat/pip_compat.py
sirosen: My only concern would be that it might only ever have this one function.
What about a `path_compat.py` so that it's not only scoped to `pathlib`?

I wouldn't move other things over right away, but any helpers for handling drive letters and such could go there.

sirosen: I snooped around and immediately found a second helper to put in here!

I still prefer `path_compat.py`, so I'll go with that for the moment, but I'm happy to rename as well.

sirosen: I unfortunately ended up with a little bit of extra change needed ( https://github.com/jazzband/pip-tools/pull/2260/commits/3df3efa92a4316d2066d639110fba943cc56f9aa ) because my desire to put two related helpers together created an import cycle. I can revert if this feels like too much extraneous change.

changelog.d/2231.bugfix.md
webknjaz: I wonder if we can get an LLM help out with this. I've been looking into configuring some LLM behaviors (`CLAUDE.md` / `AGENTS.md`) recently and this should be a func exercise.. Although, we could probably also try Vale.

piptools/_compat/pip_compat.py
webknjaz: @sirosen circular imports is not a new problem. They've been annoying people trying to organize helpers for quite some time: https://github.com/jazzband/pip-tools/pull/2108.

I think we should postpone moving other things into the new helper module as to keep the PR atomic and prevent scope creep growing infinitely.

piptools/_compat/path_compat.py
```diff
@@ -0,0 +1,36 @@
...
+    # prefer `pathlib.Path.relative_to` where available
+    if sys.version_info >= (3, 12):
+        return path.relative_to(start, walk_up=True)
+
+    str_result = os.path.relpath(path, start=start)
+    return pathlib.Path(str_result)
+
+
+def is_path_relative_to(path1: pathlib.Path, path2: pathlib.Path) -> bool:
+    """Return True if ``path1`` is relative to ``path2``."""
```
webknjaz: The literal can also be linked.
```suggestion
    """Return :data:`True` if ``path1`` is relative to ``path2``."""
```

But as I said elsewhere, let's not include this helper.

piptools/_compat/path_compat.py
```diff
@@ -0,0 +1,36 @@
...
+    # TODO: remove this function in favor of Path.is_relative_to()
+    #       when we drop support for Python 3.8
```
webknjaz: Both 3.8 and 3.9 are both EOL already: https://endoflife.date/python. And testing against pip is becoming difficult in older runtimes.

I think we could plan dropping Python 3.8 once #2257 is implemented and released. With that, there would be no need to even think about hosting this function in compat modules.

piptools/_compat/path_compat.py
```diff
@@ -0,0 +1,36 @@
+"""
+Compatibility helpers for working with paths and ``pathlib`` across platforms
```
webknjaz:
```suggestion
Compatibility helpers for working with paths and :mod:`pathlib` across platforms
```

piptools/_compat/path_compat.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+
+from __future__ import annotations
+
+import os.path
+import pathlib
+import sys
+
+
+def relative_to_walk_up(path: pathlib.Path, start: pathlib.Path) -> pathlib.Path:
```
webknjaz, line 15: Let's be explicit about what's expected to be importable
```suggestion
__all__ = ("relative_to_walk_up",)


def relative_to_walk_up(path: pathlib.Path, start: pathlib.Path) -> pathlib.Path:
```

piptools/_compat/pip_compat.py
sirosen: I think you're right, especially coming back to this after a couple of days.
I'm going to partially revert, cutting it back to just this one helper in a `path_compat.py` module, as part of my rebase right now.

piptools/_compat/path_compat.py
sirosen: Since I reverted any move of this helper, I'm not changing the docstring, but :+1: to marking literals like this when possible.

changelog.d/2231.bugfix.md
sirosen: I'm interested in Vale, I think it looks very cool.

Less interested in using an LLM for this, personally. I'm waiting for the hype to die down, and plan to be a late adopter of the tech (after some not-great experiences trying it so far) assuming I can square it with my ethics.

webknjaz: I think it's important to have these files in the repo because people are going to use agentic help regardless of what we tell them. And if there's some guidelines, the chance of them producing rubbish would be decreased. With that, they wouldn't waste as much compute retrying or generating nonsense. So in my book helping such tools be efficient is definitely positive if we're talking about the environmental impact.

We just got such configuration in https://github.com/ansible/ansible/pull/85841 just now and I'm interested in exploring making such configs for projects better.

webknjaz: @sirosen re:vale — I have this old issue where I collect various prose checkers @ https://github.com/aio-libs/multidict/issues/276. If we succeed in integating any of them, I'd like to scale that to other projects.

webknjaz: > I think it's important to have these files in the repo because people are going to use agentic help regardless of what we tell them. And if there's some guidelines, the chance of them producing rubbish would be decreased. With that, they wouldn't waste as much compute retrying or generating nonsense. So in my book helping such tools be efficient is definitely positive if we're talking about the environmental impact.

I opened a discussion for this @ https://github.com/jazzband/pip-tools/discussions/2278

PR #2265, opened 2025-11-07
🧪 Improve Codecov integration stability

webknjaz:
The patch aims to ensure Codecov reports are uploaded on failures. This is achieved by checking a `cancelled()` state which makes sure the default `success()` isn't applied.

It also bumps the Codecov action to v5 and makes the Codecov config dot-prefixed.

Finally, an explicit notification mechanism is configured to tell Codecov about upload completions. This should to prevent Codecov from processing coverage reports too early.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/changelog.d/2265.contrib.md
+++ b/changelog.d/2265.contrib.md
@@ -1,0 +1,7 @@
+The integration with Codecov has been updated to ensure that reports
+are uploaded to the service even on failures -- by {user}`webknjaz`.
+
+GitHub Actions is now configured to also send an explicit notification
+to Codecov about the completion of previously initiated uploads.
+
+Additionally, the configuration file is now {file}`.codecov.yml`.


.codecov.yml
```diff
@@ -0,0 +1,11 @@
...
+
+codecov:
+  notify:
+    manual_trigger: true  # prevent notifications until we notify Codecov
+
+  require_ci_to_pass: false
+
+comment: false # avoid spamming reviews
+
+...
```
sirosen, line 11: Is an ellipsis a special kind of YAML node? I don't recall this one, but there are so many exotic features in YAML that I'm slow to assume that it's a mistake.

tests/test_pip_compat.py
```diff
@@ -21,4 +21,4 @@ def test_parse_requirements_preserve_editable_relative_path(tmp_path, repository
     )
 
     assert install_requirement.link.url == test_package_path
-    assert install_requirement.link.file_path == test_package_path
+    assert install_requirement.link.file_path == ""
```
sirosen: Just to get a failure temporarily, for testing?

sirosen (commented):
👍 for the core ideas. I've but used this codecov mechanism before, but I'm happy to learn.

.codecov.yml
webknjaz: It's just a marker of the end. Totally optional. I just like it there and it got copied from another project.

A multi-document YAML file might look like this:
```yaml
---

doc1: stuff

---

doc2: things

---

doc3: the rest

...
```

tests/test_pip_compat.py
webknjaz: yep

webknjaz:
> I've ~but~ *(edited) **not*** used this codecov mechanism before, but I'm happy to learn.

I've only started integrating it recently. I've added it to [ansible/ansible](https://github.com/ansible/ansible/pull/85968) and [aio-libs/yarl](https://github.com/aio-libs/yarl/pull/1577) only, I think.

Previously, I was adding the exact number of uploads to the config to prevent Codecov from reporting too early. But that's annoying to maintain as it should match the number of CI jobs which changes with the matrix.

So explicit notifications is sorta the same but not precisely. AFAIU, this delays Codecov's processing in their backend until notified. However, I think it still may sometimes report checks (statuses/comments) to GH earlier than it finished processing all the reports it'd received. The linked PRs have a bunch of references that I explored — they made me conclude it works like this.

-----

In other news, I started this because the codecov upload step gets skipped on failures and I'm currently trying to fix that — GHA seems to evaluate the conditional to `false` and I'm trying to understand why. Perhaps, #2266 will help shed some light on it as I think it's capable of analysing expressions.

webknjaz:
@sirosen I've confirmed that this works and reverted the artificial failure. This should be ready for merge. Once it's in, we should make sure #2253 is rebased so that we get more visibility into what might not be marked as covered under Python 3.8 in there.

PR #2267, opened 2025-11-07
Bump docs constraints pip-tools runtime to 3.13

webknjaz:
This is needed because Dependabot and other GH Code Security tooling reads the file header to determine the Python version to use.

Currently, it's unable to bump pip in the docs env [[1]] supposedly due to this.

[1]: https://github.com/jazzband/pip-tools/security/dependabot/22

##### Contributor checklist

##### Maintainer checklist


sirosen:
Can we also set `base_python` for the tox env which updates this file?

webknjaz:
> Can we also set `base_python` for the tox env which updates this file?

Not sure. `base_python` has proven difficult to maintain in the past. Plus it might not be available for contributors while Sphinx would be perfectly functional on other versions otherwise. I think that having the version set for the job in GHA already achieves what we want.

If anything, we could work on a warning emitted from a `toxfile.py` on failure with a hint if the current Python version differs from what's in GHA/RTD configs.

webknjaz:
Yes, I know what you meant. It's just that I tried doing this in the past and hated it eventually.

webknjaz:
FTR, this didn't help with dependabot. Perhaps it needs an explicit config..

PR #2268, opened 2025-11-09
Add conditional coverage exclusions to trivial places

webknjaz:
The patch adds version-dependent pragmas where no actual coverage can be provided.
They sometimes are applied to conditionals checking the deps versions and won't work perfectly but it's a start.

The change also excludes tests marked as `xfail` from coverage — they are either chronically uncovered up to a certain line or have flaky coverage. Either way, having them in is rather harmful for measurements.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,5 +4,7 @@ omit =
     piptools/_compat/*
 
 [report]
+exclude_also =
+    ^\s*@pytest\.mark\.xfail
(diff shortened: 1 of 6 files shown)


sirosen:
These were merged in #2347, but because it was rebased, didn't auto-close.
But this is merged! :tada:

PR #2270, opened 2025-11-12
Release version 7.5.2

sirosen:
These changes were applied via `towncrier build --version v7.5.2`.
An additional minor edit was made to wrap some text in the changelog.

##### Contributor checklist

- [x] ~A change note is created in `changelog.d/` (see [`changelog.d/README.md`](https://github.com/jazzband/pip-tools/blob/main/changelog.d/#readme) for instructions) or the PR text says "no changelog needed".~

##### Maintainer checklist

-----

##### Changelog Preview

https://pip-tools--2270.org.readthedocs.build/en/2270/changelog/#v7-5-2


sirosen:
You had mentioned it, but we didn't discuss it much and decide on anything. In my head it's filed under "try to establish a policy around what the version number of pip-tools means" -- it's a good idea but not a top priority IMO.
Maybe as part of `pip-tools[supported]` we would establish some meaning there as well?

webknjaz:
> Maybe as part of `pip-tools[supported]` we would establish some meaning there as well?

Perhaps. Although bumping that bit of core packaging metadata is probably not something that would count as adding a new API feature under SemVer. Maybe just the first addition of the extra, but not updates.

Issue #1134, opened 2020-05-04
Test the dists


atugushev:
When you run tox, it works with the source in the project dir and even creates some temporary dists. But they are not what you're going to publish to PyPI hence you test not exactly the same as users will get. [Also, you may be testing not what's installed for various reasons.](https://blog.ganssle.io/articles/2019/08/test-as-installed.html)
Better way would be to divide this process into stages:
```
  (1) build dists -> (2) test dists from [1] -> (3) upload *exactly* those dists from [1] that have been tested
```

_Originally posted by @webknjaz in https://github.com/jazzband/pip-tools/pull/1012#issuecomment-623373257_

PR #2274, opened 2025-11-15
📦 Build release artifacts in CI workflow

webknjaz:
This is a narrowly scoped initial integration of `reusable-tox.yml` into the project's CI/CD infra. It makes use of the following external action and reusable workflow repositories, some are optional:
* `codecov/codecov-action` (transitively, via `tox-dev/workflow`)
* `irongut/CodeCoverageSummary` (transitively, via `tox-dev/workflow`)
* `re-actors/cache-python-deps` « (directly; and also transitively, via `tox-dev/workflow`)
* `re-actors/checkout-python-sdist` « (transitively, via `tox-dev/workflow`)
* `test-summary/action` (transitively, via `tox-dev/workflow`)
* `tox-dev/workflow` « (directly)

The marked projects are under my control.

This patch plugs the main CI workflow as reusable into the release one. The CI workflow gets two new jobs — pre-setup that computes a number of variables, and build that makes the artifacts and stores them for later consumption by the PyPI upload job. The build job also double-checks that the artifacts are created with expected names.

The invocation of `pypa/build` moved to tox so that it's possible to run the same process locally.

In follow-up PRs, these dists will also be used in testing.

This partially solves #1134.

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/changelog.d/2274.contrib.md
+++ b/changelog.d/2274.contrib.md
@@ -1,0 +1,6 @@
+The CI/CD is now set up so that the distribution build job
+is a part of the test pipeline. That pipeline is included in
+the release workflow which sources the artifact in produces.
+The tests must now pass for the release to be published to PyPI.
+
+-- by {user}`webknjaz`


webknjaz:
Going forward requires @jezdez to add the respective reusable components to allowlist: https://github.com/jazzband/help/issues/413.

webknjaz:
@sirosen yes, the idea is to route generic contributions upstream + maybe experiment with more ideas too. Currently, it *must be* pinned to a commit hash for this reason alone. I was hoping to get you more involved at some point. I've been using it in probably like 5 projects for a while now and have some ideas of what to change already.

webknjaz:
*(the last rebase is just a conflict resolution — nothing new)*

jezdez:
This is now unblocked.

webknjaz:
Thanks!

webknjaz:
Oh no.. The pattern is wrong in the settings. I'll reopen the request.

webknjaz:
> Oh no.. The pattern is wrong in the settings.

Another fix needed: https://github.com/jazzband/help/issues/413#issuecomment-3587567752

webknjaz:
@sirosen this now reflects what I've got in other projects in terms of the dist build job. We'll have to watch out for changes during release since they aren't really testable with tag-as-a-trigger that's currently set up. Obviously, more follow-up changes to come but I wanted something relatively small and digestable first. And iterate later, in small portions.
We can gather ideas while integrating and then have mass-refactoring across many projects. Until then, I intend to keep the infra as close as possible to the rest of my projects.

.github/actions/cache-keys/action.yml
```diff
@@ -0,0 +1,48 @@
+---
+
+name: placeholder
+description: placeholder
```
sirosen, line 4: Do these appear somewhere in the output? I've never used the composite actions feature, so I'm not super familiar with it. I'd think these need updating, but maybe it doesn't matter?

.github/actions/cache-keys/action.yml
```diff
@@ -0,0 +1,48 @@
...
+      print(f'Computed file-derived hash is {files_derived_hash}.')
+
+      with Path(environ['GITHUB_OUTPUT']).open(
+              mode=FILE_APPEND_MODE,
+      ) as outputs_file:
+          print(
+              f'files-hash-key={files_derived_hash}',
+              file=outputs_file,
+          )
+    shell: python
```
sirosen, line 46: So far, this embedded script, and the other ones like it in `ci.yml`, are the thing which I find most surprising in this PR.
If I'm reading correctly, the near-equivalent bash would be this?
```shell
      hash='${{
          hashFiles(
            'tox.ini',
            'pyproject.toml',
            '.pre-commit-config.yaml',
            'pytest.ini',
            'dependencies/**',
            'dependencies/*/**',
            'setup.cfg'
          )
      }}'
      echo "computed hash is $hash"
      echo "files-hash-key=$hash" >> "${GITHUB_OUTPUT}"
```

I have no objection to using Python over bash as the shell for a step, but I was reading it with the expectation that I'd see some Python-specific feature here. It is more verbose in this case and I'm not sure I understand why you made this choice -- am I missing something important that I should note?

If the explanation is merely a matter of preference, then let's mark this resolved, that works for me. But I didn't want to overlook some essential detail here.

.github/reusables/tox-dev/workflow/reusable-tox/hooks/post-tox-run/action.yml
```diff
@@ -0,0 +1,82 @@
...
+            fromJSON(
+              fromJSON(
+                inputs.job-dependencies-context
+              ).pre-setup.outputs.release-requested
+            )
+            && 90
+            || 30
+        }}
+
+...
```
sirosen, line 82: I haven't fully worked this out yet, but I think we should look for a way to make the hook mechanism easier to follow. The "double delegation" pattern (we delegate to reusable-tox, which delegates back to this action) results in some difficult to understand logic here in order to extract values.

i.e. We need `fromJSON(inputs.job-dependencies-context).pre-setup.outputs.release-requested`, which is a pretty long expression, in order to get an input which we had in-hand in a workflow in this repo (when we did the `pre-setup` step).

I think I would prefer if it were the responsibility of the `ci.yml` workflow to specify data to pass to the hooks, and then thread that down into this job. I haven't explored if this is possible, but something like...

```yaml
# in ci.yml
 build:
    name: >-
      📦 Build dists
    needs:
    - pre-setup
    uses: tox-dev/workflow/.github/workflows/reusable-tox.yml@...
    with:
      # is it possible to make this something we pass in? (I'm guessing not, but it's an idea)
      post-tox-action: .github/workflows/reusables/reusable-tox/post-tox/action.yml
      # this is the important part: passing a blob of action inputs
      post-tox-action-call-args:
        release-requested: ${{ jobs.pre-setup.outputs.release-requested }}
        ...

# in action.yml , using that data
      retention-days: >-
        ${{
            fromJSON(
              fromJSON(inputs.call-args).release-requested
            )
            && 90
            || 30
        }}
```

It's not more _powerful_ than having full access to the `needs` job context data, but I think the explicit arg-passing style is much easier to follow. I'm not sure how much of this is possible within the constraints of GitHub Actions -- it's been a while since I tinkered with custom actions, but I'm aware that there are some limitations on the system which could make this difficult.

---

All of the above is more of a comment on `reusable-tox` than it is about the integration here. I've followed enough to understand that we're getting context data from the original `ci.yml` passed down into here and we can move forward with this. But it was difficult to follow and I think it may be hard for future contributors to read.

.github/workflows/ci.yml
```diff
@@ -44,8 +73,221 @@ env:
...
+      run: |
+        from os import environ
+        from pathlib import Path
+
+        FILE_APPEND_MODE = 'a'
+
+        with Path(environ['GITHUB_OUTPUT']).open(
+                mode=FILE_APPEND_MODE,
+        ) as outputs_file:
+            print('release-requested=true', file=outputs_file)
```
sirosen, line 149: The two scripts above seem significantly longer than their bash equivalents. More than in the case of the hashing action, I wonder if these should be bash shell steps, and then the tomllib and setuptools_scm scripts below can set `shell: python`.

.github/workflows/ci.yml
```diff
@@ -44,8 +73,221 @@ env:
...
+
+        ver = setuptools_scm.get_version(local_scheme='dirty-tag')
+        with Path(environ['GITHUB_OUTPUT']).open(
+                mode=FILE_APPEND_MODE,
+        ) as outputs_file:
+            print(f'dist-version={ver}', file=outputs_file)
+            print(
+                f'dist-version-for-filenames={ver.replace("+", "-")}',
+                file=outputs_file,
+            )
```
sirosen, line 226: Seeing this snippet makes me think of a tool I've been working on which might interest you:
https://mddj.readthedocs.io/en/latest/

The advantage over calling into `setuptools_scm` directly is that it's generic, so if you wanted a version of this script which you can port across projects, it might look like...
```python
# /// script
# dependencies = ["mddj==0.4.2"]
# ///
from mddj.api import DJ
dj = DJ()
ver = dj.read.version()
... # rest of script follows
```

And that will read `project.version` if it's static, or call into `build` if it's dynamic.

The project still needs a lot of work; it's just an alpha right now. But I find I frequently want to be able to script around project metadata, so it's an interesting space to try to make a good utility.

.github/workflows/ci.yml
```diff
@@ -44,8 +73,221 @@ env:
...
+        ${{
+          fromJSON(needs.pre-setup.outputs.release-requested)
+          && 1
+          || 0
+        }}
+      job-dependencies-context: >-  # context for hooks
+        ${{ toJSON(needs) }}
+      python-version: 3.12
+      runner-vm-os: ubuntu-latest
+      timeout-minutes: 2
```
sirosen, line 287: 2 minutes seems a bit tight, if there's any slowdown (e.g., network flakiness). Should we maybe loosen it a little, e.g. to 5 minutes?

tox.ini
```diff
@@ -191,3 +201,64 @@ commands_pre =
...
+  twine
+commands =
+  {envpython} \
+    {[python-cli-options]byte-errors} \
+    {[python-cli-options]max-isolation} \
+    {[python-cli-options]warnings-to-errors} \
+    -m twine \
+      check \
+      --strict \
+      dist{/}*
```
sirosen, line 262: This reminds me, I'd like to incorporate `check-sdist` as well. We can do that in a follow-up.

tox.ini
```diff
@@ -8,6 +8,16 @@ envlist =
     readme
 skip_missing_interpreters = True
 
+[python-cli-options]
+byte-warnings = -b
+byte-errors = -bb
+max-isolation = -E -s -I
```
sirosen, line 14: Current docs indicate that `-I` implies `-E` and `-s`. So I think we can cut this down to just `-I`?

tox.ini
```diff
@@ -8,6 +8,16 @@ envlist =
...
+# some-isolation = -I
+# FIXME: Python 2 shim. Is this equivalent to the above?
+some-isolation = -E -s
+warnings-to-errors = -Werror
```
sirosen, line 18: I have somewhat mixed feelings about this block of config. On the one hand, I can see how it documents the short flags we use. And I didn't know you could have a custom section in `tox.ini` loaded like this, which is neat!

But I also find that it makes the usage sites _harder_ to read in some sense, since you have to lookup this table to figure out what the command is. (adding a bit of indirection)
The comparison is between
```
{envpython} \
    -I -bb -Werror
```
and
```
{envpython} \
    {[python-cli-options]byte-errors} \
    {[python-cli-options]max-isolation} \
    {[python-cli-options]warnings-to-errors}
```

I'm okay with it, at least to try out, but I'm just noting that I'm not fully convinced as of yet.

sirosen (commented):
I've reviewed the various pieces here, but I need a bit more time with it before I can approve. I might have another comment or two, but I'm most of the way there on this! (I know it's taken a while.)

.github/actions/cache-keys/action.yml
webknjaz: No, these are used when publishing standalone actions (in the repo root) to Marketplace. It's not limited actions being composite.

I think the missing fields were making either jsonschema or actionlint unhappy so I added dummy strings. They aren't actually needed for the action to function.

.github/actions/cache-keys/action.yml
webknjaz: You're right. There's nothing special about this. I think that originally, I was trying to unify bits of caching, even before I distilled some into https://github.com/marketplace/actions/cache-python-deps, I was trying to separate components and reduce job sizes (it seemed necessary at the time). So I came up with a part that computes the file-based portion of the cache key and the Python runtime-derived one. Eventually, it turned out that this only needs to be called once but it was already separated and I was synchronizing changes across projects so I left it as is.

The Python syntax side of this likely came from me copying some scaffolding from other steps where I was already using Python. By the way, that second cache key portion does access version data through Python imports. I don't remember but maybe they were combined at some point prior to refactoring.

Anyway, I was thinking that it'd be good to replace GHA's `hashFiles()` w/ pure Python (and filter out the non-existing file path wildcards dynamically), turning this into an external action later — maybe we'd just have a list of all most used file patterns and it would be called w/o inputs even.

.github/reusables/tox-dev/workflow/reusable-tox/hooks/post-tox-run/action.yml
webknjaz: Yeah, we can work on this in the reusable workflow. Some things (like passing an arbitrary hook path) are indeed impossible within the native constraints of GHA. They could probably be hacked through manual repo cloning but that's a can of worms.

I wasn't sure about passing the hook args but maybe we can think of something later. Your example of a nested input might not work natively and you'd have to turn it into JSON (or a YAML string, via `|`, perhaps).

.github/workflows/ci.yml
webknjaz: Yes, bash is often shorter.

This job is about computing a lot of things so I wanted to make use of Python more w/o being caught in potential Bash traps. Plus it's in sync with other repos, which would mean having to do this in other places at the same time.

It may be a good idea to look into making it more reusable too at some point.

.github/workflows/ci.yml
webknjaz: Yeah, this looks fun. The thing is that this snippet is supposed to predict what version is to be expected from Git so that it can be used for sanity checks in other jobs that actually grab the metadata. Perhaps, it's fine to move to reading the metadata but up until recently, `setuptools-scm` didn't have an easy way to let us influence its settings from outside the config. In some projects, I was modifying `pyproject.toml` on disk, in multiple jobs and would `--assume-unchanged` it in Git. There's 3 version modes this (in combination with other steps) was supposed to achieve:
* release (an exact matching tag on the same commit; sometimes emulated — in case of `workflow_dispatch` the tag is created locally but then actually pushed to GitHub way later)
* merge to the main branch (a version that has the `.dev` counter but no local portion)
* PR run (both `.dev` counter and the local version portion)

The first two are publishable to Test(PyPI). But since I wasn't copying all the bits from other repos just yet, this may not be as obvious.

P.S. It's always scary to modify these steps because I often mess up there with the context spread across multiple lines. So I'm usually trying to avoid that until I can turn it into something reusable.

.github/workflows/ci.yml
webknjaz: For this one, this shouldn't be necessary. It takes 30s and having 4x more time is already more than enough. 1-2 extra mins for network flakiness has shown good results in the past. But I'm usually playing with longer timeouts for tests and linters, though. For example, things like sphinx's linkcheck.

tox.ini
webknjaz: I started implementation of this in Cheroot originally, when I was trying to handle Python 2.7 too. This needs to be cleaned up across multiple projects but I never took time to compare implications so I'm copying this around.

Can we do the cleanup in a follow-up?

tox.ini
webknjaz: Yeah, I just wanted the same things reused across the config. Also, when running `tox run -v`, you'll see the rendered commands so this should be enough, despite indirection. Additionally, there's `tox config` that shows all the settings in the resolved state.

When copying these flags, I find it helpful to see visually that I haven't forgotten about them and it's a part of the idiom for invoking Python, with all the indents and EOL escaping in the config.

I think that it may be a bit confusing when seen first but later, it feels rather natural to see all the invocations across the config "protected". I don't remember a time when I needed to look up which flags actually get rendered, unless something wasn't working, at which time I'd have to go into a deep inspection mode anyway...

tox.ini
webknjaz: Sounds good. I haven't used that one so would need a demo PR or smth.

webknjaz:
@sirosen do you think we can get this in soon so it doesn't accumulate conflicts?

sirosen (approved):
Yeah, let me go ahead and approve! I have the level of understanding that I need on this to feel good about merging. I got behind on various projects and this review ended up lagging on my side.

---

A small conflict has crept in. And I've seen some GitHub Actions outages today with jobs failing to start, so just a warning that if you rebase today, CI might misfire.

webknjaz:
Yeah, I assumed those were only in nightlies. We'll see.

webknjaz:
Yep.. GitHub's drunk today: https://github.com/jazzband/pip-tools/actions/runs/21608455077. ~Will restart sometime later.~ Ah, that's from another PR.

.github/actions/cache-keys/action.yml
wizardfriend2018: It's ok thanks

PR #2276, opened 2025-11-17
Normalize the option `--unsafe-package`

### LINKED ISSUE #2150: `--unsafe-package` doesn't normalize hyphens/underscores
Pip resolves `package-name` and `package_name` to the same dependency. However, the unsafe package specifier doesn't taket his into account on the latest released version.

Example: `pip-compile --unsafe-package opencv_python --no-allow-unsafe` still installs `opencv-python` in the final output.

shifqu:
Normalize the option `--unsafe-package`.

Resolves #2150 
##### Contributor checklist

- [X] Included tests for the changes.
- [X] A change note is created in `changelog.d/` (see [`changelog.d/README.md`](https://github.com/jazzband/pip-tools/blob/main/changelog.d/#readme) for instructions) or the PR text says "no changelog needed".

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -467,9 +467,7 @@ def cli(
             for find_link in dedup(repository.finder.find_links):
                 log.debug(redact_auth_from_url(find_link))
 
-    unsafe_package = tuple(
-        str(canonicalize_name(pkg_name)) for pkg_name in unsafe_package
-    )
+    unsafe_package = tuple(canonicalize_name(pkg_name) for pkg_name in unsafe_package)
 
     resolver_cls = LegacyResolver if resolver_name == "legacy" else BacktrackingResolver
     try:


shifqu:
Not sure why linkcheck-docs was cancelled. :/

EDIT: Re-running the step made it pass..

changelog.d/2150.bugfix.md
```diff
@@ -0,0 +1,2 @@
+The option --unsafe-package is now normalized
```
webknjaz: RST would render this as an em-dash unless it's double-backticked:

```suggestion
The option ``--unsafe-package`` is now normalized
```

changelog.d/2150.bugfix.md
```diff
@@ -0,0 +1,2 @@
+The option ``--unsafe-package`` is now normalized
+-- by :user:`shifqu`.
```
webknjaz: I just realized that this was supposed to be Markdown (got confused by parts of RST when looking from mobile). So this should actually be:
```suggestion
The option `--unsafe-package` is now normalized
-- by {user}`shifqu`.
```

piptools/scripts/compile.py
```diff
@@ -13,6 +13,7 @@
 from click.utils import LazyFile, safecall
 from pip._internal.req import InstallRequirement
 from pip._internal.utils.misc import redact_auth_from_url
+from pip._vendor.packaging.utils import canonicalize_name
```
webknjaz, line 16: Should we start depending on `packaging` instead? And use the original https://github.com/pypa/packaging/blob/f451a9fc5172584cc51f70f5f8da456cf7ebd833/src/packaging/utils.py#L46-L51.

cc @sirosen

piptools/scripts/compile.py
```diff
@@ -466,6 +467,10 @@ def cli(
             for find_link in dedup(repository.finder.find_links):
                 log.debug(redact_auth_from_url(find_link))
 
+    unsafe_package = tuple(
+        str(canonicalize_name(pkg_name)) for pkg_name in unsafe_package
```
webknjaz: Is the cast needed here? The type seems compatible: https://github.com/pypa/packaging/blob/f451a9fc5172584cc51f70f5f8da456cf7ebd833/src/packaging/utils.py#L15C44-L15C47

piptools/scripts/compile.py
shifqu: Let me know if you'd like me to add packaging as a dep -- I initially chose to use the vendored version because this is how our `utils.py` currently uses it (https://github.com/jazzband/pip-tools/blob/a2d936373f67eed3d39f69441f3ebb12c33f4803/piptools/utils.py#L75-L89).

piptools/scripts/compile.py
shifqu: It might indeed not be needed. This is mostly because in our utils, when we call this function, we also cast a str, so to remain consistent with the rest of the code, I opted to add this cast.

I could remove it if you'd like

EDIT: I removed it as indeed the cast is not needed. In the codebase, we often times cast (either by using `typing.cast` or effectively cast by using `str`) the result of `canonicalize_name`. Should these casts be updated/removed or not? If not, should a separate issue/pr be created to do this?

changelog.d/2150.bugfix.md
shifqu: To avoid useless `update X` commit, I will just apply this suggestion to the last commit, it's already co-authored, so we lose nothing with it.

piptools/scripts/compile.py
shifqu: I do think changing to use the `packaging` package as a dependency is out of scope for this PR. I'd be happy to open a new PR for it, but a quick search on `from pip._vendor.packaging` in the codebase shows that there's 21 results over 12 files.. Which would make the diff of this PR quite big and unrelated to the normalization of the option.

sirosen: I agree that it's out of scope for this PR.

My initial take is that we should not pull in `packaging` directly, and we should continue to use the `pip` vendored copy. I am not clear on what the advantage of depending on it directly, and I can imagine some obscure scenario in the future in which being out of sync with `pip` could lead to bugs.

If there's some specific risk (e.g., `pip` moving their vendored copy, which they could do freely at any time) which we want to protect against, I'd be happy to discuss that further in a separate issue.

piptools/scripts/compile.py
sirosen: Ooh! I think this is my first time seeing `NewType` used in the wild! :grin: 

If you want to remove the casts in a separate PR, I'm happy to review it. They should all be no-ops.
I suspect that they were introduced at some point by someone whose toolchain flagged that `canonicalize_name` returns a `NormalizedName`, and they misinterpreted that as meaning that it _wasn't_ a `str`.

If you touch `cast` usages too, and find that mypy fails, I'll note that I have started to prefer careful use of `type: ignore[<code>]` comments over `typing.cast`. (Because there's no runtime cost, and the two are equivalent as long as the type ignore is appropriately narrow.)

shifqu: > and they misinterpreted that as meaning that it *wasn't* a str.

Exactly what happend here, I didn't check the `NormalizedName` implementation and *assumed* it was a complex type. Good thing we have @webknjaz reviewing thoroughly :D

> I'll note that I have started to prefer careful use of `type: ignore[<code>]` comments over typing.cast

Agreed! For me the conditions are that:
- you should include `[<code>]` (even though `pyright` does not care about them, in projects where I use `pyright`, I still add the code)
- They should be used sparingly, first try to resolve the error, rather than defaulting to ignoring it.

piptools/scripts/compile.py
webknjaz: I think we have a few other deps that are vendored in pip but we depend on directly. One reason could be — what pip stops vendoring a thing? That would cause a scenario where some versions would have it while other wouldn't.

I guess it's fine to say that it's out of the scope, though.

piptools/scripts/compile.py
webknjaz: Yep. Agreed. I also like to configure MyPy to error out if `[<code>]` is missing and it's a blanket ignore.

tests/test_cli_compile.py
```diff
@@ -1709,6 +1709,55 @@ def test_unsafe_package_option(pip_conf, monkeypatch, runner, unsafe_package, ex
     assert out.stdout == expected
 
 
+@pytest.mark.parametrize(
+    "unsafe_package",
+    (
+        pytest.param("small-fake-with-deps", id="kebab-case"),
+        pytest.param("small_fake_with_deps", id="snake_case"),
+        pytest.param("Small_Fake_With_Deps", id="mixed-case"),
+    ),
```
webknjaz, line 1718: For lists of 3 items or less, it's usually fine to go for `ids=`. It's a bit more compact than explicit `pytest.param`. But for longer lists, this is indeed the way.

webknjaz (approved):

PR #2279, opened 2025-11-23
Remove support for Python 3.8

sirosen:
I wanted to wait for "one last release" before dropping support, but I think we're now at a point where the cost/benefit tradeoff favors removing 3.8.

---

- Remove it from tox and CI configs
- Update code to py3.9+ paradigms, aided by pyupgrade
- Update `project.requires-python` and classifiers
- Replace the `is_path_relative_to` utility (per its comment about 3.8)
- Remove the 3.8 coverage workaround in CI

##### Contributor checklist

- [x] ~Included tests for the changes.~
- [x] A change note is created in `changelog.d/` (see [`changelog.d/README.md`](https://github.com/jazzband/pip-tools/blob/main/changelog.d/#readme) for instructions) or the PR text says "no changelog needed".

##### Maintainer checklist

- [x] If no changelog is needed, apply the `bot:chronographer:skip` label.
- [x] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #2280, opened 2025-11-23
Remove setup.cfg from cache key hashing

sirosen:
EDIT: the true cause is https://github.com/actions/runner-images/issues/13341 , but the listing of `setup.cfg` as a file to hash is still a minor fix I think we should make.

In CI builds for #2279, we have failures ~due to this nonexistent file being hashed.~ which I believed were caused by nonexistent `setup.cfg` file hashing.
The `hashFiles` call has been the same for >4 years without issue, even after the file was deleted, but has just started failing.
Although the exact reason is unclear, we don't need to try to hash on `setup.cfg` at this stage anyway.

Also, quote `GITHUB_*` variables to prevent actionlint from failing with errors about unquoted variables & potential word-splitting.
This was also noted in #2279, as in both PRs I've seen local issues with actionlint which are not failing in CI.
Specifically, local runs of actionlint flag shellcheck SC2086: "Double quote to prevent globbing and word splitting".
Although it's safe in this case (GitHub does not produce paths with spaces for these variables), it's simplest to quote the variables to pacify shellcheck.

Because this is a purely internal fix which isn't meant to impact the contributor experience (other than fixing things), I've elected to go without a changelog on this one.

##### Contributor checklist

- ~[x] Included tests for the changes.~

##### Maintainer checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,13 +103,13 @@ jobs:
         id: pip-cache
         shell: bash
         run: |
-          echo "dir=$(pip cache dir)" >> $GITHUB_OUTPUT
+          echo "dir=$(pip cache dir)" >> "$GITHUB_OUTPUT"
       - name: Pip cache
         uses: actions/cache@v4
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
           key: >-
-            ${{ runner.os }}-pip-${{ hashFiles('setup.cfg') }}-${{
+            ${{ runner.os }}-pip-${{
             hashFiles('pyproject.toml') }}-${{ hashFiles('tox.ini') }}-${{
             hashFiles('.pre-commit-config.yaml') }}
           restore-keys: |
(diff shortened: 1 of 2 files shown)


sirosen:
Hm. Perhaps GitHub is just having issues right now? Builds in this PR fail with
> Error: The template is not valid. .github/workflows/ci.yml (Line: 111, Col: 16): hashFiles('pyproject.toml') failed. Fail to hash files under directory '/Users/runner/work/pip-tools/pip-tools'

I had thought that perhaps the `setup.cfg` hashing was failing in a new and interesting way. Maybe it's just broken...

webknjaz (commented):
Could you move quoting into a separate PR in the form of `"${GITHUB_OUTPUT}"` — I like that curly braces make it more obvious where the env var ends in some cases.

webknjaz:
As for caching, I usually keep about the same file list for portability across repos. GHA normally just ignores that they don't exist.
Although, in some places I'd just hash things w/ python. In fact, I have an action for hashi the Python version portion this way.

sirosen:
Hm, maybe I should just close this?

- I'll do a dedicated PR for the actionlint quoting
- my guess was wrong, this was a GitHub bug (and they're doing the rollback right now) 

I don't really like having `setup.cfg` in there since it's not a match for this project, but I understand it has some practical benefits for you maintaining parity with other repos. And I don't feel strongly.

webknjaz:
Yeah, I think let's close it and perhaps revisit on a higher level once reusable-tox is in. We could see if this could be made more robust (I'm pretty sure we could rewrite that in Python, for example).

PR #2282, opened 2025-11-24
Update actionlint integration to include shellcheck, and fix shellcheck failures

### LINKED ISSUE #2281: Update pre-commit actionlint hook to always install shellcheck, ensuring consistent behavior
`actionlint` has a standing issue that the pre-commit configuration won't install `shellcheck`, but the tool will be picked up from your system if it is installed: https://github.com/rhysd/actionlint/issues/477
This leads to discrepancies between local pre-commit runs, where, for example, I have `shellcheck` installed in `~/bin/`, and pre-commit.ci, where `shellcheck` is missing.

There's a PR up which would (incorrectly) pull in shellcheck's latest version into the actionlint hook config, but [an excellent comment from Alex Waygood](https://github.com/rhysd/actionlint/pull/482#issuecomment-2545874061) explains why it's not quite right and links to Astral's [config in the `ruff` repo](https://github.com/astral-sh/ruff/blob/9eaca50f43d3fe1aa1048e7d2d60d08a68c14fb8/.pre-commit-config.yaml#L104-L120).

I plan to both update the variable quoting issue I noted in our workflows (which makes shellcheck fail, even though the quoting is fine in practice) and update the pre-commit config based on the `ruff` one.

sirosen:
closes #2281
This is also noted in #2280 and #2279 .

My local actionlint runs were flagging things which were missing in CI. The cause of the discrepancy: a local shellcheck install, and missing shellcheck integration in CI.
The first change here adds protective quoting so that my local runs pass, and the second updates our actionlint hook config to always include shellcheck.

Because the config is copied from https://github.com/astral-sh/ruff/commit/ad0018649780d8cb72bcaaed7a6c690baf70e4c8, and I've even taken part of the comment from there, I created it as a co-authored commit.
( :wave: @AlexWaygood, thanks for the config! )

##### Contributor checklist

- ~[x] Included tests for the changes.~

##### Maintainer checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -106,5 +106,10 @@ repos:
     rev: v1.7.8
     hooks:
       - id: actionlint
+        additional_dependencies:
+          # actionlint has a shellcheck integration which extracts shell scripts in `run:` steps from GitHub Actions
+          # and checks these with shellcheck.
+          # The integration only works if shellcheck is installed.
+          - "github.com/wasilibs/go-shellcheck/cmd/shellcheck@v0.11.1"
 
 ...
--- a/changelog.d/2281.contrib.md
+++ b/changelog.d/2281.contrib.md
@@ -1,0 +1,1 @@
+Fix `actionlint` hook usage to always include `shellcheck` integration -- by {user}`sirosen`.


.pre-commit-config.yaml
```diff
@@ -106,5 +106,10 @@ repos:
     rev: v1.7.8
     hooks:
       - id: actionlint
+        additional_dependencies:
+          # actionlint has a shellcheck integration which extracts shell scripts in `run:` steps from GitHub Actions
+          # and checks these with shellcheck.
+          # The integration only works if shellcheck is installed.
+          - "github.com/wasilibs/go-shellcheck/cmd/shellcheck@v0.10.0"
```
AlexWaygood: > Because the config is copied from https://github.com/astral-sh/ruff/commit/ad0018649780d8cb72bcaaed7a6c690baf70e4c8, and I've even taken part of the comment from there, I created it as a co-authored commit.
> ( 👋 @AlexWaygood, thanks for the config! )

Very kind, thank you!

In some of my other projects I've bumped the shellcheck pin to the latest version; I should probably go do that with Ruff as well 😅

```suggestion
          - "github.com/wasilibs/go-shellcheck/cmd/shellcheck@v0.11.1"
```

sirosen: Thanks for this; good catch! I'll apply and then squash together.

I also maintain a little pre-commit additional_dependencies update tool, and [filed an issue](https://github.com/sirosen/upadup/issues/10) for figuring out how to script this update, since it seems like all of the information to do it should be available.

sirosen:
Yeah, I think it's sort of a shame but understandable that we need `shellcheck` bundled up for different consumers.

My understanding of the golang wrapper is that a WASI build of `shellcheck` itself is included in that repo, and the Go wrapper (which pre-commit consumers have to compile on the fly) is just a frontend to invoke it.

Issue #2285, opened 2025-11-25
Reduce circular dependencies between `utils.py` and `_compat/` by migrating pip-related utilities into a new `_pip_api`


sirosen:
I've been bitten a couple of times by the circular relationship between "utils" and "_compat", and we have some old PRs which got snagged on it as well. (I was trying to port the changes in #2108 forward.) The current module structure is not right in terms of defining dependencies between components -- if you try to migrate any component from "utils" into "compat", you create cycles.
"utils" is also a generally disliked component name, since it doesn't contain important information about its contents.

Tinkering a bit in the repo today, and trying to decide on a path forward, I started to focus on the fact that `utils` and `_compat.pip_compat` are both concerned with providing an API for `pip` usages, like the `PIP_VERSION` constant.

I propose that we pursue a gradual migration:
1. We introduce `piptools._pip_api`, which is _our API_ for usage of `pip`. It is, of course, internal-only.
2. `piptools._pip_api` is defined to not have any dependencies on any other parts of `piptools` for now. i.e. We aren't going to allow the reintroduction of this kind of circular problem -- we could add smaller modules which `_pip_api` would be able to consume in the future, but the one we have to strictly rule out is `utils`.
3. As much of `utils` as is reasonable is ported into `_pip_api`, and consumption is updated.
4. Testing is also dir structured, into `tests/pip_api/test_{module_name}.py` (we have too many tests in a flat dir at present, IMO. I want some better organization to the testsuite so that I can find things).
5. Once `piptools._pip_api` is defined, `piptools._compat.pip_compat` is migrated into it.
6. Celebrate! Then reassess and look for more ways to reduce circularity and improve module structure.

@webknjaz, looking for your feedback on this in particular. I have some initial work I'll turn into a draft PR to demo the idea.

webknjaz:
> 1\. We introduce `piptools._pip_api`, which is _our API_ for usage of `pip`. It is, of course, internal-only.

I think I'd like a more agressive approach and would put it into `piptools._private._pip_api` or `piptools._internal._pip_api`.

> 2\. We aren't going to allow the reintroduction of this kind of circular problem

For this, I'd like to have a port of https://github.com/aio-libs/aiohttp/blob/91108c9/tests/test_circular_imports.py in our test suite.

Additionally, may I ask you to play with https://import-linter.rtfd.io/en/stable/readme.html? I've been meaning to start using it but never got to actually playiing with it..



[...]

PR #2286, opened 2025-11-25
Create initial internal `pip_api` package

sirosen:
For motivation and context, see: #2285

These are just a couple of first steps:

- Introduce `piptools._internal._pip_api` w/ version utilities
- Begin handling `InstallRequirement` in _pip_api

##### Contributor checklist

##### Maintainer checklist


piptools/_internal/_pip_api/__init__.py
webknjaz, line 1: Instead of re-exporting names, I'd have private modules inside too (this is something we've started adopting in ansible/ansible recently). Plus in the issue I suggested this to be a bit more nested. With that, I'd like the imports to be of modules. So it'd be `from ....._private import _helper_module` with following use as `_helper_module` on the calling side. This would help reduce leaking re-imported things as public names.

webknjaz:
@sirosen can we start with having a circular import test I mentioned in https://github.com/jazzband/pip-tools/issues/2285#issuecomment-3578142117? It'd be a good idea to merge it right away.

piptools/_internal/_pip_api/__init__.py
sirosen: I'm going to get started on this. To confirm that I understand, I expect usage to look like:
```python
from piptools._internal.pip_api import _pip_version

if _pip_version.PIP_VERSION[:2] < (25, 3): ...
```

Users of the internal API only import the modules, not their members. If I'm not quite getting it right, please show me an example. :sweat_smile:

webknjaz: More like if there's `piptools/_internal/_pip_api.py`, then imports would be
```python
from piptools._internal import _pip_api

if _pip_api.PIP_VERSION[:2] < (25, 3): ...
```

So we'd import modules and access constants within through attributes.

---

P.S. I'd also prefer having relative imports instead of absolute:

```python
from .._internal import _pip_api

if _pip_api.PIP_VERSION[:2] < (25, 3): ...
```

---


re: `_pip_api.PIP_VERSION[:2] < (25, 3)` — we might want to have a bunch of pre-computed booleans in some compat module since many of those version-checks would happen in multiple places an it'd be good for consistency. Like `from .._internal import _pip_runtime_compat; if _pip_runtime_compat.PIP_VERSION_BELOW_25_3: ...` or `if _pip_runtime_compat.PIP_SUPPORTS_FEATURE_X`. Maybe, we'd be able to tie such constants into the conditional coverage plugin.

sirosen: Okay, that makes for a nicer interface than what I was working on. I'm going to adjust.
And I can try to move towards relative imports everywhere that I touch.

Regarding the version constant, I was thinking something similar. I have `PIP_VERSION_MAJOR_MINOR = PIP_VERSION[:2]` right now and I'm happy for us to expand the suite of things over time. For the first draft I'm going to try to keep it small and we can add more constants in the future.

webknjaz: Agreed. Another thought: when looking into the coverage plugin, I wonder if we could cover cetain code branches by monkey-patching these contstants instead of having another job in the test suite with an additional pip version (on top of min+max pip) — how would the plugin cover such cases then?

changelog.d/2285.contrib.md
```diff
@@ -0,0 +1 @@
+Utilities for interacting with `pip` have started to move into an internal `_pip_api` subpackage -- by {user}`sirosen`.
```
webknjaz: I suppose its
```suggestion
Utilities for interacting with `pip` have started to move into the :py:mod:`piptools._internal._pip_api` subpackage -- by {user}`sirosen`.
```
now?

piptools/_internal/__init__.py
webknjaz, line 1: I prefer avoiding `__init__.py` and even lint against having it in some places to prevent it from accidentally becoming an entry point with logic that would run during import time when reaching for neighboring submodules: https://github.com/ansible/awx-plugins/blob/274b32a7ab9eb09493b05e28eb5cad7441ec08bf/.pre-commit-config.yaml#L7-L33

If an entrypoint is needed, I add a module called `api.py` or similar for external consumers.

tests/test_circular_imports.py
```diff
@@ -0,0 +1,84 @@
...
+            for _, pkg_path, _ in pkgutil.walk_packages(
+                (str(pkg_dir_path),),
+                prefix=f"{pkg_pref}.",
+            )
+        )
+
+
+# This test is set to skip on lower pip versions, and only run on the latest.
+# Doing so avoids deprecations and other warnings caused by older pip code or vendored
+# libraries within pip.
```
webknjaz: Or.. we could add another `-Wdefault::DeprecationWarning` under older pips instead. I can imagine that older versions might trigger different import cascades, potentially causing loops but only under those envs. WDYT?

tests/test_circular_imports.py
webknjaz, line 1: @sirosen does this module succeed on `main` if added separately? If yes, we could merge it earlier than this PR.

tests/unit/__init__.py
webknjaz, line 1: I rather dislike having inits in tests (elsewhere too, actually). Can we agree on not having them?

tests/unit/_internal/pip_api/test_install_requirements.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import annotations
+
+import pytest
+
+from piptools._internal import _pip_api
+
+
+@pytest.mark.skipif(
+    _pip_api.PIP_VERSION_MAJOR_MINOR < (25, 3), reason="test requires pip>=25.3"
+)
```
webknjaz, line 10: Marks used multiple times can be defined once and reused as follows:
```suggestion
requires_pip_25_3 = pytest.mark.skipif(
    _pip_api.PIP_VERSION_MAJOR_MINOR < (25, 3),
    reason="test requires pip>=25.3",
)


@requires_pip_25_3
```

piptools/repositories/local.py
```diff
@@ -12,7 +12,8 @@
 from pip._internal.req import InstallRequirement
 from pip._internal.utils.hashes import FAVORITE_HASH
 
-from piptools.utils import as_tuple, key_from_ireq, make_install_requirement
+from piptools._internal import _pip_api
+from piptools.utils import as_tuple, key_from_ireq
```
webknjaz:
```suggestion
from .._internal import _pip_api
from ..utils import as_tuple, key_from_ireq
```

piptools/_internal/_pip_api/pip_version.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import annotations
+
+import subprocess  # nosec
```
webknjaz: Try to avoid importing+using `subprocess` directly in favor of `piptools.subprocess_utils`.
There's https://github.com/jazzband/pip-tools/blob/main/piptools/subprocess_utils.py. The idea behind it is that we allow `# nosec` comments in there but make sure that every change and suppression is double-checked. Then, the linters are supposed to disallow these imports elsewhere (we should probably an additional linter with forbidden imports — I know `pylint` has a builtin check for this, for example).
This creates a centralized space for potentially insecure invocation with additional handling/precautions so that the rest of the code base wouldn't have to care about that. And we wouldn't be letting new insecure uses squeeze in undedected due to blanket suppressions being present. Hence only allowing these suppressions in that module.

webknjaz:
```suggestion
from ... import subprocess_utils as _subprocess_utils
```

piptools/_internal/_pip_api/pip_version.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def get_pip_version_for_python_executable(python_executable: str) -> Version:
+    """Return pip version for the given python executable."""
+
+    str_version = subprocess.check_output(  # nosec
+        [python_executable, "-c", "import pip; print(pip.__version__)"],
+        shell=False,
+        text=True,
+    )
```
webknjaz:
```suggestion
    str_version = _subprocess_utils.run_python_snippet(
        python_executable,
        "import pip; print(pip.__version__)",
    )
```

piptools/_internal/_pip_api/install_requirements.py
```diff
@@ -0,0 +1,93 @@
...
+
+import copy
+from typing import Any
+
+from pip._internal.req import InstallRequirement
+from pip._internal.req.constructors import install_req_from_line
+from pip._vendor.packaging.utils import canonicalize_name
+from pip._vendor.packaging.version import Version
+
+from .pip_version import PIP_VERSION_MAJOR_MINOR
```
webknjaz: We don't want to export `PIP_VERSION_MAJOR_MINOR` under this module.
```suggestion
from . import _pip_version
```

Alternatively, add something like `__all__ = ()`.

tests/test_circular_imports.py
sirosen: I'm pretty sure it will work on `main`. Let me pull it out into a separate PR and we can iterate on it further.

Per your comment above, I think I _do_ want to allow it to run on all `pip` versions. I thought about it, and it makes it a better test. So I'll need to do some tinkering with the warning filters.

webknjaz: Yeah.. That list can get quite long: https://github.com/sanitizers/octomachinery/blob/09b5e5ca2f7eb1f7d7b6f974392456a25411a11a/tests/circular_imports_test.py#L72-L133.

piptools/_internal/__init__.py
sirosen: I can adjust to this, yes! I still have the old habits, from before implicit namespace packages. 😁

piptools/_internal/_pip_api/pip_version.py
sirosen: Okay, I can go back to depending on it!

My reason for duplicating the logic was that I wanted `_pip_api` to be independent from the rest of `pip-tools`, as an easy way to guarantee that we don't get back into the situation we're coming from with `utils.py` and `_compat`.

I think that the "layers" lint in importlinter is going to help us here in the future. I played around with it in a small project and that part seemed like a good fit for this sort of case.

I may want to move `subprocess_utils` to `piptools._internal._subprocess` if that seems okay.

tests/unit/_internal/pip_api/test_install_requirements.py
sirosen: I was going to apply this, but then I noticed that one test is using `< (25, 3)`, and the other is using `>= (25, 3)`.

piptools/_internal/_pip_api/pip_version.py
webknjaz: Sounds reasonable.

tests/unit/_internal/pip_api/test_install_requirements.py
webknjaz: Ah, I missed that. Marking as resolved.

sirosen:
I just rebased this to clean it up and think it's ready for (final? :crossed_fingers: ) review. I gave it a final once over and it's all adjusted per feedback.

The one wrinkle is that I found that I could not remove one of the `__init__.py` files because it caused import errors. However, given that I've just now failed to reproduce those errors locally, I'm investigating a bit more.

sirosen:
After some quick tests, I've restored that `__init__.py`. It seems to be causing pytest collection to fail and creates an orphaned page when module docs are build (which then causes a warning). I'm open to refining this later, but I don't want to add more technical challenges to this PR.

PR #2292, opened 2025-11-29
Introduce a version of the circular import test

sirosen:
This test ensures that we don't have circular imports allowed in the project.
(ported from #2286 )

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/tests/test_circular_imports.py
+++ b/tests/test_circular_imports.py
@@ -49,7 +49,7 @@ def _discover_path_importables(pkg_pth: Path, pkg_name: str) -> Iterator[str]:
         if pkg_dir_path.parts[-1] == "__pycache__":
             continue
 
-        if all(Path(_).suffix != ".py" for _ in file_names):
+        if all(Path(_).suffix != ".py" for _ in file_names):  # pragma: no cover
             continue
 
         rel_pt = pkg_dir_path.relative_to(pkg_pth)


webknjaz (approved):

tests/test_circular_imports.py
```diff
@@ -0,0 +1,110 @@
...
+
+def _discover_path_importables(pkg_pth: Path, pkg_name: str) -> Iterator[str]:
+    """Yield all importables under a given path and package."""
+    for dir_path, _d, file_names in os.walk(pkg_pth):
+        pkg_dir_path = Path(dir_path)
+
+        if pkg_dir_path.parts[-1] == "__pycache__":
+            continue
+
+        if all(Path(_).suffix != ".py" for _ in file_names):
```
webknjaz: This is normally not hit, IIRC.
```suggestion
        if all(Path(_).suffix != ".py" for _ in file_names):  # pragma: no cover
```

webknjaz:
@sirosen we should probably manually watch the patch coverage until we can enforce it for real, right?

sirosen:
> @sirosen we should probably manually watch the patch coverage until we can enforce it for real, right?

Yeah, I've been wondering about that. I got sidetracked when trying to put together a coverage plugin config which would work for us. In the meantime, it would be good to avoid these spurious failures.

tests/test_circular_imports.py
```diff
@@ -0,0 +1,110 @@
...
+            for _, pkg_path, _ in pkgutil.walk_packages(
+                (str(pkg_dir_path),),
+                prefix=f"{pkg_pref}.",
+            )
+        )
+
+
+def _allowed_deprecation_warning_filters() -> list[str]:
+    """
+    Yield filters which allow for deprecation warnings based on the current
```
webknjaz, line 68: (this would've actually been more elegant as a generator FWIW)

sirosen: Oh, haha, clearly from the doctoring I intended to write it as a generator! It's fine, but I might come back and fix it sometime...

PR #2300, opened 2025-12-26
ci: remove redundant v prefix from release job name

### LINKED ISSUE #2271: Update GitHub Workflows to remove redundant 'v' from release jobs
This is a good, low-urgency item for anyone who wants to get involved with the project.

We transitioned from tags named `X.Y.Z` to `vX.Y.Z`, but the CI job for publication was setup to include a prefix `v`. It's a purely cosmetic defect in the build.
Example:
Image

To fix, find the source of the errant extra `v` and remove it.

anandvenugopal-tech:
This PR removes the redundant `v` prefix from the release workflow job name.

Since `github.ref_name` already includes the tag prefix, this avoids duplicated
version strings such as `vvX.Y.Z`.

No changelog needed for this CI-only change.

Closes #2271

##### Contributor checklist

##### Maintainer checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -115,7 +115,7 @@ jobs:
   publish-pypi:
     name: >-
       📦
-      Publish v${{ github.ref_name }} to PyPI
+      Publish ${{ github.ref_name }} to PyPI
     needs:
       - build
     if: >-
@@ -142,7 +142,7 @@ jobs:
           path: dist/
       - name: >-
           📦
-          Publish v${{ github.ref_name }} to PyPI
+          Publish ${{ github.ref_name }} to PyPI
           🔏
         uses: pypa/gh-action-pypi-publish@release/v1
       - name: Clean up the publish attestation leftovers


sirosen:
Thanks for doing this!

Could you create a `contrib` category change note for this? It can be very short.
You can see doc on the process and examples here: https://github.com/jazzband/pip-tools/tree/main/changelog.d

> No changelog needed for this CI-only change.

Although I would normally agree, the entire point of this being an issue (rather than a quick fix) is to onboard a new contributor. Creating a `contrib` changelog is an important part of that process.

anandvenugopal-tech:
Thanks! I’ve added the contrib changelog entry as requested.
Please let me know if anything else is needed.

changelog.d/2300.contrib.md
```diff
@@ -0,0 +1 @@
+Remove a redundant 'v' prefix from the CI release workflow job name -- by {user} anandvenugopal-tech.
```
sirosen:
```suggestion
Remove a redundant 'v' prefix from the CI release workflow job name -- by {user}`anandvenugopal-tech`.
```

anandvenugopal-tech:
Thanks for the catch!
I’ve fixed the punctuation and pushed the update.
Please let me know if anything else is needed.

sirosen:
I just double-checked the changeset and realized that this is only fixing one of several instances of the errant `v`. It appears multiple times in that workflow, for both jobs and for at least one step name. All of these need to be corrected, not just the first one.

anandvenugopal-tech:
Thanks for pointing that out — good catch.

I’ve now updated the workflow to remove the redundant `v` prefix from all job and step names and pushed the changes. Please let me know if anything else needs adjustment.

sirosen (approved):
Thanks for working this up!

Issue #2220, opened 2025-08-04
`--extra-index-url` not included in generated requirements.txt


smarnach:
Version 7.5.0 of pip-tools no longer passes `--extra-index-url` from `requirements.in` on to `requirements.txt` when using `--generate-hashes`. Version 7.4.1 still did.

#### Environment Versions

1. Linux
1. Python version: 3.12.3
1. pip version: 24.0
1. pip-tools version: 7.5.0

#### Steps to replicate

Here's a [minimal shell session reproducing the issue](https://gist.github.com/smarnach/75c12d912dd8a8b551e81d74eb28863e).

1. Write a `requirements.in` file containing a `--extra-index-url` line.
2. Run `pip-compile --generate-hashes`

#### Expected result

The `--extra-index-url` line should be included in the generated `requirements.txt` file.

#### Actual result

It's not.

webknjaz:
I seem to remember a security-related request to stop leaking possibly embedded creds. Search the tracker for the details..

webknjaz:
A quick search from mobile didn't return anything useful. I might retry from the laptop later..

smarnach:
I can't find anything in the [commit history between the two versions](https://github.com/jazzband/pip-tools/compare/7.4.1...v7.5.0). If this is intentional, I'd suggest a prominent notice in the [changelog](https://github.com/jazzband/pip-tools/blob/main/CHANGELOG.md#v750) and probably a major version upgrade for such a major breaking change.

sirosen:
I'm not certain that there is an intentional change. I've gone over the changes for 7.5.0 a couple of times because of all of the work we had to do in order to reassemble the changelog and don't recall anything relevant.

I find it very strange that `--generate-hashes` is key to this behavior. With it, the option is dropped. Without it, things work as before. I tested with
```
echo '--extra-index-url https://www.example.com/' | pip-compile --no-header --strip-extras --generate-hashes -o - - > /dev/null
```
as a simple reproducer.

I have traced this out by bisecting the history of `main` to the merge of #2195 .
When I snooped around, it seems to be caused by the replacement of the package finder. Reverting that change and running with a sufficiently old `pip` makes `--extra-index-url` pop back up in the output.

So I found out _where_ this changed but I have a very limited understanding of _why_ right now. I'm marking this as an accepted bug.

webknjaz:
> I have a very limited understanding of why right now.

[...]

PR #2305, opened 2026-01-07
Fix bugs related to losing index URL information when clearing finder cache

### LINKED ISSUE #2294: IndexError when upgrading to Python 3.11 with rlextra
I just tried upgrading from Python 3.9 because it's EOL and it runs into an exception when running `pip-compile` with `rlextra` on Python 3.11

#### Environment Versions

1. Debian Trixie
1. Python version: 3.11.14
1. pip version: 24.0
1. pip-tools version: 7.5.2

#### Steps to replicate

1. Put `rlextra` in a requirements.in file using Python 3.9
2. Build requirements.txt with `pip-compile`
3. Upgrade to Python 3.11
4. Run `pip-compile`
5. Observe traceback

#### Expected result
Would work on Python 3.11 like 3.9

#### Actual result
`IndexError` happens when processing `rlextra`. I believe this is because the versions available in `pypi` don't cover 3.11 and the private repo is needed to get the actual package, but `pip-compile` assumes that something will be available in `pypi` and that causes the `IndexError`

```python-traceback
  File "/usr/src/venv/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/usr/src/venv/lib/python3.11/site-packages/click/core.py", line 1485, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/venv/lib/python3.11/site-packages/click/core.py", line 1406, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/src/venv/lib/python3.11/site-packages/click/core.py", line 1269, in invoke
  File "/usr/src/venv/lib/python3.11/site-packages/click/core.py", line 824, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/venv/lib/python3.11/site-packages/click/decorators.py", line 34, in new_func
  File "/usr/src/venv/lib/python3.11/site-packages/piptools/scripts/compile.py", line 482, in cli
    hashes = resolver.resolve_hashes(results) if generate_hashes else None
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/venv/lib/python3.11/site-packages/piptools/resolver.py", line 171, in resolve_hashes
    return {ireq: self.repository.get_hashes(ireq) for ireq in ireqs}
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/venv/lib/python3.11/site-packages/piptools/resolver.py", line 171, in <dictcomp>
    return {ireq: self.repository.get_hashes(ireq) for ireq in ireqs}
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/venv/lib/python3.11/site-packages/piptools/repositories/pypi.py", line 329, in get_hashes
    return self._get_req_hashes(ireq)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/venv/lib/python3.11/site-packages/piptools/repositories/pypi.py", line 336, in _get_req_hashes
    matching_candidates = self._get_matching_candidates(ireq)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/venv/lib/python3.11/site-packages/piptools/repositories/pypi.py", line 393, in _get_matching_candidates
    return candidates_by_version[matching_versions[0]]
                                 ~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

sirosen:
Rather than rebuilding the package finder, which seems to have a number of unexpected and undesirable side-effects, implement a safe way to clear the cache on a package finder.

A new regression test for #2220 validates the fix, demonstrating that we preserve the `--extra-index-url` provided via `requirements.in`.

This also fixes #2294 , for similar reasons.

This change was also manually tested on the reproducers from #2220 and #2294, where it has the correct effect.

---

The originating work where this bug was introduced comes from #2176, which was fixed in #2178 (and incorporated in #2195).
We were aware at the time that it wasn't clear what side-effects rebuilding the finder might have, but I considered it safe.
It has since turned out (#2220 and #2294) that we lose important information in the process -- we could reconstitute that information but this approach seems simpler.

Importantly, the new test would catch any future regression in this behavior.

##### Contributor checklist

##### Maintainer checklist


changelog.d/2220.bugfix.md
webknjaz, line 1: symlink this PR too?

webknjaz (approved):

PR #2307, opened 2026-01-14
docs: clarify that pip-compile reads existing output file

### LINKED ISSUE #2122: Add information that pip-compile takes into consideration previously generated output-file

During solving one of dependency conflicts I've discovered that pip-compile takes into consideration previously generated output file. I had following error when executing pip-compile to update jupyter/requirements-tools.txt file:

```
Command failed: pip-compile --no-emit-index-url --output-file=jupyter/requirements-tools.txt jupyter/requirements-tools.in requirements.in --upgrade-package=twisted==24.7.0
  ERROR: Cannot install -r requirements.in (line 30) because these package versions have conflicting dependencies.
Discarding incremental==22.10.0 (from -r jupyter/requirements-tools.txt (line 300)) to proceed the resolution
  ERROR: Cannot install opentelemetry-instrumentation-django and opentelemetry-instrumentation-psycopg2 because these package versions have conflicting dependencies.
Discarding opentelemetry-instrumentation==0.45b0 (from -r jupyter/requirements-tools.txt (line 453)) to proceed the resolution
  ERROR: Cannot install opentelemetry-instrumentation-django and opentelemetry-instrumentation-psycopg2  because these package versions have conflicting dependencies.
...
```
After many tries and errors I've discovered that if I just removed the previous jupyter/requirements-tools.txt issue was solved, I could narrow it to the specific internal library in jupyter/requirements-tools.txt but the point is that I had no idea that pip-compile is taking in any consideration previous output file and by intuition if I see name --output-file I don't consider it to have impact on input. Looked through [documentation](https://pip-tools.readthedocs.io/en/latest/) and couldn't find information about that. I must admit that I'm not sure if I understand logic behind that also. I presume that pip-compile tries to create minimum update of file to be compliant with requirements.in or other constraints files but in case of conflicts I'm suprising that it chooses to raise error instead of updating more packages. 

In documentation it should be better explained that pip-compile takes previous output-file into consideration and why is doing that. Additionally it would be good to apply flag that allows to force-update this file. We are using renovate with pip-compile in our repositories and it failes often because of this issue.

#### Alternative Solutions

We can modify renovate execution to always clean previous output file of pip-compile but that it's workaround, not solution to the issue. 

#### Additional context

I didn't go into the depth of working of pip-compile and pip-tools, so maybe I miss some general basic logic here. Maybe it's not output-file per se that is taking into consideration but something else that it's somehow connected to it. If that's the case I'm your humble student and ask for explanation of how it works or where should I look into the code/documentation for better understanding.

maliktafheem:
## Summary
Adds a new 'Understanding output file behavior' section to the README explaining that:

- `pip-compile` uses the existing output file as a constraint
- This can cause dependency conflicts if the file contains stale/incompatible pins
- The `--upgrade` flag or deleting the output file enables fresh resolution

This addresses user confusion about why dependency conflicts occur when an output file already exists.

Fixes #2122

## Changes
- Added new 'Understanding output file behavior' subsection after 'Updating requirements' in README.md
- Documents the three key behaviors: version preservation, potential conflicts, and resolution options

## Notes
- Documentation-only change, no code changes
- Follows existing README style and structure

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,27 +230,16 @@ $ pip-compile --upgrade --upgrade-package 'requests<3.0'
 
 When you run `pip-compile`, it reads any existing output file (e.g.,
 `requirements.txt`) and uses the versions specified there as constraints.
-This behavior has some important implications:
+This ensures that the output is stable -- `pip-compile` will prefer to use existing pinned versions.
 
-1. **Existing pins are preserved**: If a package version in your output file
-   still satisfies your input requirements, `pip-compile` will keep that version
-   rather than upgrading to a newer one. This provides stability by avoiding
-   unnecessary upgrades.
+When upgrading packages or adding new packages to the input,
+it is possible for the existing pins to conflict with new packages. In such cases,
+`pip-compile` will update the pins and emit a warning
+(e.g., `Discarding foo=1.2.3 to proceed`).
 
-2. **Dependency conflicts can occur**: If your output file contains outdated or
-   incompatible package versions, you may encounter resolution errors. The error
-   messages may reference packages from the output file (e.g.,
-   `Discarding package==X.Y.Z (from -r requirements.txt)` when `pip-compile`
-   attempts to resolve conflicts).
-
-3. **Fresh resolution**: To compile dependencies from scratch without considering
-   the existing output file, either:
-   - Delete the output file before running `pip-compile`, or
-   - Use the `--upgrade` / `-U` flag to upgrade all packages, or
-   - Use `--upgrade-package <package>` to upgrade specific packages
-
-This behavior is intentional and provides stability by avoiding unnecessary
-upgrades while still allowing controlled updates when needed.
+In order to resolve dependencies afresh, `pip-compile` provides the following options:
+- `--upgrade` / `-U` flag to upgrade all packages
+- `--upgrade-package <package>`/`-P <package>` to upgrade specific packages
 
 ### Using hashes
 


README.md
```diff
@@ -226,6 +226,32 @@ packages whilst constraining requests to the latest version less than 3.0:
 $ pip-compile --upgrade --upgrade-package 'requests<3.0'
 ```
 
+### Understanding output file behavior
+
+When you run `pip-compile`, it reads any existing output file (e.g.,
+`requirements.txt`) and uses the versions specified there as constraints.
```
sirosen, line 232: You note the rationale for this behavior after a listing of implications, but I think it's better to *start* with a statement that this is part of the defined behavior of the tool:

```suggestion
`requirements.txt`) and uses the versions specified there as constraints.
This ensures that the output is stable -- `pip-compile` will prefer to use existing pinned versions.

```

README.md
```diff
@@ -226,6 +226,32 @@ packages whilst constraining requests to the latest version less than 3.0:
...
+
+3. **Fresh resolution**: To compile dependencies from scratch without considering
+   the existing output file, either:
+   - Delete the output file before running `pip-compile`, or
+   - Use the `--upgrade` / `-U` flag to upgrade all packages, or
+   - Use `--upgrade-package <package>` to upgrade specific packages
+
+This behavior is intentional and provides stability by avoiding unnecessary
+upgrades while still allowing controlled updates when needed.
+
```
sirosen: In concert with my suggestion of an addition at the start, this can go:
```suggestion
```

README.md
```diff
@@ -226,6 +226,32 @@ packages whilst constraining requests to the latest version less than 3.0:
...
+   messages may reference packages from the output file (e.g.,
+   `Discarding package==X.Y.Z (from -r requirements.txt)` when `pip-compile`
+   attempts to resolve conflicts).
+
+3. **Fresh resolution**: To compile dependencies from scratch without considering
+   the existing output file, either:
+   - Delete the output file before running `pip-compile`, or
+   - Use the `--upgrade` / `-U` flag to upgrade all packages, or
+   - Use `--upgrade-package <package>` to upgrade specific packages
+
```
sirosen: I don't think that instructions for how to update already-pinned packages should be part of a list of "implications" -- rather, I think it's better to dedent this and just make it a set of instructions for updating pins.

I also do not consider it worth documenting that deleting a file means that `pip-compile` cannot read it. I strongly prefer that we drop that note.
The result would be:
```suggestion

In order to resolve dependencies afresh, `pip-compile` provides the following options:
- `--upgrade` / `-U` flag to upgrade all packages
- `--upgrade-package <package>`/`-P <package>` to upgrade specific packages

```

README.md
```diff
@@ -226,6 +226,32 @@ packages whilst constraining requests to the latest version less than 3.0:
...
+### Understanding output file behavior
+
+When you run `pip-compile`, it reads any existing output file (e.g.,
+`requirements.txt`) and uses the versions specified there as constraints.
+This behavior has some important implications:
+
+1. **Existing pins are preserved**: If a package version in your output file
+   still satisfies your input requirements, `pip-compile` will keep that version
+   rather than upgrading to a newer one. This provides stability by avoiding
+   unnecessary upgrades.
```
sirosen: I think this is a restatement of the stability guarantee, but more verbose. Is it providing additional information?
I'm open to hearing that this is important to restate. Perhaps I have a bias from being in close contact with the tool. But the current phrasing isn't putting emphasis on anything that is not already captured in the above statement that `pip-compile` prefers to keep existing pins.

I recommend removing it -- and by implication we'd have a list with exactly one element -- so I'm including removal of the list-leading phrase in my suggestion for removal.
```suggestion
```

If we do keep it, I'd like to understand why.

README.md
```diff
@@ -226,6 +226,32 @@ packages whilst constraining requests to the latest version less than 3.0:
...
+
+2. **Dependency conflicts can occur**: If your output file contains outdated or
+   incompatible package versions, you may encounter resolution errors. The error
+   messages may reference packages from the output file (e.g.,
+   `Discarding package==X.Y.Z (from -r requirements.txt)` when `pip-compile`
+   attempts to resolve conflicts).
+
```
sirosen: Because I'm recommending a refinement of this doc addition which eliminates this list, and generally aims for brevity, I feel it's right to suggest a way of handling this list item. This one is mostly good as-is, but I think we can make it a bit simpler:
```suggestion
When upgrading packages or adding new packages to the input,
it is possible for the existing pins to conflict with new packages. In such cases,
`pip-compile` will update the pins and emit a warning
(e.g., `Discarding foo=1.2.3 to proceed`).
```

maliktafheem:
Thanks for the detailed feedback @sirosen! I've applied all your suggestions:

- Added stability statement upfront after the first sentence
- Removed the numbered list structure and redundant "implications" phrasing
- Simplified the conflict warning explanation
- Replaced the verbose "Fresh resolution" section with a concise bullet list for upgrade options
- Removed the note about deleting the output file

The section is now much more direct and succinct. Let me know if any further adjustments are needed!

maliktafheem:
You're right - I did use an LLM as an assistant for this contribution. I reviewed and verified all changes myself, but should have disclosed this upfront. I'll be transparent about this in future PRs. Thanks for the feedback and for merging!

PR #2311, opened 2026-01-30
Fix pip-sync --python-executable marker evaluation

wingding12:
## Summary

`pip-sync --python-executable` was evaluating environment markers (like `python_version < '3.12'`) against the **current** Python environment instead of the **target** environment specified by `--python-executable`.

### Problem

Running from Python 3.12 and targeting Python 3.11:

```bash
echo 'typing-extensions==4.12.2 ; python_version < "3.12"' > req.txt
pip-sync --python-executable=.venv311/bin/python req.txt
```

Would incorrectly say **'Everything up-to-date'** even though `typing-extensions` should be installed in the Python 3.11 environment.

The marker `python_version < '3.12'` was being evaluated against Python 3.12 (the running environment) returning `False`, rather than against Python 3.11 (the target environment) where it would return `True`.

### Root Cause

The `merge()` and `diff()` functions in `sync.py` use `ireq.match_markers()` which evaluates markers against `pip._vendor.packaging.markers.default_environment()` - the current Python's environment, not the target.

### Fix

**3 source files changed:**

1. **`piptools/utils.py`** - Added `get_environment_for_python_executable()` utility function to retrieve PEP 508 environment markers from the target Python executable

2. **`piptools/sync.py`** - Added `_match_markers()` helper function and modified `merge()` and `diff()` to accept an optional `environment` parameter for marker evaluation

3. **`piptools/scripts/sync.py`** - Obtains target environment using the new utility function and passes it to `merge()` and `diff()` when `--python-executable` is specified

## Test Plan

Added comprehensive tests in:
- `tests/test_sync.py` - Tests for `merge()` and `diff()` with custom environment
- `tests/test_utils.py` - Test for `get_environment_for_python_executable()`

Fixes #2115

PR #2312, opened 2026-01-30
Fix src_files config option for pip-sync

### LINKED ISSUE #2187: Specifying `src_files` for `pip-sync` in `pyproject.toml` is ignored.
#### Environment Versions

1. OS Type: `Ubuntu 24.04.2 LTS`
1. Python version: `Python 3.13.3`
1. pip version: `pip 25.1.1 `
1. pip-tools version: `pip-compile, version 7.4.1`

#### Steps to replicate

Specifying `src_files` for `pip-sync` in `pyproject.toml` as follows is ignored and always synchronizes with `requirements.txt`.

```pyproject.toml
[tool.pip-tools.sync]
src_files = ['requirements_lock.txt']
```

And the following error occurs.

```
 $ pip-sync
No requirement files given and no requirements.txt found in the current directory
```

#### Expected result

Expect the same behavior as the following command.

```shell
pip-sync requirements_lock.txt
```

#### Actual result

The behavior is actually equivalent to the following command.

```shell
pip-sync requirements.txt
```

#### My suggetion

The following fix will result in the expected behavior.

https://github.com/jazzband/pip-tools/compare/e4ed0c1e028d1ca73673a51722ba153f0c02b0c6...takuminakaso:pip-tools:ba97901e4b31731259f9da1d2a8bddcd527adcae

wingding12:
## Summary

`pip-sync` was not reading the `src_files` option from config file (`pyproject.toml` or `.pip-tools.toml`), causing it to always look for the default `requirements.txt` even when `src_files` was specified in `[tool.pip-tools.sync]`.

### Problem

```toml
[tool.pip-tools.sync]
src_files = ['requirements_lock.txt']
```

Running `pip-sync` without arguments would fail with:
```
No requirement files given and no requirements.txt found in the current directory
```

### Root Cause

`src_files` is defined as a `click.argument` (not `click.option`), and click's `default_map` mechanism only applies to options, not arguments. The config file is parsed correctly, but the value was never applied to the argument.

Interestingly, `pip-compile` already had a workaround for this (manually checking `ctx.default_map` for `src_files`), but `pip-sync` was missing it.

### Fix

**3 source files changed:**

1. **`piptools/utils.py`** - Added `get_src_files_from_config()` utility function to extract `src_files` from click context's `default_map`

2. **`piptools/scripts/sync.py`** - Use the utility to read `src_files` from config, with `@click.pass_context` decorator

3. **`piptools/scripts/compile.py`** - Refactored to use the shared utility (was using inline code before)

## Test Plan

Added tests in:
- `tests/test_cli_sync.py` - Tests for src_files from config and CLI override
- `tests/test_utils.py` - Unit tests for `get_src_files_from_config()`

Fixes #2187

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -26,6 +26,7 @@ from ..resolver import BacktrackingResolver, LegacyResolver
 from ..utils import (
     dedup,
     drop_extras,
+    get_src_files_from_config,
     is_pinned_requirement,
     key_from_ireq,
 )
@@ -180,11 +181,10 @@ def cli(
         ctx.color = color
     log.verbosity = verbose - quiet
 
-    # If ``src-files` was not provided as an input, but rather as config,
-    # it will be part of the click context ``ctx``.
-    # However, if ``src_files`` is specified, then we want to use that.
-    if not src_files and ctx.default_map and "src_files" in ctx.default_map:
-        src_files = ctx.default_map["src_files"]
+    # If ``src_files`` was not provided as an input, check config.
+    # Since src_files is a click argument (not option), it's not automatically
+    # populated from the default_map, so we handle it explicitly.
+    src_files = get_src_files_from_config(ctx, src_files)
 
     if all_build_deps and build_deps_targets:
         raise click.BadParameter(
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -23,6 +23,7 @@ from ..repositories import PyPIRepository
 from ..utils import (
     flat_map,
     get_required_pip_specification,
+    get_src_files_from_config,
     get_sys_path_for_python_executable,
 )
 from . import options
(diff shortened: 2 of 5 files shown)


sirosen:
You have opened 6 PRs in a very short timespan (#2311, #2312, #2313, #2314, #2315, #2316).
Not all of them are viable (e.g., #2313 may work but IMO is not the right way to solve) and at least two of them overlap (#2312 and #2314 appear to address the same issue).
Some are making more changes at once than are comfortable for review of a new contributor's work (#2316).

Development and review resources on `pip-tools` are very limited, as it's purely a volunteer project.
To protect that time, I am closing all of these PRs in a batch, without prejudice[^1] but also without carefully examining all of the code here.
I suspect there are some good and usable changes within these, but the manner of contribution matters -- as provided, the project can't responsibly accept these.

If you are really interested in contributing, and I hope you are, please choose one small area to contribute and only open that one PR at first to get ramped up on the project.
Try to follow existing conventions, e.g. do not introduce test classes into a testsuite which primarily does not use them.

[^1]: Meaning, any of them could be reopened.

PR #2313, opened 2026-01-30
Fix pip-compile hang on named pipes as input

### LINKED ISSUE #2232: [regression][6.10.0] `pip-compile` hangs on named pipes as input due to double-read

#### Environment Versions

1. pip version: `25.2`

#### Steps to replicate

*(\*NIX instructions)*

1. `mkfifo -m0600 named-pipe`

#### Expected result

It shouldn't hang.

#### Actual result

It does.

-----------

```pycon
-> breakpoint()
(Pdb) args
strategy = 'preserve'
(Pdb) pp locals()
(Pdb) c
-> breakpoint()
(Pdb) pp locals()
 'fname': './named-pipe',
 'strategy': 'preserve'}
False
False
False
False
(Pdb) pp os.path.isdir(fname)
False
False
(Pdb) pp os.path.isabs(fname)
False
(Pdb) import stat
True
True
```

-----------

wingding12:
## Summary

When using named pipes (FIFOs) as input files, `pip-compile` would hang during output generation. This is a regression introduced in pip-tools v6.10.0.

### Problem

```bash
mkfifo named-pipe
pip-compile -o- ./named-pipe  # hangs indefinitely
```

### Root Cause

The `_determine_linesep()` function attempts to read files to detect existing line endings, but reading from a named pipe blocks indefinitely waiting for a writer.

### Fix

**2 source files changed:**

1. **`piptools/utils.py`** - Added `is_regular_file()` utility function that properly checks for regular files (returns False for stdin, FIFOs, directories, sockets, etc.)

2. **`piptools/scripts/compile.py`** - Use `is_regular_file()` in `_determine_linesep()` to skip non-regular files before attempting to read them

This prevents the hang while preserving the line ending detection behavior for regular files.

## Test Plan

Added tests in:
- `tests/test_utils.py` - Unit tests for `is_regular_file()` (regular files, stdin, FIFOs, directories)
- `tests/test_cli_compile.py` - Integration tests for `_determine_linesep()` with FIFOs

Tests are skipped on Windows where named pipes work differently.

Fixes #2232

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -441,7 +441,6 @@ def omit_list_value(lst: list[_T], value: _T) -> list[_T]:
     return [item for item in lst if item != value]
 
 
-
 def is_regular_file(path: str) -> bool:
     """
     Check if the given path is a regular file (not stdin, FIFO, socket, etc.).
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -4206,13 +4206,12 @@ def test_that_self_referential_pyproject_toml_extra_can_be_compiled(
 
 
 @pytest.mark.skipif(
-    sys.platform == "win32",
-    reason="Named pipes (FIFOs) are Unix-specific"
+    sys.platform == "win32", reason="Named pipes (FIFOs) are Unix-specific"
 )
 def test_determine_linesep_skips_fifo(tmpdir_cwd):
     """
     Test that _determine_linesep doesn't block on named pipes (FIFOs).
-    
+
     Regression test for https://github.com/jazzband/pip-tools/issues/2232
     """
     import os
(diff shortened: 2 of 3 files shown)


sirosen:
You have opened 6 PRs in a very short timespan (#2311, #2312, #2313, #2314, #2315, #2316).
Not all of them are viable (e.g., #2313 may work but IMO is not the right way to solve) and at least two of them overlap (#2312 and #2314 appear to address the same issue).
Some are making more changes at once than are comfortable for review of a new contributor's work (#2316).

Development and review resources on `pip-tools` are very limited, as it's purely a volunteer project.
To protect that time, I am closing all of these PRs in a batch, without prejudice[^1] but also without carefully examining all of the code here.
I suspect there are some good and usable changes within these, but the manner of contribution matters -- as provided, the project can't responsibly accept these.

If you are really interested in contributing, and I hope you are, please choose one small area to contribute and only open that one PR at first to get ramped up on the project.
Try to follow existing conventions, e.g. do not introduce test classes into a testsuite which primarily does not use them.

[^1]: Meaning, any of them could be reopened.

PR #2317, opened 2026-01-30
Increase the GHA testsuite timeout to 15m

sirosen:
The existing 9m mark can actually be reached on Windows runners, when they are slower than usual.

I observed this recently while trying to get reruns of CI to pass after a GitHub outage:
  https://github.com/jazzband/pip-tools/actions/runs/21367431586/job/62022523913

No changelog for this, as it's purely repo CI config.

##### Contributor checklist

- [x] ~A change note is created in `changelog.d/` (see [`changelog.d/README.md`](https://github.com/jazzband/pip-tools/blob/main/changelog.d/#readme) for instructions) or the PR text says "no changelog needed".~

##### Maintainer checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
   test:
     name: ${{ matrix.os }} / ${{ matrix.python-version }} / ${{ matrix.pip-version }}
     runs-on: ${{ matrix.os }}-latest
-    timeout-minutes: 9
+    timeout-minutes: 15
     strategy:
       fail-fast: false
       matrix:


sirosen:
I considered doing something like that. I think that as a general rule, the timeout should either be set based on the slowest job in the matrix or set per-job.

Making it configurable didn't seem worth it in exchange for the reduced readability. If it continues to be a problem I'll definitely revisit.

---

Related topic: pypy gradual EOL.

It's only Windows-pypy that is very slow like this. It currently looks like we should plan to drop pypy support at some point: I saw a thread in which a pypy maintainer said that they're not continuing to work on it.

I haven't formed much opinion yet, other than that I roughly plan to drop pypy support from projects when they drop Python 3.11 (the last version implemented in pypy).

We might drop pypy from CI sooner if it becomes an issue, but I'm not advocating for that yet.

webknjaz:
I usually prefer tight timeouts for responsiveness but allow a minute or two for network flakiness / retries.

----

Yep, Antonio mentioned the same on the C-API Summit at EuroPython last summer. I'm not too concerned about it and usually only run it on Ubuntu in CI. pip-tools is on the fundamental side of things in the ecosystem so I wouldn't rush unless it causes problems.

Issue #2264, opened 2025-11-06
Start using a `coverage` plugin to handle conditional coverage omissions


sirosen:
In #2253, it came up that we wanted to conditionally exclude a block from coverage reporting on Python 3.8.

In order to support this, we'd need some mechanism beyond the baseline suite of features in `coverage`. Either our own workarounds (e.g., multiple configs) or a `coverage` plugin. A plugin seems preferable, as it keeps all of our config in one place and builds upon open and prior art.

### Existing Plugins

Going off of the plugins which have gotten themselves added to [the `coverage` docs here](https://coverage.readthedocs.io/en/7.11.0/other.html), there are primarily two options.

[coverage-conditional-plugin](https://github.com/wemake-services/coverage-conditional-plugin) offers conditional configs, which encompasses this use case. The project has few updates and appears mostly inactive, but not necessarily dead.

[coverage-simple-excludes](https://github.com/haukex/coverage-simple-excludes) appears to be a perfect fit for our needs. It's newer and more active, but appears to be a solo effort without community traction as of yet.

### Our Own Plugin

Writing our own plugin for this purpose, especially an in-tree one (e.g., `piptools._internal.extensions.coverage`) would be very easy. It would not give back to the community necessarily, except in the form of showing how this can be done. Using the existing plugins and filing issues if we have trouble would potentially feed forward, and I therefore treat it as a preferable path.

I consider this an excellent *backup* option, if using the existing plugins is problematic.

### Action Items

- Experiment with one or both of the listed plugins
- Configure one, of the implementer's choosing, for use in `pip-tools`
- Examine any existing no-cover pragmas to see if they can be updated to be conditional
- Report back on your findings, here or in a new PR

---

I have tried to phrase this as a good exploratory issue for any motivated new contributor. However, we should discuss this idea -- to ensure maintainers are aligned -- before anyone implements.

webknjaz:
Here's a few plugin references: https://github.com/python-trio/trio/issues/3163. I don't necessarily want to choose the same plugins (supply chain suspicion) but there's a few to look into.

webknjaz:
> [coverage-conditional-plugin](https://github.com/wemake-services/coverage-conditional-plugin)

This one is not being tested past Python 3.11

> [coverage-simple-excludes](https://github.com/haukex/coverage-simple-excludes)

And this dropped Python 3.8 already (although, using an older version for that env is fine too).

---

[...]

PR #2347, opened 2026-03-05
Introduce a local coverage plugin + add conditional exclusions

sirosen:
This PR is a continuation from #2268 .
As I mentioned in #2264 , I've worked up a local plugin and tried out using `PYTHONPATH` to make it importable under `coverage`, and locally my testing of it seems to work.
 
The plugin is defined in `plugins/coverage/` as a module, and injects coverage excludes of the form

    pragma: pip{comparator}{version} no cover

I've added it on top of the commits from #2268, rebased, and updated to use the new "pip-version pragmas" where possible.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist


sirosen:
`codecov` was unhappy -- my guess about the issue, based on the output, was that it was treating the coverage plugin as "uncovered", so I've pushed a fix for that.

If the coverage still looks bad, I may have to look more deeply and potentially mark this draft until I can fix.

tests/test_cli_compile.py
```diff
@@ -4101,7 +4101,7 @@ def test_second_order_requirements_relative_path_in_separate_dir(
     output_path = test_files_collection.get_path_to("requirements2.in")
 
     # for older pip versions, recompute the output path to be relative to the input path
-    if not pip_produces_absolute_paths:
+    if not pip_produces_absolute_paths:  # FIXME: figure out how to cover piplowest
```
webknjaz: Could you elaborate?

tests/conftest.py
```diff
@@ -230,7 +230,8 @@ def from_editable():
 
 @pytest.fixture
 def runner():
-    if Version(version_of("click")) < Version("8.2"):
+    # Coverage is excluded because we only test with the latest Click
```
webknjaz, line 252: Should we fix that? Or maybe make it not branchy by updating deps in the meta?

.coveragerc
```diff
@@ -1,8 +1,14 @@
 [run]
-plugins = covdefaults
+plugins =
+    covdefaults
+    piptools_coverage
 omit =
     piptools/_compat/*
+    # any local plugins we use aren't part of normal testing/coverage
```
webknjaz, line 7: Honestly, I'd argue that our own tooling is subject to being tested as well. Especially since it may grow into something that could eventually be ready to be moved out. So if possible, I'd add a test for the plugin.

plugins/coverage/piptools_coverage.py
webknjaz, line 1: FWIW, I'm not sure I like a top-level public name like `plugins/`. Wonder if it could live in `tests/_plugins/` or smth like that since it's a part of our test toolchain.

plugins/coverage/piptools_coverage.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class PipVersionPragmas(CoveragePlugin):  # type: ignore[misc]
+    def configure(self, config: CoverageConfig) -> None:
+        exclude = set(config.get_option("report:exclude_lines"))
+        exclude.update(_compute_pip_version_exclude_pragmas())
+        config.set_option("report:exclude_lines", sorted(exclude))
+
+
+def coverage_init(registry: Plugins, options: dict[str, _t.Any]) -> None:
```
webknjaz: MyPy recommends `object` instead of `_t.Any`. Can we use that?

tests/test_cli_compile.py
```diff
@@ -1885,9 +1885,9 @@ def test_forwarded_args_filter_deprecated(PyPIRepository, runner, pip_args):
 
     (first_posarg, *_tail_args), _kwargs = PyPIRepository.call_args
 
-    if _pip_api.PIP_VERSION_MAJOR_MINOR >= (25, 3):  # pragma: >=3.9 cover
+    if _pip_api.PIP_VERSION_MAJOR_MINOR >= (25, 3):  # pragma: pip<25.3 no cover
```
webknjaz: This feels a bit contradictary: the check is `<=` and the pragma's inversed. And then the opposite branch adds inversion that looks like what's in the if-expression down below. So I'd suggest supporting `cover`, not just `no cover`.

Althouh, can we analyse expressions with the known constants like `_pip_api.PIP_VERSION_MAJOR_MINOR` / `_pip_api.PIP_VERSION` and apply coverage inclusion/exclusion based on that automatically even w/o the comments? The `else`-branch could be handled the same way.

tox.ini
```diff
@@ -34,6 +34,8 @@ deps =
     piplatest: pip
     pipmain: https://github.com/pypa/pip/archive/main.zip
 setenv =
+    # local plugin inclusion
+    coverage: PYTHONPATH=./plugins/coverage/
```
webknjaz: Do we know if pytest's [`pythonpath` setting](https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-pythonpath) would be enough instead of this?

plugins/coverage/piptools_coverage.py
```diff
@@ -0,0 +1,74 @@
...
+from coverage.config import CoverageConfig
+from coverage.plugin_support import Plugins
+
+# only support versions which are explicitly listed
+# support starts from our lowest supported major version and includes all minor versions
+# pip uses calver, so we can count years from 2023 up to the current year
+CURRENT_YEAR_EPOCH = datetime.date.today().year - 2000
+_PRAGMA_SUPPORTED_PIP_VERSIONS: list[tuple[int, int]] = [
+    (major, minor)
+    for major in range(22, CURRENT_YEAR_EPOCH + 1)
```
webknjaz: Can we read the lowest supported from `pyproject.toml`?

webknjaz (commented):
Cool that you got to looking into this! I've left a few comments inline.

tests/test_cli_compile.py
sirosen: Oh, that wasn't supposed to still be there! :flushed:

I think that's from before I had the plugin working.
But it's wrong. Let me replace with a pragma.

tests/conftest.py
sirosen: I think we should go after a fix, but that it doesn't fit within the scope of this PR.

I've been considering the possible argument for _vendoring_ `click`, on account of the fact that it would allow `pip-tools` to be used in environments with really old `click` versions.

Like `pip`, I think we should not have Python package dependencies, other than `pip` itself. (Also like `pip`, I think this will pose a licensing, SBOM, and transparency challenge! )

.coveragerc
sirosen: Although I agree in principle, I think we'll have trouble unless we add a dedicated test job in tox and CI, with a dedicated and distinct `.coveragerc`. The reason is that `coverage` imports this code on startup, so it's imported _too early_ to measure. I can't think of a way to measure it right now, though I was also considering reaching out to Ned (he's active in the main Python discord) to see if he could help point me in the right direction.

I'm happy to write tests for it, but I think it might belong outside of coverage reporting until I can figure that out?
I was hoping we could include it early, and I could continue to work on this.

plugins/coverage/piptools_coverage.py
sirosen: I can move it around. `tests/_plugins/_coverage/` works for me, if that sounds good?

tests/test_cli_compile.py
sirosen: Why don't I work on `cover` first, and we can think about supporting comparisons with `_pip_api.PIP_VERSION` attributes as a future bit of work?

plugins/coverage/piptools_coverage.py
sirosen: When I circle back on this to do more work, I'll give it a try. :+1:

tox.ini
sirosen: I can give it a try! :+1:

plugins/coverage/piptools_coverage.py
sirosen: Oh, that's a good idea. Yes, we can extract it from there.

sirosen:
It looks like codecov is still displeased with me. :grinning:

I've marked as a draft while I continue to work on this. I'll mark it ready when I feel that I've gotten it to a good place again. I have lots of lovely feedback to act upon, noted above, so changes to be made.

As for codecov, I'm not totally sure. I clearly misdiagnosed the "loss of coverage" it's flagging, but the reporting it shows doesn't quite match my expectations there. I think it *might* be that it sees the missing `else` branches of various conditionals as "untested" because they aren't explicit and flagged with appropriate `pragma`s. That will be one of my next experiments.

sirosen:
I moved the plugin into `tests/_plugins/_coverage/`, and tweaked our pytest config to put it in pythonpath but exclude it from test discovery. That works!

If we go with this approach, I plan to write a short blog post to share the technique of a local coverage plugin. It's not something I've seen anyone write about before, and it wasn't _hard_ but it did have some non-obvious requirements (most notably, pythonpath manipulation).

---

I've enhanced the plugin with `pragma: pip{comparator}X.Y cover` pragmas, and added a suite of unit tests to help get those bits correct. I find these easy to get confused when implementing -- there's some reasoning about double-negatives involved in checking this. I've been doing some local test runs and looking at coverage reporting, and it seems to be working as desired.

I've also explicitly expanded a lot of `else` branches for pip-version conditionals, almost all of which are `else: pass`, so that they can be marked with the appropriate pragmas. That fixed coverage branch reporting showing up as partial, and seems to have gotten codecov to pass. As part of that body of work, I aligned the pragmas with the comparisons being done. They now typically read like so:

```python
if _pip_api.PIP_VERSION_MAJOR_MINOR >= (22.3):  # pragma: pip>=22.3 cover
    ...
else:  # pragma: pip>=22.3 no cover
    ...
```

That is, I've been using the matching comparison in both branches, but with `cover` vs `no cover`.

---

Deriving the minimum pip version from `pyproject.toml` involved making a few decisions:
1. How do we find `pyproject.toml`? I went with `cwd() / pyproject.toml`, since `tox` will always execute from the repo root.
2. What do we do if there are 0 or multiple `pip` versions listed? I treat this as an error.
3. How sophisticated do we want to be in parsing the `pip` version specifier? Again, I treated many cases as errors.

My reasoning around bailing out early with errors is that we definitely don't want to try to solve the general case of "what is the minimum version of package X which satisfies this dependency list?" as it's a hard problem. If we make some simplifying assumptions -- and treat anything unexpected as an error -- we get an easy problem for now and can improve it if we ever need to.

---

I'm watching CI to see if it comes up healthy. If everything looks okay, I feel like I can mark this ready.

.coveragerc
webknjaz: Right.. I imagine we could maybe have some tests w/ the [`pytester` fixture](https://docs.pytest.org/en/stable/reference/reference.html#pytester) invoking them in a subprocess, wrapped in a `coverage run` maybe, shielded from auto-importing coverage and something along the lines of `-p no:pytest_cov`, perhaps.

I'll think about getting this in early.

sirosen:
I've been working on self-testing this plugin and getting full coverage reporting, including the plugin itself.

I intend to open it up as a separate draft PR for follow-up work. I'm not certain that we'll easily agree about exactly how to do that. (I've tried a couple of things and have something working.)

Because it's more fraught, I'd like to separate that from the PR here which introduces the plugin and its usage. I'm rebasing now to cleanup conflicts.

piptools/_internal/_pip_api/cli_options.py
```diff
@@ -16,5 +16,7 @@ def postprocess_cli_options(options: optparse.Values) -> None:
     After CLI parsing, pip processes options further to check various constraints and
     coalesce values. Emulate and/or invoke those same behaviors.
     """
-    if _pip_version.PIP_VERSION_MAJOR_MINOR >= (26, 0):
+    if _pip_version.PIP_VERSION_MAJOR_MINOR >= (26, 0):  # pragma: pip>=26.0 cover
         cmdoptions.check_release_control_exclusive(options)
+    else:  # pragma: pip>=26.0 no cover
```
webknjaz: Any way to apply this to the implicit else-branches automatically?

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,128 @@
...
+    # the path to pyproject.toml is taken always from the CWD
+    # since tests should always run under tox, this is always expected to be the repo root
+    pyproject_path = pathlib.Path.cwd() / "pyproject.toml"
+    with pyproject_path.open("rb") as pyproject_file:
+        pyproject_toml = tomllib.load(pyproject_file)
+    return pyproject_toml
+
+
+def get_min_supported_pip_major_version() -> int:
+    pyproject_toml = read_project_pyproject_toml()
```
webknjaz: I'm wondering if it'd be possible to use a `.coveragerc.toml` instead.

webknjaz: Ah, never mind. We don't have own config.

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,128 @@
...
+    ("==", operator.eq),
+]
+
+
+# pip uses 2-digit calver, so the current year is the latest possibly supported version
+def get_max_pip_major_version() -> int:
+    return datetime.date.today().year - 2000
+
+
+def read_project_pyproject_toml() -> dict[str, _t.Any]:
```
webknjaz: What if we returned the deps from this function? The typing would be better.
Alternatively, we could use a typed dict that has `project⇾dependencies` defined. This would help us avoid `typing.Any`.

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,128 @@
...
+
+def get_min_supported_pip_major_version() -> int:
+    pyproject_toml = read_project_pyproject_toml()
+    parsed_dependencies = [
+        Requirement(d) for d in pyproject_toml["project"]["dependencies"]
+    ]
+    pip_requirements = [r for r in parsed_dependencies if r.name == "pip"]
+
+    # we will presently assume that there's exactly one declared dependency on 'pip'
+    # which sets the lower bound
```
webknjaz, line 57: @sirosen now that we have that `stable` extra, we should probably parse it for the upper bound too.

webknjaz: Though, I guess `get_max_pip_major_version()` takes care of that..

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,128 @@
...
+    result: list[str] = [
+        rf"# pragma: pip=={current_major}.{current_minor} no cover\b",
+    ]
+
+    for major, minor in list_supported_pip_versions():
+        for opname, opfunc in COMPARATORS:
+            if opfunc((current_major, current_minor), (major, minor)):
+                result.append(rf"# pragma: pip{opname}{major}.{minor} no cover\b")
+            else:
+                result.append(rf"# pragma: pip{opname}{major}.{minor} cover\b")
```
webknjaz: How about 
```suggestion
            nope = "no " if opfunc((current_major, current_minor), (major, minor)) else ""
            result.append(rf"# pragma: pip{opname}{major}\.{minor} {nope}cover\b")
            if not opfunc((current_major, current_minor), (major, minor)):
                result.append(fr"^\s*if\s*_pip_api\.PIP_VERSION_MAJOR_MINOR\s*{opname}\s*({major},\s*{minor}):\b")
```
?

Though, the addition of native `_pip_api.PIP_VERSION_MAJOR_MINOR` checks reveals the need to have a mapping of opposite `COMPARATORS` names...

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,128 @@
...
+
+    return result
+
+
+class PipVersionPragmas(CoveragePlugin):  # type: ignore[misc]
```
webknjaz: Could you add a comment on what's ignored exactly?

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,128 @@
...
+    base_version = parse_version(pip.__version__).base_version
+    major_str, minor_str = base_version.split(".")[:2]
+
+    return int(major_str), int(minor_str)
+
+
+def compute_pip_version_exclude_pragmas() -> list[str]:
+    current_major, current_minor = get_pip_major_minor()
+
+    result: list[str] = [
```
webknjaz: Why not use a `set` right away, seeing that the caller works with sets?

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,128 @@
...
+                result.append(rf"# pragma: pip{opname}{major}.{minor} cover\b")
+
+    return result
+
+
+class PipVersionPragmas(CoveragePlugin):  # type: ignore[misc]
+    def configure(self, config: CoverageConfig) -> None:
+        exclude = set(config.get_option("report:exclude_lines"))
+        exclude.update(compute_pip_version_exclude_pragmas())
+        config.set_option("report:exclude_lines", sorted(exclude))
```
webknjaz: Shouldn't we use `exclude_also` for this?

webknjaz: If the helper were to return a set, this could be something like
```suggestion
        exclude_also_line_regexes = (
            compute_pip_version_exclude_pragmas()
            | set(config.get_option("report:exclude_also"))
        )
        config.set_option("report:exclude_also", sorted(exclude_also_line_regexes))
```

tests/unit/_internal/pip_api/test_install_requirements.py
```diff
@@ -9,7 +9,9 @@
     _pip_api.PIP_VERSION_MAJOR_MINOR < (25, 3), reason="test requires pip>=25.3"
 )
 @pytest.mark.parametrize("use_pep517", (True, False))
-def test_copy_install_requirement_removes_pip_25_3_unsupported_opts(use_pep517):
+def test_copy_install_requirement_removes_pip_25_3_unsupported_opts(
+    use_pep517,
+):  # pragma: pip<25.3 no cover
```
webknjaz, line 14: We could also have marks like `@pytest.mark.requires_pip_25_3_plus` (or just `requires_pip_25_3_plus` defined as `requires_pip_25_3_plus = pytest.mark.skipif(
    _pip_api.PIP_VERSION_MAJOR_MINOR < (25, 3), reason="test requires pip>=25.3"
)` for runtime) that would work for coverage skips too and would be better integrated.
This is how I skip `xfail` elsewhere, for example: https://github.com/aio-libs/yarl/blob/e25e8d23e6912db52a23513ef1f6a17f889751ef/.coveragerc#L18C3-L18C27.

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def compute_pip_version_exclude_pragmas() -> list[str]:
+    current_major, current_minor = get_pip_major_minor()
+
+    result: list[str] = [
+        rf"# pragma: pip=={current_major}.{current_minor} no cover\b",
+    ]
+
+    for major, minor in list_supported_pip_versions():
```
webknjaz, line 115: What if we also generate entries matching `if _pip_api.PIP_VERSION_MAJOR_MINOR <= (major, minor):`?

piptools/_internal/_pip_api/cli_options.py
sirosen: It may or may not be possible, but I don't have a theory as to how as of today.
It would need a different mechanism from the current excludes -- I looked a little at the coveragepy plugin interfaces, but I don't know how to do it (yet?)!

I wouldn't want to block on this, but I agree that it's desirable.

tests/_plugins/_coverage/piptools_coverage.py
sirosen: Yeah, I can make that tweak.

tests/_plugins/_coverage/piptools_coverage.py
sirosen: `[stable]` isn't merged yet, but I'd be happy to rebase and switch to it if we get it in. (Right now it's marked as a draft, but I don't think there is any blocker for merging it? I'd have to re-check.)

tests/_plugins/_coverage/piptools_coverage.py
sirosen: I think either works equivalently in this case, because we preserve the pre-existing value. Let me try switching it to `report:exclude_also` and I'll confirm back here if it works.

tests/unit/_internal/pip_api/test_install_requirements.py
sirosen: Most of the other suggested changes (e.g., switch the return type in the plugin to `set`) are things I'm game to tinker with, but I'd like to leave this one out at least for a first draft. I'm not sure if it should be `@pytest.mark.requires_pip_25_3_plus` or maybe more like... `@pytest.mark.requires_pip(">=25.3")` or something else. I'd like to have more time and space to explore this without blocking the rest of the work.

tests/_plugins/_coverage/piptools_coverage.py
sirosen: We can do that, yeah! There's a bit of a question for me of how many patterns we want to have here at the start. I'll add all of the comparators with `_pip_api.PIP_VERSION_MAJOR_MINOR` as the left-hand-side.

sirosen: I tried this for a couple of hours and I'm not sure why, but it didn't work. I'm happy to try again, but I don't want to include it in this PR.

nedbat:
Sorry I'm late to this, @sirosen showed this to me at PyCon. I'm wondering if there is a way to get what you need without code. For example, [this tox.ini](https://github.com/pyga/ebb-lint/blob/master/tox.ini) defines environment variables:
```
[testenv]
setenv =
    py27: COVERAGE_MAJOR = 2
    py27: COVERAGE_MAJORMINOR = 27
    py{34,35}: COVERAGE_MAJOR = 3
    py34: COVERAGE_MAJORMINOR = 34
    py35: COVERAGE_MAJORMINOR = 35
```
then uses [coverage settings](https://github.com/pyga/ebb-lint/blob/master/.coveragerc) to define pragmas:
```
[report]
exclude_lines =
    ✘py${COVERAGE_MAJOR}
    ✘py${COVERAGE_MAJORMINOR}
    pragma: no ?cover
```

sirosen:
Thanks for the tips, @nedbat ! I'm pretty interested in this topic. I think this would expand out into a really large configuration in our case.

I'll try to produce a worked example, maybe in a separate repo, and I'll share it with you when I'm able. I think the abstract version of the plugin I've written here is pretty simple. It's pip-tools-specific stuff that makes it more complicated looking!

**Update**:
I shared this in the coverage-py channel of the Python discord, but I worked out the plugin [as a standalone example](https://github.com/sirosen/jigs/blob/a4bad9cfd3d01ba2a78286a490ce069f88c7c67d/python/testing/version-detecting-coverage-plugin/detecting_coverage_plugin.py) to make it easier to dsicuss. I'm happy to talk about it here or there; wherever is convenient. :grinning:

piptools/_internal/_pip_api/cli_options.py
webknjaz: Would https://coverage.readthedocs.io/en/7.15.2/config.html#report-partial-also be helpful?

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,128 @@
...
+            "piptools_coverage is unable to determine the pip lower bound. "
+            "Please update the plugin to handle changes to piptools metadata."
+        )
+
+    version = Version(pip_specifier.version)
+    return version.major
+
+
+def list_supported_pip_versions() -> list[tuple[int, int]]:
+    return [
```
webknjaz, line 85: Feels like we could do some `zip()` here, perhaps.

sirosen: I can't think of a way which would be more readable than what I have written. Maybe I'm not thinking of some clever usage.

Is it hard to read? If so, I can try to address it, but I'd need some pointer about what to improve. To me it looks okay.

webknjaz:
So I pointed Claude at this with the following prompt:
```
Inspect PR 2347 and figure out if it's possible to avoid adding else-blocks just for having those  for custom no-cover pragmas.
```

It came back to me with
```
Yes — it's possible, and I've implemented and verified it.

The else: pass blocks in PR #2347 exist only so a # pragma: ... no cover comment has somewhere to live — coverage.py's branch-coverage checker otherwise flags the untaken arm of a bare if as a "partial branch." The fix (which you'd actually already flagged, unanswered, in an inline PR review comment) is to tell coverage.py that lines carrying our pip-version pragma are allowed to have an unbalanced branch, via the report:partial_branches config option — the same mechanism covdefaults uses internally for its own Python-version pragmas.

One correction to your original suggestion: report:partial_also (the newer option) doesn't work when set from a plugin — I confirmed via an isolated repro that coverage.py merges partial_also into the effective list inside config.post_process(), which runs before plugin configure() hooks fire, so anything a plugin sets there is silently dropped. report:partial_branches (read-merge-write, same pattern already used for exclude_lines) is the one that actually works.

Changes made (tests/_plugins/_coverage/piptools_coverage.py + 8 call sites):
- Added a static regex PARTIAL_BRANCH_PIP_VERSION_PRAGMA and merged it into report:partial_branches in PipVersionPragmas.configure().
- Removed the now-unneeded else: pass pairs in cli_options.py, install_requirements.py (×3), resolver.py, _deprecations.py, tests/conftest.py, and tests/test_circular_imports.py (×2).
- Added unit tests for the new regex and the plugin's partial_branches merging behavior.

Verified in a scratch venv (pip 26.1, coverage 7.15.4): full test suite passes (1040 passed), and a --cov-report=term-missing run shows all 8 touched files at 100% branch coverage with no partial-branch warnings — matching baseline exactly, just without the boilerplate. Also confirmed clean via black and flake8 (this repo's actual formatters, not ruff).
```

and here's what's it got on my disk at a result:
```diff
$ git diff
diff --git a/piptools/_internal/_pip_api/cli_options.py b/piptools/_internal/_pip_api/cli_options.py
index 80f10f9b..98551956 100644
--- a/piptools/_internal/_pip_api/cli_options.py
+++ b/piptools/_internal/_pip_api/cli_options.py
@@ -18,5 +18,3 @@ def postprocess_cli_options(options: optparse.Values) -> None:
     """
     if _pip_version.PIP_VERSION_MAJOR_MINOR >= (26, 0):  # pragma: pip>=26.0 cover
         cmdoptions.check_release_control_exclusive(options)
-    else:  # pragma: pip>=26.0 no cover
-        pass
diff --git a/piptools/_internal/_pip_api/install_requirements.py b/piptools/_internal/_pip_api/install_requirements.py
index 31c9048e..28dd94e7 100644
--- a/piptools/_internal/_pip_api/install_requirements.py
+++ b/piptools/_internal/_pip_api/install_requirements.py
@@ -60,21 +60,15 @@ def copy_install_requirement(
         # Ref: https://github.com/jazzband/pip-tools/issues/2252
         kwargs["use_pep517"] = template.use_pep517
         kwargs["global_options"] = template.global_options
-    else:  # pragma: pip<25.3 no cover
-        pass
     kwargs.update(extra_kwargs)
 
     if _pip_version.PIP_VERSION_MAJOR_MINOR >= (25, 3):  # pragma: pip>=25.3 cover
         # Ref: https://github.com/jazzband/pip-tools/issues/2252
         kwargs.pop("use_pep517", None)
         kwargs.pop("global_options", None)
-    else:  # pragma: pip>=25.3 no cover
-        pass
 
     if _pip_version.PIP_VERSION_MAJOR_MINOR <= (23, 0):  # pragma: pip<=23.0 cover
         kwargs["install_options"] = template.install_options
-    else:  # pragma: pip<=23.0 no cover
-        pass
 
     # Original link does not belong to install requirements constructor,
     # pop it now to update later.
diff --git a/piptools/resolver.py b/piptools/resolver.py
index 23642134..1a5680ea 100644
--- a/piptools/resolver.py
+++ b/piptools/resolver.py
@@ -611,8 +611,6 @@ class BacktrackingResolver(BaseResolver):
             if _pip_api.PIP_VERSION_MAJOR_MINOR < (25, 3):  # pragma: pip<25.3 cover
                 # Ref: https://github.com/jazzband/pip-tools/issues/2252
                 extra_resolver_kwargs["use_pep517"] = self.options.use_pep517
-            else:  # pragma: pip<25.3 no cover
-                pass
 
             resolver = self.command.make_resolver(
                 preparer=preparer,
diff --git a/piptools/scripts/_deprecations.py b/piptools/scripts/_deprecations.py
index b631b91d..b0ead496 100644
--- a/piptools/scripts/_deprecations.py
+++ b/piptools/scripts/_deprecations.py
@@ -19,8 +19,6 @@ def filter_deprecated_pip_args(args: list[str]) -> list[str]:
     """
     if _pip_api.PIP_VERSION_MAJOR_MINOR >= (25, 3):  # pragma: pip>=25.3 cover
         args = _filter_args_deprecated_in_pip_25_3(args)
-    else:  # pragma: pip>=25.3 no cover
-        pass
     return args
 
 
diff --git a/tests/_plugins/_coverage/piptools_coverage.py b/tests/_plugins/_coverage/piptools_coverage.py
index b6067aed..55ee0089 100644
--- a/tests/_plugins/_coverage/piptools_coverage.py
+++ b/tests/_plugins/_coverage/piptools_coverage.py
@@ -98,6 +98,12 @@ def get_pip_major_minor() -> tuple[int, int]:
     return int(major_str), int(minor_str)
 
 
+# unlike the exclude pragmas below, this doesn't depend on the currently installed
+# pip version: any line carrying one of our pragmas has an intentionally-unbalanced
+# branch, regardless of which arm is excluded in a given test run
+PARTIAL_BRANCH_PIP_VERSION_PRAGMA = r"# pragma: pip(<=?|>=?|==)\d+\.\d+ (no )?cover\b"
+
+
 def compute_pip_version_exclude_pragmas() -> set[str]:
     current_major, current_minor = get_pip_major_minor()
 
@@ -123,6 +129,20 @@ class PipVersionPragmas(CoveragePlugin):  # type: ignore[misc]
         )
         config.set_option("report:exclude_lines", sorted(exclude_lines))
 
+        # Tell coverage that an `if` carrying one of our pragmas is allowed to have
+        # an untaken branch, so a bare `if` (no `else`) doesn't need a dummy
+        # `else: pass` arm just to host a matching "no cover" pragma.
+        #
+        # NOTE: this must merge into "report:partial_branches", not the newer
+        # "report:partial_also". `partial_also` is folded into the effective
+        # partial-branches list by `config.post_process()`, which coverage.py
+        # runs *before* plugin `configure()` hooks fire -- setting it here would
+        # be silently ignored.
+        partial_branches = set(config.get_option("report:partial_branches") or ()) | {
+            PARTIAL_BRANCH_PIP_VERSION_PRAGMA
+        }
+        config.set_option("report:partial_branches", sorted(partial_branches))
+
 
 def coverage_init(registry: Plugins, options: dict[str, object]) -> None:
     registry.add_configurer(PipVersionPragmas())
diff --git a/tests/conftest.py b/tests/conftest.py
index a0cd9ffb..2369377c 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -218,8 +218,6 @@ def from_line():
             options = kwargs.pop("options", {})
             options["hashes"] = hash_options
             kwargs["options"] = options
-        else:  # pragma: pip<=23.0 no cover
-            pass
         return install_req_from_line(*args, **kwargs)
 
     return _from_line
diff --git a/tests/test_circular_imports.py b/tests/test_circular_imports.py
index 2affff05..ccf59ddf 100644
--- a/tests/test_circular_imports.py
+++ b/tests/test_circular_imports.py
@@ -76,8 +76,6 @@ def _allowed_deprecation_warning_filters() -> list[str]:
         flags.extend(
             ("-W", "ignore:pkg_resources is deprecated as an API.:DeprecationWarning:")
         )
-    else:  # pragma: pip<25.3 no cover
-        pass
 
     if _pip_api.PIP_VERSION_MAJOR_MINOR <= (22, 2):  # pragma: pip<=22.2 cover
         flags.extend(
@@ -95,8 +93,6 @@ def _allowed_deprecation_warning_filters() -> list[str]:
                 ),
             )
         )
-    else:  # pragma: pip<=22.2 no cover
-        pass
 
     return flags
 
diff --git a/tests/test_local_coverage_plugin.py b/tests/test_local_coverage_plugin.py
index 28841d2b..5c2c12e0 100644
--- a/tests/test_local_coverage_plugin.py
+++ b/tests/test_local_coverage_plugin.py
@@ -9,6 +9,7 @@ plugin itself, since it is imported and run before coverage collection begins.
 from __future__ import annotations
 
 import datetime
+import re
 from unittest import mock
 
 import pytest
@@ -130,3 +131,50 @@ def test_computed_pragmas_state_no_cover_below_previous_major_version():
     assert nocover_eq_pragma not in computed_pragmas
     assert cover_le_pragma in computed_pragmas
     assert nocover_le_pragma not in computed_pragmas
+
+
+@pytest.mark.parametrize(
+    "pragma_comment",
+    (
+        "# pragma: pip>=26.0 cover",
+        "# pragma: pip>=26.0 no cover",
+        "# pragma: pip<25.3 cover",
+        "# pragma: pip<25.3 no cover",
+        "# pragma: pip<=23.0 cover",
+        "# pragma: pip<=23.0 no cover",
+        "# pragma: pip==25.3 no cover",
+    ),
+)
+def test_partial_branch_pragma_matches_pip_version_pragmas(pragma_comment):
+    assert re.search(
+        piptools_coverage.PARTIAL_BRANCH_PIP_VERSION_PRAGMA, pragma_comment
+    )
+
+
+def test_partial_branch_pragma_does_not_match_unrelated_pragmas():
+    assert not re.search(
+        piptools_coverage.PARTIAL_BRANCH_PIP_VERSION_PRAGMA, "# pragma: no cover"
+    )
+    assert not re.search(
+        piptools_coverage.PARTIAL_BRANCH_PIP_VERSION_PRAGMA, "# pragma: >=3.9 cover"
+    )
+
+
+def test_configure_extends_existing_partial_branches_setting():
+    mock_config = mock.Mock()
+    mock_config.get_option.side_effect = lambda name: {
+        "report:exclude_lines": [],
+        "report:partial_branches": ["pragma: no branch"],
+    }[name]
+
+    piptools_coverage.PipVersionPragmas().configure(mock_config)
+
+    partial_branches_calls = [
+        call
+        for call in mock_config.set_option.call_args_list
+        if call.args[0] == "report:partial_branches"
+    ]
+    assert len(partial_branches_calls) == 1
+    partial_branches_value = partial_branches_calls[0].args[1]
+    assert "pragma: no branch" in partial_branches_value
+    assert piptools_coverage.PARTIAL_BRANCH_PIP_VERSION_PRAGMA in partial_branches_value
```

@sirosen maybe try applying this once you rebase and see if there's anthing useful in here? Some robot-generated test bits seem dumb but the plugin updates look somewhat interesting at glance. Although, I haven't looked too closely, I'm trying to do my rounds across PRs needing attention..

sirosen:
I'm about to rebase, cleanup any conflicts, and I'll try to see if there's some way of using partial-branch coverage options. However, if I can't find a working setup based on this setting, I don't want to block this PR looking for a perfect solution.

I'm very much happy to try improvements, but if nothing works, I will push for us to merge the current, working, plugin, and continue to allow for future iterations on this topic.
pip-tools CI is still much too unstable, and at present codecov is a contributor to that instability.

sirosen:
I was able to use the partial-branches configuration (introduced in a co-authored commit).

I had remembered this changeset as achieving 100% coverage, but after a rebase and minor fixes, there are still some gaps. I prefer to go after those in follow-up PRs if this moves forward.

~Windows CI runs appear to be failing on network errors. I'm hoping this is something which can be resolved by rerunning these jobs.~ Update: resolved by rerunning.

.coveragerc
```diff
@@ -1,15 +1,15 @@
...
-
-# custom excludes extend baselines from coverage + covdefaults
-# for reference, see
-# coverage: https://coverage.readthedocs.io/en/latest/excluding.html#default-exclusions
-# covdefaults: https://github.com/asottile/covdefaults/blob/8d8712c26ad505f1269851a7c8431b8c1fedaa62/covdefaults.py#L86-L105
 exclude_also =
+    ^\s*@pytest\.mark\.xfail
     ^\s*if _t\.TYPE_CHECKING:
+include = piptools/*, tests/*
+fail_under = 99
```
webknjaz, line 15: Can we keep these in the same place above w/o reordering? Looks like the values haven't changed? Should we try setting `fail_under` to 100 or deleting it since `covdefaults` does that internally?

tests/test_sync.py
```diff
@@ -183,7 +183,7 @@ def test_diff_should_not_uninstall(fake_dist):
     # on Python 3.12 and above, `ensurepip` and `venv` do not default to installing
     # 'setuptools' -- as such, `pip` changes behavior on many Python 3.12 environments
     # to use isolated builds
-    if sys.version_info < (3, 12):
+    if sys.version_info < (3, 12):  # pragma: <3.12 cover
```
webknjaz, line 186: Isn't this something `covdefaults` already does under the hood w/o a pragma comment?

tests/test_cli_compile.py
```diff
@@ -1942,9 +1946,9 @@ def test_forwarded_args_filter_deprecated(PyPIRepository, runner, pip_args):
 
     (first_posarg, *_tail_args), _kwargs = PyPIRepository.call_args
 
-    if _pip_api.PIP_VERSION_MAJOR_MINOR >= (25, 3):  # pragma: >=3.9 cover
+    if _pip_api.PIP_VERSION_MAJOR_MINOR >= (25, 3):  # pragma: pip>=25.3 cover
```
webknjaz, line 1949: Can we arrive at automatically handling this pattern now?
```suggestion
    if _pip_api.PIP_VERSION_MAJOR_MINOR >= (25, 3):
```

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,146 @@
...
+
+    result: set[str] = {
+        rf"# pragma: pip=={current_major}.{current_minor} no cover\b",
+    }
+
+    for major, minor in list_supported_pip_versions():
+        for opname, opfunc in COMPARATORS:
+            comparator_applies = opfunc((current_major, current_minor), (major, minor))
+            designator = "no cover" if comparator_applies else "cover"
+            result.add(rf"# pragma: pip{opname}{major}.{minor} {designator}\b")
```
webknjaz, line 119: Would something like this get the job done for the cases that match our exact version check style?
```suggestion
            result.add(rf"# pragma: pip{opname}{major}.{minor} {designator}\b")
            if not comparator_applies:
                result.add(rf"if\s*_pip_api\.PIP_VERSION_MAJOR_MINOR\s*{opname} \({major},\s*{minor}\):\b")
```
(but not the else-branch, of course)

tests/_plugins/_coverage/piptools_coverage.py
```diff
@@ -0,0 +1,146 @@
...
+            # allow us to specify coverage only for the max or min versions of pip
+            if major == min_major:
+                result.add(rf"# pragma: pip{opname}MIN.{minor} {designator}\b")
+            elif major == max_major:
+                result.add(rf"# pragma: pip{opname}MAX.{minor} {designator}\b")
+
+    return result
+
+
+# mypy flags CoveragePlugin as being of type Any, type ignore "subclassing Any"
```
webknjaz: Can we consider the local fake stubs approach instead?
* https://github.com/ansible/awx_plugins.interfaces/tree/b52a9f8095330a94daa4adcafab24d2ed8770b94/_type_stubs
* https://github.com/ansible/awx_plugins.interfaces/blob/b52a9f8095330a94daa4adcafab24d2ed8770b94/.mypy.ini#L36C68-L36C104

.coveragerc
sirosen: Sorry, I'm not clear on what reordering you're referring to? Happy to tweak this config.

We can't set `fail_under = 100` (or let `covdefaults` set it) because unfortunately in the time since this PR opened things have drifted again, and we're no longer at 100%. I'll need to do it in follow-up PRs.

tests/test_sync.py
sirosen: I don't believe so: I don't see evidence of it looking at the `covdefaults` source.

tests/test_cli_compile.py
sirosen: We don't have support for this in the plugin yet, and I very strongly do not want to block this PR on developing such support. I tried and it didn't work correctly.

I want to try again and explore this, but only after this initial body of work lands. If we get this in, I can start getting us to 100% in small PRs, and I can work on improving the plugin in parallel with that effort.

tests/_plugins/_coverage/piptools_coverage.py
sirosen: This looks very similar to what I tried.

I feel certain that the regex was correct, but that there was some unexpected nuance that tripped me up. I bet the fix is something small but subtle. As noted above, I want to try again, but not as a blocker.

tests/_plugins/_coverage/piptools_coverage.py
sirosen: This prompted me to look into whether or not there are `coverage` types. Turns out, `coverage` is publishing a `py.typed` but our `mypy` runs don't have it installed. Installing it fixes this, so I'm working through the type-checking issues this flags. (None are runtime issues.)

sirosen: This is done in https://github.com/jazzband/pip-tools/pull/2347/changes/73603e7084ff00078e805626f8de8418b21a1533 .

(I'm going to have to rerun some of the CI jobs. It looks like GitHub is having network stability issues again.)

tests/test_cli_compile.py
webknjaz: I sorta wanted to make the diff in the files outside the plugin itself smaller, hence the desire to have this from the beginning.

----

Are you saying we're not at 100% with this now? Could you expand?

tests/test_sync.py
webknjaz: Ah, I guess that was just wishful thinking.

tests/_plugins/_coverage/piptools_coverage.py
webknjaz: ack. marking as resolved.

tests/_plugins/_coverage/piptools_coverage.py
webknjaz: When I was doing this in-browser, in the back of my mind, I was quite unsure about whether it's `if not comparator_applies:` or `if comparator_applies:`, not just the regex, tbh. However, I do feel that this is an important feature.

I'll send Claude out to investigate and if it comes back with something relatively tolerable, we can consider including it in the scope of this PR, otherwise, we'll postpone and declare it a follow-up only research.

tests/_plugins/_coverage/piptools_coverage.py
webknjaz: Never mind. I guess I was just throwing ideas around while doing the review passes.  I'll mark this as resolved.

tests/_plugins/_coverage/piptools_coverage.py
webknjaz: Claude pointed out a few observations I'd missed (points generated, followed by my commentary):

1. **Misses half the call sites.** Modules inside `piptools/_internal/_pip_api/` (`cli_options.py`, `command_methods.py`, `package_finder.py`, `install_requirements.py`) alias the value as `_pip_version`, not `_pip_api`, to avoid circular imports. The suggested regex only matches `_pip_api.`, so it'd silently skip most of the sites this PR annotates.

   *(we could probably figure this out by doing `_pip_api = _pip_version` in there short-term; Claude also lost some suggestions it'd made after I'd asked it to be concise, but it was worried about regexes being fragile and insisted that checking the AST in the plugin would be more robust but I dunno how'd that work)*

2. **Not every match should be excluded.** `compile.py:304` has no pragma because it's already fully exercised via `mock.patch.object(_pip_api, "PIP_VERSION_MAJOR_MINOR", ...)` in `test_cli_compile.py`. A blanket text match can't tell "untestable on this pip version" apart from "already tested via mocking," and would quietly exclude coverage on a branch we're deliberately testing.

   *(not sure how I feel about this one — @sirosen you might want to actually add a pragma there if we're not including smart autoexclusions into the plugin; seems like a good catch to me)*

3. **No symmetric `else:` handling**. `package_finder.py`/`pypi.py` pragma their `else:` clauses too, and a bare `else:` has no version info to regex-match — so we'd end up with implicit `if`s next to explicit `else`s, which is more confusing than the status quo.

   *(I'd be fine with this, I think Claude is worrying too much here)*

4. It's also more fragile/invisible than a comment — any reformat (alias rename, wrapped parens, reflow) silently stops matching with no error, which is basically how the `_pip_api`/`_pip_version` split in (1) happened in the first place.

  *(Claude's probably wrong here because with 100% target enforced, the CI would sound the alarm, and we could still fall back for the explicit pragma syntax, I think)*

---

@sirosen all in all, I think you're right to push back and request this to be postponed to a follow-up stage.

piptools/scripts/_deprecations.py
```diff
@@ -17,9 +17,14 @@ def filter_deprecated_pip_args(args: list[str]) -> list[str]:
...
     - ``--build-option``
     """
-    if _pip_api.PIP_VERSION_MAJOR_MINOR < (25, 3):  # pragma: <3.9 cover
-        return args
+    if _pip_api.PIP_VERSION_MAJOR_MINOR >= (25, 3):  # pragma: pip>=25.3 cover
+        args = _filter_args_deprecated_in_pip_25_3(args)
+    return args
 
+
+def _filter_args_deprecated_in_pip_25_3(
```
webknjaz, line 25: (not *entirely* sure whether this *needed* to be in **this** PR, FTR, but whateva)

webknjaz (approved):
Now's better than never 😜

https://github.com/jazzband/pip-tools/pull/2347#discussion_r3784568277

tests/test_cli_compile.py
sirosen: We are not. I have several new PRs up to close specific coverage gaps. They're all very small and should be really easy and pleasant to review! 😁
There will be a few more, when I get more time for the project.

I'm not sure if we were at 100% when I wrote this. I thought we were, but looking at where the gaps are, I don't think that's possible.

There are also *new* untested lines, which will continue to creep in if we don't try to push to 100%.

> I sorta wanted to make the diff in the files outside the plugin itself smaller, hence the desire to have this from the beginning.

I understand, and I did try, but I don't want to keep on trying this and let it block the current solution which I consider already good.

tests/_plugins/_coverage/piptools_coverage.py
sirosen: The mock based test case is a particularly interesting one. I'm not sure if we should have such a test. It came from an outside contributor, and I let it in with some comments about it being an interesting surprise to me. It's sort of clever, but with this plugin I might prefer to put the pragmas on the target code? I haven't revisited it yet, and I'm not sure what I think. 

When we talk about cases like line wrapping, my feeling is that it's better to have the comment than it is to make the plugin super complicated. But maybe it would result in a lot of non-uniformity over time, as lines wrap and the comments get added.

We can refer back to this if we decide to work on improvements to the plugin in the future! 🚀

webknjaz: Yep, that's why I thought I'd share Claude's output — some of its points touch areas that aren't immediately obvious or discussed explicitly.

changelog.d/2264.contrib.md
webknjaz, line 1: @sirosen looks like we forgot to symlink the PR number to this?

PR #2367, opened 2026-04-03
Add better self-testing of the local coverage plugin and meta-coverage reporting

sirosen:
*This is a draft on top of #2347; see that PR for more context.*

Add a separate tox environment and CI job for doing the testing of the coverage
plugin itself.

Because the plugin uses a separate coverage config for this test, both it and
normal coverage can be maintained at 100%, and codecov can aggregate the two.

---

Because this is a draft right now, I'm not clear on whether or not it should get
a separate changelog? I feel like it should fold into the one for #2347.

##### Contributor checklist

- [ ] A change note is created in `changelog.d/` (see [`changelog.d/README.md`]

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,7 +8,4 @@ codecov:
 
 comment: false # avoid spamming reviews
 
-ignore:
-  - "tests/_plugins/_coverage/*"
-
 ...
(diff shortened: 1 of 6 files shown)


sirosen:
I'm closing and opening fresh because the zizmor alerts are mostly stale, but there doesn't seem to be a way to clear them "due to rebase" and ask it to run afresh.

PR #2371, opened 2026-04-04
docs: add shell completion instructions to README

### LINKED ISSUE #1702: [advice] pip-{compile,sync} have completions, why not add some document about it to README.md?

```sh
```

tysoncung:
## Problem

`pip-compile` and `pip-sync` support shell completion via Click, but this isn't documented anywhere in the README. Users have no way to discover this feature.

Closes #1702

## Solution

Added a "Shell completion" section to the README with setup instructions for Bash, Zsh, and Fish shells, following Click's shell completion documentation.

## Changes

- `README.md`: Added "Shell completion" section with instructions for Bash, Zsh, and Fish

### DIFF
--- a/README.md
+++ b/README.md
@@ -682,6 +682,47 @@ This will be indicated in the output with one of the following suffixes:
 - `(pyproject.toml::build-system.backend::sdist)`
 - `(pyproject.toml::build-system.backend::wheel)`
 
+### Shell completion
+
+`pip-compile` and `pip-sync` support shell completion via
+[Click's shell completion](https://click.palletsprojects.com/en/stable/shell-completion/).
+To enable it, you need to register the completion for your shell.
+
+#### Bash
+
+```console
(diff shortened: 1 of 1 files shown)


sirosen:
Thanks for writing this up; these instructions look good. I want to double-check that they render in docs correctly before approving.

Please restore the PR template -- per our contributing doc, we do not generally accept contributions which remove the template. You should also add a doc changelog item.

WhyNotHugo:
Thanks for mentioning this, I had no idea this feature existing.

Added shell completions for downstream Alpine package, might serve as reference for other packagers: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/100135

sirosen:
It appears that this was an automated PR, which is a direct violation of project policy. I'm closing.

I would genuinely like to see shell completion written down in docs, and I'm getting started on a significant doc refactor. So we'll likely get this eventually. It's just unfortunate that rather than a genuine desire to contribute, what we have here is slop.

PR #2376, opened 2026-04-15
Give pypy jobs a 15m timeout (was 9m)

sirosen:
Trivial CI tweak, so I've dropped the template. No changelog for this.


webknjaz (approved):
Personally, I like tighter timeouts but this is fine I guess..

sirosen:
I like the timeout to be close to the projected real runtime, but I think that the harm of going too low is worse, in this case, than the harm of going a bit too high.

We can ratchet it tighter in the future, but I value our human time on the project very highly. If we have to re-run CI even a few times in a week, those delays add up.

Probably the best way to decide these things is to script out harvesting runtime data from past runs. I may try my hand at this sometime.

PR #2380, opened 2026-05-01
✨ feat(lock): add pip-lock command for PEP 751 pylock.toml

### LINKED ISSUE #2124: Implement PEP 751 "A file format to list Python dependencies for installation reproducibility" (a.k.a. lock files | `pylock.toml` | `pylock.*.toml`)

Proper ecosystem-standard lock files are getting traction yet again and we have to keep up. This will allow us to work with more ecosystem tooling than just pip, uniformly covering more use cases. And personally, I'm just thrilled to see this PEP implemented across the Python Packaging ecosystem.

Implementation of whatever's possible, including both lock file types. The feature should be marked as experimental until [PEP 751](https://peps.python.org/pep-0751/) is marked as accepted. But it should give people opportunity to try it out and perhaps feed the feedback back into the upstream discussion.

#### Alternative Solutions

Not doing anything..?

#### Additional context

This is a follow-up for #1526 that @brettcannon opened years ago.

The initial discussion is @ https://discuss.python.org/t/lock-files-again-but-this-time-w-sdists/46593.

The current open discussion is @ https://discuss.python.org/t/pep-751-lock-files-again/59173.

gaborbernat:
PEP 751 standardizes a single lockfile format that any compliant Python installer can consume. 🔒 The classic pip-tools output captures pinned versions, but it does not record the dist-file hashes, sizes, upload times, source URLs, or VCS commits a non-pip installer needs to reproduce a build. Without those fields, projects that adopt the new format have to choose between losing pip-tools' resolver or maintaining two parallel lockfiles. This branch adds a third top-level command, `pip-lock`, that produces a spec-conformant `pylock.toml` from the same authoring inputs that drive `pip-compile`, plus PEP 735 dependency-group parsing introduced for the new command. `pip-compile` and `pip-sync` keep their previous behavior; broader PEP 735 support inside those commands stays tracked under #2062.

## What we are trying to achieve

A typical Python project today targets several platforms (Linux x86_64, Linux aarch64, Windows AMD64, macOS arm64) across several Python versions (3.10 through 3.13), with several extras (`dev`, `docs`, `gpu`), dependency groups (`test`, `lint`, `ci`), and at times multiple interpreter implementations (`cpython`, `pypy`, `graalpy`). The default `pip-lock` invocation should hand back a single byte-stable lockfile that captures every variant the project supports, in roughly the time a developer is willing to wait. CI should be able to ask "is this lockfile up to date?" without rewriting any files. Mutually exclusive extras (a CPU and a GPU build of the same library, for example) should be expressible without forcing the project to maintain a separate lock per variant. The disjointness rule the spec mandates should never silently fail closed; a collision must surface a witness environment plus a remediation hint.

That is the design space. Two structural problems show up the moment you try to hit those targets at realistic scale.

## Why generating PEP 751 lockfiles is hard

The first problem is correctness across platforms. A graph like `pywin32 ; sys_platform == 'win32'` only surfaces during a Windows resolution. A discovery scan that resolves once for the host would miss every platform-conditional dependency, while N parallel per-platform resolutions blow up cold-start time and duplicate work for graphs that happen to converge.

The second problem is the disjointness check the spec requires for same-name `[[packages]]` entries. Two entries for the same package must use markers that no environment plus opt-in combination can satisfy simultaneously. The naive verifier iterates `|envs| × 2^|extras| × 2^|groups|` combinations. Real projects that lock with `--all-extras --all-groups` across N platforms hit millions of combinations before finishing. The naive check hangs on inputs the user expects to work.

A correct implementation has to solve both without forcing the user to scale back what they ask for.

## How the resolver scales: marker-driven cohorting

```mermaid
flowchart LR
  A[Authoring inputs<br/>pyproject groups, requirements files, args]:::inputs --> B[Build target environments]:::inputs
  B --> C[Discovery scan<br/>platform markers blinded to True]:::scan
  C --> D[Walk every platform-conditional<br/>dependency in one pass]:::scan
  D --> E[Partition target envs into cohorts<br/>that share a dependency closure]:::partition
  E --> F[Resolve once per cohort<br/>parallel worker processes]:::resolve
  F --> G[Merge variants<br/>verify marker disjointness]:::validate
  G --> H[Emit byte-stable pylock.toml]:::output
  classDef inputs fill:#dbeafe,stroke:#1e40af,color:#1e3a8a;
  classDef scan fill:#dcfce7,stroke:#166534,color:#14532d;
  classDef partition fill:#fef3c7,stroke:#92400e,color:#78350f;
  classDef resolve fill:#fde68a,stroke:#a16207,color:#713f12;
  classDef validate fill:#fed7aa,stroke:#c2410c,color:#7c2d12;
  classDef output fill:#ede9fe,stroke:#6d28d9,color:#4c1d95;
```

The first stage rewrites the marker evaluator so platform comparisons evaluate to `True` for the duration of one resolution. ✨ A single scan walks every platform-conditional dependency, including transitive ones the host machine would never see. Python-version and implementation comparisons stay honored, which preserves the partitioner's ability to keep python-conditional and pypy-conditional cohorts apart from cpython-conditional ones. The scan trades correctness on each individual platform for completeness across all of them, and the next stage spends per-cohort resolutions to recover correctness.

Target environments then collapse into cohorts that share a dependency closure. The cohort key comes from the marker-driven equivalence classes the discovery scan revealed: two environments collapse iff every package the scan reached selects the same marker disjunction for both. Seventeen platforms with identical resolutions cost one resolver invocation, not seventeen. The cohorts resolve in parallel under a process pool that uses the spawn start method, so worker processes do not inherit a forked SSL session from the parent (an intermittent source of network failures on Linux where a parent and child reuse the same TLS context).

## How disjointness becomes tractable

```mermaid
flowchart LR
  M[Marker AST]:::input --> S{Recognized shape?<br/>opt-in disjunction AND env clause}:::decision
  S -- yes --> L[Linear sweep over env axis<br/>extras / groups handled symbolically]:::fast
  S -- no --> P[Bounded powerset<br/>conflict-aware skip]:::fallback
  L --> R[Disjoint or witness collision]:::result
  P --> R
  classDef input fill:#dbeafe,stroke:#1e40af,color:#1e3a8a;
  classDef decision fill:#fef3c7,stroke:#92400e,color:#78350f;
  classDef fast fill:#dcfce7,stroke:#166534,color:#14532d;
  classDef fallback fill:#fed7aa,stroke:#c2410c,color:#7c2d12;
  classDef result fill:#ede9fe,stroke:#6d28d9,color:#4c1d95;
```

`pip-lock` emits markers in a constrained shape: an opt-in disjunction (extras, groups) joined by `and` to an environment clause (platform, Python version, implementation). The validator decomposes the marker AST and exploits that shape. Extras and groups never make these markers disjoint on their own (a user can always request the union), so disjointness reduces to a linear sweep over the environment axis. The bounded powerset stays as a safety net for marker shapes the symbolic decomposer cannot recognize, with a configurable iteration ceiling so a pathological input fails loudly rather than hangs. The validator is conflict-aware: the validator skips subsets that violate a declared `[tool.pip-tools].conflicts` group because the user can never request that combination.

When two same-name entries collide on a real install, the error names the witness environment, lists the offending extras and groups, and points at the conflicts declaration when the overlap is something the project can declare away. The failure mode is actionable; the user always knows what to change.

## Why we depend on `packaging >= 26.1`

PEP 751 reading, writing, and disjointness validation rely on `packaging.pylock`, which only ships from `packaging` 26.1. PEP 735 group resolution leans on `packaging.dependency_groups` for the same reason. Both modules come from the top-level `packaging` distribution, which the branch pins via the runtime requirement so the modules are present regardless of which pip version is in use.

Every other routine API (`canonicalize_name`, `Version`, `Marker`, `Requirement`, `SpecifierSet`, `Tag`) imports from `pip._vendor.packaging`. Pip's resolver and `InstallRequirement` constructor `isinstance`-check against vendored types, so a top-level `Marker` or `Requirement` passed into pip code can silently fail at the type guard. Holding the vendored convention here protects the boundary; the only places where the top-level distribution remains the source of truth are the two modules the vendored copy on the lowest-supported pip lacks.

## Three pip-helper caches keep cold-start cost reasonable

The parallel worker fan-out exposes three pip hot paths that re-do the same work on every cohort. The branch ships memoizing wrappers for all three, scoped to one `pip-lock` invocation.

A session-scoped finder cache lets each cohort reuse the configured pip session rather than re-reading the user's pip configuration on every resolution. A per-package metadata cache shares JSON-API hits between the resolver and the dist-file collector so the same release is fetched once and used many times; an in-memory `Distribution` wrapper decouples the parsed metadata from the resolver's temp-dir lifetime so cached entries survive across cohorts. An LRU-bounded wheel-filename parser rebinds across every `from X import Y` consumer in the running process (a rebind sweep, since `Wheel.__init__` would otherwise keep calling the unwrapped copy), so pip's hot path inside the wheel installer benefits from the cache too. The LRU bound caps the working set, so a long-lived `--jobs auto` worker cannot grow it without limit.

A separate on-disk hash cache shortens the second-and-later run for content-addressable URLs (`*.pythonhosted.org`). Private indexes that re-publish the same URL with different bytes fall outside the host allow-list, so a stale digest never serves from cache.

## Reproducibility and parallelism

The lockfile is byte-stable across runs given the same inputs. Listing order, hash ordering, and field ordering are deterministic, so diffs in CI carry signal rather than churn. Wheels sort newest-Python-first inside each package entry; the `sdist` block follows wheels in TOML emission; `.tar.gz` wins over `.zip` when an index ships both; a second sdist for the same release logs at WARNING and names the dropped artifact.

The provenance block records the pip-tools version, the pip version, the command line, the generated-at timestamp, and the resolved target environments. `pip-lock` honors `SOURCE_DATE_EPOCH` for `generated_at` so reproducible-build pipelines can pin the timestamp without surrendering the rest of the provenance block, and every field stays selectively suppressible for projects that want frozen byte equality across machines.

The download directory swaps stale roots through an atomic rename before cleanup, so a concurrent worker never observes a half-deleted tree.

## CI integration and drift detection

CI generally wants to verify that the lockfile is up to date without writing any files. `--check` re-resolves in memory and exits non-zero when the result differs from the on-disk lockfile. `--dry-run` streams the rendered TOML to stdout for preview. The two compose with every other flag, so a project can validate any combination (universal, restricted to a subset of platforms, with or without extras and groups) on any branch.

For projects that adopt the format gradually, the existing `pip-compile` and `pip-sync` workflows continue to work unchanged. `pip-lock` runs alongside them, reading the same authoring inputs.

## CLI surface and behavioral impact

The new command takes target axes via `--platform`, `--python-version`, and `--implementation`. The third axis lets a project that ships both CPython and PyPy wheels lock each implementation's resolution without splitting the lockfile. Default groups declared in `pyproject.toml` flow into the top-level `default-groups` key in the lockfile, so installers that respect the key auto-include them. Find-links artifacts that happen to live on a configured index URL stay attributed to that index in the lockfile rather than collapsing to a URL-only entry, because the index URL carries audit-relevant context that the bare URL drops.

`pip-compile` and `pip-sync` keep their previous behavior. The minimum supported pip moves to `22.3`, the lowest version whose simple-API content-type negotiation matches modern PyPI. New runtime dependencies cover the PEP 751 model and validator (`packaging >= 26.1, < 27`), TOML round-tripping that preserves `decimal.Decimal` (`tomli-w >= 1.2.0`), and the test suite (`pytest-mock >= 3.15.1`). The CLI help wraps at 120 columns so the option matrix reads on a modern terminal without the historical 80-column truncation.

---

## AI usage disclosure

Parts of this branch were drafted with Claude Opus 4.7 under human review. Every commit was inspected by the maintainer before landing, and every change here was validated against the project's existing test suite plus an end-to-end cross-check against `uv`'s `pylock.toml` output on a real-world target. Reviewers are encouraged to read the diff carefully and to flag anything that looks off; the goal is to land a feature the community can trust, not to short-circuit review.

---

Closes [#2124](https://github.com/jazzband/pip-tools/issues/2124) and partially addresses [#2062](https://github.com/jazzband/pip-tools/issues/2062) (PEP 735 dependency groups are read by `pip-lock` only; `pip-compile` and `pip-sync` coverage stays open under that issue).

Upstream pip performance optimizations that benefit `pip-lock`'s resolver hot path: [pypa/pip#13986](https://github.com/pypa/pip/pull/13986) skips the URL roundtrip when the link's path is already safe; [pypa/pip#13987](https://github.com/pypa/pip/pull/13987) skips the fragment regex when no fragment marker is present.

Help-wanted thread on Python Discourse for cross-implementation testing: <https://discuss.python.org/t/help-me-shake-out-pip-lock/107232>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -429,6 +429,35 @@ jobs:
             VM-${{ matrix.runner-vm }},
             Py-${{ steps.python-install.outputs.python-version }},
             Pip-${{ matrix.pip-version }}
+      - name: Rename coverage data file
+        # ``actions/download-artifact@v4`` with ``merge-multiple: true`` would
+        # overwrite the bare ``.coverage`` from every matrix slot at the
+        # combine step, leaving the combine job a single file (which
+        # ``coverage combine`` reports as "No data to combine"). Rename to a
+        # unique suffix per slot so the merged directory carries one file
+        # per slot and the combiner sees them all.
+        if: >-
+          !cancelled()
+          && !inputs.cpython-pip-version
+          && runner.os == 'Linux'
+        env:
+          RUNNER_VM: ${{ matrix.runner-vm }}
+          PY_VERSION: ${{ matrix.python-version }}
+          PIP_VERSION: ${{ matrix.pip-version }}
+        run: mv .coverage ".coverage.${RUNNER_VM}-${PY_VERSION}-${PIP_VERSION}"
+        shell: bash
+      - name: Upload coverage data
+        if: >-
+          !cancelled()
+          && !inputs.cpython-pip-version
+          && runner.os == 'Linux'
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4
+        with:
+          include-hidden-files: true
+          name: >-
+            coverage-data-${{ matrix.runner-vm }}-${{ matrix.python-version }}-${{ matrix.pip-version }}
+          path: .coverage.*
+          retention-days: 3
 
   pypy:
     name: ${{ matrix.runner-vm }} / ${{ matrix.python-version }} / ${{ matrix.pip-version }}
@@ -512,18 +541,53 @@ jobs:
           && exit 1
         shell: bash
 
-  coverage-summary:
-    name: Coverage processing
+  coverage-combine:
+    name: 📊 Combine coverage
     if: >-
       !cancelled()
     runs-on: ubuntu-24.04
-    timeout-minutes: 1
+    timeout-minutes: 5
     needs:
+      - pre-setup
       - test
     steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          # The combine job only reads the workspace; without this, ``actions/checkout``
+          # stashes a credential in ``.git/config`` that subsequent steps could leak
+          # into uploaded artifacts.
+          persist-credentials: false
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
+        with:
+          python-version: "3.14"
+      - run: python -m pip install tox
+      - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093  # v4
+        with:
+          pattern: coverage-data-*
+          merge-multiple: true
+      - run: tox -e coverage-combine
+      # Surface the combined coverage report on the workflow run summary so
+      # PR reviewers can see line coverage at a glance without downloading
+      # the html artifact. ``coverage report --format=markdown`` is what the
+      # combine env emits; capture it again here and route to the summary.
+      - name: Publish coverage to step summary
+        if: >-
+          !cancelled()
+        run: |
+          {
+            echo "## Coverage report"
+            echo ""
+            .tox/coverage-combine/bin/python -m coverage report --format=markdown
+          } >> "$GITHUB_STEP_SUMMARY"
+        shell: bash
+      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4
+        with:
+          name: html-coverage-report
+          path: htmlcov
       - name: Notify Codecov that all coverage reports have been uploaded
         if: >-
           !cancelled()
+          && fromJSON(needs.pre-setup.outputs.is-upstream-repository)
         uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2  # v6.0.0
         with:
           fail_ci_if_error: true
@@ -545,6 +609,7 @@ jobs:
       - pypy
       - test
       - zizmor
+      - coverage-combine
 
     runs-on: ubuntu-24.04
 
@@ -554,4 +619,5 @@ jobs:
       - name: Decide whether the needed jobs succeeded or failed
         uses: re-actors/alls-green@05ac9388f0aebcb5727afa17fcccfecd6f8ec5fe  # v1.2.2
         with:
+          allowed-skips: coverage-combine
           jobs: ${{ toJSON(needs) }}
--- a/piptools/_internal/_pip_caches.py
+++ b/piptools/_internal/_pip_caches.py
@@ -1,19 +1,19 @@
 """Process-wide caches for pip helper functions hot during pylock.
 
-Pylock iterates the same package set across multiple resolution cohorts, but pip
-is built for long-running sessions and re-parses index responses and wheel
-filenames each time through by design. This module installs memoizing wrappers
-that short-circuit those repeats for the duration of a pip-tools command.
+Pylock iterates the same package set across multiple resolution cohorts, but pip is built for
+long-running sessions and re-parses index responses and wheel filenames each time through by
+design. This module installs memoizing wrappers that short-circuit those repeats for the
+duration of a pip-tools command.
 
-For ``parse_wheel_filename`` we also walk ``sys.modules`` to rebind every module
-that imported it at module load time. ``from X import Y`` binds ``Y`` as a name
-in the importer, not as a reference into ``X``, so patching the source module
-alone would leave pip's ``Wheel.__init__`` calling the unwrapped copy.
+For ``parse_wheel_filename`` we also walk ``sys.modules`` to rebind every module that imported
+it at module load time. ``from X import Y`` binds ``Y`` as a name in the importer, not as a
+reference into ``X``, so patching the source module alone would leave pip's ``Wheel.__init__``
+calling the unwrapped copy.
 
-Memory profile: the URL-keyed dicts grow with the number of distinct (link,
-metadata, index-page) URLs the resolver touches and are bounded by the
-``scope()`` lifetime. The wheel-filename cache caps at a configurable bound so
-long-lived ``--jobs auto`` workers cannot grow the working set without limit.
+Memory profile: the URL-keyed dicts grow with the number of distinct (link, metadata, index-
+page) URLs the resolver touches and are bounded by the ``scope()`` lifetime. The wheel-filename
+cache caps at a configurable bound so long-lived ``--jobs auto`` workers cannot grow the working
+set without limit.
 """
 
 from __future__ import annotations
@@ -39,12 +39,10 @@ from ..logging import log
 class _InMemoryImportlibDistribution(_importlib_md.Distribution):
     """Decouples a parsed ``Distribution`` from the on-disk temp dir.
 
-    Pip's stock backend wraps a ``PathDistribution`` that re-reads
-    ``METADATA`` on every property access, against a temp dir whose
-    lifetime ends with one resolver invocation. Caching that wrapper
-    across passes raises ``FileNotFoundError`` on later reads. Reading
-    from bytes held on the instance lets the wrapper survive across
-    passes within a single lock command.
+    Pip's stock backend wraps a ``PathDistribution`` that re-reads ``METADATA`` on every property
+    access, against a temp dir whose lifetime ends with one resolver invocation. Caching that
+    wrapper across passes raises ``FileNotFoundError`` on later reads. Reading from bytes held on
+    the instance lets the wrapper survive across passes within a single lock command.
     """
 
     def __init__(self, metadata_bytes: bytes) -> None:
@@ -60,9 +58,8 @@ class _InMemoryImportlibDistribution(_importlib_md.Distribution):
 
     @property
     def name(self) -> str:
-        # ``Distribution.name`` is a property from Python 3.10 on; on 3.9
-        # pip's compat layer reaches for ``dist.name`` directly and crashes
-        # without this fallback.
+        # ``Distribution.name`` is a property from Python 3.10 on; on 3.9 pip's compat layer
+        # reaches for ``dist.name`` directly and crashes without this fallback.
         return _t.cast(str, self.metadata.get("Name"))
 
     @property
@@ -94,20 +91,18 @@ if _t.TYPE_CHECKING:
     )
 
 
-# Cache backing stores. ``Final`` marks the *binding* immutable; the dict
-# contents are populated and cleared in-place. Caching the wrapped
-# Distribution alongside the bytes amortises ``email.feedparser`` across the
-# per-cohort, per-extra, and per-group resolver invocations a single
-# ``pip-lock`` command fans out (the bytes alone do not, since pip's
-# importlib backend re-parses on every property access).
+# Cache backing stores. ``Final`` marks the *binding* immutable; the dict contents are populated
+# and cleared in-place. Caching the wrapped Distribution alongside the bytes amortises
+# ``email.feedparser`` across the per-cohort, per-extra, and per-group resolver invocations a
+# single ``pip-lock`` command fans out (the bytes alone do not, since pip's importlib backend
+# re-parses on every property access).
 _parsed_links_by_url: _t.Final[dict[str, list[Link]]] = {}
 _index_content_by_url: _t.Final[dict[str, IndexContent | None]] = {}
 _metadata_bytes_by_url: _t.Final[dict[str, bytes | None]] = {}
 _metadata_dist_by_url: _t.Final[dict[str, _t.Any]] = {}
-# Originals captured at import time so we swap them in and out around an
-# installed scope. ``_fetch_metadata_using_link_data_attr`` ships in pip 22.3+
-# which is our declared lower bound, so the method is present and the wrapper
-# installs.
+# Originals captured at import time so we swap them in and out around an installed scope.
+# ``_fetch_metadata_using_link_data_attr`` ships in pip 22.3+ which is our declared lower bound,
+# so the method is present and the wrapper installs.
 _original_parse_links: _t.Final[_ParseLinks] = _pkgfinder.parse_links
 _original_parse_wheel_filename: _t.Final[_ParseWheelFilename] = (
     _packaging_utils.parse_wheel_filename
@@ -117,10 +112,9 @@ _original_fetch_metadata_using_link_data_attr: _t.Final[_FetchMetadata] = (
     _RequirementPreparer._fetch_metadata_using_link_data_attr
 )
 _installed = False
-# RLock so an outer scope's install/clear/uninstall serializes with any
-# concurrent thread's view of the patches. Without it two threads using
-# pip-tools as a library race the install flag and produce one half-patched,
-# one half-cleared session.
+# RLock so an outer scope's install/clear/uninstall serializes with any concurrent thread's view
+# of the patches. Without it two threads using pip-tools as a library race the install flag and
+# produce one half-patched, one half-cleared session.
 _scope_lock: _t.Final[threading.RLock] = threading.RLock()
 _scope_depth = 0
 
@@ -129,10 +123,9 @@ _scope_depth = 0
 def scope() -> Iterator[None]:
     """Install pip helper caches for the duration of the block.
 
-    Reverts and clears the caches on exit. Nested entries are idempotent:
-    the outermost entry installs and uninstalls; inner entries are no-ops.
-    Thread-safe via an ``RLock`` and a depth counter so in-process library
-    callers can spawn parallel locks safely.
+    Reverts and clears the caches on exit. Nested entries are idempotent: the outermost entry
+    installs and uninstalls; inner entries are no-ops. Thread-safe via an ``RLock`` and a depth
+    counter so in-process library callers can spawn parallel locks safely.
 
     :returns: Context manager that installs and reverts the pip helper caches.
     """
@@ -154,9 +147,8 @@ def scope() -> Iterator[None]:
 def install() -> None:
     """Patch pip's helper functions process-wide with memoising wrappers.
 
-    Idempotent. Prefer the symmetric scope context for in-process callers;
-    this entry point serves worker processes that exit on pool teardown
-    without running cleanup.
+    Idempotent. Prefer the symmetric scope context for in-process callers; this entry point
+    serves worker processes that exit on pool teardown without running cleanup.
     """
     global _installed
     if _installed:
@@ -208,32 +200,28 @@ def uninstall() -> None:
 
 
 def _cached_parse_links(page: IndexContent) -> list[Link]:
-    cached = _parsed_links_by_url.get(page.url)
-    if cached is None:
+    if (cached := _parsed_links_by_url.get(page.url)) is None:
         cached = list(_original_parse_links(page))
         _parsed_links_by_url[page.url] = cached
     return cached
 
 
-# Bound the wheel-filename cache so a high cardinality of distinct wheel
-# filenames cannot push the working set past a fixed ceiling. Under
-# ``scope()`` the cache drops on exit, so the bound protects long-lived
-# workers in ``--jobs auto``. ``OrderedDict`` gives LRU semantics whose
-# lifetime ``clear()`` controls. ``functools.lru_cache`` would store its
-# data on the function object and survive module re-imports in ways hard
-# to reason about under multi-process pools. Override via
-# ``PIP_TOOLS_PARSED_WHEEL_FILENAME_BOUND`` to lift the ceiling when the
-# resolver sees more distinct filenames than the default holds.
+# Bound the wheel-filename cache so a high cardinality of distinct wheel filenames cannot push
+# the working set past a fixed ceiling. Under ``scope()`` the cache drops on exit, so the bound
+# protects long-lived workers in ``--jobs auto``. ``OrderedDict`` gives LRU semantics whose
+# lifetime ``clear()`` controls. ``functools.lru_cache`` would store its data on the function
+# object and survive module re-imports in ways hard to reason about under multi-process pools.
+# Override via ``PIP_TOOLS_PARSED_WHEEL_FILENAME_BOUND`` to lift the ceiling when the resolver
+# sees more distinct filenames than the default holds.
 _PARSED_WHEEL_FILENAME_BOUND = int(
     os.environ.get("PIP_TOOLS_PARSED_WHEEL_FILENAME_BOUND", "10000")
 )
 _parsed_wheel_filename_cache: _t.Final[
     OrderedDict[str, tuple[NormalizedName, Version, BuildTag, frozenset[Tag]]]
 ] = OrderedDict()
-# When the LRU evicts more than ~1% of capacity in a single resolution pass,
-# every evicted-then-re-requested filename re-pays the parse cost. Surface a
-# one-time hint so the user can lift the bound; otherwise the signal is
-# "this lock is slow" with no pointer to the knob.
+# When the LRU evicts more than ~1% of capacity in a single resolution pass, every evicted-then-
+# re-requested filename re-pays the parse cost. Surface a one-time hint so the user can lift the
+# bound; otherwise the signal is "this lock is slow" with no pointer to the knob.
 _EVICTION_WARN_THRESHOLD: _t.Final = max(1, _PARSED_WHEEL_FILENAME_BOUND // 100)
 _parsed_wheel_filename_evictions = 0
 _eviction_warning_emitted = False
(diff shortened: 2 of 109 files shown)


sirosen:
Thank you so much for putting this together!

I don't know when I'll have the time to dive deep (it might have to wait until during or after PyCon), but this is extremely exciting to me and I'm looking forward to diving in on all of this. The high-level architecture and doing multiple resolution passes seems like an approach we can maintain even as `packaging` and `pip` continue to evolve.

Marker manipulation logic may get easier in future `packaging` versions (there's WIP to expose the markers as a tree), and some other things will become available over time like `packaging.dependency_groups`. Lots to think about!

gaborbernat:
I'll iterate a bit more until I'll make it ready for review next week. But testing it on the few projects I have around seems to work fine, and matches with what I added it https://github.com/astral-sh/uv/pull/14728 so in theory equivalent with a potential uv implementation (minus this is a bit slower :D).

webknjaz:
> from the inputs pip-tools already understands: PEP 735 dependency groups in `pyproject.toml`

This is definitely not true and needs a standalone PR with an implementation: #2062.

gaborbernat:
> This is definitely not true and needs a standalone PR with an implementation: #2062.

I meant for pip-lock, updated the formulation in my PR description text.

piptools/pylock/_marker_ast.py
```diff
@@ -0,0 +1,357 @@
...
+            if sub_env:
+                # An ``or`` group that mixes extras with env clauses is outside
+                # pip-tools' emitted shape; refuse to decompose so the caller
+                # falls back to the powerset.
+                if op_seen == "or" and (sub_extras or sub_groups):
+                    return False
+                if not env_nodes:
+                    env_nodes.append(node)
+                else:
+                    env_nodes.extend([op_seen or "and", node])
```
thatch: I don't think we can get here without `op_seen` set to something (because env_nodes contains something, that means we're on an even index 2+, and we would have saved (or returned) on index 1).  Is this level of defensiveness something that should be pervasive?

piptools/pylock/_marker_ast.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+def _walk_top_level(
+    nodes: list[_t.Any],
+    extras: set[str],
+    groups: set[str],
+    env_nodes: list[_t.Any],
+    *,
+    top: bool,
+) -> bool:
```
thatch: High-level comment: it took a while to realize that the return value is inverted (well, approximately) from `_has_mixed_or_axes` -- given the generic "walk" name it might be worth giving the meaning of this return bool (and maybe even flipping one of these so you don't need the inversion)

piptools/pylock/_marker_ast.py
```diff
@@ -0,0 +1,357 @@
...
+    # Strip nested top-level wrappers so ``Marker("(extra == 'a')")`` decomposes
+    # the same way as the unwrapped form.
+    while top and len(nodes) == 1 and isinstance(nodes[0], list):
```
thatch: I think this is safe regardless of `top`?

piptools/pylock/_marker_ast.py
```diff
@@ -0,0 +1,357 @@
...
+    if not env_nodes:
+        return MarkerShape(frozenset(extras), frozenset(groups), None)
+    env_marker = Marker.__new__(Marker)
+    env_marker._markers = env_nodes
+    return MarkerShape(frozenset(extras), frozenset(groups), env_marker)
+
+
+def _walk_top_level(
+    nodes: list[_t.Any],
+    extras: set[str],
```
thatch: it also took a while to grok that these sets are modified -- are they meaningful only for a True return value?

piptools/pylock/_marker_ast.py
```diff
@@ -0,0 +1,357 @@
...
+                if op_seen == "or" and (sub_extras or sub_groups):
+                    return False
+                if not env_nodes:
+                    env_nodes.append(node)
+                else:
+                    env_nodes.extend([op_seen or "and", node])
+            extras |= sub_extras
+            groups |= sub_groups
+            continue
+        if not isinstance(node, tuple):
```
thatch: It seems like you're trying to detect malformed ast and not raise, but there seem to be many places where that could still happen (like a short tuple).  The control flow here is almost like the False return _should_ have been an exception (since they just bubble up manually), and True doesn't carry any explicit information.  Am I missing some larger coding style?

piptools/pylock/_marker_ast.py
gaborbernat: On another read, you're right and switched to your proposal.

piptools/pylock/_marker_ast.py
gaborbernat: On another read, you're right and switched to your proposal.

piptools/pylock/_marker_ast.py
gaborbernat: On another read, you're right and switched to your proposal.

piptools/pylock/_marker_ast.py
gaborbernat: On another read, you're right and switched to your proposal.

piptools/pylock/_marker_ast.py
gaborbernat: On another read, you're right and switched to your proposal.

piptools/pylock/_merge.py
```diff
@@ -0,0 +1,309 @@
...
+        return
+    negations: list[str] = []
+    negations.extend(f"'{e}' not in extras" for e in sorted(excluded_extras))
+    negations.extend(f"'{g}' not in dependency_groups" for g in sorted(excluded_groups))
+    parts: list[str] = []
+    if base.marker:
+        # Wrap the existing marker if it contains an ``or`` so the trailing
+        # ``and`` doesn't bind to a single disjunct and quietly change the
+        # marker's truth set.
+        parts.append(f"({base.marker})" if " or " in base.marker else base.marker)
```
thatch: I don't think the whitespace is a required part of the 508 grammar -- has it been normalized by this point?

.github/workflows/ci.yml
```diff
@@ -554,4 +619,5 @@ jobs:
       - name: Decide whether the needed jobs succeeded or failed
         uses: re-actors/alls-green@05ac9388f0aebcb5727afa17fcccfecd6f8ec5fe  # v1.2.2
         with:
+          allowed-skips: coverage-combine
           jobs: ${{ toJSON(needs) }}
```
sirosen, line 623: I'm starting to look for pieces that we can break off of this PR, so that we can get the size down to something more digestible.

Is it valid to pull this CI change out into a dedicated PR? (I think so, unless I missed something.) If we can turn this into a stream of small changes, we can do a much more rapid-fire approve+merge cycle.

piptools/_compat/pip_compat.py
```diff
@@ -213,11 +213,14 @@ def create_wheel_cache(cache_dir: str, format_control: str | None = None) -> Whe
...
+    try:
+        from pip._internal.commands.freeze import _dev_pkgs
+    except ImportError:
         from pip._internal.commands.freeze import DEV_PKGS
 
         return _t.cast(set[str], DEV_PKGS)
-
-    from pip._internal.commands.freeze import _dev_pkgs
-
     return _t.cast(set[str], _dev_pkgs())
```
sirosen, line 223: This change catches my eye. The previous code was correct, but I guess some test broke it? But that makes me want to look at what test needed this to change.

FWIW, I'm sympathetic to the view that the `try-except` is more robust and therefore better. Sometimes the version checks turn out to be way easier for us, but this doesn't seem like such a case. So I'm okay with it, but it's attention-grabbing for me.

piptools/_compat/pip_compat.py
```diff
@@ -34,7 +34,7 @@
     def canonicalize_name(name: str) -> str: ...
 
 else:
-    from pip._vendor.packaging.utils import canonicalize_name  # noqa: F401
+    from packaging.utils import canonicalize_name  # noqa: F401
```
sirosen: I'd like us to stick with `pip._vendor.packaging` where possible, until/unless we make an intentional decision to switch to using `packaging` directly from the surrounding environment.

My standing worry, as we've considered pulling in `packaging` before, has been about what happens if `pip` is using its vendored copy and it's ever meaningfully different from the unvendored copy. e.g., If we were to pass a `packaging.markers.Marker` into `pip` code using `pip._vendor.packaging.markers.Marker`, I would expect things could go wrong.

I think `canonicalize_name` is pretty safe, since the type is `str`... Maybe this case is okay, but I at least want to be really careful about this.

---

I'm aware that for new features (`dependency_groups`, `pylock`), there's no choice but to use the unvendored copy. My note here only applies to the APIs which are available in the vendored copy.

piptools/_internal/_pip_api/install_requirements.py
```diff
@@ -3,10 +3,10 @@
 import copy
 import typing as _t
 
+from packaging.utils import canonicalize_name
+from packaging.version import Version
 from pip._internal.req import InstallRequirement
 from pip._internal.req.constructors import install_req_from_line
-from pip._vendor.packaging.utils import canonicalize_name
-from pip._vendor.packaging.version import Version
```
sirosen, line 9: As above, let's stick with `pip._vendor.packaging` for now? I'm amenable to switching in a separate PR (it may be a prereq to this? if so, that would have bearing), but don't want to mix it in here.

I'll refrain from noting the various other cases of this.

piptools/_internal/_pip_caches.py
```diff
@@ -0,0 +1,372 @@
...
+        if getattr(mod, attr, None) is original:
+            setattr(mod, attr, replacement)
+
+
+__all__ = [
+    "clear",
+    "install",
+    "scope",
+    "uninstall",
+]
```
sirosen, line 350: I recognize that this module is mostly needed for `pip-lock`, but is there any part of it we could benefit from under `pip-compile` today? (Again, looking for pieces we can break off for smaller review chunks.)

piptools/pylock/cli/_commands.py
```diff
@@ -0,0 +1,397 @@
...
+    if not environ.get("PIP_TOOLS_HIDE_EXPERIMENTAL_WARNING"):
+        # Marking pip-lock as experimental gives the CLI room to evolve while
+        # users start trying it; mirror what pip 26.1 does for ``pip lock``.
+        # The env-var escape lets CI suppress the noise once the team has
+        # acknowledged the contract is fluid.
+        log.warning(
+            "pip-lock is experimental: the CLI and the [tool.pip-tools] block "
+            "may change between releases. "
+            "Set PIP_TOOLS_HIDE_EXPERIMENTAL_WARNING=1 to silence."
+        )
```
sirosen, line 185: 👍 I like having this warning, with the opt-out. No notes. 😁

piptools/pylock/cli/_targets.py
```diff
@@ -0,0 +1,255 @@
...
+        try:
+            resolved = resolver.resolve(group_name)
+        except ExceptionGroup as eg:
+            raise BadParameter(
+                "; ".join(str(e) for e in eg.exceptions), param_hint="--group"
+            ) from eg
+        group_constraints[group_name] = [
+            _pip_api.create_install_requirement_from_line(str(req)) for req in resolved
+        ]
+    return group_constraints, groups
```
sirosen, line 76: This handler seems good, and now that `packaging.dependency_groups` is available, I feel like the ingredients are all here.

Can we lift this into a reformulated #2175 ? e.g., I could add a variant of this + a requirement for `packaging >= 26.1` into a co-authored commit on #2175 and then if we can get that merged, we'd have a smaller body of work here.

piptools/pylock/_marker_ast.py
```diff
@@ -0,0 +1,357 @@
...
+) -> bool:
+    # Strip nested top-level wrappers so ``Marker("(extra == 'a')")`` decomposes
+    # the same way as the unwrapped form.
+    while top and len(nodes) == 1 and isinstance(nodes[0], list):
+        nodes = nodes[0]
+    # An ``or`` joining an extras opt-in with an env clause cannot collapse to
+    # ``(extras_in) and (env)``: the original fires when *either* side is true,
+    # while the collapsed form needs *both*. Detect mixed-axis ``or`` at this
+    # level and refuse to decompose so the caller falls back to the powerset.
+    if not top and _has_mixed_or_axes(nodes):
```
thatch: Seems like you handle both cases of `top` here, was there supposed to be another conditional or can these be combined?

piptools/pylock/redact.py
```diff
@@ -0,0 +1,69 @@
...
+        if "://" in arg:
+            cleaned.append(redact_auth_from_url(arg))
+            continue
+        cleaned.append(arg)
+    return [_CONTROL_CHARS.sub("�", arg) for arg in cleaned]
+
+
+__all__ = [
+    "redact_command",
+]
```
sirosen, line 66: Is there some generic version of this which we could use here and for `pip-compile` headers? It feels like the same work done twice. (Yet another possible chance to pull something forward/separate it out.)

It would probably need to be customizable since `pip-compile` doesn't care about `tomli_w` requirements.

piptools/repositories/pypi.py
```diff
@@ -24,26 +28,34 @@
...
     lookup_table,
 )
+from . import _hash_cache
 from .base import BaseRepository
 
-FILE_CHUNK_SIZE = 4096
+# Override via ``PIP_TOOLS_FILE_CHUNK_SIZE`` to tune for high-bandwidth
+# environments where a 4 KiB read is too small (chunks per second x syscall
+# overhead dominates throughput).
+FILE_CHUNK_SIZE = int(os.environ.get("PIP_TOOLS_FILE_CHUNK_SIZE", "4096"))
```
sirosen: I feel like I could approve a separate feature PR for this in seconds. 😀 

I'm looking for every last line we can possibly pull forward.

piptools/repositories/pypi.py
```diff
@@ -104,7 +116,20 @@ def __init__(self, pip_args: list[str], cache_dir: str):
...
+        # lookup either still sees the old dir or sees ``no dir`` cleanly,
+        # never a half-deleted tree. The temp is then rmtree'd opportunistically.
+        if not os.path.exists(self._download_dir):
+            return
+        stale = f"{self._download_dir}.stale-{os.getpid()}"
+        try:
+            os.replace(self._download_dir, stale)
+        except OSError:
+            return
+        rmtree(stale, ignore_errors=True)
```
sirosen, line 128: This is also a bugfix for parallel invocations of `pip-compile`, isn't it? (another to pull into a separate PR)

piptools/repositories/local.py
```diff
@@ -100,6 +100,14 @@ def get_hashes(self, ireq: InstallRequirement) -> set[str]:
...
         return self.repository.get_hashes(ireq)
 
+    def get_distribution_files(
+        self, ireq: InstallRequirement
+    ) -> list[PackageWheel | PackageSdist]:
+        return self.repository.get_distribution_files(ireq)
+
+    def get_requires_python(self, ireq: InstallRequirement) -> str | None:
+        return self.repository.get_requires_python(ireq)
+
```
sirosen, line 110: Even though the changes in this module wouldn't be used in `main` today, they are sufficiently harmless that I'd be happy with them as things to pull forward, with basic unit tests.

piptools/scripts/compile.py
```diff
@@ -136,8 +136,9 @@ def _determine_linesep(
 @options.emit_index_url
 @options.emit_options
 @options.unsafe_package
-@options.config
+# Order matters; see ``override_defaults_from_config_file``.
 @options.no_config
+@options.config
```
sirosen, line 139: Separable bugfix?

piptools/scripts/sync.py
```diff
@@ -66,8 +66,9 @@
 @options.client_cert
 @options.src_files
 @options.pip_args
-@options.config
+# Order matters; see ``override_defaults_from_config_file``.
 @options.no_config
+@options.config
```
sirosen, line 69: As in pip-compile, I think we can pull this forward?

piptools/sync.py
```diff
@@ -146,16 +146,27 @@ def diff_key_from_ireq(ireq: InstallRequirement) -> str:
...
+    the supported pip range.
+    """
+    info = getattr(direct_url, "archive_info", None)
+    if info is None:
+        info = getattr(direct_url, "info", None)
+    if not isinstance(info, ArchiveInfo):
+        return False
+    if hashes := getattr(info, "hashes", None):
+        return bool(hashes)
+    return bool(getattr(info, "hash", None))
```
sirosen: This modification seems simply like an improvement over what we had before? So another one to separate into a new PR?

sirosen (commented):
I focused my review time on the source tree; I think tests can mostly follow from that.

Right now this is just too big. I don't have a great methodology for reviewing a PR of this size and complexity. As a first step, I'd like us to trim parts out as much as possible, pulling them into distinct bugfix and feature changes.

A lot of the minor items here seem like they're standing bugs currently in `main`! So that's awesome, in my book (not the having of bugs, but the knowing of bugs 😉 ). Even if the work goes no further, we've got a number of things we can fix. I've tried to flag as many as I can as things I'd like to see in standalone PRs. That should reduce the size of the source changes a bit, and we can start to focus on the main parts of `pip-lock`.

piptools/pylock/_merge.py
gaborbernat: You're right — the substring check encoded an undocumented producer assumption. PEP 508 grammar allows zero whitespace around `and` / `or`, so a marker round-tripped through any non-canonical formatter would slip past `" or " in base.marker`, and the substring also produces false positives on parenthesised `or` (`(A or B) and C`) where the wrap is unnecessary.

Replaced with a `has_top_level_or` helper in `_marker_ast.py` that inspects the parsed marker's AST in `d0293b3`. Centralising the inspection in `_marker_ast` keeps all `Marker._markers` private-API contact in one file, matching the principle behind the earlier marker-AST refactor.

piptools/pylock/_marker_ast.py
gaborbernat: You're right that both branches handled the same case identically — the duplicate was dead code (both returned `False`).

Already fixed in `955495b`: collapsed to a single unconditional `if _has_mixed_or_axes(nodes): return None`, and the `top` parameter was dropped entirely (the wrapper-strip while-loop above is correctness-preserving at any depth, which was the broader observation in your earlier review round). GitHub's flagging this comment Outdated because the exact substring no longer exists in the file — `git log -S 'if not top and _has_mixed_or_axes'` confirms the removal in that commit. Just confirming the fix landed before the comment surfaced.

gaborbernat:
> Right now this is just too big. I don't have a great methodology for reviewing a PR of this size and complexity. As a first step, I'd like us to trim parts out as much as possible, pulling them into distinct bugfix and feature changes.

I feel you on this. This is a big feature, with complicated stories to it. Unless we want to iterate it over 10 releases there's no better way than a chunkier PR. My strategy was to keep as much as possible of the new code in separate files, and touch as little as possible the old one. My personal strategy on reviewing features of this size is to focus on the functionality and high level architecture, we can always come back and change parts of code as follow-up and refactors once we know (including with user feedback) that what we have works well in practice. That way the benefit is upfront and significant and there's no churn of small PRs that take months to fully ingest. 

I'm very wary that trying to break this up is simply more work than worth it. Perhaps, that's just me.

piptools/_internal/_pip_api/install_requirements.py
gaborbernat: > As above, let's stick with `pip._vendor.packaging` for now? I'm amenable to switching in a separate PR (it may be a prereq to this? if so, that would have bearing), but don't want to mix it in here.

Ough, some of these are performance optimizations 🤔 so backing out stuff like this will make the feature work slower.

piptools/_compat/pip_compat.py
gaborbernat: > My standing worry, as we've considered pulling in `packaging` before, has been about what happens if `pip` is using its vendored copy and it's ever meaningfully different from the unvendored copy. e.g., If we were to pass a `packaging.markers.Marker` into `pip` code using `pip._vendor.packaging.markers.Marker`, I would expect things could go wrong.

I wouldn't worry about this too much, we have a type checker and a CI, that both would flag such incompatibilities.

piptools/_internal/_pip_caches.py
gaborbernat: > I recognize that this module is mostly needed for `pip-lock`, but is there any part of it we could benefit from under `pip-compile` today? (Again, looking for pieces we can break off for smaller review chunks.)

Dunno, didn't really focus on that part at all, felt orthogonal to this feature, and purposefully wanted to keep this unrelated as much as possible to the rest.

piptools/repositories/pypi.py
gaborbernat: Perhaps, but was really added to address a pip-lock bug 😆

piptools/_internal/_pip_api/install_requirements.py
sirosen: Okay, that makes sense. packaging did just get a lot faster.

I still want to try to address our use of packaging outside of this PR, to lay the groundwork for this.

Can we maybe try a change to use (unvendored) packaging everywhere, in a dedicated PR, and see what breaks? If everything works on our lowest pip version, we could get this all sorted out as a prereq. And if there's only some specific case where we have to use the vendored copy, we could understand and document it.

gaborbernat:
Pivoting to use https://github.com/notatallshaw/nab

Issue #2012, opened 2023-10-31
pip-compile should accept special files (pipes, fifo, pty, sockets, ...) as output files


manuco:
When running `pip-compile -o /dev/stdout`, the command blocks indefinitely. It shouldn't and should write the requirement file in the given file, even if it's not a bare file. 

Please note that `pip-compile | <other program>` doesn't work either (the other program doesn't receive anything).

Please also note that `stderr` is cluttered by some eventual warnings not parseable by pip and is so not usable as a viable content source.

#### Environment Versions

1. > Linux VirtualBox 5.4.0-165-generic # 182-Ubuntu SMP Mon Oct 2 19:43:28 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
1. > Python version: Python 3.8.10
1. > pip version: pip 23.3.1
1. > pip-tools version: pip-compile, version 7.3.0

#### Steps to replicate

1. `git clone <your favourite "using setup.py" project> && cd in`
2. `pip-compile setup.py -o /dev/stdout`
3. nothing happens

#### Expected result

output on stdout (**not stderr**)
warnings (if any) on stderr

#### Actual result
\<nothing and process blocked>

#### Steps to replicate (advanced)
TTY 1 :
1. `mkfifo /tmp/mylovelyfifo`
2. `cat /tmp/mylovelyfifo`

TTY 2 :
1. `git clone <your favourite "using setup.py" project> && cd in`
2. `pip-compile setup.py -o /tmp/mylovelyfifo`

#### Expected result

output on TTY 1
warnings (if any) on TTY 2

#### Actual result
\<nothing and process blocked>

manuco:
The main problem is that `pip compile` try to **read** the **output file** in order to find previous requirement and update it. `pip compile` should ensure that previous requirements is only read from a **real file** (with [stat.S_ISREG(mode)](https://docs.python.org/3/library/stat.html)), but should not trust that reading from this file won't block.

For example, setting `/dev/stdout` as an output file will make `pip compile` try to read this file. And it won't generate any error because `/dev/stdout` as a link to your terminal (`/dev/pts/*n*`) is also readable (since `/dev/pts/*n*` is also stdin and stderr). But it won't give any data until you type on your keyboard (which is not what's expected).

manuco:
Also the documentation should emphasize that `-` is a legit output filename redirecting it to `stdout`.

chrysle:
[...]

PR #2382, opened 2026-05-07
📝 docs: restructure under Diátaxis

### LINKED ISSUE #1129: Document explicitly how to deal with system packages

I have a Dockerfile in a project.  On that project I do development sometimes in a virtualenv, sometimes in docker, according to whatever is convenient for what I'm up to.  The Dockerfile provides some python packages as system packages (e.g. debian or alpine packages), rather than as PyPI dependencies or VCS links (to github etc.).  There may be reasons for that -- e.g. production stability is thought to be better with a particular system package, build dependencies are heavy/tricky in the chosen docker environment, reducing docker build time, etc.  I assume most system packages do include egg metadata (I haven't checked).

When I'm working with a virtualenv, I want to use the same python package versions as I have in the docker image.  So, some packages should not be installed by pip in the docker image (so that the system packages are used), but should be installed in the virtualenv (so that the packages are present, with the same version as used in docker).  Also, versions that pip-compile selects should be compatible with what is installed by the system packages in the docker image.

I'm not clear from just the pip-tools docs what, if any, are the recommended solutions to this.

I imagine there might be multiple ways to deal with this, and the solution might just be documentation.

In particular, I'm not sure from just the docs whether pip-compile is supposed to see that a package is already installed and, in the absence of something more explicit in requirements.in, take that as a kind of implicit declaration that we're going to use that version?

If so, perhaps I just need to run pip-compile in the docker environment, and then use the resulting requirements.txt to pip-sync both in the Dockerfile and when installing dependencies in a virtualenv?

#### Alternative Solutions

Perhaps there are other solutions involving multiple requirements files.  I won't think through that further here because since I don't know the intent of pip-tools, and haven't tested the actual behaviour either, I don't know whether it's necessary!

### LINKED ISSUE #1314: docs update to explain how to add a dependency from a git repo commit

There does not appear to be any documentation that explains how to add a dependency based on a github commit url

Adding this to the README.md by someone who knows what the format in `requirements.in` should look like, should be enough to address this.

### LINKED ISSUE #1524: Improve Readme to explain how pip-compile is better than pip freeze
It would be interesting to explain the README the benefit of pip-compile vs. doing a pip install in a fresh environment followed by a pip freeze to dump out the pinned dependency versions.

Is it faster? In what usage patterns does it save time?

### LINKED ISSUE #1528: Document that pyproject.toml dependencies are supported

Issues like: #1510, general confusion

At the moment, the docs include phrases like:

> [The pip-compile command lets you compile a requirements.txt file from your dependencies, specified in either setup.py or requirements.in.](https://pip-tools.readthedocs.io/en/latest/#example-usage-for-pip-compile)

> [pip-compile generates a requirements.txt file using the latest versions that fulfil the dependencies of setup.py or requirements.in.](https://github.com/jazzband/pip-tools#updating-requirements)

To me, and I assume the person who opened #1510, this reads like only setup.py or requirements.in are supported.

### LINKED ISSUE #1702: [advice] pip-{compile,sync} have completions, why not add some document about it to README.md?

### LINKED ISSUE #1794: pip-tools is missing a lot of practical documentation
Coming from having experience with `poetry` and `bundler` from Ruby, I'm finding it frustrating to get started with `pip-tools`. As far as I can tell, the only docs available are:
 * https://pip-tools.readthedocs.io/en/stable/
 * https://github.com/jazzband/pip-tools/blob/main/README.rst

 A synopsis of my experience follows.

#### Environment Versions

1. OS Type: OSX
1. Python version: `Python 3.11.1`
1. pip version: `pip 22.3.1`
1. pip-tools version: `pip-compile, version 6.12.1`

#### Steps to replicate

I started out trying to use a `pyproject.toml`, as I heard `pip-tools` now supports it. I wasn't even able to get a minimal working project up and running:

```toml
[project]
name = "sandbox"
# dependencies = ["more-itertools"]
```

Which fails if the last line is uncommented, with no guidance as what to do:
image

Adopting a `pyproject.toml` generated via poetry doesn't work either.

I was more successful with `requirements.in`, but have been unable to find mentions of or ways to do:
 * separate prod, test, dev dependency constraints
 * basic constraint operators, e.g.
   * poetry explaining things like `^`: https://python-poetry.org/docs/dependency-specification/
   * bundler explaining things like `~>`: https://bundler.io/guides/gemfile.html

#### Expected result

...

#### Actual result

...

### LINKED ISSUE #1845: Refactor README

### LINKED ISSUE #2044: Backtracking tries to download all versions of package
Let's assume, i have already this `requirements.txt` file:

```
urllib3==2.1.0
```

Now i add `boto3` to `requirements.in`:

```
urllib3
boto3
```

Now command `pip-compile --resolver=backtracking -v requirements.in` tries to download every version of package or every metadata file (if i am using pip servers directly, not caching servers on my local network):

```
pip-compile --resolver=backtracking -v requirements.in 
Using indexes:
  https://pypi.org/simple

                          ROUND 1                           
  Collecting urllib3 (from -r requirements.in (line 1))
    Obtaining dependency information for urllib3 from https://files.pythonhosted.org/packages/96/94/c31f58c7a7f470d5665935262ebd7455c7e4c7782eb525658d3dbf4b9403/urllib3-2.1.0-py3-none-any.whl.metadata
    Using cached urllib3-2.1.0-py3-none-any.whl.metadata (6.4 kB)
  Collecting boto3 (from -r requirements.in (line 2))
    Obtaining dependency information for boto3 from https://files.pythonhosted.org/packages/e3/f7/93a4ba1cd2cc4ee95f871b0890e4ed60e52365110a074e7265279750a736/boto3-1.34.18-py3-none-any.whl.metadata
    Using cached boto3-1.34.18-py3-none-any.whl.metadata (6.6 kB)
  Collecting botocore<1.35.0,>=1.34.18 (from boto3->-r requirements.in (line 2))
    Obtaining dependency information for botocore<1.35.0,>=1.34.18 from https://files.pythonhosted.org/packages/ec/79/cc5e52bfc3cf7c26ba452c348fe6a765a888730b692d783e64a175243572/botocore-1.34.18-py3-none-any.whl.metadata
    Using cached botocore-1.34.18-py3-none-any.whl.metadata (5.6 kB)
  Collecting jmespath<2.0.0,>=0.7.1 (from boto3->-r requirements.in (line 2))
    Using cached jmespath-1.0.1-py3-none-any.whl (20 kB)
  Collecting s3transfer<0.11.0,>=0.10.0 (from boto3->-r requirements.in (line 2))
    Obtaining dependency information for s3transfer<0.11.0,>=0.10.0 from https://files.pythonhosted.org/packages/12/bb/7e7912e18cd558e7880d9b58ffc57300b2c28ffba9882b3a54ba5ce3ebc4/s3transfer-0.10.0-py3-none-any.whl.metadata
    Using cached s3transfer-0.10.0-py3-none-any.whl.metadata (1.7 kB)
  Collecting python-dateutil<3.0.0,>=2.1 (from botocore<1.35.0,>=1.34.18->boto3->-r requirements.in (line 2))
    Using cached python_dateutil-2.8.2-py2.py3-none-any.whl (247 kB)
  INFO: pip is looking at multiple versions of botocore to determine which version is compatible with other requirements. This could take a while.
  Collecting boto3 (from -r requirements.in (line 2))
    Obtaining dependency information for boto3 from https://files.pythonhosted.org/packages/3b/34/0bdbd20d688438a46f8e9255e9e9a06ef350689a05e9a7233babff554978/boto3-1.34.17-py3-none-any.whl.metadata
    Using cached boto3-1.34.17-py3-none-any.whl.metadata (6.6 kB)
  Collecting botocore<1.35.0,>=1.34.17 (from boto3->-r requirements.in (line 2))
    Obtaining dependency information for botocore<1.35.0,>=1.34.17 from https://files.pythonhosted.org/packages/4a/ed/9f3ec1754d1a444997d7d18aeaef2611a3694783b6baf062c2e74627c4b3/botocore-1.34.17-py3-none-any.whl.metadata
    Using cached botocore-1.34.17-py3-none-any.whl.metadata (5.6 kB)
  Collecting boto3 (from -r requirements.in (line 2))
    Obtaining dependency information for boto3 from https://files.pythonhosted.org/packages/8b/dc/26c1c654cb6a177fc0b7ca7f916cd61daf045a42ca091fce44906d65be9f/boto3-1.34.16-py3-none-any.whl.metadata
    Using cached boto3-1.34.16-py3-none-any.whl.metadata (6.6 kB)
  Collecting botocore<1.35.0,>=1.34.16 (from boto3->-r requirements.in (line 2))
    Obtaining dependency information for botocore<1.35.0,>=1.34.16 from https://files.pythonhosted.org/packages/6d/84/36a78ba9d992baf3ed48dc0ad2bdb711d27033de0b088d71f4cfcd698bde/botocore-1.34.16-py3-none-any.whl.metadata
    Using cached botocore-1.34.16-py3-none-any.whl.metadata (5.6 kB)
...
```

In this case it takes hours of downloading.

From persective of resolver, is correct to try other versions.

I don't know if this can be solved in some smart way. I might suggest to not check older versions if installed package version is already higher than maximal version defined in botocore (in this case), but this would not be correct.

This is more discussion topic, than real bug report, because i think, `pip-tools` behaves correctly. Maybe best solution would be extend pip server metadata to include all dependencies of package on single request and correctly grouped to version ranges, so it would be not one gigantic file. Instead of this, it would contain sections like `'>=1.0,<1.2': {'dependencies': ...`.

### LINKED ISSUE #2117: Poor interaction between isolated mode and `--python-executable`
Running `pip-sync` with `--python-executable`, I was baffled as to why it was attempting to remove items from the local directory (any environment) and the user site directory (python-build-standalone environments with no virtual env present).

I eventually realised it was due to the following problem: even though `pip-sync` was being executed from a Python runtime configured to use isolated mode (no implied sys.path[0], no user site directory), the same was not true when `pip-sync` invoked the given Python executable to check the value of `sys.path`.

This means actually performing the sync in an isolated environment requires setting the relevant Python environment variables rather relying solely on the `-I` command line flag:

```
PYTHONNOUSERSITE=1
PYTHONSAFEPATH=1
```

gaborbernat:
The `README.md` is 750 lines covering install, every flag, layered files, hashes, pre-commit, cross-environment caveats, and deprecations. New users scroll past everything to find the install command; experienced users have no anchored reference. The Sphinx site duplicates the README and adds a flat CLI dump.

Restructure under [Diátaxis](https://diataxis.fr/). 📝 The `README.md` keeps a one-paragraph summary, two examples, and a link to the docs. The docs split into a tutorial, how-to recipes grouped by command, a CLI reference auto-generated by `sphinx-click` against the live `click` commands, and an explanation section. The explanation covers the resolver, caching, the local-pin proxy, build-system integration, the upcoming `--strip-extras` and `--allow-unsafe` default flips, and a comparison with `uv`, `poetry`, and `pdm`. Mermaid diagrams replace prose where a graph helps. Tabs cover platform variants and parallel approaches (`uv` vs `pipx` vs `pip`; index URL via flag, `.in` file, `pip.conf`, or env var). Admonitions replace inline `WARNING:` blocks and "Reading next" bullet lists.

Pages name only the commands that ship today, so a follow-up adding `pip-lock` for [PEP 751](https://peps.python.org/pep-0751/) lockfiles slots in as additive content. ✨ Nested toctrees with sub-captions keep the sidebar compact at the root and surface structure on expand. `.pymarkdown.yml` raises the line-length cap to 120 to match the project's prose wrap.

📖 **RTD preview**: https://pip-tools--2382.org.readthedocs.build/en/2382/

Each closure links to the page that satisfies it on the preview build, so a reviewer clicks straight through.

- Closes #1845: [the new slim `README.md`](https://github.com/gaborbernat/pip-tools/blob/docs/README.md).
- Closes #1794: [the new how-to section](https://pip-tools--2382.org.readthedocs.build/en/2382/how-to/).
- Closes #1524: [`pip-compile` vs `pip freeze` and other tools](https://pip-tools--2382.org.readthedocs.build/en/2382/explanation/comparison/).
- Closes #1528: [Compile from `pyproject.toml`](https://pip-tools--2382.org.readthedocs.build/en/2382/how-to/compile-from-pyproject/).
- Closes #1702: [Enable shell completions](https://pip-tools--2382.org.readthedocs.build/en/2382/how-to/shell-completions/).
- Closes #1314: [Pin to a Git commit, branch, or URL](https://pip-tools--2382.org.readthedocs.build/en/2382/how-to/vcs-and-url-deps/).
- Closes #1129: [Coexist with system-installed packages](https://pip-tools--2382.org.readthedocs.build/en/2382/how-to/system-packages/).
- Closes #2117: [Sync an environment that isn't yours, with the isolated-mode warning](https://pip-tools--2382.org.readthedocs.build/en/2382/how-to/sync-other-env/#caveats). The OP wrote up a discovery, not a fix request.
- Closes #2044: ["Why backtracking downloads many versions"](https://pip-tools--2382.org.readthedocs.build/en/2382/explanation/resolvers/#why-backtracking-downloads-many-versions). The OP framed the issue as a discussion topic.

Refs #2012, [the `--output-file=-` workaround](https://pip-tools--2382.org.readthedocs.build/en/2382/how-to/compile-from-stdin/#caveats). The issue asks `pip-compile` to accept `/dev/stdout` and FIFO targets. Closing it needs a code change to how `click.File(atomic=True)` opens the path, which this PR does not make. The workaround is in `compile-from-stdin`; the feature stays open.

Refs #989, [the `--allow-unsafe` deprecation plan](https://pip-tools--2382.org.readthedocs.build/en/2382/explanation/deprecations/). The issue asks to deprecate `--allow-unsafe` and flip the default. Closing it needs a code change this PR does not make. The plan is in `deprecations` and `unsafe-packages`; the issue stays open until the flip ships.


sirosen:
I'm going to read the whole thing, so it will take me a bit of time. Mostly I'm very happy with what I'm seeing.

This is so large and landed so fast, I assume it was tool assisted. Which is fine by me but means I want to be careful to read even where I would often fully trust the author. With a change this size, it's hard to calibrate review properly. I think even if there's a low rate of mistakes that slip past review (counting a presumed self-review by Bernat), this is still a sufficiently large improvement that it's a big step forward.[^1]

[^1]: I'm still not really prepared for review in the tool-assisted era when someone I generally trust (vs unknown sources) can produce a large PR so quickly. I'm not sure how we need to approach review, but I'm starting out with a commitment that I _want_ this PR to land.

There are a couple of things I think we should drop from this, based on my initial pass:

(1) Several of the diagrams are nice, but [this one in the getting started doc](https://pip-tools--2382.org.readthedocs.build/en/2382/tutorial/getting-started/) isn't very compelling for me. The text which follows strikes me as actually easier to consume and more accessible than the diagram:
> You write top-level dependencies in requirements.in. pip-compile resolves them and writes requirements.txt. pip-sync makes the venv match. To add a new package, edit requirements.in and run the loop again.

Simply swap "run the loop again" with "run `pip-compile` again" and the text stands on its own.
I think that's better? Screen readers can consume the text, and the page becomes smaller and easier to navigate.

(2) ["variables you might wish existed"](https://pip-tools--2382.org.readthedocs.build/en/2382/reference/environment-variables/#variables-you-might-wish-existed) sort of feels strange, almost like we're critiquing ourselves? I'd be happier to take or make PRs for `PIP_TOOLS_CONFIG` and `PIP_TOOLS_OUTPUT_FILE` than I would be to publish a doc that notes their non-existence.[^2]

[^2]: It's slightly OT, but I'd like every CLI flag to be settable via an environment variable _except_ those which are passed through to `pip`, since `pip` already has `PIP_*` env vars. So my opinion is that `PIP_TOOLS_CONFIG` _should_ exist.

(3) The logo. I have mixed feelings about this. We could use a project logo. And the one in the PR doesn't look half-bad. But I have friends who are artists, so if this was machine generated, it just doesn't sit right with me. If we want a logo, I'd like to do some outreach and see if I can find a designer interested in contributing or in a commission I could afford. I'd be willing to go with a machine-made logo _if_ I make an effort to get a human-made one and can't get it done (possibly by some deadline). Would you be alright with setting it aside, at least for now? If it's _not_ machine generated, please LMK, of course.

The last thing that's on my mind after a first pass is that some of these examples might prove hard to maintain over time? But, TBH, that might not be that different from the current sprawl in the readme.

These notes are a little critical, so I want to circle back and re-emphasize: as I said at the start, I'm overall quite happy with this. I was planning a pretty slow migration to a better doc layout, and haven't been able to work on it. Having this starting point for refined docs puts us much closer to my vision state.

gaborbernat:
> Would you be alright with setting it aside, at least for now? If it's _not_ machine generated, please LMK, of course.

Sure, yeah was machine generated with multiple rounds of feedback on it, I'd on a gut feeling go ahead and then replace when we get better but I'm fine dropping it for now. For reference these are some of the original brainstorm ideas I started out with: 

image

If some (human) artist can make some nice SVGs  based on one of this (or have a better idea) that would be epic!
 
Addressing the other points too.

sirosen:
Just so it's no surprise, I have decided on a way of doing this review which is a little unconventional but seems appropriate to the context.

I'd like to do a full pass over every file, making minor edits if I see any need, and push it as a new PR (preserving changelog attribution + commits) for @webknjaz to review. That will constitute
- a good forcing function to make me read everything in full
- a chance to double-check and make tweaks (limited to anything which I think is needful, not too many discretionary things)
- a good clear signal about when that full review is done (when that PR gets opened)

sirosen:
I have started trying to do this in the manner that I suggested and I am realizing increasingly that I am not going to be able to do it. There are simply too many changes which need to go through review process on a case-by-case basis here.

Every change to `docs/conf.py`, use of mermaid diagrams, introduction of custom CSS, removal of the current SVG, migrating away from `program-output`, etc. Several of these are things I am in principle in agreement with or at least accept, but trying to pick this apart into reviewable parts is turning out to be more work than just starting over again.

This has still been immensely useful as a demonstration of what the docs could look like after a significant refactor, but unfortunately it's not in a form that I feel I can use. As such, I'm using my discretion to close this, treating the hours I've put into trying to rework and pick it apart not as sunk cost but as a valuable learning experience.

A form of this will very likely resurface, slowly, in reviewable parts.

Issue #2379, opened 2026-04-27
Fix incompatibilities with pip v26.1


sirosen:
Minimally, `pip==26.1` makes tests fail under tox. Some of the issues may be active user-facing problems, and some may be internal or niche things caught by the unit tests.

Here's what I saw in my first check on python 3.14:

1. `AttributeError: 'DirectUrl' object has no attribute 'info'` hit when `pip-sync` runs with URL reqs
2. `tests/test_cli_sync.py::test_python_executable_option` does not call mocked run func

#### Steps to replicate

Update the supported version of `pip` in `tox.ini` and run tests.

ffernand:
> Some of the issues may be active user-facing problems

We're on pip 26.1 + pip-tools 7.5.3, Python 3.10, and pip-sync crashes against any venv that has a locally-installed package (i.e. anything pulled in via a `file:` requirement or through other means).

The only reason we found out was to deal with the [CVE-2026-6357](https://nvd.nist.gov/vuln/detail/CVE-2026-6357) for `pip<26.1` and tried to update pip-tools for compatibility with pip 26.x

Reproduction if it's useful to anyone...

```python-traceback
$ python -m venv /tmp/repro && /tmp/repro/bin/pip install 'pip==26.1' 'pip-tools==7.5.3'

$ mkdir -p /tmp/pkg && cat > /tmp/pkg/pyproject.toml <<'EOF'
[project]
name = "demo"
version = "0.0.1"
EOF

$ echo "file:/tmp/pkg" > /tmp/requirements.txt
$ /tmp/repro/bin/pip install -r /tmp/requirements.txt
$ /tmp/repro/bin/pip-sync /tmp/requirements.txt   # crashes

Traceback (most recent call last):
  File "/tmp/repro/bin/pip-sync", line 8, in <module>
    sys.exit(cli())
  File "/tmp/repro/lib/python3.10/site-packages/click/core.py", line 1514, in __call__
    return self.main(*args, **kwargs)
  File "/tmp/repro/lib/python3.10/site-packages/click/core.py", line 1435, in main
    rv = self.invoke(ctx)
  File "/tmp/repro/lib/python3.10/site-packages/click/core.py", line 1298, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/tmp/repro/lib/python3.10/site-packages/click/core.py", line 853, in invoke
    return callback(*args, **kwargs)
[...]

PR #2383, opened 2026-05-08
🐛 fix(sync): tolerate pip 26.1 DirectUrl shape rename

gaborbernat:
`pip>=26.1` renamed `DirectUrl.info` to `DirectUrl.archive_info` and `ArchiveInfo.hash` to `ArchiveInfo.hashes`. `pip-sync` reads both names when computing the diff key for URL-pinned distributions and crashes with `AttributeError: 'DirectUrl' object has no attribute 'info'` against the new pip. Issue #2379 tracks this as the first of the pip 26.1 regressions.

The patch probes `archive_info`/`hashes` first, falls back to `info`/`hash`, and keeps the existing `isinstance(_, ArchiveInfo)` guard. The dual-shape probe sits in a small helper to keep the diff-key call site readable.

Refs #2379.

### DIFF
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -8,10 +8,18 @@ from collections import Counter
 from unittest import mock
 
 import pytest
+from pip._internal.models.direct_url import ArchiveInfo, DirectUrl
 from pip._internal.utils.urls import path_to_url
+from pytest_mock import MockerFixture
 
 from piptools.exceptions import IncompatibleRequirements
-from piptools.sync import dependency_tree, diff, merge, sync
+from piptools.sync import (
+    _direct_url_has_archive_hash,
+    dependency_tree,
+    diff,
+    merge,
+    sync,
+)
 
 from .constants import PACKAGES_PATH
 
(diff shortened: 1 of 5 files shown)


sirosen:
This looks good (as do several of the other PRs). It's pretty late in my local time; I can get through reviews and merging the ready ones tomorrow or this weekend, I think.

Right now, this is missing pytest-mocker. I'm perfectly happy for us to add it here though.

changelog.d/2379.bugfix.md
webknjaz, line 1: We've historically put pip compat things under features: https://pip-tools.rtfd.io/en/latest/changelog/#v7-5-3

changelog.d/2379.bugfix.md
```diff
@@ -0,0 +1,4 @@
+Restored ``pip-sync`` compatibility with ``pip>=26.1`` by accepting both the
```
webknjaz: Can we link this?

tests/test_sync.py
```diff
@@ -327,6 +334,54 @@ def test_diff_with_unequal_url_hash(fake_dist, from_line):
...
 
 
+def test_direct_url_has_archive_hash_new_shape_with_hashes(mocker):
+    archive_info = mocker.Mock(spec=ArchiveInfo)
+    archive_info.hashes = {"sha256": "abc"}
+    direct_url = mocker.Mock(archive_info=archive_info)
+    assert _direct_url_has_archive_hash(direct_url) is True
+
+
+def test_direct_url_has_archive_hash_new_shape_empty_hashes(mocker):
```
webknjaz: Looks like this is the same test with different params. Use the parametrization marker instead.

webknjaz (commented):
Looks like this can be made smaller. And perhaps use real stubs instead of mocks?

webknjaz:
Also, can we avoid those conversational commit prefixes? We have a problem change log rendering integration and don't need to pollute the commit titles with those. Git log is not a change log as it serves a different purpose for a different audience.

gaborbernat:
> Also, can we avoid those conversational commit prefixes?

What do you mean specifically?

sirosen:
I think that was a typo for "conventional commit". (e.g., `fix(sync): ...`)
I don't really like the Conventional Commits style, personally, but I don't want discussion of that to detract from focus on the diffs.

---

I just checked out the branch and ran with `pip==26.1`. It looks like it should resolve #2379 .

I'd really like us to have a changelog entry which is: "support `pip` version 26.1". That's what a user cares about, much more than how we achieved compatibility. I think the changelog right now is too low-level/detailed.

For the last compat change, we put it in features and the text was/is:
> pip-tools is now compatible with pip 26.0 

To complete the story and fully close #2379, I'd like us to also bump the `pip` pin in `tox.ini` to `26.1`, so we're testing on it. I just tried this change, and it worked for me locally:
```
    pipsupported: pip == 26.0 ; python_version < "3.10"
    pipsupported: pip == 26.1 ; python_version >= "3.10"
```

@gaborbernat , could you add that here please? I very much want to get this in and do a release.

webknjaz:
Yep, that was autocorrect. I did mean "conventional".

sirosen (approved):
Thanks for putting in the final tweaks! I'll probably queue up a release during PyCon, in one of the evenings.

(As you're aware, I'm mostly offline. But I'm stealing some time here and there to do some reviews.)

PR #2384, opened 2026-05-08
Avoid half-deleted cache trees on concurrent pip-compile --rebuild

gaborbernat:
Two `pip-compile --rebuild` invocations sharing a `cache_dir` race inside `PyPIRepository.clear_caches`. Both call `shutil.rmtree(self._download_dir, ignore_errors=True)` on the same path, and the second resolver walks a tree the first started removing. `ignore_errors=True` masks the error but not the corruption.

The patch swaps the in-place delete for an atomic `os.replace` to a per-PID sibling directory, then deletes the renamed copy. `os.replace` is atomic on POSIX and NTFS, so a concurrent peer sees the original directory or no directory, never a half-deleted tree.

Scope is intentionally narrow per discussion with @sirosen below: this fixes one observable symptom of #2393, not the wider race where a process is already using the cache and a peer `--rebuild` later renames the directory out from under it. Full concurrent-rebuild safety needs locking around the resolve and is left for future work.

Refs #2393.


tests/test_repository_pypi.py
```diff
@@ -169,6 +169,41 @@ def test_pip_cache_dir_is_empty(from_line, tmpdir):
     assert not pypi_repository.options.cache_dir
 
 
+def test_clear_caches_removes_existing_download_dir(tmpdir):
```
webknjaz: Use `tmp_path` instead of the deprecated fixture.

changelog.d/+clear-caches-parallel.bugfix.md
```diff
@@ -0,0 +1,5 @@
+``pip-compile --rebuild`` is now safe to invoke concurrently against the same
```
webknjaz: Plz use roles like `:command:` for command examples.

changelog.d/+clear-caches-parallel.bugfix.md
webknjaz, line 1: The note should be linked to an issue or a PR, I don't think user-facing changes should be w/o a reference.

cc @sirosen

webknjaz (commented):
Not sure how I feel about adding more mocks. The general preference is having more integration tests..

changelog.d/+clear-caches-parallel.bugfix.md
sirosen: I don't think there's an existing issue for this one. I'm normally fine with fixes coming in without associated issues, but do prefer to have one for more intricate changes.

This is part of pulling work out of #2380 -- which I asked for (we can discuss the merits/faults of breaking it apart over there).
I want to make sure I do everything I can to make the process smooth. So I'm filing an issue right now, which we can use as a referent. :slightly_smiling_face:

sirosen: Okay, #2393 is open, we can use that.

sirosen:
Something I'm slightly unclear about is the ordering on this and safety offered. Isn't it a problem if the second process does the rename operation while the first process is using the cache directory?
Did I miss some detail which makes the collision I describe below impossible?

Since it's a concurrency issue, I put together a sequence diagram.

<details><summary>sequence diagram source</summary>

```
title \-\-rebuild collision

participant Process1
participant Process2
participant SharedCacheDir
participant Proc1CacheDir
participant Proc2CacheDir

activate SharedCacheDir
box over Process1:start
Process1->SharedCacheDir: clear_caches
SharedCacheDir->Proc1CacheDir: rename (by proc1)
deactivate SharedCacheDir
Process1->Proc1CacheDir: rmtree
Process1->SharedCacheDir: start using cache
activate SharedCacheDir
box over Process2:start
Process2->SharedCacheDir: clear_caches
SharedCacheDir-[#red]->Proc2CacheDir: rename (by proc2)
```

</details>

image

My thought is that once a process is running and using the cache directory, the second process would still move it, not being aware. We've guaranteed that concurrent removals appear to be atomic, but not that concurrent use of a shared cache directory is really safe.

gaborbernat:
You're right, the PR title oversells what the change actually delivers. The atomic rename only collapses the specific window where ``rmtree`` partially deletes a tree another resolver is reading; it does not protect a process that has already started using the freshly-cleared cache from a peer that arrives later, calls ``clear_caches`` again, and renames the directory out from under it.

The wider concurrent-``--rebuild`` story really wants a lock around the whole resolve, not just around the cache wipe. That's a bigger change than this PR scope and I'd rather not bolt it on here.

Three options I see, happy to take direction:

1. **Re-scope this PR** to just "make ``rmtree`` atomic so fork CI / test runs cannot trip over partial trees", retitle accordingly, and open a separate issue/PR for proper file-based locking around the whole resolve.
2. **Close this** and bundle the proper fix into the locking work later.
3. **Expand here** to add ``fcntl`` / ``msvcrt.locking`` around ``clear_caches`` plus the resolve pass so ``--rebuild`` is genuinely concurrent-safe.

I lean toward (1) because it keeps the PR small and reviewable while still fixing a real (if narrower) symptom; (3) is the right end-state but a meaningfully larger surface. Which would you prefer?

sirosen:
> I lean toward (1) because it keeps the PR small and reviewable while still fixing a real (if narrower) symptom; (3) is the right end-state but a meaningfully larger surface. Which would you prefer?

I fully agree. Let's clarify that this is not a full fix for the identified issue, and we can go after a complete solution in the future.

changelog.d/2393.bugfix.md
webknjaz, line 1: @gaborbernat could you symlink `2384.bugfix.md` to this file, so both are linked @ https://pip-tools--2384.org.readthedocs.build/en/2384/changelog/#bug-fixes?

changelog.d/2393.bugfix.md
```diff
@@ -0,0 +1,6 @@
+Atomically rename the download directory before deleting it in
```
webknjaz: Instead of using [imperative mood], rephrase this to use a comparison with the previous release (`X now does Y`) or the past tense (`X was done/changed to do Y`). [Imperative mood] is good for commit messages but in a change log, it makes it look like a TODO-list, which is not at all what we want.

[imperative mood]: https://en.wikipedia.org/wiki/Imperative_mood

piptools/repositories/pypi.py
```diff
@@ -104,7 +104,20 @@ def __init__(self, pip_args: list[str], cache_dir: str):
...
+        # started removing. It does not serialise concurrent ``--rebuild``
+        # runs as a whole; full concurrent-rebuild safety needs locking
+        # around the resolve, which is out of scope here.
+        if not os.path.exists(self._download_dir):
+            return
+        stale = f"{self._download_dir}.stale-{os.getpid()}"
+        try:
+            os.replace(self._download_dir, stale)
+        except OSError:
+            return
```
webknjaz, line 124: This probably needs to be logged in the very verbose mode. Especially since any arbitrary `OSError` is suppressed w/o checking for specific error codes that may change over time.

cc @sirosen WDYT?

piptools/repositories/pypi.py
```diff
@@ -104,7 +104,20 @@ def __init__(self, pip_args: list[str], cache_dir: str):
...
-        rmtree(self._download_dir, ignore_errors=True)
+        # Atomic rename + opportunistic rmtree narrows one specific race:
+        # two ``pip-compile --rebuild`` runs on the same ``cache_dir`` no
+        # longer leave the second resolver staring at a tree the first
+        # started removing. It does not serialise concurrent ``--rebuild``
+        # runs as a whole; full concurrent-rebuild safety needs locking
+        # around the resolve, which is out of scope here.
+        if not os.path.exists(self._download_dir):
+            return
+        stale = f"{self._download_dir}.stale-{os.getpid()}"
```
webknjaz: Can we use a more specific var name? `stale` is quite generic. `old_downloads_dir`, perhaps?

tests/test_repository_pypi.py
```diff
@@ -169,6 +171,48 @@ def test_pip_cache_dir_is_empty(from_line, tmpdir):
     assert not pypi_repository.options.cache_dir
 
 
+@pytest.fixture
+def repo_with_cache_dir(tmp_path: Path) -> PyPIRepository:
+    return PyPIRepository([], cache_dir=str(tmp_path / "cache"))
```
webknjaz, line 176: Side note: `cache_dir` could've accepted pathlib objects already if we were to just relax typing in the initializer… An idea for a follow-up.

tests/test_repository_pypi.py
```diff
@@ -169,6 +171,48 @@ def test_pip_cache_dir_is_empty(from_line, tmpdir):
...
+
+    assert not Path(repo_with_cache_dir._download_dir).exists()
+
+
+def test_clear_caches_swallows_replace_failure(
+    repo_with_cache_dir: PyPIRepository, mocker: MockerFixture
+) -> None:
+    Path(repo_with_cache_dir._download_dir).mkdir(parents=True, exist_ok=True)
+    mocker.patch(
+        "piptools.repositories.pypi.os.replace",
```
webknjaz, line 207: Since you're already integrating `pytest-mock`, perhaps also apply a `mocker.spy()` on top, to confirm the args being passed?

pyproject.toml
```diff
@@ -58,6 +58,7 @@ changelog = "https://github.com/jazzband/pip-tools/releases"
 [project.optional-dependencies]
 testing = [
   "pytest >= 7.2.0",
+  "pytest-mock >= 3.15.1",
```
webknjaz: Is this version really the lowest compatible? Let's find out which one is necessary and set it to that. Or drop the lower bound if any version works.

webknjaz (changes requested):
The overall shape looks good, but I've got a few more suggestions and requests below.

webknjaz:
> [!tip]
> @sirosen did you know you can use mermaid on GH natively? https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams#creating-mermaid-diagrams

pyproject.toml
gaborbernat: This is the only version we tested agaisnt and know to work, unless we have a CI that runs against lowest version we're really just hoping that those low relaxed versions work.

webknjaz:
Is https://github.com/jazzband/pip-tools/actions/runs/25874542161/job/76038161294?pr=2384#step:9:1111 a flaky run?

pyproject.toml
webknjaz: Yeah, but the lower boundaries should reflect APIs relied upon (or known broken things). Without such knowledge, we shouldn't assume that something won't work on older versions, this would only cause problems for downstream packagers for no good reason. After all, the depresolvers would optimize for latest releases anyway.

On a related note, we should look into switching from extras to depgroups once that's supported (I know tox does support these, but I wanted to wait until `pip-tools` itself does too).

changelog.d/2393.bugfix.md
webknjaz: Verified that the render looks good. Marking this as resolved.

pyproject.toml
gaborbernat: I heard that https://github.com/jazzband/pip-tools/pull/2380 can help with that 😉

piptools/repositories/pypi.py
```diff
@@ -104,7 +104,25 @@ def __init__(self, pip_args: list[str], cache_dir: str):
...
+        # longer leave the second resolver staring at a tree the first
+        # started removing. It does not serialise concurrent ``--rebuild``
+        # runs as a whole; full concurrent-rebuild safety needs locking
+        # around the resolve, which is out of scope here.
+        if not os.path.exists(self._download_dir):
+            return
+        old_downloads_dir = f"{self._download_dir}.stale-{os.getpid()}"
+        try:
+            os.replace(self._download_dir, old_downloads_dir)
+        except OSError as exc:
```
webknjaz: Could you use an `os_err` instead of a generic var name?

piptools/repositories/pypi.py
```diff
@@ -104,7 +104,25 @@ def __init__(self, pip_args: list[str], cache_dir: str):
...
+            # A racing peer renamed the tree first, or the OS refused the
+            # move (cross-device, permissions). The directory is no longer
+            # ours to clear; log at -v since the suppressed OSError is
+            # otherwise invisible and its error codes shift across platforms.
+            log.debug(f"clear_caches skipped {self._download_dir!r}: {exc}")
```
webknjaz: Should this be

```suggestion
            log.debug(f"clear_caches skipped {self._download_dir!s}: {os_err}")
```

? We probably don't need to call that repr interface.

tests/test_cli_compile.py
```diff
@@ -872,6 +872,15 @@ def test_relative_file_uri_package(pip_conf, runner):
     assert "file:small_fake_with_deps-0.1-py2.py3-none-any.whl" in out.stderr
 
 
+@pytest.mark.skipif(
```
webknjaz: Is this directly related to the patch? How? Why do you think it's deterministic?

pyproject.toml
gaborbernat: BTW I did lower it to 3.3.0 that's the first compatible version.

tests/test_cli_compile.py
webknjaz, line 1: https://github.com/jazzband/pip-tools/actions/runs/25870600925/job/76024490776#step:9:1111 proves that this test change is not related to this PR as it was failing on `main` before. Meaning this shouldn't be included in the scope.

pyproject.toml
webknjaz: That's too big to consider at the moment + I was hoping #2175 would go in.

gaborbernat: 😊 one can dream and hope 😊

pyproject.toml
```diff
@@ -58,7 +58,7 @@ changelog = "https://github.com/jazzband/pip-tools/releases"
 [project.optional-dependencies]
 testing = [
   "pytest >= 7.2.0",
-  "pytest-mock >= 3.15.1",
+  "pytest-mock >= 3.3.0",
```
webknjaz, line 61: Move to a separate PR

gaborbernat: But is needed for this PR. I don't know man this feels to me like we're being pedantic and generating lots of new work for negligable benefit 🤷

webknjaz:
The spy question stands. Also waiting for Stephen's feedback on questions. Perhaps, this should be marked as a draft while being constantly force-pushed?

gaborbernat:
> Perhaps, this should be marked as a draft while being constantly force-pushed?

It's ready from POV, I'm just addressing your constant feedback, hence constant force-push.

webknjaz:
FYI, it's a good tone to follow https://hynek.me/til/easier-crediting-contributors-github/.

pyproject.toml
webknjaz: A good Git hygiene means being able to `git revert` things w/o additional edits by having atomic changes bundled. The addition of that dep with a wrong version spec slipped into `main` from another PR, so this PR doesn't need to change it, but the change still needs to happen separately.

I guess we could keep it in this PR if it moves to an atomic commit within the branch.

gaborbernat: While I can sympathize with the sentiment here, at the same time there is a trade-off to be made about chasing perfection and balancing the return on the time investment we are asking of people. My personal stance is that we should aim for good enough. Not everything needs to be a revert, and often, when something goes wrong in my experience, it's easier to just follow it up with the pull request that either fixes that issue or reverses the part of a bigger PR that didn't work out. This very easily could be just a difference of style when contributing and maintaining projects between us.

webknjaz: I understand, it's just not good enough. Good shape of Git is important to me. Otherwise, we'd be allowing fake squash merges and other questionable approaches.

webknjaz:
This was almost ready, I'd rather see it completed (maybe by my or Stephen).

PR #2385, opened 2026-05-08
🐛 fix(compile): redact secrets in --pip-args header capture

gaborbernat:
`get_compile_command` records the invocation argv in the requirements-file header so reviewers can reproduce the resolution. URL basic-auth on top-level options like `--extra-index-url` already runs through pip's `redact_auth_from_url`, but the contents of `--pip-args` slip through. `pip-compile --pip-args='--cert /etc/ssl/private.pem'` and `pip-compile --pip-args='--config-settings token=hunter2 --extra-index-url https://user:pw@host/simple'` land in the committed `requirements.txt` and leak the contents to anyone with read access to the repository.

The patch walks the `shlex`-split tokens of `pip_args_str` and replaces the value following `--cert`, `--client-cert`, `--proxy`, `--config-settings`, and `-C` with `<REDACTED>`. The flag itself stays so the header documents which knobs were tuned. URL tokens still go through `redact_auth_from_url`, and C0/C1 control characters get normalised to `�` so a malicious filename cannot smuggle a CR that re-orders the rendered header. The redaction lives in a small private helper next to `get_compile_command`; the existing top-level URL redaction path stays unchanged.

### DIFF
--- a/changelog.d/2385.bugfix.md
+++ b/changelog.d/2385.bugfix.md
@@ -1,0 +1,4 @@
+{program}`pip-compile` now redacts credential-bearing tokens (`--cert`,
+`--client-cert`, `--proxy`, `--config-settings`, `-C`, embedded URL
+basic-auth) inside `--pip-args` before recording the command in the
+requirements-file header -- {user}`gaborbernat`.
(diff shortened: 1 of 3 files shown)


webknjaz:
@sirosen do you think anything's worth picking up from this patch?

sirosen:
I think basically the whole thing is worth doing -- I'm not even sure any changes are needed.

I need to go over the behavior for before-and-after (I want to make sure I can see what each of these flags does today), and then I was going to open a fresh PR and make sure to preserve the original authorship here.

PR #2386, opened 2026-05-08
Recognize the project's _t typing alias in coverage config

gaborbernat:
`covdefaults` ships a default `exclude_lines` regex for `if typing.TYPE_CHECKING:` and `if t.TYPE_CHECKING:`, but pip-tools aliases `typing` as `_t` and the default regex misses those blocks. The lines inside are unreachable at runtime, so they show up as missed in the coverage report and noise the diff for any contributor whose change touches a module with a type-only import block.

The `exclude_also` entry covers the project's `_t` convention. The two `sys.version_info`-conditional import branches in `piptools.build` and `piptools.writer` get `# pragma:` markers so `covdefaults` recognises the unreachable side on a given Python version instead of counting it as missing.

Per @webknjaz's review on the prior #2386 scope, the CI workflow rework stays out of this PR; webknjaz will land the GHA changes separately by syncing setup from sibling projects.

---

No changelog fragment: not a user-facing change.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,5 +6,10 @@ omit =
 [report]
 include = piptools/*, tests/*
 fail_under = 99
+
+# custom excludes extend baselines from coverage + covdefaults
+# for reference, see
+# coverage: https://coverage.readthedocs.io/en/latest/excluding.html#default-exclusions
+# covdefaults: https://github.com/asottile/covdefaults/blob/8d8712c26ad505f1269851a7c8431b8c1fedaa62/covdefaults.py#L86-L105
 exclude_also =
     ^\s*if _t\.TYPE_CHECKING:
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,7 +7,4 @@ omit =
 include = piptools/*, tests/*
 fail_under = 99
 exclude_also =
-    # The covdefaults regex matches `if (t|te|typ|typi|typin|typing)\.TYPE_CHECKING:`
-    # but the project aliases `typing` as `_t`, which falls outside that pattern.
-    # Also exclude it from coverage measurements.
     ^\s*if _t\.TYPE_CHECKING:
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,3 +6,8 @@ omit =
 [report]
 include = piptools/*, tests/*
 fail_under = 99
+exclude_also =
+    # The covdefaults regex matches `if (t|te|typ|typi|typin|typing)\.TYPE_CHECKING:`
+    # but the project aliases `typing` as `_t`, which falls outside that pattern.
+    # Also exclude it from coverage measurements.
+    ^\s*if _t\.TYPE_CHECKING:
--- a/changelog.d/2386.contrib.md
+++ b/changelog.d/2386.contrib.md
@@ -1,2 +1,2 @@
-Improve coverage reporting by skipping the project's `_t = typing` alias
-in `TYPE_CHECKING` blocks -- by {user}`gaborbernat`.
+coverage reporting in pip-tools now skips ``_t.TYPE_CHECKING`` blocks -- by
+{user}`gaborbernat`.
--- a/changelog.d/2386.contrib.md
+++ b/changelog.d/2386.contrib.md
@@ -1,0 +1,2 @@
+Improve coverage reporting by skipping the project's `_t = typing` alias
+in `TYPE_CHECKING` blocks -- by {user}`gaborbernat`.
--- a/changelog.d/2386.contrib.md
+++ b/changelog.d/2386.contrib.md
@@ -1,2 +1,2 @@
-coverage reporting in pip-tools now skips ``_t.TYPE_CHECKING`` blocks -- by
+Coverage reporting in `pip-tools` now skips `_t.TYPE_CHECKING` blocks -- by
 {user}`gaborbernat`.
--- a/piptools/build.py
+++ b/piptools/build.py
@@ -27,11 +27,11 @@ PYPROJECT_TOML = "pyproject.toml"
 _T = _t.TypeVar("_T")
 
 
-if sys.version_info >= (3, 10):
+if sys.version_info >= (3, 10):  # pragma: >=3.10 cover
     from importlib.metadata import PackageMetadata
-else:
+else:  # pragma: <3.10 cover
 
-    class PackageMetadata(_t.Protocol):
+    class PackageMetadata(_t.Protocol):  # pragma: <3.10 cover
         @_t.overload
         def get_all(self, name: str, failobj: None = None) -> list[_t.Any] | None: ...
 
--- a/piptools/build.py
+++ b/piptools/build.py
@@ -31,7 +31,7 @@ if sys.version_info >= (3, 10):  # pragma: >=3.10 cover
     from importlib.metadata import PackageMetadata
 else:  # pragma: <3.10 cover
 
-    class PackageMetadata(_t.Protocol):  # pragma: <3.10 cover
+    class PackageMetadata(_t.Protocol):
         @_t.overload
         def get_all(self, name: str, failobj: None = None) -> list[_t.Any] | None: ...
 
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -512,18 +541,53 @@ jobs:
...
+      # the html artifact. ``coverage report --format=markdown`` is what the
+      # combine env emits; capture it again here and route to the summary.
+      - name: Publish coverage to step summary
+        if: >-
+          !cancelled()
+        run: |
+          {
+            echo "## Coverage report"
+            echo ""
+            .tox/coverage-combine/bin/python -m coverage report --format=markdown
```
webknjaz: This is a case for `tox exec`.

webknjaz (commented):
The coveragepy config looks like a separate change. I'd look into getting it in on its own.

As for the CI changes, I need to sync some of the setup from other projects before we proceed. It'll include per-job reports in GHA job summaries.

.coveragerc
```diff
@@ -6,3 +6,8 @@ omit =
 [report]
 include = piptools/*, tests/*
 fail_under = 99
+exclude_also =
+    # The covdefaults regex matches `if (t|te|typ|typi|typin|typing)\.TYPE_CHECKING:`
+    # but the project aliases `typing` as `_t`, which falls outside that pattern.
+    # Whitelist it so static type-only imports stay excluded.
```
WhyNotHugo: "whitelist" is ambiguous (the term is often used with the opposite meaning):


```suggestion
    # Also exclude it from coverage measurements.
```

changelog.d/2386.contrib.md
```diff
@@ -0,0 +1,7 @@
+The coverage configuration now whitelists ``if _t.TYPE_CHECKING:`` guards used
+across the codebase (the project aliases ``typing`` as ``_t``, which falls
+outside the default ``covdefaults`` regex), and tags the
+``sys.version_info``-conditional imports in :mod:`piptools.build` and
+:mod:`piptools.writer` with ``# pragma:`` markers so version-specific branches
+are reported as covered on the runtimes that exercise them -- by
+:user:`gaborbernat` in :pr:`2386`.
```
WhyNotHugo: Do we really need a changelog entry here? Would users of this tool ever care? This sounds like an internal-facing change.

piptools/build.py
```diff
@@ -27,11 +27,11 @@
...
 
-if sys.version_info >= (3, 10):
+if sys.version_info >= (3, 10):  # pragma: >=3.10 cover
     from importlib.metadata import PackageMetadata
-else:
+else:  # pragma: <3.10 cover
 
-    class PackageMetadata(_t.Protocol):
+    class PackageMetadata(_t.Protocol):  # pragma: <3.10 cover
         @_t.overload
```
WhyNotHugo, line 35: We have CI jobs for <3.10, so we should probably have one submitting coverage reports too. But that's honestly beyond the scope of this PR.

WhyNotHugo (approved):
LGTM.

changelog.d/2386.contrib.md
```diff
@@ -0,0 +1,2 @@
+Improve coverage reporting by skipping the project's `_t = typing` alias
```
webknjaz: Let's avoid making this a TODO-list entry..

.coveragerc
```diff
@@ -6,3 +6,8 @@ omit =
 [report]
 include = piptools/*, tests/*
 fail_under = 99
+exclude_also =
+    # The covdefaults regex matches `if (t|te|typ|typi|typin|typing)\.TYPE_CHECKING:`
+    # but the project aliases `typing` as `_t`, which falls outside that pattern.
+    # Also exclude it from coverage measurements.
```
webknjaz: Technically, this doesn't exclude it from measurements, just from the generated reports. The data is still in the `.coverage` database. So this feels somewhat inaccurate.

webknjaz:
I think we should brush this up get it in.

sirosen:
@webknjaz, can you give this a final once-over? I've just cleaned it up and think it's ready to merge.

changelog.d/2386.contrib.md
```diff
@@ -0,0 +1,2 @@
+coverage reporting in pip-tools now skips ``_t.TYPE_CHECKING`` blocks -- by
```
webknjaz: Missing capitalization and RST bits? Also, we often highlight the project name (do we need a style guide for that?)
```suggestion
Coverage reporting in `pip-tools` now skips `_t.TYPE_CHECKING` blocks -- by
```

https://pip-tools--2386.org.readthedocs.build/en/2386/changelog/#contributor-facing-changes

----

@sirosen also, do we need to mention those pragma changes that aren't related to typing? And extend the PR title?

.coveragerc
```diff
@@ -6,3 +6,5 @@ omit =
 [report]
 include = piptools/*, tests/*
 fail_under = 99
+exclude_also =
+    ^\s*if _t\.TYPE_CHECKING:
```
webknjaz, line 15: @sirosen perhaps, we should still have some sort of note and a ref to covdefaults covering other common convention variants w/o the hallucination bits?

piptools/build.py
webknjaz: @WhyNotHugo they all do submit coverage. The pragma markers just tell a specific coveragepy's process in a specific job not to consider certain lines as “uncovered” because they are known to be never expected to run under that those specific envs. Different jobs will pretend different lines don't exist. But Codecov will merge the reports, overlaying them and showing coverage in both branches.

changelog.d/2386.contrib.md
webknjaz: The contrib category is for things that are internal but would affect other contributors. These notes are encouraged but optional. It's possible to tell the bot to stop requiring them (by applying a label).

changelog.d/2386.contrib.md
sirosen: Oof! I need to work on a way of catching myself writing RST in `.md` docs. I'll apply the suggestion.

WRT the title and adding a note about those unrelated pragmas... I think it's actually best to let it slide with no note. If we made those changes in a standalone PR, I'd advocate for no changelog for it. When a changelog is _overpopulated_ with details, it becomes useless. (If I want commit messages, I'll use `git log`.)

.coveragerc
sirosen: I'd rather not enumerate conventions or things. I'll add one simply noting that we inherit `coverage` defaults + `covdefaults` rules. We can always look those up as needed anyway.

webknjaz: Yep, I'm fine with not duplicating the implementation details (especially as they can change upstream over time). I'm rather talking about including motivation+clues in comments — the “why”.

webknjaz: Also, thanks for noticing the original hallucination — I've completely overlooked it initially and believe it having assumed that Bernát checked what went into the patch.

.coveragerc
```diff
@@ -6,3 +6,10 @@ omit =
 [report]
 include = piptools/*, tests/*
 fail_under = 99
+
+# custom excludes extend baselines from coverage + covdefaults
+# for reference, see
+# coverage: https://coverage.readthedocs.io/en/latest/excluding.html#default-exclusions
+# covdefaults: https://github.com/asottile/covdefaults/blob/8d8712c26ad505f1269851a7c8431b8c1fedaa62/covdefaults.py#L86-L105
```
webknjaz, line 13: Nit: it's usually helpful to shorten `readthedocs` to `rtfd`, and the commit hashes in persistent GH URLs to 7–8 chars, which makes the lines fit on screen better. We may also want persistent RTD URLs by referencing version-specific docs instead of latest which makes them resistant to upstream redesigning the structure and causing HTTP 404.

changelog.d/2386.contrib.md
webknjaz: Agreed. The change log is for high-level effects on the end users mainly, while Git commit messages have different style and target the contributors. We may want to consider splitting out the contrib notes into a separate document render in the future.

webknjaz: As for detecting RST, this should be solvable by a pygrep hook in pre-commit for easy cases.

webknjaz (approved):

PR #2388, opened 2026-05-08
Drop stray whitespace before !s in f-string conversion flags

gaborbernat:
A handful of f-strings in `piptools.utils`, `piptools.scripts.compile`, and `piptools.scripts.sync` carry a stray space between the value and the conversion flag (`f"...{value !s}..."`). PEP 498 forbids whitespace before `!s`/`!r`/`!a` and `ruff format` rewrites the form when it normalises a file, so the next time a formatter pass touches one of these modules the contributor's diff would carry an unrelated reformat hunk.

Cleaning the literals up now keeps the files canonical with no behaviour change. The rendered output is identical because Python today still parses the form (it just normalises away the whitespace internally).

### DIFF
--- a/changelog.d/2388.contrib.md
+++ b/changelog.d/2388.contrib.md
@@ -1,0 +1,7 @@
+A handful of f-string conversions in :mod:`piptools.utils` and the CLI
+entry points in :mod:`piptools.scripts.compile` and
+:mod:`piptools.scripts.sync` carried a stray space before ``!s`` (e.g.
+``f"...{value !s}..."``). PEP 498 forbids the whitespace before the
+conversion flag and tooling differs on whether the form parses; the
+literals are now in canonical ``{value!s}`` shape -- by
+:user:`gaborbernat` in :pr:`2388`.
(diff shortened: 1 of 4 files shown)


sirosen (approved):

changelog.d/2388.contrib.md
```diff
@@ -0,0 +1,7 @@
+A handful of f-string conversions in :mod:`piptools.utils` and the CLI
+entry points in :mod:`piptools.scripts.compile` and
+:mod:`piptools.scripts.sync` carried a stray space before ``!s`` (e.g.
+``f"...{value !s}..."``). PEP 498 forbids the whitespace before the
```
webknjaz, line 4: Missing `:pep:` role.

changelog.d/2388.contrib.md
webknjaz, line 1: @sirosen missed that this should use MyST-Parser Markdown, not RST.

@gaborbernat care to fix the note in a follow-up?

sirosen: Ach! Yes, I did miss that. I'll file a fixup in the next couple of days if nobody beats me to it.

PR #2389, opened 2026-05-08
Clarify pip-compile CLI help for --extra, --upgrade-package, --no-build-isolation

gaborbernat:
The CLI help for `--extra`, `--all-extras`, `--upgrade-package`, and `--no-build-isolation` reads as four disconnected single sentences. `--extra` does not mention that `--all-extras` exists, `--upgrade-package` does not explain what happens to the pins of unrelated packages, and `--no-build-isolation` only says build deps "must be already installed" without naming the symptom when a user forgets. The same questions surface on the issue tracker.

The patch cross-references `--extra` and `--all-extras` so each entry points at the other, spells out for `--upgrade-package` that other packages keep their seeded pins, and for `--no-build-isolation` calls out the concrete tools the user must pre-install plus the symptom they see when a build dep is missing.

Pure docs change. The option choices, defaults, parsing, and runtime behaviour stay the same; only the `help=` strings change.

---

No changelog fragment: docs polish, no behaviour change.

### DIFF
--- a/piptools/scripts/options.py
+++ b/piptools/scripts/options.py
@@ -106,14 +106,17 @@ extra = click.option(
     "--extra",
     "extras",
     multiple=True,
-    help="Name of an extras_require group to install; may be used more than once",
+    help=(
+        "Name of an extras_require group to install; may be used more than "
+        "once. Pass ``--all-extras`` to include every declared extra."
+    ),
 )
 
 all_extras = click.option(
     "--all-extras",
     is_flag=True,
     default=False,
-    help="Install all extras_require groups",
+    help="Install all extras_require groups (see also: ``--extra`` for one at a time)",
 )
 
 find_links = click.option(
@@ -214,7 +217,11 @@ upgrade_package = click.option(
     "upgrade_packages",
     nargs=1,
     multiple=True,
-    help="Specify a particular package to upgrade; may be used more than once",
+    help=(
+        "Re-resolve the named package against the index, ignoring any pin "
+        "from the existing requirements file. May be used more than once. "
+        "Other packages keep their seeded pins."
+    ),
 )
 
 output_file = click.option(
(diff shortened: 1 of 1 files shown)


piptools/scripts/options.py
```diff
@@ -106,14 +106,17 @@ def _get_default_option(option_name: str) -> _t.Any:
...
+        "once. Pass ``--all-extras`` to include every declared extra."
+    ),
 )
 
 all_extras = click.option(
     "--all-extras",
     is_flag=True,
     default=False,
-    help="Install all extras_require groups",
+    help="Install all extras_require groups (cf. ``--extra`` for one at a time)",
```
sirosen: Is `cf.` well enough known? I consider it a little bit obscure. I tend to use the longer "see also".

This is super weakly-held by me. Feel free to reject.

```suggestion
    help="Install all extras_require groups (see also: ``--extra`` for one at a time)",
```

piptools/scripts/options.py
```diff
@@ -291,9 +298,12 @@ def _get_default_option(option_name: str) -> _t.Any:
     is_flag=True,
     default=True,
     help=(
-        "Enable isolation when building a modern source distribution. "
-        "Build dependencies specified by PEP 518 must be already installed "
-        "if build isolation is disabled."
+        "Enable isolation when building a modern source distribution. With "
+        "``--no-build-isolation`` you must pre-install every PEP 518 build "
+        "dependency (``setuptools``, ``wheel``, ``hatchling``, etc.) into "
+        "the active environment; missing build deps surface as confusing "
```
sirosen: I'd like to strike "confusing", as it's a judgement on how the error handling goes in practice. I'd like to work on making it less confusing. pip actually has some pretty-print handling for this, but `pip-compile` ends up losing that -- we need to add our own version of it.

```suggestion
        "the active environment; missing build deps surface as "
```

sirosen (changes requested):

sirosen (approved):

PR #2390, opened 2026-05-08
Neutralize PIP_* environment variables in the runner fixture

gaborbernat:
Tests that drive `pip-compile` or `pip-sync` through `click.testing.CliRunner` inherit the contributor's shell environment, including `PIP_INDEX_URL`, `PIP_EXTRA_INDEX_URL`, and `PIP_TRUSTED_HOST`. On developer machines or self-hosted CI runners with a corporate Artifactory mirror configured at the shell level, those variables leak into the recorded pip argument list and tests asserting on exact pip arguments flake.

The fixture drops the three URL-pointing variables for the duration of each CLI invocation. Production code paths see no difference; the change touches only the test fixture.

### DIFF
--- a/changelog.d/2390.contrib.md
+++ b/changelog.d/2390.contrib.md
@@ -1,0 +1,5 @@
+Drop `PIP_INDEX_URL`, `PIP_EXTRA_INDEX_URL`, and `PIP_TRUSTED_HOST` from
+the environment in the `runner` test fixture so CLI tests asserting on
+exact pip argument lists no longer flake on contributor machines with a
+corporate index mirror configured at the shell level --
+{user}`gaborbernat`.
--- a/changelog.d/2390.contrib.md
+++ b/changelog.d/2390.contrib.md
@@ -1,5 +1,7 @@
-Drop `PIP_INDEX_URL`, `PIP_EXTRA_INDEX_URL`, and `PIP_TRUSTED_HOST` from
-the environment in the `runner` test fixture so CLI tests asserting on
-exact pip argument lists no longer flake on contributor machines with a
-corporate index mirror configured at the shell level --
-{user}`gaborbernat`.
+A session-scoped autouse fixture in {file}`tests/conftest.py` now drops every
+`PIP_*` environment variable before any other fixture runs, so the test
+suite no longer flakes on contributor machines whose shell exports
+`PIP_INDEX_URL`, `PIP_FIND_LINKS`, `PIP_TRUSTED_HOST`, and the rest of
+the family for a corporate mirror. Tests that need a specific variable
+keep their function-scoped {meth}`pytest:pytest.MonkeyPatch.setenv` and
+supersede the session-level deletion -- by {user}`gaborbernat`.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -229,13 +229,13 @@ def from_editable():
     return install_req_from_editable
 
 
-# _isolate_pip_env is important for direct runs of the testsuite when contributors have
-# pip configurations in their env (e.g., for corporate index servers)
 @pytest.fixture(scope="session", autouse=True)
 def _isolate_pip_env() -> _c.Iterator[None]:
     """
     Automatically drop all ``PIP_*`` environment variables during test execution.
     """
+    # this is important for direct runs of the testsuite when contributors have
+    # pip configurations in their env (e.g., for corporate index servers)
     with pytest.MonkeyPatch.context() as mp:
         for env_var in (name for name in os.environ if name.startswith("PIP_")):
             mp.delenv(env_var)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import collections.abc as _c
 import json
 import os
 import platform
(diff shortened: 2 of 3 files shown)


changelog.d/2390.contrib.md
```diff
@@ -0,0 +1,5 @@
+Drop `PIP_INDEX_URL`, `PIP_EXTRA_INDEX_URL`, and `PIP_TRUSTED_HOST` from
```
webknjaz: Plz use the past tense or something like that. Here's the preview: https://pip-tools--2390.org.readthedocs.build/en/2390/changelog/#contributor-facing-changes.

Also, are these linkable via the `:envvar:` role?

tests/conftest.py
```diff
@@ -229,7 +229,26 @@ def from_editable():
 
 
 @pytest.fixture
-def runner():
+def runner(
+    monkeypatch: pytest.MonkeyPatch,
+    minimal_wheels_path: Path,
+) -> _t.Generator[CliRunner, None, None]:
```
webknjaz: Iterator is more elegant for such cases:
```suggestion
) -> _t.Iterator[CliRunner]:
```
*(although, it should probably be reached via `collections.abc` instead of `typing` these days)*

tests/conftest.py
```diff
@@ -229,7 +229,26 @@ def from_editable():
...
+    # sees no index URL and the download fails. Declaring it as a
+    # dependency forces the session fixture to initialise before the
+    # ``delenv`` below ever fires.
+    del minimal_wheels_path
+    # Pip honours ``PIP_INDEX_URL`` and friends from the contributor's shell;
+    # corporate Artifactory mirrors leak through and break tests that assert
+    # exact pip argument lists. Drop the variables for the duration of the
+    # CLI invocation. ``PIP_CONFIG_FILE`` is intentionally left alone because
+    # the cross-platform ``os.devnull`` device file (``/dev/null`` vs ``nul``)
+    # exposed pip path-resolution flakes on Windows in lowest-pip CI.
```
webknjaz: Can't it be made pip-version dependent, then? Or point to an empty file in `tmp_path`? I don't fully understand where the devnull implication is coming from here. Care to elaborate?

tests/conftest.py
```diff
@@ -7,6 +7,7 @@
 import subprocess
 import sys
 import typing as _t
+from collections.abc import Iterator
```
webknjaz: Let's use the `import as _c` convention similar to the typing import.

tests/conftest.py
```diff
@@ -229,7 +230,37 @@ def from_editable():
...
+    # Pip honours ``PIP_INDEX_URL`` and friends from the contributor's shell;
+    # a corporate Artifactory mirror leaks through and breaks tests that
+    # assert exact pip argument lists. Drop the variables for the duration of
+    # the CLI invocation. ``PIP_CONFIG_FILE`` is left alone on purpose: the
+    # ``pip_conf`` fixture owns that variable and points it at a generated
+    # ``pip.conf``; a second writer here would race ``pip_conf`` depending on
+    # fixture-setup order. Pointing it at ``os.devnull`` instead was tried and
+    # reverted because the device file (``/dev/null`` vs ``nul``) tripped pip
+    # path-resolution flakes on Windows in lowest-pip CI.
+    for env_var in (
```
webknjaz: Looking at this again, I realized that we should iterate over all the env vars and drop anything starting with `PIP_` instead of picking a subset of settings to shield the test env from.
Additionally, a session fixture with autouse would be a better fit for this. Example: https://github.com/ansible/tox-ansible/blob/2c79bf8c9d685151c3e2c642ae49ad95257657e2/tests/conftest.py#L116. Pip does an even deeper isolation: https://github.com/pypa/pip/blob/f7bfe280f00831b249534fc8e8a549cb48b3d166/tests/conftest.py#L313

webknjaz: @sirosen WDYT?

changelog.d/2390.contrib.md
```diff
@@ -0,0 +1,7 @@
+A session-scoped autouse fixture in `tests/conftest.py` now drops every
```
webknjaz:
```suggestion
A session-scoped autouse fixture in {file}`tests/conftest.py` now drops every
```

changelog.d/2390.contrib.md
```diff
@@ -0,0 +1,7 @@
...
+`PIP_*` environment variable before any other fixture runs, so the test
+suite no longer flakes on contributor machines whose shell exports
+`PIP_INDEX_URL`, `PIP_FIND_LINKS`, `PIP_TRUSTED_HOST`, and the rest of
+the family for a corporate mirror. Tests that need a specific variable
+keep their function-scoped `monkeypatch.setenv` and supersede the
```
webknjaz: Can we reference this via intersphinx? (needs an addition to `docs/conf.py`)

tests/conftest.py
```diff
@@ -228,8 +229,28 @@ def from_editable():
...
+    # session so contributor pip configuration (corporate Artifactory mirror,
+    # find-links, trusted hosts, and so on) cannot leak into CLI tests that
+    # assert on exact pip argument lists. Session-scoped + autouse so the
+    # cleanup fires before any other fixture; subprocesses spawned by session
+    # fixtures (the ``pip download`` in ``setuptools_wheel_path``) also run
+    # with a clean environment. Function-scoped ``monkeypatch.setenv`` in a
+    # specific test still overrides the deletion for that test, so cases like
+    # ``test_find_links_envvar`` keep working.
+    mp = pytest.MonkeyPatch()
+    for env_var in [name for name in os.environ if name.startswith("PIP_")]:
```
webknjaz: No need to create an in-memory throw-away list. Favor generator expression:
```suggestion
    for env_var in (name for name in os.environ if name.startswith("PIP_")):
```

tests/conftest.py
```diff
@@ -228,8 +229,28 @@ def from_editable():
...
+        mp.delenv(env_var, raising=False)
```
webknjaz: Raising shouldn't be needed as we run this against known-set vars:
```suggestion
        mp.delenv(env_var)
```

tests/conftest.py
```diff
@@ -228,8 +229,28 @@ def from_editable():
...
+    try:
+        yield
+    finally:
+        mp.undo()
+
+
 @pytest.fixture
-def runner():
+def runner() -> _c.Iterator[CliRunner]:
```
webknjaz: Could you move this to a separate PR as this is a separate scope?

tests/conftest.py
```diff
@@ -228,6 +229,26 @@ def from_editable():
     return install_req_from_editable
 
 
+@pytest.fixture(scope="session", autouse=True)
+def _isolate_pip_env() -> _c.Iterator[None]:
+    # Drop every ``PIP_*`` environment variable for the duration of the test
```
webknjaz: Part of this should probably be a docstring (what this does) while corner cases may remain in a comment. `pytest --fixtures` will show the docstring in CLI output so it's contributor-facing (includes downstream distributors debugging test runs in their envs).

webknjaz:
If you don't want to add improvements, that's fine — I'll try to add them into this branch myself. Let's keep it open.

tests/conftest.py
sirosen: It looks like `pytest --fixtures` skips fixtures with private (`_`) names.

Would you like me to make the adjustment you requested here, so you can review? Or leave it as-is.

The change looks good to me at this stage.

tests/conftest.py
sirosen: It looks like this got done before I could get to this PR -- it got pushed back in my review queue. I do like the version we have in the PR now, so :+1: for this note (after the fact)!

tests/conftest.py
webknjaz: @sirosen I didn't know this and couldn't find the confirmation in the docs. I only saw elsewhere that added verbosity exposes more of the docstring.

So I've gone to the source and if I'm reading this right, `-v` also unhides the “_private” fixtures:
https://github.com/pytest-dev/pytest/blob/543b11f2534bb0edc85b86767ed239da9c5bd140/src/_pytest/fixtures.py#L2061C12-L2061C24

With that in mind, I think that we do need a docstring and I'm pretty confident that `pytest --collect-only --fixtures -v` will reveal it.

----

As for possibly making this fixture have a public name, it's probably unnecessary since it's autouse. Up to you, though, I don't have strong opinions.

----

@sirosen since you volunteered, I'll let you write that docstring, if you don't mind.

tests/conftest.py
```diff
@@ -228,6 +229,26 @@ def from_editable():
...
+    # session so contributor pip configuration (corporate Artifactory mirror,
+    # find-links, trusted hosts, and so on) cannot leak into CLI tests that
+    # assert on exact pip argument lists. Session-scoped + autouse so the
+    # cleanup fires before any other fixture; subprocesses spawned by session
+    # fixtures (the ``pip download`` in ``setuptools_wheel_path``) also run
+    # with a clean environment. Function-scoped ``monkeypatch.setenv`` in a
+    # specific test still overrides the deletion for that test, so cases like
+    # ``test_find_links_envvar`` keep working.
+    mp = pytest.MonkeyPatch()
```
webknjaz: Note on this: I wonder if upstream is ready to accept a `monkeypatch_session` now. https://github.com/pytest-dev/pytest/issues/363 suggests that here were some problems with it in the past, but it's not clear to me what happened...

tests/conftest.py
sirosen: Done! I also fixed it to use the monkeypatch context manager, which is more maintainable than try-finally-`undo()`.

webknjaz: Brilliant, thanks!

webknjaz (approved):
@sirosen feel free to merge. But shouldn't the comment part live inside the function for maintaining coupling? Under the docstring, perhaps?

sirosen:
> But shouldn't the comment part live inside the function for maintaining coupling? Under the docstring, perhaps?

Sure, let me switch that around! :+1: 
I'll hold that your approval still applies, so I'll merge after that update.

PR #2391, opened 2026-05-08
Add intersphinx mappings for packaging, pip, click, build, importlib_metadata

gaborbernat:
The autodoc-rendered docstrings reference types from five upstream projects (``pip._internal.req.InstallRequirement``, ``click.Context``, ``packaging.markers.Marker``, ``build.ProjectBuilder``, and so on), but the Sphinx ``intersphinx_mapping`` only points at the standard library. Every external cross-reference falls back to ``nitpick_ignore`` suppression and the rendered HTML drops the hyperlink, so readers cannot click through to the upstream docs from a pip-tools method signature.

Adding the five inventories restores the links. The mappings target the canonical stable docs URLs for each project (``packaging.pypa.io/en/stable``, ``pip.pypa.io/en/stable``, ``click.palletsprojects.com/en/stable``, ``build.pypa.io/en/stable``, ``importlib-metadata.readthedocs.io/en/latest``).

While in the file, the deprecated ``sphinx.util.console.bold`` import on the two version-banner lines is dropped; the colour markup never rendered in CI logs and the helper is on track to be removed in upcoming Sphinx releases.

### DIFF
--- a/changelog.d/2391.contrib.md
+++ b/changelog.d/2391.contrib.md
@@ -1,0 +1,8 @@
+The Sphinx configuration now resolves cross-references for
+:mod:`packaging`, :mod:`pip`, :mod:`click`, :mod:`build`, and
+:mod:`importlib_metadata` via intersphinx, and the deprecated
+``sphinx.util.console.bold`` helper is replaced with a plain
+``logger.info`` call (the colour markup never rendered in CI logs
+anyway). Cross-references in docstrings that previously fell back to
+nitpick suppression now link to the upstream documentation -- by
+:user:`gaborbernat` in :pr:`2391`.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -9,7 +9,6 @@ from pathlib import Path
 
 from sphinx.application import Sphinx
 from sphinx.util import logging
-from sphinx.util.console import bold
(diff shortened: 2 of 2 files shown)


sirosen (approved):
I don't want to read too much or too little into these PRs being opened as drafts? I'm ready to merge this, but draft status is telling me "author doesn't consider it ready yet" (?).

gaborbernat:
Draft mostly means here that I'm still double checking I'm happy with them 😊 e.g. validate CI passes on my fork and such.

changelog.d/2391.contrib.md
```diff
@@ -0,0 +1,8 @@
+The Sphinx configuration now resolves cross-references for
+:mod:`packaging`, :mod:`pip`, :mod:`click`, :mod:`build`, and
```
webknjaz, line 2: @sirosen @gaborbernat another case to fix for MyST-Parser Markdown instead of RST ^

sirosen: Yeah, I just put up a fix PR a few minutes ago (#2396), but it seems to be failing to handle intersphinx correctly now. I'll tinker on it between various talks and such.

PR #2392, opened 2026-05-08
Let CI completion Codecov notify vote upstream

gaborbernat:
`codecov/codecov-action` used to be called with `fail_ci_if_error: true` in the `Coverage processing` CI job unconditionally. This works well when a project for the repository is created on Codecov but fails in the context of external repositories that aren't (like typical forks).

A contributor could create a Codecov-side project for their fork and the CI calls to Codecov config would work in that context, but most people don't need that.

This patch makes the `fail_ci_if_error` context-aware and only fails the CI job step when the context is the upstream repository (including `pull_request` runs from forks to upstream). This keeps the door open a for forks that might still want to have their own Codecov project instance for whatever reason.

When run in a foreign context, Codecov CLI still logs the `Token required - not valid tokenless upload` error message but sets the successful return code, making the step and job look green.

If this happens upstream, this will still bubble up in CI as a loud failure.

The same treatment is applied to the Codecov action invocation in the test jobs for consistency.

PR #2400, opened 2026-05-18
Fix pypy+windows CI

sirosen:
We're seeing pypy test failures in CI which block anything from merging.
~I'm hopeful that it's something which is wrong with the runner image itself, and that downgrading (temporarily?) will fix it.~

**Update**
I figured out that the cause is the click v8.4.0 release, _for some reason_. I don't know precisely why.
I think we should move forward with this tweak, however, because broken CI has such a strong negative impact.

Per #2399, I would actually like to drop pypy entirely.

**Update 2**
It turns out that this was not correct either. See discussion below for details.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -441,11 +441,14 @@ jobs:
           - ubuntu-24.04
           - macos-26
           - macos-26-intel
-          - windows-2025
         python-version:
           - pypy-3.11
         pip-version:
           - supported
+        include:
+          - runner-vm: windows-2025
+            python-version: pypy-3.11
+            pip-version: lowest  # PR 2400: legacy depresolver fails
     env:
       TOXENV: pip${{ matrix.pip-version }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -443,7 +443,7 @@ jobs:
           - macos-26-intel
           - windows-2025
         python-version:
-          - pypy-3.10
+          - pypy-3.11
         pip-version:
           - supported
     env:
--- a/changelog.d/+96cdce2b.contrib.md
+++ b/changelog.d/+96cdce2b.contrib.md
@@ -1,0 +1,1 @@
+`pip-tools` now tests against PyPy 3.11 -- by {user}`sirosen`.
--- a/changelog.d/+96cdce2b.contrib.md
+++ b/changelog.d/+96cdce2b.contrib.md
@@ -1,1 +1,1 @@
-`pip-tools` now tests against PyPy 3.11 -- by {user}`sirosen`.
+`pip-tools` now tests against PyPy 3.11 -- by {user}`sirosen` and {user}`webknjaz`.


webknjaz (approved):

sirosen:
I feel like someone is playing tricks on me. CI is now failing again on the same build (windows + pypy), even with this adjustment, although it passed yesterday. I'll need to look into this again.

webknjaz:
@sirosen I mentioned in some other PR that there was flakiness, and sometimes it was on macos too, I think.

webknjaz:
Though, maybe that one's different: https://github.com/jazzband/pip-tools/pull/2392#issuecomment-4467369610

webknjaz:
@sirosen could you also sign off on https://github.com/jazzband/pip-tools/pull/2392 ?

sirosen:
> @sirosen could you also sign off on #2392 ?

Yeah, happily! I'll go do that.

But I think it will also be blocked on CI. Each time I've put this in the merge queue, it fails with the same errors as before, and I therefore no longer think this patch has a useful effect (although it's not terribly harmful). So I don't understand how it passed here when I last made changes.

webknjaz:
> But I think it will also be blocked on CI. Each time I've put this in the merge queue, it fails with the same errors as before, and I therefore no longer think this patch has a useful effect (although it's not terribly harmful). So I don't understand how it passed here when I last made changes.

In the past, such cases (conceptually) would happen due to `push` and `merge_queue` events picking up different cache entries than `pull_request`.
As a test, we could invalidate everything under https://github.com/jazzband/pip-tools/actions/caches?query=key%3AWindows-+ and see if that makes PRs fail too, for example.

@sirosen would you like to try that?

webknjaz:
I've also seen Alex mentioning going to allow failures on PyPy jobs in trio/anyio linking some upstream bug.

There was a release a few weeks ago: https://doc.pypy.org/release-v7.3.22.html#pypy-v7-3-22-release-of-python-2-7-3-11-released-2026-04-26 — perhaps it'd hit downstreams more recently.

webknjaz:
Oh, wait.. PyPy v3.7.19 was the last version shipping Python 3.10, we should've switched to `pypy-3.11` long ago.

tox.ini
```diff
@@ -34,6 +34,10 @@ deps =
 
     piplatest: pip
     pipmain: https://github.com/pypa/pip/archive/main.zip
+
+    # specifically for pypy builds, pin back the version of `click` in use
+    # NOTE: the cause of failures on this version is not currently clear
+    pypy: click<8.4.0
```
webknjaz: Just realized that GHA passes `pypy3` into `$TOXENV`, not `pypy`. So this change didn't actually contribute to the PR runs, not to the merge queues. Though, I imagine that the PR runs picked up older cache from `main`

### failing job (`merge_queue`)
```console
Cache restored from key: Windows-pip--b6ae944a4f2824a978e48a24f277353596b88e0baba52580e575288b804e1d40-aef95a4d383c36cd50738524aa41a394c7e56e41dbda99138a15c081aff56d93-fe11e2b76c38ca0106a0edff5d0631aff168d5c054cb5a99a40365e49e7b2d51
```
*(takes this cache entry from the recent `main`)*

### this PR (`pull_request`)
```console
Cache restored from key: Windows-pip--b6ae944a4f2824a978e48a24f277353596b88e0baba52580e575288b804e1d40-e5f96bf9819bf9093eebfcfaa2047077cb4d7c911f0787125dfebba01a195d40-fe11e2b76c38ca0106a0edff5d0631aff168d5c054cb5a99a40365e49e7b2d51
```
*(takes this cache entry from the PR-specific ephemeral merge pointer at `refs/pull/2400/merge` that is known to get updates asynchronously and get stale when there are no pushes to the PR branch)*

### possible solution

(unless we want to rename on the GHA side)
```suggestion
    pypy3: click<8.4.0
```

sirosen:
I'm going to push a change here which bumps us up to pypy3.11 and fixes the `pypy3` factor in tox.ini (oop! :grin: ).

If that doesn't work, I'll clear caches and try again. The last merge to main passed, so it's not a strictly _bad_ cache, but it could be somehow tripping things up.

webknjaz:
So now that a proper version of click is being picked up, we can be pretty sure that its version change is unrelated: https://github.com/jazzband/pip-tools/actions/runs/26172327888/job/76993181908#step:7:32.

Observation: all of these failures are connection errors likely coming from `urllib3` that is vendored in pip. And all are happening in the legacy resolver that pip is working on getting rid of lately. I'm wondering if this is hitting something that pip stopped testing…

webknjaz:
Also, the tests hit URLs like https://github.com/jazzband/pip-tools/archive/7d86c8d3.zip — GitHub's been rate-limiting unauthenticated requests in many ways lately. This hits Sphinx's linkcheck and cibuildwheel (the old versions — the new ones pass a header with a token).

So perhaps, this is a part of the problem — GH might be sending the response slowly which would cascade into interrupted connections.

.github/workflows/ci.yml
```diff
@@ -443,11 +443,11 @@ jobs:
           - macos-26-intel
           - windows-2025
         python-version:
-          - pypy-3.10
+          - pypy-3.11
         pip-version:
           - supported
```
webknjaz: @sirosen sounds like this could be https://github.com/pypa/pip/issues/13965 / https://github.com/urllib3/urllib3/pull/2514.

Let's try
```suggestion
          - lowest
```

sirosen:
> So now that a proper version of click is being picked up, we can be pretty sure that its version change is unrelated: https://github.com/jazzband/pip-tools/actions/runs/26172327888/job/76993181908#step:7:32.

I agree. I refrained from backing that out at first to reduce the number of variables I was changing at once.
I don't really want to include that change (the minor CI config change is still fine, IMO), so I will remove it.

> So perhaps, this is a part of the problem — GH might be sending the response slowly which would cascade into interrupted connections.

I like this explanation, but it's very strange that we see passing CI on the branch build but not in the merge queue. I was thinking something similar, but it doesn't explain that discrepancy on its own.

---

FYI, I cleared some of the caches for the merge queue builds. If that was related, we may see future builds pass.

.github/workflows/ci.yml
sirosen: There's also a possibility that the pip bug report is related to a bad cache, and I just cleared some of the caches, per below. Would you be alright with holding off on this change until/unless we see a failure on the next try? (I'd prefer we test on our "supported" target if possible.)

webknjaz: Sure. That's why I didn't commit it into the branch. I'm expecting that we need to understand what's going on before declaring any solution good enough, anyway.

.github/workflows/ci.yml
```diff
@@ -443,11 +443,11 @@ jobs:
...
     env:
-      TOXENV: pip${{ matrix.pip-version }}
+      TOXENV: pip${{ matrix.pip-version }}-pypy3
```
webknjaz:
```suggestion
      TOXENV: pip${{ matrix.pip-version }}
```

sirosen:
I've just pushed a small change to make two tweaks:
- removed use of `click<8.4.0` on pypy from tox.ini
- added @webknjaz to the news fragment (I wasn't sure which commit could be marked co-authored, but this is definitely a collaborative effort I want to credit)

Because caches are now cleared, I plan to try to put this through the merge queue (if it passes CI on the branch) before checking the results when using our "piplowest" target, although that is an option.

.github/workflows/ci.yml
sirosen: I was inclined to leave this, as a micro-improvement, but we can remove it, yes. Applying! :slightly_smiling_face:

.github/workflows/ci.yml
sirosen: The latest build failed, so let's try this. I wouldn't be happy with it long term, but not being able to merge anything is a bigger problem.

*Marked unresolved for the moment to improve visibility*

sirosen: @webknjaz, do you feel comfortable trying to merge this fix, and seeing if it passes in the queue?

I don't want to hit merge unless you're onboard with it, but there's no other straightforward way to test the merge-queue builds.

StanFromIreland (approved):

.github/workflows/ci.yml
webknjaz: > I don't want to hit merge unless you're onboard with it, but there's no other straightforward way to test the merge-queue builds.

Okay.. Trying that now.

webknjaz:
> > So perhaps, this is a part of the problem — GH might be sending the response slowly which would cascade into interrupted connections.
> 
> I like this explanation, but it's very strange that we see passing CI on the branch build but not in the merge queue. I was thinking something similar, but it doesn't explain that discrepancy on its own.

@sirosen yeah, I'm feeling lost here too. Let's see if the merge passes. I was rather hoping that the PR builds would start failing w/o cache, and it looks like they did, followed by a recovery with the pip version change.

If the merge proceeds, I'd maybe try to scope `lowest` to just Windows, maybe, in a follow-up.

.github/workflows/ci.yml
```diff
@@ -443,9 +443,9 @@ jobs:
           - macos-26-intel
           - windows-2025
```
webknjaz: @sirosen so the affected job in the merge queue run [succeeded](https://github.com/jazzband/pip-tools/actions/runs/26191242415/job/77059890908), but another flaky job failed, so I interrupted it.

Now that we know that this fixes the CI stability, perhaps, we could make it more granular like this?
```suggestion
        python-version:
          - pypy-3.11
        pip-version:
          - supported
        include:
          - runner-vm: windows-2025
            python-version: pypy-3.11
            pip-version: lowest  # PR 2400: legacy depresolver fails
```

Leaving this uncommitted so you could hit accept and clean up commits/changelog as needed when you get to it yourself.

sirosen: That looks great!

I'm finally settled at home and have a bit of time to work through this. I'll get a version of this in, and I feel ready to merge assuming I don't hit any new surprises.

PR #2405, opened 2026-05-22
Move configuration docs to new reference section

sirosen:
**Quick Review Link:** [PR doc build](https://pip-tools--2405.org.readthedocs.build/en/2405/).

1.  Move `docs/cli/` to `docs/reference/`

2.  Move configuration documentation from the readme into
    `docs/reference/configuration.md`

    - In the process, leave behind doc on `CUSTOM_COMPILE_COMMAND`, which
      can move separately, under a new header

3.  Revise the configuration documentation, generally maintaining the
    same content.

    - As a notable exception, we now explicitly recommend against using
      `_` in config keys -- notionally, we may remove support for that in
      the future, as it allows for conflicting keys unnecessarily.

See also #2382 as a reference point for how this can be presented.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist


changelog.d/+97e08771.doc.md
```diff
@@ -0,0 +1,5 @@
+`pip-tools`' documentation now features a section titled "Reference" covering
+the CLI commands and configuration. Configuration documentation has been removed
+from the readme.
+
+-- by {user}`sirosen`.
```
webknjaz:
```suggestion
-- by {user}`sirosen`
```

changelog.d/+97e08771.doc.md
```diff
@@ -0,0 +1,5 @@
+`pip-tools`' documentation now features a section titled "Reference" covering
+the CLI commands and configuration. Configuration documentation has been removed
```
webknjaz, line 2: There's a chance to use `{ref}` / `{doc}` links here.

docs/reference/configuration.md
```diff
@@ -0,0 +1,65 @@
...
+writing them to a configuration file in the same directory as your requirements
+input files (or the current working directory if piping input from stdin).
+
+Each command-line flag is also a valid configuration key.
+
+## Lookup Order
+
+1. Any path passed to `--config`
+2. `.pip-tools.toml` in the directory of the input file
+3. `pyproject.toml` in the directory of the input file
```
webknjaz, line 13: Is this accurate? I thought this was relative to CWD for some reason..

docs/reference/configuration.md
```diff
@@ -0,0 +1,65 @@
...
+[tool.pip-tools.pip-sync]
+dry-run = true
+```
+
+## Key and Value Syntax
+
+Keys are canonically the same as long option names, but with the leading `--` removed.
+Therefore, `--resolver` is configured under the name `resolver`.
+
+> [!NOTE]
```
webknjaz, line 53: Should this be something stronger, like a warning?

webknjaz (approved):
Short of the inline notes, I like it. Decide if you'll incorporate them and go for the merge.

docs/reference/configuration.md
sirosen: It's CWD by default (e.g., stdin input), but it reads `src_files` paths if present.

I think the logic is actually a little bit confusing for new users (`select_config_file()` in `piptools/utils.py` today, if you're looking for it).
I might want to change it in the future, but that would be a separate discussion.

docs/reference/configuration.md
sirosen: I'd like to start it as a note because nothing will _break_ if you use underscores. We can slowly escalate it -- at some point I'd like to deprecate and remove that behavior. IMO there should be only one way to spell each option.

changelog.d/+97e08771.doc.md
sirosen: Yeah, I'll give it a quick try. It should work but I haven't always had refs work properly in myst. :shrug:

sirosen: I looked at it again; the issue is that there's no ref target named `pip-tools`. And we don't have a reference named `index` or similar either. I'll skip for now.

Maybe we should add `.. _pip-tools:` to our index doc?

PR #2406, opened 2026-05-22
Fix malformed nitpick ignore rules; add ignores to let doc builds pass

sirosen:
Based on #2394 , this changeset includes fixes to the nitpick ignore regexes.

---

The nitpick ignore regex list included `pip.*`. Because this is a simple regex match, it matches `piptools...` as well as `pip...`.

Fixing the regex to be correctly scoped reveals many nitpick errors inside of piptools, which are suppressed here in order to make it possible to track them down and fix later.

In addition to generally fixing the regexes,

- `InvalidName` and `NormalizedName`, from within `packaging`, are added to nitpick ignores

- the regexes are sorted and organized into sections

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/changelog.d/+fcb3437a.contrib.md
+++ b/changelog.d/+fcb3437a.contrib.md
@@ -1,0 +1,1 @@
+Fixed incorrect nitpick ignore regexes -- by {user}`sirosen`.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -95,16 +95,24 @@ linkcheck_ignore = [
 ]
 
 nitpick_ignore_regex = [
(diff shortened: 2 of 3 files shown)


webknjaz (approved):
It's weird to see stdlib in ignores but I guess that's out of the scope.. I also think we should link as many things as possible instead of ignoring. But that's something for the future.

sirosen:
Yep, full agreement. I think of this as an unfortunate necessity.

I have a suspicion that it will not be possible to completely eliminate the ignores, but we should do the best we can. At least now they're accurate.

PR #2407, opened 2026-05-25
Refactor docs to cover env vars and introduce How-Tos

sirosen:
*Quick Review Link*: [PR Build](https://pip-tools--2407.org.readthedocs.build/en/2407/)

There are two changes included here, both in service of the same short-term
goal: remove `CUSTOM_COMPILE_COMMAND` from the readme.

Longer term, we are setting up two new pieces of documentation we can build
upon:
- a page for environment variables
- a How-Tos section

Commits:

- **Introduce a new page for documenting env vars**
- **Add a How-To doc for controlling headers**

Inspired by #2382

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist


sirosen:
@webknjaz, if you can make time to review this, I would greatly appreciate it.

I want to get this in and then cut 7.6.0 -- I think including this new doc before releasing is worth our time, and our other changes can wait.

I had initially aspired to do this all close to PyCon US, but time got away from me and the release ended up sitting.

As always, I'm happy to take feedback on any part of this! :grin:

sirosen:
Because the OP for #1702 just self-closed, and we've gotten a lot of slop PRs for that issue, I took action and wrote the doc for completions the way I wanted it, included now in this PR. I updated the changelog note.

I pulled in `sphinx-design` because the appropriate layout uses tabs.

@webknjaz, can I bump this for your attention? I'd really like to get it merged, as it will unblock more doc improvements.

docs/reference/environment-variables.md
```diff
@@ -0,0 +1,35 @@
...
+As a result, all of the environment variables supported by `pip` are supported
+when using `pip-tools`.
+For example, `PIP_BUILD_CONSTRAINT` can be used to control the environment used
+for package builds.
+
+See the
+[`pip` documentation on environment variables][pip-env-vars]
+for full details.
+
+[pip-env-vars]: https://pip.pypa.io/en/stable/topics/configuration/#environment-variables
```
webknjaz, line 19: Something like this may work:
```suggestion
{external+pip:ref}`` `pip` documentation on environment variables <0-environment-variables> ``
```
(I haven't checked)

docs/reference/environment-variables.md
```diff
@@ -0,0 +1,35 @@
...
+
+`pip-tools` supports a number of environment variables as ways of customizing behavior.
+
+## `pip` Environment Variables
+
+{external+pip:doc}`pip <index>` is used inside of `pip-tools`.
+
+As a result, all of the environment variables supported by `pip` are supported
+when using `pip-tools`.
+For example, `PIP_BUILD_CONSTRAINT` can be used to control the environment used
```
webknjaz: Not exactly the same but `` :std:ref:`build constraints` `` does exist in there (https://webknjaz.github.io/intersphinx-untangled/pip.pypa.io/) so https://pip.pypa.io/en/latest/user_guide/#build-constraints is kinda linkable.

*(should we convince pip to actually start having the `:envvar:` definitions?)*

docs/reference/environment-variables.md
```diff
@@ -0,0 +1,35 @@
...
+
+See the
+[`pip` documentation on environment variables][pip-env-vars]
+for full details.
+
+[pip-env-vars]: https://pip.pypa.io/en/stable/topics/configuration/#environment-variables
+
+## `pip-tools` Environment Variables
+
+`CUSTOM_COMPILE_COMMAND`
```
webknjaz: Let's use the `envvar` directives  for these:
* https://www.sphinx-doc.org/en/master/usage/domains/standard.html#directive-envvar
* https://myst-parser.readthedocs.io/en/latest/syntax/roles-and-directives.html#syntax-directives

docs/reference/environment-variables.md
```diff
@@ -0,0 +1,35 @@
...
+: This setting controls the command printed in the header for files generated
+and updated by `pip-compile`. Set it in order to replace the default behavior --
+which is to use the command which was run -- with a custom command.
+
+`PIP_TOOLS_CACHE_DIR`
+: The directory to use for `pip-tools`' caching. Defaults to a directory named
+`pip-tools/` inside of the user's cache directory.
+This can also be set by the `--cache-dir` option.
```
webknjaz, line 32: One day, we'll be able to use https://www.sphinx-doc.org/en/master/usage/referencing.html#role-option here too.

docs/how-to/index.md
```diff
@@ -0,0 +1,10 @@
+# How-Tos
+
+These documents guide you through specific workflows using `pip-tools`!
```
webknjaz, line 3: Instead of hardcoding `pip-tools`, we could use substitutions FYI.

For RST, this works via epilog: https://github.com/ansible/awx-plugins/blob/7c6a7cf97ec1e2ed3c36c0dd2e0dbb29436dea78/docs/conf.py#L66-L69
But for MD, there's a separate MyST setting: https://github.com/ansible/awx-plugins/blob/7c6a7cf97ec1e2ed3c36c0dd2e0dbb29436dea78/docs/conf.py#L121-L126

docs/how-to/customize-compiled-headers.md
```diff
@@ -0,0 +1,52 @@
...
+
+Users have several options which allow them to customize the generation of this
+header.
+
+## Suppress the Header
+
+If you don't want the header at all, pass `--no-header` and `pip-compile` will
+omit it.
+
+```shell
```
webknjaz: Isn't this
````suggestion
```console
````
?

webknjaz (approved):
Looks nice overall, I've marked a few possibilities to improve things, though. Up to you. Feel free to postpone into follow-ups.

docs/how-to/customize-compiled-headers.md
sirosen: Aghghgh! 😂

Yes, it is, let me apply as part of the general tweaks and final tune-up.

docs/how-to/index.md
sirosen: I'm not totally averse to it, but I don't see the advantage on this page, since it's not generic / reusable?

I'm going to defer any change for this until later, but am open to it in principle if you can help me see the benefit. 🙂

docs/how-to/customize-compiled-headers.md
webknjaz: There's also a `shell-session` alias:
https://pygments.org/docs/lexers/#pygments.lexers.shell.BashSessionLexer

docs/reference/environment-variables.md
sirosen: In this particular case, we could link the text `` `PIP_BUILD_CONSTRAINT` can be used ``. I think that it might actually be better as-is, because the point of this section is to document that env vars are passed through, not how to use this particular one. (I am undecided on this. The link is sort of nice? But also a slight distraction. I will try it both ways and look at the rendered result.)

I'm not sure how best to improve `pip`'s docs for envvars. The only missing thing that I'd want is a page in the reference section for it, with all of the vars listed. It would need to take the same text that is currently found on the various command pages, and link back to those various places. I'll raise the question informally with them, but don't want to file an issue without some level of buy-in about it.

docs/reference/environment-variables.md
sirosen: 🤞 I'll try it. I'm not sure if it will work, but there's one surefire way to find out!

docs/reference/environment-variables.md
sirosen: Are we not able to today? I was going to add the role -- I guess I'll see if there's a reason it doesn't work?

webknjaz: We don't declare our own CLI options in Sphinx properly. Only once we start doing that: https://www.sphinx-doc.org/en/master/usage/domains/standard.html#directive-program. https://pip-tools--2407.org.readthedocs.build/en/2407/reference/pip-compile/ is just a snippet with stdout of the `--help` output. Sphinx does not know of any options. I think tox has some plugin for extracting those from argparser or from click. Might need exploration.

docs/reference/environment-variables.md
sirosen: It can be made to work with a small tweak (remove the space between `>` and ` `` `), but it points to the wrong location.

I looked at the pip doc source and there is no explicit reference being provided, so I don't think that I should change this. Pointing directly to the stable docs anchor seems like the best path right now.

docs/reference/environment-variables.md
sirosen: I tried it, but it generated a cross-link to the pip CLI option. I'm leaving it out for now, but I assume that there's some way to use `{option}` and explicitly avoid intersphinx.

EDIT: The reply above (45m old!) just rendered for me. I get it now!

PR #2409, opened 2026-06-07
Avoid rereading inputs for preserved newlines

### LINKED ISSUE #2232: [regression][6.10.0] `pip-compile` hangs on named pipes as input due to double-read

marko1olo:
Fixes #2232.

This avoids rereading requirement input files when `--newline=preserve` chooses the writer newline. Requirement file contents are captured while pip is already parsing them, then reused as the fallback newline source after the output file and metadata files have been checked. Non-regular files and `-` are skipped in the direct filename pass, so FIFOs/stdout aliases are not opened just for newline detection.

This keeps the existing precedence for an existing output file, preserves metadata-file behavior, and avoids changing requirement parsing paths or relative include semantics.

##### Contributor checklist

- [x] A change note is created in `changelog.d/`.

Checks run locally:

- `python -m pytest tests\test_cli_compile.py::test_preserve_newline_from_parsed_input_without_reopening_source tests\test_cli_compile.py::test_determine_linesep_skips_non_regular_files -q`
- `python -m pytest tests\test_cli_compile.py -k "newline or linesep" -q`
- `pre-commit run --color=never --files piptools/scripts/compile.py tests/test_cli_compile.py changelog.d/2232.bugfix.md`
- `python -m compileall -q piptools\scripts\compile.py tests\test_cli_compile.py`
- `git diff --check`

### DIFF
--- a/changelog.d/2232.bugfix.md
+++ b/changelog.d/2232.bugfix.md
@@ -1,0 +1,3 @@
+Fixed `pip-compile --newline=preserve` rereading requirement input files while
+choosing the output newline, which could hang when the input was a named pipe
+-- by {user}`marko1olo`.
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 import itertools
 import os
 import shlex
@@ -10,7 +11,7 @@ from pathlib import Path
 import click
 from build import BuildBackendException
 from click.utils import LazyFile, safecall
-from pip._internal.req import InstallRequirement
+from pip._internal.req import InstallRequirement, req_file
 from pip._internal.utils.misc import redact_auth_from_url
 
 from .._compat import canonicalize_name, parse_requirements, tempfile_compat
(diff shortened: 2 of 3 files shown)


sirosen:
Monkeypatching pip internals is questionable at the best of times. I see multiple additional signals that this is slop.

We have had multiple patches for this bug of inferior quality. It's on my TODO list but I might move it up simply to get rid of the spam.

PR #2410, opened 2026-06-08
Update to the latest version of reusable-tox + codecov-action

sirosen:
- Update to the latest version (pinned)
- Note in the comment that we're tracking 'unstable/v1'

---

@webknjaz, I think this is what you were referring to in #2392 ?
I looked at the bump in the reusable-tox workflow and I don't expect we need any other changes.

I don't believe it needs a contrib doc fragment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -277,7 +277,7 @@ jobs:
     needs:
     - pre-setup  # transitive, for accessing settings
 
-    uses: tox-dev/workflow/.github/workflows/reusable-tox.yml@617ca35caa695c572377861016677905e58a328c  # yamllint disable-line rule:line-length
+    uses: tox-dev/workflow/.github/workflows/reusable-tox.yml@1bb961580e20073f66ccb9024f248357f8c8fecb  # unstable/v1  # yamllint disable-line rule:line-length
     with:
       cache-key-for-dependency-files: >-
         ${{ needs.pre-setup.outputs.cache-key-for-dep-files }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -417,7 +417,7 @@ jobs:
         if: >-
           !cancelled()
           && !inputs.cpython-pip-version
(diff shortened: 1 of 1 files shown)


sirosen:
I see we have direct use of `codecov` in this repo too. I'll make some further changes in a bit. I'm setting up and testing out `pinact`, just to have another option in my tool-box for how we handle these updates.

webknjaz:
I wanted to look at David's `gha-update` too at some point too. FWIW, I imagine the direct bump would be separate from the reusable workflow..

webknjaz (approved):
The fact that it's on the unstable branch isn't very important. Just because there's no final interface at the moment, I'm not versioning the workflow for now, and so it must be pinned to hashes no matter what..

sirosen:
> The fact that it's on the unstable branch isn't very important. Just because there's no final interface at the moment, I'm not versioning the workflow for now, and so it must be pinned to hashes no matter what..

Agree and disagree (mostly agree). The branch note is a useful hint because it says "go here to get the latest hash". Otherwise, it's not clear what the hash is meant to be tracking. So it's important because it helps me (or theoretical new maintainers) know where to look for updates.

> I wanted to look at David's `gha-update` too at some point too. FWIW, I imagine the direct bump would be separate from the reusable workflow..

I already use `gha-update` and have mostly good experiences with it. I'm keepin it around.

It's not "surgical" though, if I want to update only one action. I found that `pinact` is handy for this, e.g.:
```
$ pinact run -u -i 'codecov'
.github/workflows/ci.yml:420
-         uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2  # v6.0.0
+         uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f  # v7.0.0
.github/workflows/ci.yml:535
-         uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2  # v6.0.0
+         uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f  # v7.0.0
```

I just applied this.

PR #2413, opened 2026-06-15
🧪🚑 Enable `security-events` @ cron+release

webknjaz:
This patch sets the top-level `security-events: write` privilege in the nightly and release GHA CI/CD workflows where they call `ci.yml`, which in turn calls the upstream Zizmor workflow requiring it.

The change intends to fix the following error [[1]] we've recently started encountering due to internal changes within GH:

```console
Invalid workflow file: .github/workflows/cron.yml#L9
The workflow is not valid. .github/workflows/cron.yml (Line: 9, Col: 3): Error calling workflow 'jazzband/pip-tools/.github/workflows/ci.yml@91636f5a1b49f8ff55b96088e43aa1a3a5dd4b2b'. The nested job 'zizmor' is requesting 'security-events: write', but is only allowed 'security-events: none'.
```

[1]: https://github.com/jazzband/pip-tools/actions/runs/27537892831/workflow

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,6 +59,10 @@ jobs:
         github.event.action == 'published'
         && 'release' || 'nightly'
       }}]
+
+    permissions:
+      security-events: write  # Needed by the nested Zizmor workflow
+
     uses: ./.github/workflows/ci.yml
     with:
       release-version: ${{ github.ref_name }}
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -8,6 +8,10 @@ on:
 jobs:
   main:
     name: CI
+
+    permissions:
+      security-events: write  # Needed by the nested Zizmor workflow
+
     uses: ./.github/workflows/ci.yml
     with:
       cpython-versions: >-


.github/workflows/release.yml
```diff
@@ -59,6 +59,10 @@ jobs:
         github.event.action == 'published'
         && 'release' || 'nightly'
       }}]
+
+    permissions:
```
webknjaz, line 63: @sirosen since this redefines default permissions and doesn't extend them, we may end up having to list some more. Not sure which ones exactly.

webknjaz:
@sirosen looks like the pytest 9.0.3 -> 9.1.0 bump revealed some duplicate parametrization issues, eh? https://github.com/jazzband/pip-tools/actions/runs/27544376656/job/81413891673?pr=2413#step:9:3049

Here's the difference between the last `main` commit's merge queue run and this PR:
```diff
- pipsupported: build==1.5.0,cffi==1.18.0.dev0,click==8.4.1,execnet==2.1.2,flit_core==3.12.0,greenlet==0.4.13,hpy==0.9.1.dev61+g79fb330d,iniconfig==2.3.0,packaging==26.2,pip==26.1,pip-tools @ file:///home/runner/work/pip-tools/pip-tools/.tox/.tmp/package/1/pip_tools-0.1.dev1.tar.gz#sha256=f0eb3a18726f4bd5e71c54264f4d97daa477a4e2c6957dc853c8a8bc0eada67b,pluggy==1.6.0,poetry-core==2.4.1,Pygments==2.20.0,pyproject_hooks==1.2.0,pytest==9.0.3,pytest-mock==3.15.1,pytest-rerunfailures==16.3,pytest-xdist==3.8.0,setuptools==80.0.0,tomli_w==1.2.0,wheel==0.47.0
+ pipsupported: build==1.5.0,cffi==1.18.0.dev0,click==8.4.1,execnet==2.1.2,flit_core==3.12.0,greenlet==0.4.13,hpy==0.9.1.dev61+g79fb330d,iniconfig==2.3.0,packaging==26.2,pip==26.1,pip-tools @ file:///home/runner/work/pip-tools/pip-tools/.tox/.tmp/package/1/pip_tools-0.1.dev1.tar.gz#sha256=6a5f90c05e92a362fbebae2ef3b44e46c69ad63d0655944bb331bbf643e5ecda,pluggy==1.6.0,poetry-core==2.4.1,Pygments==2.20.0,pyproject_hooks==1.2.0,pytest==9.1.0,pytest-mock==3.15.1,pytest-rerunfailures==16.3,pytest-xdist==3.8.0,setuptools==80.0.0,tomli_w==1.2.0,wheel==0.47.0
```

webknjaz:
This is probably related to the `legacy_resolver_only` and `backtracking_resolver_only` definitions and `current_resolver` being parametrized...

I'm not going to have time to look into this, but would something like I did here https://github.com/aio-libs/multidict/blob/8b7c4d8e827928b27aa4d7fbe5f93148cd64a805/tests/conftest.py#L159-L208 help?

.github/workflows/release.yml
sirosen: 👍 yep, I'm happy to with out what we need for this. I've been meaning to refresh my memory of what the valid permissions values are.

I'll do some reading as part of handling the issues you've uncovered here.

sirosen (commented):
I'll dig in and see what I can figure out here.
(I'm on a train; it's hard to read all the details on my phone) 

I'm hoping that I can write a separate PR which resolves the pytest issues, and then a third PR which merges it together with this one.

sirosen (approved):
I thought we might need `contents: read`, but I now don't believe so -- in other repos where I've set `permissions: {}`, I'm still able to clone. As you say, we might need more permissions, but it's hard to verify up front. I'm pretty sure this is good for normal builds. I may want to review the release workflow again before trying a release.

pytest 9.1.1 fixed the issue in 9.1.0 . I worked up a fix anyway, but I'm not sure if we want it, so I put it in a draft PR.

webknjaz:
> I thought we might need `contents: read`, but I now don't believe so


Technically, it's needed for retrieving data+repo from GH but it's implicit for public repos. We'd hit a problem if the repo was private (private forks might face it still).

PR #2415, opened 2026-06-20
Replace indirect parametrization with custom marker (fix tests on pytest==9.1.0)

sirosen:
Indirect fixture parametrization was used to define
"legacy resolver only" and "backtracking resolver only" decorators, which
controlled the "current_resolver" fixture. Under pytest 9.1.0 , this
broke due to a pytest bug, which was noted as fixed in 9.1.1.

Even so, using indirect parametrization in this way is really expressing
conditionality in the tests; it's not clear that this is the best way to
do so.

A custom marker + a fixture can provide the same functionality in this
case, and works on pytest 9.1.0 . It is arguable which way of expressing
the condition is clearer, but using markers to control tests is at least
somewhat normative in pytest. This does result in more reported (skipped)
tests, which seems to be necessary in order to pick up on the fixture
value.

##### Contributor checklist

- [x] ~A change note is created in `changelog.d/` (see [`changelog.d/README.md`]
      for instructions) or the PR text says "no changelog needed".~

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -29,23 +29,17 @@ from piptools.utils import COMPILE_EXCLUDE_OPTIONS
 
 from .constants import MINIMAL_WHEELS_PATH, PACKAGES_PATH
 
-legacy_resolver_only = pytest.mark.parametrize(
-    "current_resolver",
-    ("legacy",),
-    indirect=("current_resolver",),
-)
-
-backtracking_resolver_only = pytest.mark.parametrize(
-    "current_resolver",
-    ("backtracking",),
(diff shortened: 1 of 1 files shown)


sirosen:
The marker declaration doesn't seem to be working. Strange that it worked for me locally.

sirosen:
I'm self-closing as I don't think it's worth pursuing this. I was surprised that my fix didn't apply cleanly in CI, but since pytest pushed a fix I'm not going to worry about it.

If pytest 10 breaks us, I can dig into it then.

PR #2419, opened 2026-07-11
Fix the `fake_with_deps` test package

sirosen:
One of the local test packages, `fake_with_deps` has a nontrivial suite of dependencies. Among these, `pyzmq` was pinned to an older version which fails to build with the latest `scikit-build-core`, resulting in testsuite failures.

It is not clear what the impacts on testing would be of making more dramatic changes to `fake_with_deps`, so this change is limited to updating `pyzmq` to the latest version. It is given no upper bound because there is no clear value which could be used for one.

Additionally, update the test package license metadata to avoid warnings.

The `fake_with_deps` package used the deprecated `license` style for setuptools, resulting in warnings. This contaminates test output and makes debugging more difficult.

The usage was noted as needed for Python 3.8 test compatibility, but we no longer support 3.8 , so it is safe to remove.

---

Given that this impacts how the testsuite interacts with the network, it could impact repackagers or other downstream testers.

However, I don't feel that a changelog note is needed, as this is purely a testsuite change.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/tests/test_data/packages/fake_with_deps/pyproject.toml
+++ b/tests/test_data/packages/fake_with_deps/pyproject.toml
@@ -6,9 +6,7 @@ build-backend = "setuptools.build_meta"
 name = "fake_with_deps"
 description = "Fake package with dependencies"
 authors = [{name = "jazzband"}]
-# license = "0BSD"
-# Deprecated license table for Python 3.8's latest setuptools compatibility:
-license = {text = "0BSD"}
+license = "0BSD"
 version = "0.1"
 dependencies = [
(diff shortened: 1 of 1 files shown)


webknjaz (approved):
FWIW, if this may impact downstreams, a `packaging` change note could be reasonable to have.

sirosen:
If I were documenting this one, I would probably put it under `contrib`, since it's about a change to the testsuite. But I think test changes don't really belong in the changelog unless they are dramatic, since most consumers -- including downstreams -- don't care about the details.

PR #2426, opened 2026-07-22
Fix hash generation from PyPI JSON API failing when --index-url ends with a slash

### LINKED ISSUE #1669: Get hashes from PyPI JSON API fails when index URL ends with slash
When using `https://pypi.org/simple/` instead of `https://pypi.org/simple` as index URL, pip-compile falls back to hashing files.

This is caused by passing the index URL to pip's `PackageIndex` and then calling `package_index.pypi_url`: https://github.com/jazzband/pip-tools/blob/1e05d0028ce16d4b2de53ae38a33df7feb84e800/piptools/repositories/pypi.py#L264-L269

- pip itself only uses `PackageIndex` to initialize the constants [`PyPI`](https://github.com/pypa/pip/blob/e2c9c40edeb65b9d26c13a12dbe5642dc2e00fd0/src/pip/_internal/models/index.py#L25) and [`TestPyPI`](https://github.com/pypa/pip/blob/e2c9c40edeb65b9d26c13a12dbe5642dc2e00fd0/src/pip/_internal/models/index.py#L26-L28).  I don't know what the recommended way is to convert between Simple API URLs and JSON API URLs.
- [PEP 503 – Simple Repository API](https://peps.python.org/pep-0503/) says that “PyPI’s base URL is https://pypi.org/simple/”.  I don't see why the trailing slash would be considered wrong when used in an index URL.
- Sure, this shouldn't happen a lot for people who use PyPI.org.  I've noticed it because AWS CodeArtifact [generates/documents index URLs](https://docs.aws.amazon.com/codeartifact/latest/ug/python-configure-pip.html) that end with `/simple/` and then `package_index.simple_url` as used in jazzband/pip-tools#1556 also doesn't work.
  <details><summary>Code Example</summary>

  ```python
  from pip._internal.commands import create_command
  from pip._internal.models.index import PackageIndex

  command = create_command("install")
  options, _ = command.parse_args(["-i", "https://pypi.org/simple/"])
  session = command._build_session(options)
  finder = command._build_package_finder(options=options, session=session)

  for index_url in finder.search_scope.index_urls:
      package_index = PackageIndex(url=index_url, file_storage_domain="")
      print(package_index.simple_url)  # prints: https://pypi.org/simple/simple
      print(package_index.pypi_url)  # prints: https://pypi.org/simple/pypi
  ```
  </details>

#### Environment Versions

1. OS Type: Linux
1. Python version: `$ python -V`: Python 3.10.6
1. pip version: `$ pip --version`: pip 22.2.2
1. pip-tools version: `$ pip-compile --version`: pip-compile, version 6.8.0

#### Steps to replicate

1. `echo "click" > requirements.in`
2. `pip-compile --generate-hashes --verbose --index-url=https://pypi.org/simple/ requirements.in`

#### Expected result

<details><summary>Log output</summary>

```
Using indexes:
  https://pypi.org/simple/

                          ROUND 1
Current constraints:
  click (from -r requirements.in (line 1))

Finding the best candidates:
  found candidate click==8.1.3 (constraint was <any>)

  click==8.1.3              requires -

Generating hashes:
  click

#
# This file is autogenerated by pip-compile with python 3.10
# To update, run:
#
#    pip-compile --generate-hashes --index-url=https://pypi.org/simple/ requirements.in
#
click==8.1.3 \
    --hash=sha256:7682dc8afb30297001674575ea00d1814d808d6a36af415a82bd481d37ba7b8e \
    --hash=sha256:bb4d8133cb15a609f44e8213d9b391b0809795062913b383c62be0ee95b1db48
    # via -r requirements.in
```
</details>

#### Actual result

<details><summary>Log output</summary>

```
Using indexes:
  https://pypi.org/simple/

                          ROUND 1
Current constraints:
  click (from -r requirements.in (line 1))

Finding the best candidates:
  found candidate click==8.1.3 (constraint was <any>)

  click==8.1.3              requires -

Generating hashes:
  click
    Couldn't get hashes from PyPI, fallback to hashing files
    Hashing click-8.1.3-py3-none-any.whl
      |████████████████████████████████| 100%
    Hashing click-8.1.3.tar.gz
      |████████████████████████████████| 100%

#
# This file is autogenerated by pip-compile with python 3.10
# To update, run:
#
#    pip-compile --generate-hashes --index-url=https://pypi.org/simple/ requirements.in
#
click==8.1.3 \
    --hash=sha256:7682dc8afb30297001674575ea00d1814d808d6a36af415a82bd481d37ba7b8e \
    --hash=sha256:bb4d8133cb15a609f44e8213d9b391b0809795062913b383c62be0ee95b1db48
    # via -r requirements.in
```
</details>

agu2347:
Fixes #1669.

`PyPIRepository._get_project()` passes each configured index URL to pip's `PackageIndex`, then reads its `.pypi_url` property to build the JSON API request URL. `PackageIndex` derives `pypi_url` via `urllib.parse.urljoin(url, "pypi")`, whose result differs depending on whether `url` ends with a trailing slash:

```
urljoin("https://pypi.org/simple", "pypi")
  -> "https://pypi.org/pypi"            (correct)
urljoin("https://pypi.org/simple/", "pypi")
  -> "https://pypi.org/simple/pypi"      (wrong: doubled path)
```

An index URL with a trailing slash -- which PEP 503 itself uses as the canonical form ("PyPI's base URL is `https://pypi.org/simple/`"), and which e.g. AWS CodeArtifact generates by default -- therefore produces a malformed JSON API URL. The request 404s, and `pip-compile` silently falls back to the much slower path of downloading and hashing each file directly, without any indication to the user of why, exactly as shown in the issue.

**Fix**: strip trailing slash(es) from each index URL before constructing `PackageIndex`, so `urljoin` (and therefore `pypi_url`) behaves consistently and correctly regardless of how the user formatted `--index-url`. This fixes it for any PEP 503-style index, not just pypi.org specifically, without needing changes to pip itself.

**Testing**: verified directly against pip's `PackageIndex` that the fix produces the correct `https://pypi.org/pypi` URL for both trailing-slash and no-trailing-slash inputs, matching the exact `urljoin` behavior difference diagnosed in the issue.

Added a regression test that configures a `PyPIRepository` with a trailing-slash `--index-url` (matching the issue's exact repro command) and asserts the URL actually requested from the mocked session is the correct, non-doubled JSON API URL. I confirmed the test fails with the original code (requests the doubled `.../simple/pypi/...` URL) and passes with the fix. Ran the full existing `test_repository_pypi.py` suite: 29 passed (28 baseline + 1 new), no regressions.

### DIFF
--- a/changelog.d/1669.bugfix.md
+++ b/changelog.d/1669.bugfix.md
@@ -1,0 +1,4 @@
+Fix `--index-url` values ending with a trailing slash (e.g.
+`https://pypi.org/simple/`) causing hash generation to silently fall
+back to hashing downloaded files instead of fetching hashes from the
+PyPI JSON API, due to a malformed (doubled) API URL being constructed.
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -273,7 +273,15 @@ class PyPIRepository(BaseRepository):
         API reference: https://warehouse.readthedocs.io/api-reference/json/
         """
         package_indexes = (
-            PackageIndex(url=index_url, file_storage_domain="")
+            # Strip trailing slash(es): pip's PackageIndex derives
+            # pypi_url via urllib.parse.urljoin(url, "pypi"), which
+            # produces a different (and, with a trailing slash,
+            # incorrect/nested) result depending on whether url ends
+            # with a slash -- e.g. "https://pypi.org/simple/" yields
+            # ".../simple/pypi" instead of ".../pypi". Normalizing here
+            # ensures a consistent, correct pypi_url regardless of how
+            # the user formatted --index-url. See GH #1669.
+            PackageIndex(url=index_url.rstrip("/"), file_storage_domain="")
             for index_url in self.finder.search_scope.index_urls
         )
         for package_index in package_indexes:
--- a/tests/test_repository_pypi.py
+++ b/tests/test_repository_pypi.py
@@ -429,6 +429,44 @@ def test_get_project__handles_404(from_line, tmpdir, monkeypatch, pypi_repositor
     assert actual_data is None
 
 
+def test_get_project__index_url_trailing_slash(from_line, tmpdir, monkeypatch):
+    """
+    Regression test for
+    https://github.com/jazzband/pip-tools/issues/1669
+
+    An --index-url ending with a trailing slash (e.g.
+    "https://pypi.org/simple/", matching PEP 503's own description of
+    PyPI's base URL) must still produce a correct, non-doubled JSON
+    API URL like "https://pypi.org/pypi/<name>/json" -- not
+    "https://pypi.org/simple/pypi/<name>/json".
+    """
+    pypi_repository = PyPIRepository(
+        [
+            "--index-url",
+            "https://pypi.org/simple/",
+            "--use-deprecated",
+            "legacy-resolver",
+        ],
+        cache_dir=(tmpdir / "pypi-repo"),
+    )
+
+    requested_urls = []
+
+    class MockResponse:
+        status_code = 404
+
+    def mock_get(url, *args, **kwargs):
+        requested_urls.append(url)
+        return MockResponse()
+
+    monkeypatch.setattr(pypi_repository.session, "get", mock_get)
+    ireq = from_line("fake-package==0.1")
+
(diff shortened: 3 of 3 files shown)


sirosen:
Please read our contributing documentation and restore the PR template.

PR #2427, opened 2026-07-23
fix: do not echo resolved requirements to stderr by default

### LINKED ISSUE #1987: Verbose output without final result when outputting to a file

Having verbose output is nice so progress can be viewed, but when writing the result to a file, it's still output to the console

Verbose output can be used without dumping all of the result to the console, when writing to a file

#### Alternative Solutions

Flag to turn on progress output

MohammedAnasNathani:
## Summary

When `pip-compile` writes through `--output-file`, it currently also mirrors every resolved requirement line to stderr at default verbosity (`log.info`). That makes verbose/progress output hard to separate from the lockfile body and matches the confusion reported in #1987.

### Decision (aligned with maintainer discussion on #2342 / #1987)

- **Default verbosity:** write resolved lines **only** to the output file.
- **`-v` / `--verbose` (or higher):** also echo those lines to stderr via `log.debug` so the old “see the result while compiling” behavior is still available.
- **Dry-run:** unchanged — still always prints the result via `log.log`.

Implementation is a one-line behavior change in `_FileLineWriter.__call__` (`log.info` → `log.debug`), plus test updates so successful non-dry-run cases assert against the output file, and a dedicated regression test.

Fixes #1987.

### Motivation / use case

I hit this while compiling with `--output-file` and wanting clean stderr for progress/warnings only. Redirecting stdout/stderr was not enough because the lockfile body was duplicated on stderr at default verbosity.

### LLM use

I used an AI coding assistant to help draft the test updates and PR text. I reviewed the maintainer notes on #2342 (verbosity threshold should be `-v`, not only `-vv`), implemented that threshold deliberately, and ran the local `tests/test_cli_compile.py` suite (534 passed; 2 failures only from a local missing `pip` executable for `test_realistic_complex_sub_dependencies`, unrelated to this change).

##### Contributor checklist

- [x] A change note is created in `changelog.d/` (`changelog.d/1987.bugfix.md`).

##### Maintainer checklist

### DIFF
--- a/changelog.d/1987.bugfix.md
+++ b/changelog.d/1987.bugfix.md
@@ -1,0 +1,4 @@
+Resolved requirement lines written to an ``--output-file`` are no longer
+mirrored to stderr at default verbosity. Pass ``-v``/``--verbose`` (or higher)
+to echo them while compiling. Dry-run output is unchanged.
+-- by {user}`MohammedAnasNathani`
--- a/piptools/writer.py
+++ b/piptools/writer.py
@@ -357,7 +357,11 @@ class _FileLineWriter:
         )
 
     def __call__(self, line: str) -> None:
-        log.info(line)
+        # Echo resolved lines only with -v/--verbose (or higher). At default
+        # verbosity the output file is the sole destination for the result,
+        # which keeps stderr for progress/logging and avoids the "double
+        # output" confusion with -o / --output-file (see #1987).
+        log.debug(line)
         self.dst_file.write(unstyle(line))
         self.dst_file.write("\n")
 
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -148,7 +148,9 @@ def test_default_pip_conf_read(pip_with_index_conf, runner):
 
     # check that we have our index-url as specified in pip.conf
     assert "Using indexes:\n  http://example.com" in out.stderr
-    assert "--index-url http://example.com" in out.stderr
+    with open("requirements.txt") as _compiled:
+        _compiled_text = _compiled.read()
+    assert "--index-url http://example.com" in _compiled_text
 
 
 def test_command_line_overrides_pip_conf(pip_with_index_conf, runner):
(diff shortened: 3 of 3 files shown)


changelog.d/1987.bugfix.md
```diff
@@ -0,0 +1,4 @@
+Resolved requirement lines written to an ``--output-file`` are no longer
+mirrored to stderr at default verbosity. Pass ``-v``/``--verbose`` (or higher)
+to echo them while compiling. Dry-run output is unchanged.
+-- by {user}`MohammedAnasNathani`
```
webknjaz, line 4: This will render poorly w/o a line feed
```suggestion

-- by {user}`MohammedAnasNathani`
```

changelog.d/1987.bugfix.md
webknjaz, line 1: A few pieces of syntax here use RST instead of expected MD. The preview could be checked at RTD or locally.

piptools/writer.py
```diff
@@ -357,7 +357,11 @@ def __init__(self, dst_file: _t.BinaryIO, linesep: str) -> None:
         )
 
     def __call__(self, line: str) -> None:
-        log.info(line)
+        # Echo resolved lines only with -v/--verbose (or higher). At default
+        # verbosity the output file is the sole destination for the result,
+        # which keeps stderr for progress/logging and avoids the "double
+        # output" confusion with -o / --output-file (see #1987).
```
sirosen, line 363: Opinions may vary, but I consider this comment extremely low-value. It makes the code harder to read and maintain.

```suggestion
```

tests/test_cli_compile.py
```diff
@@ -148,7 +148,9 @@ def test_default_pip_conf_read(pip_with_index_conf, runner):
 
     # check that we have our index-url as specified in pip.conf
     assert "Using indexes:\n  http://example.com" in out.stderr
-    assert "--index-url http://example.com" in out.stderr
+    with open("requirements.txt") as _compiled:
+        _compiled_text = _compiled.read()
+    assert "--index-url http://example.com" in _compiled_text
```
sirosen, line 153: I don't love the way that these tests are stylistically different from the others based only on whether `tmp_path` vs `tmpdir` is used. I also find the use of underscore-prefixed names in local scope slightly strange; why is it named `_compiled`? Why not `compiled`?

Can we use the `read_text()` API in `tmpdir`-based tests as well, or switch to `tmp_dir` if that change is sufficiently small and surgical?

My preference would be to see a precursor PR to switch all of our `tmpdir` fixture usage to `tmp_path`, which is a beneficial modernization in its own right.

sirosen (changes requested):
Before we even consider merging this we should decide whether or not this change necessitates a version bump to 8.0 .

PR #2429, opened 2026-07-24
Improve normalization of index URLs for JSON API

sirosen:
Given a suite of index URLs in use, when pip-tools wants to call the JSON API, it currently is constructing an internal pip datatype -- meant to represent only 'pypi.org' and 'test.pypi.org' -- and using that to call `urllib.parse.urljoin`.

This has two significant downsides:

1. The code is much harder to understand -- rather than a well-documented stdlib function call, we need to look into pip internals.

2. Any normalization or cleaning of the input data is purely incidental.
   `urljoin()` happens to do some, but the overall effect is inconsistent behavior in resolving these URLs: `/simple` gets cleaned up, but `/simple/` does not (#1669).

By replacing this with a dedicated function, we can simplify the relevant code and add dedicated normalization logic. And because the logic is our own, we have a well-defined function to unit-test.

-----

#2426 is related but deserves no credit. The fix proposed in that PR is incorrect, but I needed to look carefully to confirm it. As a result of the review effort I put in, I understood the problem and worked on a good fix.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/changelog.d/1669.bugfix.md
+++ b/changelog.d/1669.bugfix.md
@@ -1,0 +1,4 @@
+Fix handling of index URLs ending in `/simple/` by improving URL normalization logic.
+These URLs would previously make `pip-compile` fail to use the PyPI JSON API.
+
+-- by {user}`sirosen`
--- a/changelog.d/1669.bugfix.md
+++ b/changelog.d/1669.bugfix.md
@@ -1,4 +1,5 @@
-Fix handling of index URLs ending in `/simple/` by improving URL normalization logic.
+Fixed handling of index URLs ending in `/simple/` by improving URL normalization
+logic.
 These URLs would previously make `pip-compile` fail to use the PyPI JSON API.
 
 -- by {user}`sirosen`
--- a/piptools/repositories/pypi.py
+++ b/piptools/repositories/pypi.py
@@ -6,6 +6,7 @@ import itertools
 import optparse
 import os
 import typing as _t
+import urllib.parse
 from collections.abc import Iterator
 from contextlib import contextmanager
 from shutil import rmtree
@@ -16,7 +17,6 @@ from pip._internal.commands import create_command
 from pip._internal.commands.install import InstallCommand
 from pip._internal.index.package_finder import PackageFinder
 from pip._internal.models.candidate import InstallationCandidate
-from pip._internal.models.index import PackageIndex
 from pip._internal.models.link import Link
 from pip._internal.models.wheel import Wheel
 from pip._internal.network.session import PipSession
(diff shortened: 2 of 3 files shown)


changelog.d/1669.bugfix.md
```diff
@@ -0,0 +1,4 @@
+Fix handling of index URLs ending in `/simple/` by improving URL normalization logic.
```
webknjaz:
```suggestion
Fixed handling of index URLs ending in `/simple/` by improving URL normalization logic.
```

piptools/repositories/pypi.py
```diff
@@ -520,3 +520,22 @@ def open_local_or_remote_file(link: Link, session: Session) -> Iterator[FileStre
...
+
+    It is not possible to accommodate all users perfectly -- we can ensure proper
+    behavior for pypi.org and test.pypi.org , and try not to break anyone else in the
+    process.
+    """
+    if url.endswith("/simple/"):
+        url = url[: -len("simple/")]
+    elif url.endswith("/simple"):
+        url = url[: -len("simple")]
+    return url
```
webknjaz, line 542: Would this read better?
```suggestion
    if url.endswith("/simple/"):
        return url.removesuffix("simple/")
    elif url.endswith("/simple"):
        return url.removesuffix("simple")

    return url
```

I'm trying to see if there's a simpler way to express the same.. In general, it seems like early return is better than re-assigning.
Even
```suggestion
    if url.endswith("/simple/"):
        return url[: -len("simple/")]
    elif url.endswith("/simple"):
        return url[: -len("simple")]

    return url
```
looks somewhat more appealing than a few `url =` in branches.

But what if we do
```suggestion
    url_base, simple_slug, trailing_slash = url.rpartition("/simple")

    if simple_slug and trailing_slash in {"", "/"}:
        return url_base

    return url
```
would this read nicer too?

In general, my preferences are suggestions 1 and 3.

tests/unit/repositories/test_pypi_repository.py
```diff
@@ -0,0 +1,41 @@
...
+
+@pytest.mark.parametrize(
+    "url",
+    (
+        "https://pypi.org/",
+        "https://test.pypi.org/",
+        "https://my-pypi-mirror.example.org/",
+    ),
+)
+def test_true_base_url_no_change(url: str) -> None:
```
webknjaz: I think this could as well be folded into that second test. Maybe the params could be wrapped into `pytest.param` with explicit `id=` for these. But up to you.

webknjaz (approved):
Good to go, unless you want to entertain some of my inline comments. They are totally optional.

piptools/repositories/pypi.py
sirosen: I like `removesuffix()` and early returns; both are very nice simplifications.

I much prefer (1) over (3) -- I think it's faster to read and understand what it's doing.

tests/unit/repositories/test_pypi_repository.py
sirosen: I'll take another look! One mistake I've made in past projects is trying to combine _too many_ cases into a single test. I don't think there are that many here, so I'll try it out.

sirosen: I tried it and immediately liked the result; applied!

PR #2433, opened 2026-07-25
Make pip-sync honor src_files from the config file

### LINKED ISSUE #2187: Specifying `src_files` for `pip-sync` in `pyproject.toml` is ignored.

Sanjays2402:
Fixes #2187

`pip-compile` reads `src_files` out of the click context `default_map` when none are given on the command line, but `pip-sync` never did — so `[tool.pip-tools.sync] src_files` was silently ignored and `pip-sync` always fell back to `requirements.txt` (or exited 2 when it was absent). This applies the same lookup in `pip-sync`; CLI arguments still take precedence.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/changelog.d/2187.bugfix.md
+++ b/changelog.d/2187.bugfix.md
@@ -1,0 +1,1 @@
+Made `pip-sync` respect `src_files` set in the config file, matching `pip-compile` -- by {user}`Sanjays2402`.
--- a/piptools/scripts/sync.py
+++ b/piptools/scripts/sync.py
@@ -48,6 +48,7 @@ Examples:
 
 
 @click.command(name="pip-sync")
+@click.pass_context
 @options.help_option(epilog=SYNC_EPILOG)
 @options.version
 @options.ask
@@ -69,6 +70,7 @@ Examples:
 @options.config
 @options.no_config
 def cli(
+    ctx: click.Context,
     ask: bool,
     dry_run: bool,
     force: bool,
@@ -90,6 +92,12 @@ def cli(
 ) -> None:
     """Synchronize virtual environment with requirements.txt."""
     log.verbosity = verbose - quiet
+
+    # If ``src_files`` was not provided as an input, but rather as config,
+    # it will be part of the click context ``ctx``.
+    # However, if ``src_files`` is specified, then we want to use that.
+    if not src_files and ctx.default_map and "src_files" in ctx.default_map:
+        src_files = tuple(ctx.default_map["src_files"])
 
     if not src_files:
         if os.path.exists(DEFAULT_REQUIREMENTS_FILE):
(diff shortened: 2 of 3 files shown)


sirosen:
I'm not clear on whether or not this is desirable behavior.

`src_files` has a different meaning between `pip-compile` and `pip-sync`. This behavior would result, I think, in loading `src_files` if specified in, for example, this config:
```toml
[tool.pip-tools]
src-files = ["pyproject.toml"]
```

Such a config is very clearly meant for `pip-compile`, not `pip-sync`, and loading it in this context is wrong. We may need more sophisticated config loading before the cited issue can be resolved.

sirosen:
I've noted in the originating issue exactly what is wrong here.

Sanjays2402:
You are right, and the shared-table case you called out would genuinely have been wrong with my first pass: `[tool.pip-tools] src-files = ["pyproject.toml"]` is compile config, and pip-sync had no business reading it.

Reworked to your spec: the lookup is now narrowed in `parse_config_file`, which drops `src_files` from the shared `[tool.pip-tools]` table when resolving config for `pip-sync`, so the value can only come from `[tool.pip-tools.sync]`. `pip-compile` is untouched and still reads it from either.

Tests cover both directions: `test_src_files_from_shared_config_section_is_ignored` is the negative case (top-level `src-files = ["pyproject.toml"]` is not picked up, sync falls back to `requirements.txt`), and it fails on the previous revision of this branch. The existing tests still assert `[tool.pip-tools.sync]` is honored and that CLI args win over config.

Also rewrapped the changelog fragment — pymarkdown was failing it on the 80 column line-length rule.

PR #2435, opened 2026-07-29
fix(compile): skip non-regular files in _determine_linesep (#2232)

### LINKED ISSUE #2232: [regression][6.10.0] `pip-compile` hangs on named pipes as input due to double-read

Mukller:
## Summary

Fixes #2232.

`pip-compile` hangs indefinitely when a named pipe (FIFO) is used as an input file and the output is written to stdout (`-o-` or `-`).

## Root Cause

`_determine_linesep()` with `strategy="preserve"` (the default) loops over output filenames to detect existing line endings by calling `open(fname, "rb").read()`. For regular files this is fine. For named pipes, `open()` succeeds — a FIFO is openable — but `.read()` then **blocks indefinitely** because the pipe was already fully consumed during the initial parse step and there is no new writer:

```
Step 1: pip-compile reads from named-pipe (FIFO) → pipe drained, EOF reached
Step 2: dependency resolution completes
Step 3: _determine_linesep() opens named-pipe again → blocks on .read() forever
```

The regression was introduced in #1652 (v6.10.0).

## Fix

Guard the `open()` with `Path(fname).is_file()`, which returns `True` only for regular files — `False` for FIFOs, sockets, directories, and non-existent paths. `Path` is already imported in `compile.py`.

```diff
 if strategy == "preserve":
     for fname in filenames:
         try:
+            if not Path(fname).is_file():
+                continue
             with open(fname, "rb") as existing_file:
                 existing_text = existing_file.read()
-        except FileNotFoundError:
+        except OSError:
             continue
```

The exception is also broadened from `FileNotFoundError` to `OSError` to handle permission errors, broken symlinks, and other IO failures that could occur despite the `is_file()` pre-check.

## Test Plan

- [ ] `mkfifo named-pipe && echo 'requests' > named-pipe &; pip-compile -o- named-pipe` — no longer hangs
- [ ] Existing tests for `_determine_linesep()` pass
- [ ] `pip-compile requirements.in` (regular file) — unchanged behaviour

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -56,9 +56,11 @@ def _determine_linesep(
     if strategy == "preserve":
         for fname in filenames:
             try:
+                if not Path(fname).is_file():
+                    continue
                 with open(fname, "rb") as existing_file:
                     existing_text = existing_file.read()
-            except FileNotFoundError:
+            except OSError:
                 continue
             if b"\r\n" in existing_text:
                 strategy = "CRLF"


Mukller (commented):
## Code Review

### Bug Trace

```
# User runs: mkfifo named-pipe
#            pip-compile -o- ./named-pipe
#            echo 'requests' > named-pipe  (in another shell)

# compile.py step 1 — parse:
# open('./named-pipe') → reads 'requests\n' → pipe now empty/EOF

# compile.py step 2 — resolve:
# ...dependency resolution completes...

# compile.py step 3 — write:
# _determine_linesep(filenames=('-', './named-pipe'))
#   fname='-'  → FileNotFoundError (no file named '-')  → continue
#   fname='./named-pipe' → open('./named-pipe', 'rb') succeeds
#                        → .read() BLOCKS waiting for writer
#                        → pip-compile hangs indefinitely
```

### Fix Correctness

```python
if not Path(fname).is_file():
    continue
```

`Path.is_file()` returns `True` only for regular files:

| Path | `is_file()` | Handled by |
|---|---|---|
| Regular file (exists) | `True` | open + read |
| Regular file (missing) | `False` | `continue` |
| FIFO / named pipe | `False` | `continue` ← **the fix** |
| Socket | `False` | `continue` |
| Directory | `False` | `continue` |
| `-` (stdin string) | `False` | `continue` |

### Exception Broadening

`FileNotFoundError` → `OSError`: `is_file()` already handles the not-found case, so the only remaining exceptions would be OS-level errors (permission denied, broken symlink, NFS timeouts, etc.). `OSError` is the right base class here.

### Why `is_file()` and Not `stat.S_ISFIFO()`?

`Path.is_file()` is the correct positive check: it returns `True` only for regular files and implicitly rejects all non-regular special files (FIFOs, sockets, block devices, character devices) as well as non-existent paths. Negating a specific `S_ISFIFO()` check would still allow sockets and other blocking special files through.

### Scope

- 2 lines added inside `_determine_linesep()`, no other changes.
- No new imports needed (`Path` already imported at line 8).
- Behaviour for regular files and missing files is unchanged.

PR #2441, opened 2026-07-31
docs: document shell completion for pip-compile and pip-sync

### LINKED ISSUE #1702: [advice] pip-{compile,sync} have completions, why not add some document about it to README.md?

Abhi-DevHub:
Closes #1702

Adds a \Shell completion\ section to the README explaining how to enable shell completion for \pip-compile\ and \pip-sync\ using Click's completion mechanism.


sirosen:
I've added a note to the originating issue about why I do not want us to accept this kind of change at present.

If you review the history of PRs for this same issue, you will see that there have been several attempts to do this. Some provide better detail than what is suggested here. But the underlying blocker remains: we shouldn't keep putting more info into the project readme.

Issue #2200, opened 2025-07-09
Fix type checking failures


sirosen:

webknjaz:
I usually keep MyPy in pre-commit because then it's where all the other linters are, plus https://pre-commit.ci is actually able to run it.

Here's my typical config, let's try to adopt it: https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/27bdb93/.pre-commit-config.yaml#L191-L250.

The difference you're seeing must be related to one or more of the following:
1) `tests/test_data/` is excluded
2) some deps are pinned differently
3) Python 3.9
4) it passes file paths into MyPy and you pass the entire directory (this used to be most painful for me in the past)

If you look into my config, it's explicit about testing against different Python runtime, and it doesn't pass *any* paths into the program. Instead, it always runs it with the same context, which is the entire project (kinda).

All the configuration that can possibly be shared lives in the common config @ https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/27bdb93/.mypy.ini. This is important because if any other tools were to invoke MyPy outside pre-commit, they'd get the same settings (editor/IDE integrations often do this). The folder+file paths are also stored in that config, along with the overridable defaults (like the Python version).

Actually, here's a stricter example: https://github.com/ansible/awx-plugins/blob/db2cae7/.mypy.ini#L1-L58.

Try to copy that in here when you have a minute, it'll get better. And the pre-commit wrapper is https://github.com/ansible/awx-plugins/blob/db2cae7/tox.ini#L224-L343. It also has some external Python-level strictness + outputs consumable by my `reusable-tox.yml` that I'd like to eventually bring here.

sirosen:
Hm. Okay, I can accept that it's possible to run `mypy` more successfully in pre-commit. My experiences doing this have been pretty uniformly bad. But probably a big part of that is that I haven't had a working configuration to start from, and I see that you're using `pass_filenames: false` which probably has a big impact.


As an example of the kind of `tox` config I usually setup, in `dependency-groups` we have `mypy` under different Python versions configured: https://github.com/pypa/dependency-groups/blob/29ffcd9c8b7bf692073d6ff245ff2fa364dee581/pyproject.toml#L162-L173
Also, we're trying out the `[tool.toml]` config over there, which is interesting for a small project like that one.
(I only learned after setting up that config that `mypy` supports `--python-version` to control its behavior, so I might do this differently in the future.)

sirosen:
[...]

PR #2444, opened 2026-08-04
Fix a class in tests which fails under mypy

sirosen:
_Part of #2200._

I'm working through fixing our mypy configuration.
It's drifted into an extremely bad state.

Fixing requires significant efforts, and would produce an unreviewable diff if tackled all at once.
I'll be submitting a series of small changes to move us towards healthy config.

None of these changes will have changelogs.

---

When upgrading mypy linting, this class will be flagged for having
methods which return `None` but have types which do not match.
These methods are never used, so raise errors instead, making the types
harmonious.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import collections.abc as _c
+import functools
 import json
 import os
 import platform
@@ -10,10 +11,10 @@ import sys
 import typing as _t
 from contextlib import contextmanager
 from dataclasses import dataclass, field
-from functools import partial
 from importlib.metadata import version as version_of
 from pathlib import Path
 from textwrap import dedent
+from unittest import mock
 
 import pytest
 import tomli_w
@@ -118,17 +119,17 @@ class FakeRepository(BaseRepository):
     def options(self):
         return self._options
 
-    @property
+    @functools.cached_property
     def session(self) -> PipSession:
-        """Not used"""
+        return mock.Mock()
 
-    @property
+    @functools.cached_property
     def finder(self) -> PackageFinder:
-        """Not used"""
+        return mock.Mock()
 
-    @property
+    @functools.cached_property
     def command(self) -> InstallCommand:
-        """Not used"""
+        return mock.Mock()
 
 
 def pytest_collection_modifyitems(config, items):
@@ -188,7 +189,7 @@ def resolver(depcache, repository):
     # TODO: It'd be nicer if Resolver instance could be set up and then
     #       use .resolve(...) on the specset, instead of passing it to
     #       the constructor like this (it's not reusable)
-    return partial(
+    return functools.partial(
         LegacyResolver, repository=repository, cache=depcache, existing_constraints={}
     )
 
@@ -198,7 +199,7 @@ def backtracking_resolver(depcache):
     # TODO: It'd be nicer if Resolver instance could be set up and then
     #       use .resolve(...) on the specset, instead of passing it to
     #       the constructor like this (it's not reusable)
-    return partial(
+    return functools.partial(
         BacktrackingResolver,
         repository=FakeRepository(options=FakeOptions()),
         cache=depcache,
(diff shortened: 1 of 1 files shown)


sirosen:
Clearly I hadn't actually tested this properly. It was part of a large raft of type-checking fixes I had started on, and I thought I had tested them all. I'm not closing because fixing this will be an effort in itself.

Every inch of mypy fixing is going to require work, I think.

webknjaz (approved):
Looks correct

webknjaz:
> Every inch of mypy fixing is going to require work, I think.

Have you considered pointing multiple type checkers (pyrefly, ty, pyright etc) at the files you're changing to see what they'll reveal w/o really integrating them? Might be useful for more insight.

sirosen:
I don't think it's productive to pull in additional type checkers yet, but that will eventually be the plan.

Right now, I'm using the version-bump in [this comment](https://github.com/jazzband/pip-tools/issues/2200#issuecomment-5185928006) to make it possible to reveal incorrect or missing annotations. I can then fix them and revert the `pip` version bump. Once things are close enough to pass with `pip==26.1`, I'll do a final PR which actually bumps the `pip` version used for `mypy` checking.
After that, I can start to expand out the matrix of type checker tests:
```
[mypy, ty, pyright]
x
[pip-lowest, pip-supported] (+ pip-main)
x
[python3.10, python3.14]  # or some other range
```

Because of the significant size of this matrix:
- I'm starting with `mypy` only
- I'm returning to my belief that this really ought to be in `tox`, not pre-commit

webknjaz:
> I don't think it's productive to pull in additional type checkers yet

I was mostly thinking along the lines of a few more angles providing you hints when you're facing something weird. Not pulling them in beyond that. Just as an additional input for when MyPy.

> I'm returning to my belief that this really ought to be in `tox`, not pre-commit

I'm still not sure about that. I'd need a scalable strategy that I'd be able to replicate across multiple projects that would integrate across tox+GHA+coveralls (I'm uploading type preciseness coverage to coveralls instead of codecov for separation in a few projects now: https://coveralls.io/github/aio-libs/yarl; though, maybe at some point GH-native coverage hosting evolves to allow this too).

sirosen:
> > I'm returning to my belief that this really ought to be in `tox`, not pre-commit
> 
> I'm still not sure about that. I'd need a scalable strategy that I'd be able to replicate across multiple projects that would integrate across tox+GHA+coveralls (I'm uploading type preciseness coverage to coveralls instead of codecov for separation in a few projects now: https://coveralls.io/github/aio-libs/yarl; though, maybe at some point GH-native coverage hosting evolves to allow this too).

We should definitely continue to discuss this (I think #2200 might be the right place for longer form posts)!

Even with a single type checker, type checking has two qualities which I consider unsuitable for pre-commit:
1. it needs to be a matrix of jobs
2. the package dependencies must be installed

I'm ready to do whatever work is necessary to establish a good and reusable tox configuration; I believe it is the correct decision, not only for this project but for any project using `mypy`. Our current `mypy` setup -- which installs an old version of `pip` which we do not support -- does not meet my minimum quality standards.
I can fix it without a larger reconfiguration, as I am doing with this PR and others, but I think we should fix the deeper flaws after that is done.

---

I've just made a tweak here to use `mock.Mock()` instead of raising errors. It turns out that in spite of docstrings which said "not used", those methods were used. Filling with mocks seems to be sufficient.

tests/conftest.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import collections.abc as _c
+import functools
```
webknjaz, line 4: Unrelated but I remembered in context: I've suggested that Click starts lining `import typing as t` in the past and they ended up encouraging people to port a flake8 plugin for generic import convention rules, which apparently happened at PyCon US 2026 Sprints.. I discovered this yesterday and thought that I'd like to start configuring projects for a bunch of cases and here, I'd go for `_ft`, perhaps.

PR #2446, opened 2026-08-07
Replace all use of pytest's tmpdir fixture with tmp_path

sirosen:
This change is broken down into commits as an option for review, however the total scope of work is relatively simple.
`tmpdir` is a fixture which provides a `py.path`. `tmp_path` is newer and provides a pathlib `Path`.
Make the switch for a more uniform API within the testsuite.

For the helper fixture which does a `chdir`, there are several options for effecting it -- `monkeypatch` and `contextlib`, among others.
This PR chooses `monkeypatch`, but this can easily be changed.

This is test runtime infrastructure; no changelog is needed.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -253,7 +253,10 @@ def runner():
 
 
 @pytest.fixture
-def tmp_path_cwd(tmp_path, monkeypatch):
+def tmp_path_cwd(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> _t.Iterator[Path]:
+    """Wrap ``tmp_path`` to also chdir into it."""
+    # use an explicit monkeypatch context, rather than calling monkeypatch.chdir, so
+    # that calls to `monkeypatch.undo()` won't accidentally revert
     with monkeypatch.context() as mp:
         mp.chdir(tmp_path)
         yield tmp_path
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -253,6 +253,13 @@ def runner():
 
 
 @pytest.fixture
+def tmp_path_cwd(tmp_path, monkeypatch):
+    with monkeypatch.context() as mp:
+        mp.chdir(tmp_path)
+        yield tmp_path
+
+
+@pytest.fixture
 def tmpdir_cwd(tmpdir):
     with tmpdir.as_cwd():
         yield Path(tmpdir)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -166,7 +166,7 @@ def repository():
 
 
 @pytest.fixture
-def pypi_repository(tmpdir):
+def pypi_repository(tmp_path):
     return PyPIRepository(
         [
             "--index-url",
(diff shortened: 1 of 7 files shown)


tests/test_cli_compile.py
```diff
@@ -48,6 +48,13 @@
 )
 
 
+@pytest.fixture
+def tmp_path_cwd(tmp_path, monkeypatch):
+    with monkeypatch.context() as mp:
```
webknjaz: Unclear why you needed a CM in this first commit right away. `monkeypatch.chdir(tmp_path)` should've just worked w/o it AFAICS...

tests/conftest.py
```diff
@@ -253,26 +253,27 @@ def runner():
 
 
 @pytest.fixture
-def tmpdir_cwd(tmpdir):
-    with tmpdir.as_cwd():
-        yield Path(tmpdir)
+def tmp_path_cwd(tmp_path, monkeypatch):
+    with monkeypatch.context() as mp:
```
webknjaz, line 260: Oh, so you moved this here. The previous comment still applies, then.

Let's try getting rid of this line. The monkeypatch fixture already does exactly the same internally.

So unless in complicated cases, we shouldn't need another nested CM here and could just dedent the following two lines...

tests/conftest.py
```diff
@@ -253,26 +253,27 @@ def runner():
 
 
 @pytest.fixture
-def tmpdir_cwd(tmpdir):
-    with tmpdir.as_cwd():
-        yield Path(tmpdir)
+def tmp_path_cwd(tmp_path, monkeypatch):
```
webknjaz: Since you're already changing the signature, how about maybe adding type annotations and a docstring that would show up in the `--fixtures` output?

webknjaz (approved):

tests/conftest.py
sirosen: Sure, I'm happy to remove the context manager usage! As far as I'm concerned, this is one of many sensible ways of writing an equivalent or near-equivalent fixture.

tests/conftest.py
sirosen: Yeah, that's easy enough. :+1:

tests/conftest.py
sirosen: It turns out that we need this. I didn't dig into why, fully.

I only saw one use of `monkeypatch.undo()`, which is one of the ways that relying on `monkeypatch` without an explicit context can be problematic, but I assume there was something similar at play with a number of tests which broke when I changed it. I've added a comment which explains the problem with `undo()`, at the very least.

tests/test_cli_compile.py
RonnyPfannschmidt: theres a little nuance wrt nesting patch contexts - in this particular case i'd recommend direct monkeypatch usage without the context manager

tests/conftest.py
RonnyPfannschmidt: mixing monkeypatch.undo and monkeypatch.context can be problematic in some cases as monkeypatch contexts have no interaction with onther monkeypatch instances

i recommend naming the intended behavior pattern, as that may reverse my earlier suggestion of avoiding contexts

PR #2448, opened 2026-08-10
Update changelog for v7.6.1

sirosen:
Updates applied via `towncrier build --version v7.6.1`

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,48 @@
 
 <!-- towncrier release notes start -->
 
+## v7.6.1
+
+*2026-08-11*
+
+### Bug fixes
+
+- Fixed handling of index URLs ending in `/simple/` by improving URL normalization
+  logic.
+  These URLs would previously make `pip-compile` fail to use the PyPI JSON API.
+
+  -- by {user}`sirosen`
+
+  *PRs and issues:* {issue}`1669`
+
+- Added a missing dependency on {pypi}`typing-extensions` for Python 3.9 and 3.10
+  -- by {user}`sirosen`.
+
+  *PRs and issues:* {issue}`2424`
+
+### Features
+
+- pip-tools is now compatible with pip version 26.2 -- by {user}`sirosen`.
+
+  *PRs and issues:* {issue}`2436`, {issue}`2437`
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
webknjaz, line 1: Preview: https://pip-tools--2448.org.readthedocs.build/en/2448/changelog/#v7-6-1

CHANGELOG.md
```diff
@@ -2,6 +2,48 @@
 
 <!-- towncrier release notes start -->
 
+## v7.6.1
+
+*2026-08-10*
```
webknjaz: Looks like this will need to be bumped before the final merge. Also: I've been wondering — should this be UTC or just local to whoever does the release?

CHANGELOG.md
```diff
@@ -2,6 +2,48 @@
 
 <!-- towncrier release notes start -->
 
+## v7.6.1
```
webknjaz, line 5: We should probably document somewhere that adding a pip version compat doesn't imply a pip-tools minor version bump. I keep forgetting and go back to spend some time staring at the past release notes to figure this out...

webknjaz (approved):
I had a free deja vu feeling that I'd wanted to make use of the highlights change note in the upcoming release for *something* at some point to announce a thing that I cannot remember... I think it might've been the "ai" policy. Dunno. Feel free to consider this if you want. But other than that, I don't see a reason for pumping the breaks. Onwards!

CHANGELOG.md
sirosen: Personally, I think it has questionable utility. The date can be fetched from PyPI, our git tag metadata, and GitHub Releases.

I favor removing it, but I think that is out of scope for the current release. I'll update it to today's date for now.

sirosen: Bumped in an amended commit.

CHANGELOG.md
sirosen: 👍 I also, relatedly, want to establish a clearer compatibility policy. I think this could all be documented together.

PR #2450, opened 2026-08-12
Remove support for Python 3.9

sirosen:
- Update CI, tox, and pre-commit configs
- Update package metadata (classifier and lower bound)
- Various conditional coverage pragmas were phrased incorrectly in terms of Python 3.9 -- these pragmas are all removed
- importlib.metadata.PackageMetadata had a conditional definition, now removed (as it is available in 3.10+)

---

I have tried to keep this changeset minimal. It was tempting to include the addition of 3.15, but it's better to have tighter scope.

##### Contributor checklist

- [x] A change note is created in `changelog.d/` (see [`changelog.d/README.md`] for instructions) or the PR text says "no changelog needed".

[`changelog.d/README.md`]:

##### Maintainer checklist

- [x] Assign the PR to an existing or new milestone for the target version (following Semantic Versioning).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -321,7 +321,7 @@ jobs:
               inputs.cpython-versions
               && inputs.cpython-versions
               ||
-              '["3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "3.14t"]'
+              '["3.10", "3.11", "3.12", "3.13", "3.14", "3.14t"]'
             )
           }}
         pip-version: >-
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
     rev: 26.5.1
     hooks:
       - id: black
-        args: [--target-version=py39]
+        args: [--target-version=py310]
   - repo: https://github.com/PyCQA/isort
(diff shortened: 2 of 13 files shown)


sirosen:
I'm converting this to a draft as I consider it blocked.

Right now, coverage reporting on Windows is reporting a failure:
```
FAIL Required test coverage of 99.0% not reached. Total coverage: 98.45%
```

There isn't a _real_ reduction in coverage here, but there is an apparent one by percentage.

#2347 will help, and I intend to put in other fixes to improve coverage further.

sirosen:
I have a local integration branch with all of #2453, #2458, #2460, and #2466 combined.

I've just rebased this PR on top of those fixes to see if I've reached a point where the combined improvements to coverage would unblock this change.

**Update:** It passed! If we get these various coverage improvements merged, this PR can be rebased as something clean and unblocked.

----

It's nice to be cleaning up coverage, but I feel that I'm doing it under duress. If this doesn't apply well soon, I will _strongly_ advocate for loosening the coverage requirement so that we can drop 3.9. This also impacts being able to raise the lower bound on `click` versions, which in turn impacts a number of codepaths which only exist to keep things compatible.

PR #2452, opened 2026-08-12
Remove support for the legacy resolver

sirosen:
- Remove the `--resolver` option and env var
- Remove or update docs which refer to the legacy resolver
- Update test fixtures and tests to only use the backtracking resolver

One bit of the PyPIRepository is updated to work with the backtracking resolver, although it's not clear if this is really needed outside of a dedicated unit test.

---

This is an early draft, for evaluation for v8.0.0 .

One thing I note right away is that coverage has dropped a bit. Possibly that indicates that more codepaths should be eliminated.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist


webknjaz (commented):
Should we try integrating one of those libs for tracking deprecations via decorators where you can set the exact future versions?

sirosen:
I have never used any of those tools (e.g., `deprecated`), and personally do not find them compelling enough to even try. We can discuss at greater length, but I consider it off-topic for this PR. We should have our planning somewhere outside of this PR -- plus, I think there's going to be enough to discuss just regarding the legacy resolver removal.

-----

Looking at the stats, test coverage under this change falls to 97%. I haven't done thorough analysis yet, but I'm concerned that there are a number of meaningful features -- things we want or need to retain -- which were only being tested with the legacy resolver.

I limited test removals to those which were specific to the legacy resolver, and didn't expect such a significant impact. I might have to work through a series of smaller PRs to create missing tests for the backtracking resolver. I can rebase this continually and eventually get everything tested. Overall, it's somewhat disappointing but not shocking.

A related step that I'm contemplating is adding a tox factor which disables all tests which use the legacy resolver. I can use that to incrementally improve coverage.

PR #2455, opened 2026-08-13
Omit `pytest.fail` from coverage reporting

sirosen:
Exclude lines which are invocations of `pytest.fail()`. Such lines are not meant to be hit by the testsuite -- they're there to provide useful diagnostics when a test breaks.

-----

No changelog needed. This is an element of the work being carried out in #2347 which needs to happen to stabilize CI.
In particular, this handles a case in #2453 in which coverage reports that we aren't hitting a `pytest.fail(...)` line.

Note: In #2347 , this needs to be inserted in a different location in the `.coveragerc` because these lines are moved around. I'll handle the conflict resolution after this merges.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,6 +9,7 @@ omit =
 
 [report]
 exclude_also =
+    ^\s*pytest\.fail\(
     ^\s*@pytest\.mark\.xfail
     ^\s*if _t\.TYPE_CHECKING:
 include = piptools/*, tests/*


.coveragerc
```diff
@@ -12,4 +12,5 @@ fail_under = 99
 # coverage: https://coverage.readthedocs.io/en/latest/excluding.html#default-exclusions
 # covdefaults: https://github.com/asottile/covdefaults/blob/8d8712c26ad505f1269851a7c8431b8c1fedaa62/covdefaults.py#L86-L105
 exclude_also =
+    ^\s*pytest\.fail
```
webknjaz: Do we also want to match that opening brace? Also, I typically add the xfail mark here too, FTR.

webknjaz (approved):

.coveragerc
sirosen: Yeah, I can add the paren!

I think that we have the xfail mark listed in the coverage plugin PR, so after that clears the queue, I'll rebase this and we should have both. 🙂

PR #2456, opened 2026-08-13
Remove macos-{15,26}-intel from the build matrix

sirosen:
This is intentionally a bit of a nudge, and is phrased as a PR and not an issue/discussion.

Our CI matrix is significantly oversized, and the combinations we're testing aren't all equal value.
CI wall clock times, combined with instability, pose a real problem.
Merges are frequently kicked from the merge queue spuriously, and I've even had this happen during releases -- including the last one, v7.6.1 .
The long clock times mean that re-queuing when things fail is a significant cost to pay: I need to monitor the job for a long period to make sure it succeeds.

`pip` is not testing on Intel macOS.
We are not building an extension or otherwise doing work in `pip-tools` which we should expect to be sensitive to the architecture.

I do not believe that we are getting value from doing this extra testing, and we are incurring costs.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -314,7 +314,6 @@ jobs:
           - ubuntu-24.04
           - windows-2025
           - macos-26
-          - macos-15-intel
         python-version: >-
           ${{
             fromJSON(
@@ -445,7 +444,6 @@ jobs:
         runner-vm:
           - ubuntu-24.04
           - macos-26
-          - macos-26-intel
         python-version:
           - pypy-3.11
         pip-version:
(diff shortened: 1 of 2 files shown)


webknjaz:
What if we run these on `pull_request` events but not on release or merge queues? They'd still give us some feedback at the contribution stage but would not be causing us much discomfort otherwise.

sirosen:
Removing them for merges would definitely help.
If we're only going to run them in limited places, maybe only on pushes to `main`? Other projects have such workflows.

But before we set that up... Do we actually want these in the matrix? I'm stuck on the facts that pip itself does not do this testing and pip-tools's own code should not be impacted by the Intel vs M1 difference.

There are things which our matrix currently is missing, like various pip versions, which I think would be more useful to test. But it can become infeasible due to the sheer size of the matrix. I therefore think there's value in removing things -- if we can reach agreement that it's safe to do so -- in order to make room for other kinds of testing.

PR #2459, opened 2026-08-14
Add better self-testing of the local coverage plugin and meta-coverage reporting

sirosen:
Add a separate tox environment and CI job for doing the testing of the coverage plugin itself.

Because the plugin uses a separate coverage config for this test, both it and normal coverage can be maintained at 100%, and codecov can aggregate the two.

---

I'm not clear on whether or not it should get a separate changelog. I feel like it should fold into the one for #2347 , and have therefore not included one.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,7 +8,4 @@ codecov:
 
 comment: false # avoid spamming reviews
 
-ignore:
-  - "tests/_plugins/_coverage/*"
-
 ...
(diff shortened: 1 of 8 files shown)


webknjaz:
> I'm not clear on whether or not it should get a separate changelog. I feel like it should fold into the one for #2347 , and have therefore not included one.

Could make a `2347.contrib.1.md` and/or symlink 2459 there?

sirosen:
I added two symlinks for changelog fragments, both to the existing entry for #2264 -- one for this PR and one for #2347 . I think that's the best way to document this change.

PR #2015, opened 2023-11-03
Fix for `src-files` not being used when specified in a config file

csalerno-asml:
This PR fixes https://github.com/jazzband/pip-tools/issues/2006

`src-files` is found in the click context, `ctx`, but however is not populating the variable `src_files`, so that it is an empty tuple.
For this reason, the tool will then start to select a requirement file, as shown in the linked issue.

##### Contributor checklist

- [x] Included tests for the changes.
- [x] PR title is short, clear, and ready to be included in the user-facing changelog.

##### Maintainer checklist

- [x] Verified one of these labels is present: `backwards incompatible`, `feature`, `enhancement`, `deprecation`, `bug`, `dependency`, `docs` or `skip-changelog` as they determine changelog listing.
- [ ] Assign the PR to an existing or new milestone for the target version (following [Semantic Versioning](https://blog.versioneye.com/2014/01/16/semantic-versioning/)).

PR #2461, opened 2026-08-15
Place unreachable src_files logic under pragma

sirosen:
Introduced in #2015, this logic is meant to handle `src_files` being unset, and loading the config value as a default. However, under modern versions of click, the the `src_files` are consistently populated before this point, making the relevant code unreachable.

On older click versions, this check was necessary, and we only need older `click` version support for Python 3.9 . As such, a version pragma works to describe when this does and does not run.

-----

As this is a minor internal testing change, no changelog is provided here.

Initially, I intended to remove the unreachable lines. Testing various python versions and `click` versions revealed that they are used on older `click` versions and Python 3.9 .
Finishing 3.9 removal and setting a higher lower bound for `click` will make them unnecessary.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/piptools/scripts/compile.py
+++ b/piptools/scripts/compile.py
@@ -199,10 +199,12 @@ def cli(
         ctx.color = color
     log.verbosity = verbose - quiet
 
-    # If ``src-files` was not provided as an input, but rather as config,
-    # it will be part of the click context ``ctx``.
-    # However, if ``src_files`` is specified, then we want to use that.
-    if not src_files and ctx.default_map and "src_files" in ctx.default_map:
+    # NOTE: On older `click` versions, `src_files` is not populated automatically from
+    # NOTE: config, so it has to be done explicitly here. These align with older Python
+    # NOTE: support, so the entire block is marked with a version pragma.
+    if (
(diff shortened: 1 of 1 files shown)


piptools/scripts/compile.py
```diff
@@ -199,10 +199,12 @@ def cli(
...
-    # If ``src-files` was not provided as an input, but rather as config,
-    # it will be part of the click context ``ctx``.
-    # However, if ``src_files`` is specified, then we want to use that.
-    if not src_files and ctx.default_map and "src_files" in ctx.default_map:
+    # NOTE: On older `click` versions, `src_files` is not populated automatically from
+    # NOTE: config, so it has to be done explicitly here. These align with older Python
+    # NOTE: support, so the entire block is marked with a version pragma.
+    if (
+        not src_files and ctx.default_map and "src_files" in ctx.default_map
+    ):  # pragma: <=3.9 cover
```
webknjaz, line 207: Side thought: we need a linter that would explode on pragmas or similar mentions of unsupported Pythons. Related idea: https://github.com/tox-dev/workflow/issues/6

webknjaz (approved):

PR #2466, opened 2026-08-22
Add a missing unit tests for PyPIRepository

sirosen:
The `get_dependencies` helper validates its input, raising a TypeError if its contract is violated, but internally all usage is correct and does not trigger this error. A small new unit test feeds in bad inputs and confirms that the error is raised.

`PyPIRepository.find_best_match` has handling for editables, and it is similarly unreachable. A dedicated unit test exercises that handling path and minimally verifies that it short-circuits evaluation. (It is difficult to verify that a function call doesn't do extra work, so the test is imperfect.)

-----

More test additions in pursuit of better coverage; no changelog needed.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/tests/unit/repositories/test_pypi_repository.py
+++ b/tests/unit/repositories/test_pypi_repository.py
@@ -1,8 +1,14 @@
 from __future__ import annotations
 
-import pytest
+import pathlib
+import typing as _t
+from unittest import mock
 
-from piptools.repositories.pypi import _get_true_base_from_index_url
+import pytest
+from pip._internal.req import InstallRequirement
+
+from piptools.repositories.pypi import PyPIRepository, _get_true_base_from_index_url
+from tests.constants import PACKAGES_PATH
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tests/unit/repositories/test_pypi_repository.py
```diff
@@ -48,3 +54,38 @@
...
+        pypi_repository.get_dependencies(ireq)
+
+
+def test_find_best_match_short_circuits_on_editables(
+    pypi_repository: PyPIRepository,
+    from_editable: _t.Callable[[str], InstallRequirement],
+) -> None:
+    package_path = pathlib.Path(PACKAGES_PATH) / "small_fake_a"
+    ireq = from_editable(str(package_path))
+    with mock.patch.object(
```
webknjaz: If we have `pytest-mock` I the deps already, we should prefer using the mocker fixture instead.

webknjaz (approved):

tests/unit/repositories/test_pypi_repository.py
sirosen: Done!

PR #2467, opened 2026-08-22
Add missing tests for `--generate-hashes` as an update of existing pinned dependencies

sirosen:
These tests cover previously untested scenarios in which `--generate-hashes` is used when there's already an existing `requirements.txt` which pins/constrains requirements.

Two new effective new test cases are added:

- For tests of `pip-compile` on local URI dependencies, the options for "generate hashes" were previously `True/False`. They are expanded to `True/False/"update"`.

- A dedicated test for `--generate-hashes` on an existing pinned package   (copied from an existing test which used `six`) is added.

These new tests cover untested codepaths in the LocalRepository object.

----

No changelog, as this is yet-another coverage improvement PR.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/tests/test_cli_compile.py
+++ b/tests/test_cli_compile.py
@@ -847,7 +847,7 @@ def test_url_package(runner, line, dependency, generate_hashes):
         ),
     ),
 )
-@pytest.mark.parametrize("generate_hashes", ((True,), (False,)))
+@pytest.mark.parametrize("generate_hashes", (True, False, "update"))
 def test_local_file_uri_package(
     pip_conf, runner, line, dependency, rewritten_line, generate_hashes
 ):
@@ -855,6 +855,9 @@ def test_local_file_uri_package(
         rewritten_line = line
     with open("requirements.in", "w") as req_in:
         req_in.write(line)
+    if generate_hashes == "update":
+        with open("requirements.txt", "w") as fp:
+            fp.write(rewritten_line)
     out = runner.invoke(
         cli, ["-n", "--rebuild"] + (["--generate-hashes"] if generate_hashes else [])
     )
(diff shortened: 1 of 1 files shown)


tests/test_cli_compile.py
```diff
@@ -847,14 +847,17 @@ def test_url_package(runner, line, dependency, generate_hashes):
...
+@pytest.mark.parametrize("generate_hashes", (True, False, "update"))
 def test_local_file_uri_package(
     pip_conf, runner, line, dependency, rewritten_line, generate_hashes
 ):
     if rewritten_line is None:
         rewritten_line = line
     with open("requirements.in", "w") as req_in:
         req_in.write(line)
+    if generate_hashes == "update":
+        with open("requirements.txt", "w") as fp:
```
webknjaz, line 859: Case for `Path.write_text()`?

tests/test_cli_compile.py
```diff
@@ -1215,6 +1218,30 @@ def test_generate_hashes_with_annotations(runner):
         """)
 
 
+@pytest.mark.network
+def test_generate_hashes_with_existing_pins(runner, tmp_path_cwd):
```
webknjaz, line 1222: Wonder if we could emulate this w/o network I/O at some point 🤔

webknjaz (approved):

tests/test_cli_compile.py
sirosen: Yeah, I want to rewrite these tests, but kept it locally consistent for now.

I'll do a full pass over the testsuite for cases where we should update to pathlib!

tests/test_cli_compile.py
sirosen: An experiment I've been meaning to get around to is to try [vcr](https://pypi.org/project/vcrpy/) for some of these tests. It gets tricky for tests which pull down full dists -- I'd want to come up with some clever faking -- but I bet this test is only touching the metadata and JSON APIs.

I may be able to get to it this weekend; I'll try to fit it in. :smile:

PR #2469, opened 2026-08-27
Single hash generation 

### LINKED ISSUE #1330: Provide a way to limit to a single hash per requirement when generating hashes

Currently, when the `--generate-hashes` option is passed to `pip-compile`, `pip-compile` will include potentially many hashes per requirement, even if only one is needed / desired.

This feature request is for `pip-compile` to expose a command-line option that would limit the output to including only a single hash, namely the hash of what would be installed in the environment in which `pip-compile` is being run.

This option would be useful in situations where the exact deployment target is known (e.g. when using containers). For one, the requirements files generated by `pip-compile` would be shorter and easier to review since they wouldn't include extraneous info. Secondly, I believe this would provide greater determinism / reproducibility. For example, currently, if a release that was previously being installed from a requirements file generated by `pip-compile` was [yanked](https://pypi.org/help/#yanked) from PyPI, then the result of what would be installed from that requirements file would change, even though the requirements file didn't change (because the requirements file currently includes more hashes than what was originally installed).

kamplianitis:
##### Contributor checklist

- [X] Included tests for the changes.
- [X] A change note is created in `changelog.d/` (see [`changelog.d/README.md`]

[`changelog.d/README.md`]:

##### Maintainer checklist

---
Rebase of PR #1406 

I'm copying the PR's description so that the information is transmitted here also. 

**Changelog-friendly one-liner**: Option to restrict generated hashes to the single best matching file

Closes https://github.com/jazzband/pip-tools/issues/1330

### DIFF
--- a/changelog.d/1330.feature.md
+++ b/changelog.d/1330.feature.md
@@ -1,1 +1,2 @@
-Option to restrict generated hashes to the single best matching file -- by {user} `plannigan` `kamplianitis`
+Added a new CLI option to restrict generated hashes to the single best
+matching file -- by {user} `plannigan` and {user}`kamplianitis`.
--- a/changelog.d/1330.feature.md
+++ b/changelog.d/1330.feature.md
@@ -1,0 +1,1 @@
+Option to restrict generated hashes to the single best matching file -- by {user} `plannigan` `kamplianitis`
--- a/README.md
+++ b/README.md
@@ -276,6 +276,10 @@ sqlparse==0.4.3 \
     # via django
 ```
 
+By default, this will include a hash for every release file available for the
+given version. This behavior can be changed to only include a hash for the best
+match file for the current environment with the `--single-hash` flag.
+
 ### Output File
 
 To output the pinned requirements in a filename other than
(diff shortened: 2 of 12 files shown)


webknjaz:
@kamplianitis FYI the PR template has changed over the years, you can find the new one in the repo and fill it out. You can keep the old one in a separate section after a `---` separator if you want to preserve it for history, but we'll probably want to discard most of it later anyway. Makes sure you fill out the current template to the best of your ability, though.

One other thing that's changed is that we've adopted a different method for tracking the change log entry collection — now the submitters must add a text file (or a few, sometimes through symlinks, sometimes nothing) as a part of the patch and we review that. Please check out the current contribution guidelines on RTD for the details.

If you think you'll need some time to work this out, you can also mark the PR as a Draft (there's a button somewhere on GH) and the undraft it again once it's ready for review.

changelog.d/1330.feature.md
```diff
@@ -0,0 +1,2 @@
+Option to restrict generated hashes to the single best matching file -- by
+{user} `plannigan` `kamplianitis`
```
webknjaz: Use full sentences et al. Render preview: https://pip-tools--2469.org.readthedocs.build/en/2469/changelog/#features
```suggestion
Added a new CLI option to restrict generated hashes to the single best
matching file -- by {user} `plannigan` and {user}`kamplianitis`.
```

PR #2473, opened 2026-09-02
Fix wrong default check in get_compile_command (click 8.5.0)

### LINKED ISSUE #2472: get_compile_command emits --no-index unconditionally with click 8.5.0
### Describe the Bug

With click 8.5.0, `pip-compile` writes `--no-index` into the recorded command, even when it was not passed.

This is only in the recorded command text. The actual resolve step still uses the index.

### Expected Behaviour

The recorded command should not include `--no-index` unless it was really passed.

### Steps to Reproduce

1. `pip install pip-tools "click==8.5.0"`
2. `echo requests > req.in`
3. `pip-compile req.in`
4. `head -6 req.txt`

Result:

```
#    pip-compile req.in --no-index
```

`--no-index` was never passed.

I checked `click.Option(['--no-index'], is_flag=True).default` on different click versions. It is `False` on 8.2.1, 8.3.0, 8.4.0, 8.4.2. It becomes `Sentinel.UNSET` on 8.5.0. So this is a new bug from click 8.5.0, not the one fixed in #2235 / #2236.

The cause on the click side: pallets/click#3641 (pallets/click@8f30085, in 8.5.0) removed the old eager coercion `self.default = False` for boolean flags at `__init__` time. Now `Option.default` stays `UNSET` unless resolved through `get_default()`.

`get_compile_command` in `piptools/utils.py` still compares `option.default == value` to skip options at their default. For `--no-index` (a flag with no negative form), this check never matches on click 8.5.0. So the flag name always gets added, no matter its real value.

### Environment Versions

- pip-tools 7.6.1 (still present on current `main`)
- click 8.5.0 (not on 8.2.1 / 8.3.0 / 8.4.0 / 8.4.2)
- Python 3.13

### LINKED ISSUE #2474: Fix click compatiblity for v8.5

mnencia:
Fixes #2472.

With click 8.5.0, `Option.default` is an internal sentinel for a flag with no explicit default, not `False`. `get_compile_command` compares `option.default` to the value to decide if an option is at its default. For `--no-index` (a flag with no negative form), this compare never matches. So `--no-index` gets added to the recorded command every time, even when not passed.

Fix: use `option.get_default()` instead of the raw attribute. I use a new, empty `click.Context` for this call, not the real one. The real context can hold values from a config file, and those would look like defaults too.

##### Contributor checklist

- [x] A change note is created in `changelog.d/` (see [`changelog.d/README.md`])

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/piptools/utils.py
+++ b/piptools/utils.py
@@ -362,8 +362,10 @@ def get_compile_command(click_ctx: click.Context) -> str:
         if value is None:
             continue
 
-        # Skip options with a default value
-        if option.default == value:
+        # Skip options with a default value.
+        # Use a new Context so that values from config files are
+        # not treated as defaults.
+        if option.get_default(click.Context(click_ctx.command)) == value:
             continue
 
         # Use a file name for file-like objects
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -460,6 +460,16 @@ def test_get_compile_command_with_config(tmp_path_cwd, config_file, expected_com
         assert get_compile_command(ctx) == expected_command
 
 
+def test_get_compile_command_no_index_not_added_by_default(tmp_path_cwd):
+    """
+    --no-index must not show up in the recorded command unless it was
+    actually passed. See GH-2472.
+    """
+    (tmp_path_cwd / "requirements.in").touch()
+    with compile_cli.make_context("pip-compile", ["requirements.in"]) as ctx:
+        assert get_compile_command(ctx) == "pip-compile requirements.in"
+
+
 @pytest.mark.parametrize("config_file", ("pyproject.toml", ".pip-tools.toml"))
 @pytest.mark.parametrize(
     "config_file_content",
--- a/changelog.d/2472.bugfix.md
+++ b/changelog.d/2472.bugfix.md
@@ -1,0 +1,2 @@
+Fixed `pip-compile` sometimes adding `--no-index` to the recorded command
+even when it was not passed, with click 8.5.0 -- by {user}`mnencia`.


tests/test_utils.py
```diff
@@ -460,6 +460,16 @@ def test_get_compile_command_with_config(tmp_path_cwd, config_file, expected_com
...
 
 
+def test_get_compile_command_with_config_option_value(tmp_path_cwd):
+    """
+    A value set in a config file, not on the command line, must still
+    show up in the recorded command. See GH-2472.
+    """
+    Path("pyproject.toml").write_text("[tool.pip-tools]\ngenerate-hashes = true\n")
+    with compile_cli.make_context("pip-compile", []) as ctx:
+        assert get_compile_command(ctx) == "pip-compile --generate-hashes"
```
sirosen: Is this not a bug in `pip-tools`? This test codifies it as "correct" behavior, but I'm not sure that it's correct in the first place. Do we have evidence either way?

Past maintainers and contributors may have had differing opinions on this. IMO it's questionable at best. (In fact, several of the tests for `get_compile_command()` codify behaviors that I think are extremely questionable or flat-out wrong.)

I'd rather see a test which doesn't rely on the config loading behavior.

piptools/utils.py
```diff
@@ -362,8 +362,11 @@ def get_compile_command(click_ctx: click.Context) -> str:
         if value is None:
             continue
 
-        # Skip options with a default value
-        if option.default == value:
+        # Skip options with a default value. Use get_default(), not
+        # option.default: click may set that to an internal sentinel, not
+        # the real default. Use a new Context, not click_ctx, so a value
+        # from a config file is not treated as a default too.
+        if option.get_default(click.Context(click_ctx.command)) == value:
```
sirosen: Not only the new code, but also the original, decides that explicitly setting an option to its default value will cause it to be omitted from the command.
I think that's not correct, but probably out of scope to fix right now, sadly.

Please don't over-explain `get_default()` vs `default` -- let's trim that entirely. `Option.default` is not a public API. e.g.,

```suggestion
        # Skip options with a default value.
        # Use a new Context so that values from config files are
        # not treated as defaults.
        if option.get_default(click.Context(click_ctx.command)) == value:
```

sirosen (commented):
Thanks for submitting the issue and proposed fix.

I take issue with the test included. I'd rather see one which reflects realistic usage and expectations -- your original issue is about `--no-index`, so can you demonstrate it in the form of a regression test which fails on `main` and passes with the fix?

I find the overall way that `get_compile_command()` interacts with config to be (very) poor, but I don't want that to block all improvements.

piptools/utils.py
mnencia: Applied as is.

tests/test_utils.py
mnencia: Changed. New test uses plain requirements.in, no config file. Fails on main, passes with the fix.

mnencia:
Agreed, the config interaction is out of scope here.

Added a direct --no-index test, no config involved. Also trimmed the code comment and commit message.

tests/test_utils.py
```diff
@@ -460,6 +460,17 @@ def test_get_compile_command_with_config(tmp_path_cwd, config_file, expected_com
         assert get_compile_command(ctx) == expected_command
 
 
+def test_get_compile_command_no_index_not_added_by_default(tmp_path_cwd):
+    """
+    --no-index must not show up in the recorded command unless it was
+    actually passed. See GH-2472.
+    """
+    with open("requirements.in", "w"):
+        pass
```
sirosen, line 469: As a minor tweak, let's use `(tmp_path_cwd / "requirements.in").touch()`.

I'm working on moving all tests, gradually, off of `with open(...): pass` as a pattern.

mnencia: Done.

sebfelix:
The `--no-index` being present in the outputs inevitably breaks `renovate` 's compatibility with pip due to it being unable to process its own outputs in some step.

```
DEBUG: pip-compile: Failed to run command (repository=f13/templates/microservice-template, baseBranch=148-bug-restore-renovate-pip-behaviour, branch=renovate/pip-compile-refresh)
       "err": {"message": "Option --no-index not supported (yet)"}
       Error: Option --no-index not supported (yet)
           at throwForUnknownOption (file:///usr/local/renovate/lib/modules/manager/pip-compile/common.ts:344:9)
           at extractHeaderCommand (file:///usr/local/renovate/lib/modules/manager/pip-compile/common.ts:227:5)
           at updateArtifacts (file:///usr/local/renovate/lib/modules/manager/pip-compile/artifacts.ts:119:27)
           at managerUpdateArtifacts (file:///usr/local/renovate/lib/workers/repository/update/branch/get-updated.ts:507:10)
           at getUpdatedPackageFiles (file:///usr/local/renovate/lib/workers/repository/update/branch/get-updated.ts:450:13)
           at processBranch (file:///usr/local/renovate/lib/workers/repository/update/branch/index.ts:605:19)
           at instrument.attributes (file:///usr/local/renovate/lib/workers/repository/process/write.ts:188:21)
           at writeUpdates (file:///usr/local/renovate/lib/workers/repository/process/write.ts:161:17)
           at update (file:///usr/local/renovate/lib/workers/repository/process/extract-update.ts:304:11)
           at renovateRepository (file:///usr/local/renovate/lib/workers/repository/index.ts:142:21)
           at instrument.attributes (file:///usr/local/renovate/lib/workers/global/index.ts:235:11)
           at start (file:///usr/local/renovate/lib/workers/global/index.ts:220:7)
           at file:///usr/local/renovate/lib/renovate.ts:26:22
```

Would be great to see this merged soon!

sirosen:
Note that we can't merge this, since we have numerous other failures in CI on the latest `click` version.

Quite likely a better near-term fix is for `pip-tools` to restrict its `click` version to `<8.5`. But that doesn't mean this should be discarded; it will still be useful when we do the upgrade.

PR #2475, opened 2026-09-10
Make pip-tools compatible with click version 8.5

### LINKED ISSUE #2474: Fix click compatiblity for v8.5

sirosen:
This fix builds on and closes #2473.

In addition, this changeset handles `LazyFile` objects by means of a new `_click_compat` module.
Not only will this module provide compatibility as `click` changes, but it's also a good location for basic wrappers (as in this case).

Explicitly opening the file now also includes setting up the deferred closing behavior, so the usage site is simplified to its core concern of getting an open stream to write output.

#2473 includes new dedicated tests, but no explicit testing targets `_click_compat`.
I can add some more tests on request, but I think our existing coverage gets this pretty well.

I've left two distinct news fragments in place, as I think that captures the work better than a combined one would.

##### Contributor checklist

[`changelog.d/README.md`]:

##### Maintainer checklist

### DIFF
--- a/changelog.d/2474.bugfix.md
+++ b/changelog.d/2474.bugfix.md
@@ -1,0 +1,1 @@
+Make `pip-tools` compatible with `click` version `8.5` -- by {user}`sirosen`.
--- a/changelog.d/2474.bugfix.md
+++ b/changelog.d/2474.bugfix.md
@@ -1,1 +1,1 @@
-Make `pip-tools` compatible with `click` version `8.5` -- by {user}`sirosen`.
+Make `pip-tools` compatible with {pypi}`click` version `8.5` -- by {user}`sirosen`.
--- a/piptools/_compat/_click_compat.py
+++ b/piptools/_compat/_click_compat.py
@@ -1,0 +1,41 @@
+"""
+Compatibility helpers and wrappers for working with :external+click:doc:`click <index>`.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import typing as _t
+
+import click
+
+
+def open_output_file(ctx: click.Context, filename: str) -> _t.BinaryIO:
+    """
+    Open a file with the desirable flags set for writing output.
+
+    The file will be lazy and atomic unless it's stdout.
+    """
+    # open_file() returns a typing.IO , but we know it will be binary because of the
+    # mode flag -- so type-ignore the assignment issue
+    file: _t.BinaryIO = click.open_file(  # type: ignore[assignment]
+        filename, "w+b", atomic=True, lazy=True
+    )
+    _defer_lazy_file_close(ctx, file)
+    return file
+
+
+# note that the LazyFile type is not public, so we treat the file object as "Any"
+def _defer_lazy_file_close(ctx: click.Context, fileobj: _t.Any) -> None:
+    """Setup a click "lazy file" to close on exit."""
+    ctx.call_on_close(lambda: _safe_close(fileobj))
+
+
+def _safe_close(fileobj: _t.Any) -> None:
+    """
+    Suppress *all* errors and call ``close_intelligently()``.
+
+    This will not error even if the object does not support ``close_intelligently()``.
+    """
+    with contextlib.suppress(Exception):
+        fileobj.close_intelligently()
--- a/piptools/_compat/_click_compat.py
+++ b/piptools/_compat/_click_compat.py
@@ -25,17 +25,18 @@ def open_output_file(ctx: click.Context, filename: str) -> _t.BinaryIO:
     return file
 
 
-# note that the LazyFile type is not public, so we treat the file object as "Any"
-def _defer_lazy_file_close(ctx: click.Context, fileobj: _t.Any) -> None:
+def _defer_lazy_file_close(ctx: click.Context, fileobj: _t.BinaryIO) -> None:
     """Setup a click "lazy file" to close on exit."""
     ctx.call_on_close(lambda: _safe_close(fileobj))
 
 
-def _safe_close(fileobj: _t.Any) -> None:
+def _safe_close(fileobj: _t.BinaryIO) -> None:
     """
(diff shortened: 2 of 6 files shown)


changelog.d/2474.bugfix.md
```diff
@@ -0,0 +1 @@
+Make `pip-tools` compatible with `click` version `8.5` -- by {user}`sirosen`.
```
webknjaz: Note we can use the `{pypi}` role to link these things.

piptools/_compat/_click_compat.py
```diff
@@ -0,0 +1,41 @@
...
+    # mode flag -- so type-ignore the assignment issue
+    file: _t.BinaryIO = click.open_file(  # type: ignore[assignment]
+        filename, "w+b", atomic=True, lazy=True
+    )
+    _defer_lazy_file_close(ctx, file)
+    return file
+
+
+# note that the LazyFile type is not public, so we treat the file object as "Any"
+def _defer_lazy_file_close(ctx: click.Context, fileobj: _t.Any) -> None:
```
webknjaz: Can this be
```suggestion
def _defer_lazy_file_close(ctx: click.Context, fileobj: object) -> None:
```
then?

webknjaz: Actually, since we call it with
```suggestion
def _defer_lazy_file_close(ctx: click.Context, fileobj: _t.BinaryIO) -> None:
```
Why not just put it in the signature?

changelog.d/2474.bugfix.md
sirosen: :+1: That seems worth doing. I'll update!

piptools/_compat/_click_compat.py
sirosen: I think that anything other than `Any` (with its loose behaviors) will require a type-ignore comment because `close_intelligently()` is only defined on `LazyFile` objects (and direct import/use of LazyFile is what's being deprecated).

However, I'm not particularly averse to the type-ignore. I'll update to that style (`BinaryIO`, since it's the most precise), and if you dislike it I'll revert.

mnencia:
I reviewed the diff, and the `_click_compat` refactor for the `LazyFile` handling looks clean. Approve.

mnencia (approved):

changelog.d/2474.bugfix.md
webknjaz: Marking as resolved.

piptools/_compat/_click_compat.py
```diff
@@ -0,0 +1,42 @@
...
+def _safe_close(fileobj: _t.BinaryIO) -> None:
+    """
+    Suppress *all* errors and call ``close_intelligently()``.
+
+    This will not error even if the object does not support ``close_intelligently()``.
+    """
+    # Note that the LazyFile type is not public, so we are passed a BinaryIO and we will
+    # try to use the method (which we know `click` provides).
+    with contextlib.suppress(Exception):
+        fileobj.close_intelligently()  # type: ignore[attr-defined]
```
webknjaz: Finally looked at https://github.com/pallets/click/blob/6aabf099bfdd4c1e75fe8d0e0d4241372b988ab1/src/click/utils.py#L188C9-L188C28 and realized why this was needed.

Though, now, I'm questioning whether this is something that makes sense to keep reimplemented here. This method is nothing more than a gated `fileobj.close()`, after all.


```suggestion
    if not getattr(fileobj, 'should_close'):
        return

    with contextlib.suppress(OSError):
        fileobj.close()
```

webknjaz: In https://github.com/pallets/click/issues/3099#issuecomment-3657969042 @davidism said that the interface is about the same as with normal files. He probably didn't think that anybody would be using the additional non-underscored methods like we do back then. And so when @Rowlando13 renamed it in https://github.com/pallets/click/pull/3695 nobody gave typing a second thought it seems.

tests/test_utils.py
```diff
@@ -460,6 +460,16 @@ def test_get_compile_command_with_config(tmp_path_cwd, config_file, expected_com
         assert get_compile_command(ctx) == expected_command
 
 
+def test_get_compile_command_no_index_not_added_by_default(tmp_path_cwd):
```
webknjaz:
```suggestion
def test_get_compile_command_no_index_not_added_by_default(tmp_path_cwd: Path) -> None:
```

webknjaz (approved):
@sirosen I'll leave you with a blanket approval and some suggestions inline. Feel free to either iterate some more or postpone and merge if you feel like cutting an urgent release.

piptools/_compat/_click_compat.py
sirosen: Yeah, on a related issue [I asked](https://github.com/pallets/click/issues/3847#issuecomment-5619758046) just a few days ago if the intent was for `open_file` itself to become private. It is meant to be public, which I think makes it a problem that we can't accurately describe its return type using public interfaces/types.

Lower in the thread, I suggested a public `Protocol` class. I haven't made a PR for it yet, but intend to follow up.

---

Regarding the change you suggested, I feel like it's about the same as the call to `close_intelligently()`, since `should_close` is also a special feature of the click LazyFile type. I guess it removes the need for a type-ignore. They feel pretty equivalent to me? :shrug: 

I'm a bit uncertain, so let's go with it. We can always alter it again in the future, when we get the desired interface for `open_file()` cleared up. :smile:

webknjaz:
@sirosen looks like new `setuptools` release is unhappy with our packaging config? Good thing it's a non-voting job now, and we can fix that separately.. We should think about pinning the build deps at some point...

sirosen:
> Good thing it's a non-voting job now

Oh, can I merge without fixing that? I was planning to dig into it after the end of my work day. I still want to figure it out and fix it.

And yes, let's get to pinning our build deps soon. Maybe even before our next release. :smile:

PR #2476, opened 2026-09-10
Fix Click 8.5 compatibility

### LINKED ISSUE #2472: get_compile_command emits --no-index unconditionally with click 8.5.0

### LINKED ISSUE #2474: Fix click compatiblity for v8.5

GruffElixir:
Click 8.5 deprecates the LazyFile and safecall imports used by pip-compile. Remove those imports while preserving lazy atomic output closing through the file-like interface. Handle Click's UNSET boolean default so recorded commands do not gain a spurious --no-index.

Tests:
- tests/test_utils.py with Click 8.5.0 (191 passed)
- tests/test_utils.py with Click 8.4.2 (191 passed)

Fixes #2474
Fixes #2472


sirosen:
Hard close what are you doing man.

webknjaz:
FTR, this is a sloperator: https://agentscan.tools/user/gruffelixir. I wonder if they're using another account or somebody else's email to hide spam.

PR #2478, opened 2026-09-19
Fix the build-dists job -- updated license metadata + pinned build deps

sirosen:
License metadata needed to be updated to match with PEP 639. (Newer `setuptools` versions are getting stricter about this.)

To reduce future surprises, I've also taken some steps to pin build deps for `tox run -e build-dists`.

##### Contributor checklist

- [x] A change note is created in `changelog.d/` (see [`changelog.d/README.md`] for instructions) or the PR text says "no changelog needed".

[`changelog.d/README.md`]:

##### Maintainer checklist


pyproject.toml
```diff
@@ -10,13 +10,13 @@ name = "pip-tools"
 description = "pip-tools keeps your pinned dependencies fresh."
 readme = "README.md"
 authors = [{ "name" = "Vincent Driessen", "email" = "me@nvie.com" }]
-license = { text = "BSD" }
+license = "BSD-3-Clause"
+license-files = ["LICENSE"]
```
webknjaz, line 14: I think setuptools even finds this file automatically

tox.ini
```diff
@@ -224,6 +224,28 @@ commands_post =
...
+    {[python-cli-options]byte-errors} \
+    {[python-cli-options]max-isolation} \
+    {[python-cli-options]warnings-to-errors} \
+    -m piptools compile \
+    --strip-extras \
+    --allow-unsafe \
+    --quiet \
+    --only-build-deps \
+    --build-deps-for wheel \
+    -o ./requirements/build-deps.txt \
```
webknjaz: Can we use long option names in scripts?

tox.ini
```diff
@@ -233,6 +255,14 @@ depends =
   cleanup-dists
 deps =
   build
+# set constraint files for
+# - uv
+# - pip < 26.2 (PIP_CONSTRAINT)
+# - pip >= 26.2 (PIP_BUILD_CONSTRAINT)
+setenv =
+  UV_BUILD_CONSTRAINT=./requirements/build-deps.txt
```
webknjaz: I usually make these sparse too
```suggestion
  UV_BUILD_CONSTRAINT = ./requirements/build-deps.txt
```

tox.ini
```diff
@@ -233,6 +255,14 @@ depends =
   cleanup-dists
 deps =
   build
+# set constraint files for
```
webknjaz: Maybe move the comment inside the `set_env` section so it doesn't drift away?

tox.ini
```diff
@@ -224,6 +224,28 @@ commands_post =
...
+  {envpython} \
+    {[python-cli-options]byte-errors} \
+    {[python-cli-options]max-isolation} \
+    {[python-cli-options]warnings-to-errors} \
+    -m piptools compile \
+    --strip-extras \
+    --allow-unsafe \
+    --quiet \
+    --only-build-deps \
+    --build-deps-for wheel \
```
webknjaz: We should probably not limit this to wheel. Sdist and editable also deserve the same treatment, although their trees are unlikely to ever diverge.

tox.ini
```diff
@@ -224,6 +224,28 @@ commands_post =
...
+# compile requirements.in + pyproject.toml to get pip-tools install requirements in
+# addition to doc tooling requirements *but without* putting `pip-tools` itself
+# into the output
+commands =
+  {envpython} \
+    {[python-cli-options]byte-errors} \
+    {[python-cli-options]max-isolation} \
+    {[python-cli-options]warnings-to-errors} \
+    -m piptools compile \
+    --strip-extras \
```
webknjaz: Let's start indenting more at this point
```suggestion
      --strip-extras \
```

pyproject.toml
sirosen: I think you're right (I'd want to verify if we switch it). I'm still a little bit inclined towards writing it out though, even though it might not be necessary. I'm not sure I can give a great reason for that though. It's a little more explicit; that might be all.

webknjaz: Yeah, I guess explicit might be better. I've been following jaraco's lead blindly with the implicit stuff and shared infra autodiscovery here but can't say I have strong opinions.

webknjaz (approved):

webknjaz:
@sirosen looks like pre-commit.ci pulls in different rules for pymarkdown than GHA 🤔

sirosen:
Yeah, I noticed that too. I've got #2479 up as a fix (just rebased).

I was wondering if we're missing pre-commit from our GHA setup in some way -- I didn't dig deep, just fixed the issues at the surface level.

